PR #21, opened 2017-03-02
One liners to correct register designation in rvc-instr-table

David-Horner:
No initial register value is not used by C.LI, so only rd is valid.

No initial register value is not used by C.LUI, so only rd is valid.

Both r1 and rd are set by the C.ADD expansion (to the same register id)

### DIFF
--- a/src/rvc-instr-table.tex
+++ b/src/rvc-instr-table.tex
@@ -198,7 +198,7 @@
 &
 \multicolumn{3}{|c|}{010} &
 \multicolumn{1}{c|}{imm[5]} &
-\multicolumn{5}{c|}{rs1/rd$\neq$0} &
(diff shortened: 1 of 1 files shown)


src/rvc-instr-table.tex
```diff
@@ -489,7 +489,7 @@
 &
 \multicolumn{3}{|c|}{100} &
 \multicolumn{1}{c|}{1} &
-\multicolumn{5}{c|}{rd$\neq$0} &
+\multicolumn{5}{c|}{r1/rd$\neq$0} &
```
aswaterman: should be rs1, not r1

David-Horner:
made fix (and finally brute forced the revision to Github)

aswaterman:
The bug fix didn't come through in this PR, but I applied the change manually and merged.  Thanks again.

PR #27, opened 2017-03-07
Make two's complement explicit early on (in introduction).

DSHorner:
Two's complement is implicit elsewhere in the document;
Technically, operations define the encoding type,
and two's complement is almost universal,
but it still is valuable to make the sign encoding explicit (IMO).

### DIFF
--- a/src/intro.tex
+++ b/src/intro.tex
@@ -217,8 +217,8 @@ systems (with additional supervisor-level operations), and so provides
 a convenient ISA and software toolchain ``skeleton'' around which more
 customized processor ISAs can be built.
 
-Each base integer instruction set is characterized by the width of the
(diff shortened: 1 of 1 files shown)


aswaterman:
I would say it is not a property of the registers, though, which can just as easily hold sign-magnitude or unsigned or one's complement numbers. It is a property of the instructions that interpret the bits in the registers.

I think we can get away with a single sentence somewhere else, perhaps the first time "signed" is mentioned?

kasanovic:
Agree with Andrew this isn't tied to registers, so added this sentence: The base integer
instruction sets use a two's-complement representation for signed
integer values.

DSHorner:
Thank you. You found a good place for it.

PR #31, opened 2017-03-20
Cimm

DSHorner:
A preview of what the C instruction format table will look like (offset was not included - I think for consistency it should be. But your call)
for the immediate values:
simm - sign extended (instruction bit 12 is always the sign)
uimm - unsigned zero extended
shamt - shift amount
nzsimm, nzuimm and nzshamt are corresponding values that all bits must not be zero.

I have not revised any C instruction full description other than ADDI16SPN.

I believe there is value in making these codes consistent throughout the C extension document.
But is that scope creep?

### DIFF
--- a/src/c.tex
+++ b/src/c.tex
@@ -827,7 +827,7 @@ is always 16-byte aligned.
 \multicolumn{1}{c|}{op} \\
 \hline
 3 & 8 & 3 & 2 \\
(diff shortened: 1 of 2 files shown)


DSHorner:
I don't know where the  	6cbc9f3 pull request came from - sorry about that.

aswaterman:
thanks

PR #33, opened 2017-03-20
Specify encoding of mvendorid field

aswaterman:


asb:
Perhaps I'm missing something obvious, but isn't the 8-byte id a subset of 0x01-0x7d? e.g. AMD is 0x01 in the first bank.

aswaterman:
Ah, oops, the MSB is an odd parity bit... I misremembered it as a stop bit.

aswaterman:
thanks!

PR #34, opened 2017-03-20
Any write to misa may cause UXL or SXL to change as a side effect

aswaterman:


DSHorner:
What of:
The SXL field is only implemented if S-mode is implemented, and is
 otherwise hard-wired to zero.  An implementation must enforce that
 SXL$\leq$MXL. Thus, a write to {\tt misa} changing MXL may cause 
 SXL to  change as a side effect.  

The wording "any write to {\tt misa}" seems to imply that even following Write Preserve {WPRL} precautions writes to misa can do magic things to SXL. Perhaps, though, the Write Preserve Precautions should be made more evident or perhaps explicit, not just for Reserved space in CSR but any write.

aswaterman:
misa doesn't have any WPRL fields, but yeah, the intent is that any write to misa may have side effects on SXL/UXL, even if misa is preserved.

DSHorner:
This would establish a strong precedent of co-dependence between CSR.
I understand that misa is a fundamental register, and significant in that it contains MXL.
All the extension fields are also potentially invasive to core processing, however,
none of these extension transitions need to invalidate MXL/SXL/UXL.
If the MXL or SXL/UXL functionality is implemented retaining SXL/UXL when MXL is not changed in a misa write should not be an insurmountable challenge.

Issue #54, opened 2017-04-25
ECALL and EBREAK should be documented somewhere


sorear:
The user spec supports multiple privileged architectures so it is vague about what actually happens.  The standard privileged architecture needs to define the behavior of ECALL and EBREAK in terms of trap handling.

PR #55, opened 2017-04-25
Describe ECALL/EBREAK operation in privileged architecture

aswaterman:
#54

### DIFF
--- a/src/machine.tex
+++ b/src/machine.tex
@@ -1856,7 +1856,7 @@ EBREAK  & 0 & PRIV & 0 & SYSTEM \\
 
 The ECALL instruction is used to make a request to the supporting execution
 environment.  When executed in U-mode, S-mode, or M-mode, it generates an
-environment call from U-mode, environment call from S-mode, or environment
+environment call from U-mode exception, environment call from S-mode exception, or environment
 call from M-mode exception, respectively, and performs no other operation.
 \begin{commentary}
 ECALL generates a different exception for each originating privilege mode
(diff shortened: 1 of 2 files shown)


src/machine.tex
```diff
@@ -1830,6 +1830,54 @@ \subsection{Machine Bad Bits ({\tt mtval}) Register}
...
+\multicolumn{1}{c|}{opcode} \\
+\hline
+12 & 5 & 3 & 5 & 7 \\
+ECALL   & 0 & PRIV & 0 & SYSTEM \\
+EBREAK  & 0 & PRIV & 0 & SYSTEM \\
+\end{tabular}
+\end{center}
+
+The ECALL instruction is used to make a request to the supporting execution
+environment.  When executed in U-mode, S-mode, or M-mode, it generates an
```
sorear: "Generates an environment call" is a little bit vague.  Would something more like «generates an exception of type "environment call from U-mode", "environment call from S-mode", or "environment call from M-mode", respectively» work?

Also, would it make sense to add a commentary to the user spec saying that the runtime behavior is documented elsewhere?

sorear: On that note we don't seem to have anything in the privilege spec that spells out the sequence of events for a trap in one place, although §3.1.13 (medeleg) covers most of them.

aswaterman: Thanks, that was an editing error.

aswaterman: The user spec already implies the runtime behavior is defined elsewhere

> The ABI for the system will define how parameters for the environment request...

and indeed I think it's best to leave it described as an ABI thing.

PR #61, opened 2017-05-16
Add ILEN to simplify descriptions of {m,s}tval CSRs

jcb62281:
PR as requested on isa-dev.  (I finally made a GitHub account for this.)

I had enough difficulty getting the list spacing to look reasonable and ended up with sufficiently different "spacing magic" in machine.tex and supervsior.tex that I added the "paralist" package to get its "compactitem" environment and avoid the whole mess.

I also had to think about where to actually define ILEN, and eventually decided that it makes the most sense in "Extending RISC-V" since it is influenced by support for extensions.

### DIFF
--- a/src/extensions.tex
+++ b/src/extensions.tex
@@ -52,6 +52,15 @@ example, the base ISA is defined within a 30-bit encoding space (bits
 31--2 of the 32-bit instruction), while the atomic extension ``A''
 fits within a 25-bit encoding space (bits 31--7).
 
+We use the term ILEN to refer to the maximum instruction length supported
+by an implementation, which is always a multiple of 16 bits.  For
+implementations supporting only the base instruction set, ILEN is 32 bits.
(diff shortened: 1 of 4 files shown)


aswaterman:
Thanks, Jacob; this is more clearly and succinctly expressed.  My only concern is that defining ILEN towards the end of the user spec, then relying on that definition in the privileged spec, may prove confusing.  I've moved the paragraph to the Introduction chapter, where the instruction length encoding is described, and I think it still flows naturally.  Does it look OK to you?

jcb62281:
That is fine.  I had it in "Extending RISC-V" because XLEN and FLEN are defined farther down the user spec than the Introduction and I did not want to presume adding to the very short introduction.

I would suggest finding some way to work XLEN and FLEN into the introduction as well, but that is a minor editorial issue for you to think about.

PR #62, opened 2017-05-16
Forbid S-mode execution from user memory

jcb62281:
This is an issue that I have repeatedly raised on isa-dev.  In a nutshell, the only use in a POSIX environment for S-mode to fetch instructions from user memory is exploiting the supervisor for privilege escalation.  For the rare environments that do have legitimate uses for this, aliasing remains unrestricted (such restrictions are infeasible for hardware to enforce anyway) and physical pages may appear as both user and supervisor pages at different virtual addresses.

This is an issue bad enough that x86 had to retrofit a solution; we have the chance to get this right the first time.

Since I have finally made a GitHub account, I may as well submit this. :)

### DIFF
--- a/src/machine.tex
+++ b/src/machine.tex
@@ -634,7 +634,7 @@ machine mode.
 \end{commentary}
 
 The SUM (permit Supervisor User Memory access) bit modifies the privilege with
-which S-mode loads, stores, and instruction fetches access virtual memory.
+which S-mode loads and stores access virtual memory.
 When SUM=0, S-mode memory accesses to pages that are accessible by U-mode (U=1
 in Figure~\ref{sv32pte}) will fault.  When SUM=1, these accesses are
 permitted.  SUM has no effect when page-based virtual memory is not in effect.
--- a/src/supervisor.tex
+++ b/src/supervisor.tex
@@ -207,17 +207,38 @@ either readable or executable (R=1 or X=1) will succeed.  MXR has no effect
 when page-based virtual memory is not in effect.
 
 The SUM (permit Supervisor User Memory access) bit modifies the privilege with
-which S-mode loads, stores, and instruction fetches access virtual memory.
+which S-mode loads and stores access virtual memory.
 When SUM=0, S-mode memory accesses to pages that are accessible by U-mode (U=1
-in Figure~\ref{sv32pte}) will fault.  When SUM=1, these accesses are
-permitted.  SUM has no effect when page-based virtual memory is not in effect,
-nor when executing in U-mode.
+in Figure~\ref{sv32pte}) will fault.  When SUM=1, these accesses are permitted.
+SUM has no effect when page-based virtual memory is not in effect, nor when
+executing in U-mode.  Note that S-mode can never execute instructions from user
+memory, regardless of the state of SUM.
 
 \begin{commentary}
 The SUM mechanism prevents supervisor software from inadvertently accessing
 user memory.  Operating systems can execute the majority of code with SUM clear;
 the few code segments that should access user memory can temporarily set
 SUM.
+\end{commentary}
+\begin{commentary}
+The SUM mechanism permits S-mode software to access user data, but never to
+execute instructions from user memory.  Legitimate uses cases for execution from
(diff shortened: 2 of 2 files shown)


aswaterman:
Sorry for taking so long to respond, Jacob.  I assumed this issue would be a bit more contentious, but it took a while to reach the top of my inbox again.

I looked over the old threads again, and I continue to think that the arguments in favor don't outweigh the two cons (sharing library code at same VAs between supervisor/user, which has been discussed at length; and the much more important endeavor to make no backwards incompatible changes to v1.10).  The most important concern was:

> A ROP chain that sets SUM and then returns to userspace, or even that modifies another task's saved sstatus to set SUM when that task is resumed (and modifies the saved state to "in S-mode, resume at PC value X", where X is the address of an S-mode exploit payload in a user buffer in that task) is quite plausible.

Both are plausible.  But the latter is subsumed by greater concerns within the same memory word (namely, setting the saved sstatus.SPP field to S).   The former (the SUM code itself), can be demonstrated correct statically, and if not, should be rewritten accordingly.  That leaves the trap handler, which may be invoked when SUM=1, but that needs to be carefully audited, too.

jcb62281:
Since I had raised this issue long before v1.10, and again (twice!) when the plan for v1.10 was announced, specifically in an effort to get it into v1.10, only to see it lost in the noise, I consider reluctance to break compatibility with v1.10 on this matter a red herring.

You seem to have misinterpreted my concerns about ROP chains.  The latter issue is *not* subsumed by a greater concern about sstatus.SPP.  Rather, it *includes* setting the saved SPP to S, since that will cause a preemptible supervisor to resume the task in S-mode.  (Supervisors that do not support preemption while in kernel code can protect themselves by ensuring that context switches always begin and end in U-mode.  Modern Linux supports kernel preemption and cannot do this.)  The benefit from this proposed change is that it would remove an otherwise easy option for an attacker:  running an exploit payload from a user buffer.  Currently, for this type of privilege escalation scenario, an attacker can simply spawn two cooperating tasks:  task A maps a buffer at a chosen virtual address, loads an exploit payload into the buffer, and goes to sleep waiting for task B to exploit the assumed kernel saved context corruption bug to (1) overwrite the saved sstatus for task A with SUM set and SPP==S and (2) overwrite the saved PC value for task A with the known address of the buffer, neatly defeating all KASLR protections, then continues task A, which proceeds to execute the payload in S-mode.  Depending on the extent of the saved context corruption bug, the chosen address may be important, if the attacker can know but does not fully control the overwritten PC value.  With this proposed change, the attacker needs information leaks in order to locate a kernel buffer containing the attack payload and must ensure that that kernel buffer remains valid long enough to use.

The use case you cite (sharing library code at a common virtual address between user and supervisor) is basically an information leak all by itself.  I cannot find an exact citation at the moment, but I remember reading that Mac OS X uses (or once used) a similar model of loading shared libraries into a global mapping region shared across all processes.  This led to privilege escalation exploits against root-owned processes, although the OS X kernel itself does not also share this library segment.  For POSIX systems, I believe that sharing library code at common virtual addresses between user and supervisor modes is a privilege escalation exploit waiting to happen and *should* be prevented.  I also understand that the initial RISC-V privileged ISA is intended primarily for POSIX supervisors, thus my confusion at the reluctance to adopt this change after discussions on isa-dev seemingly produced consensus for it, with the only opposition I had seen withdrawn.  (I now am sure that the issue had been raised and discussed previously before I joined isa-dev, so I had never seen your concerns and could not address them until now.)

The Linux support for x86 SMEP was added in https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=de5397ad5b9ad22e2401c4dacdf1bb3b19c05679 and essentially consists of turning SMEP on and forgetting about it.  I propose that the RISC-V equivalent be "always-on" and not require this "opt-in" step.

aswaterman:
Thanks for setting me straight on the ROP gadget.

I'm going to re-propose this to the isa-dev list as a feature for inclusion in priv-1.11.

aswaterman:
Merged!

PR #76, opened 2017-05-28
Add nag package

edqq:
Old habits die hard. All the same, there are commands, classes and packages which are outdated and superseded. The nag package provides routines to warn the user about the use of such ob­so­lete things.

The `l2tabu` packages warns you about obsolete commands and packages, and demonstrates the most severe mistakes most LaTeX users are prone to make. The orthodox option checks the LaTeX code for syntax that is technically correct, but most likely produces an unwanted result.

The `nag` package is a valuable resource to improving the quality of LaTeX documents.

### DIFF
--- a/src/riscv-spec.tex
+++ b/src/riscv-spec.tex
@@ -2,6 +2,7 @@
 % riscv-spec.tex
 %-----------------------------------------------------------------------
 
+\RequirePackage[l2tabu,orthodox]{nag}
 \documentclass[twoside,11pt]{book}
 
 \input{preamble}


aswaterman:
If this only produces warnings, it's useless from my perspective.  Building the spec generates 5500 lines of output, which is far too much to be useful: if the build succeeds, I'll never examine them.  ```nag``` adds another 9900 lines I'm also going to ignore.

If it has an option to issue errors instead of warnings (and you are willing to fix the errors :-)) then we would consider this PR.

edqq:
I don't know if you can make `nag` output errors instead of warnings, but you can always do something simple like `make > make.log && grep nag make.log` [0]

`nag` currently outputs 812 complaints in total for the manual.
Almost all of those are due to the use of deprecated commands like `\bf` and `\tt`. [1]

[0] http://sprunge.us/KSYH
[1] http://www.tex.ac.uk/FAQ-2letterfontcmd.html

PR #77, opened 2017-05-28
Add microtype package

edqq:
microtype – Subliminal refinements towards typographical perfection

This package effortlessly improves the general appearance of the text using different techniques, such as margin kerning (protrusion), extra kerning, expansion, tracking, and spacing.

Disable protrusion on the table of contents as per the documentation: http://www.ctan.org/tex-archive/macros/latex/contrib/microtype/microtype.pdf


aswaterman:
Can you point me to particular sections or tables in the spec that microtype improves, so I can get a visual idea for what this does?

PR #80, opened 2017-06-04
Fix typo

jcb62281:
Found this typo while writing a proposal for isa-dev.  A quick search of the PDF did not find any other instances of "IEP".


aswaterman:
thanks

PR #91, opened 2017-07-02
 	Clarify Float Compare description 

DSHorner:
Modelled after SLT description:
Make register compare order explicit (rs1 to rs2)
Make Boolean result explicit (1 written if true, zero otherwise).

I considered mentioning that +0 and -0 compare equal, but decided IEEE 754-2008 compliance covers it.

I have no idea why the ECALL/EBREAK is included in the pull request when it exists in master already.


aswaterman:
thanks, I've manually merged this.

PR #125, opened 2018-01-08
Fixes for two-level translation

bonzini:
Hi,

this is at last the result of my review of two-level translation in the hypervisor draft.

The first commit is, I believe, an actual bug in the specification: SPRV=1/SPV=1 is using the "S-mode meanings" of `satp` and `bsatp`, but those are reversed when running in HS-mode. Therefore, we must reverse them when HS-mode is doing accesses "as if" in S-mode.

The second commit clarifies the meaning of HS-level and S-level ASIDs. It also defines a combined HS-level + S-level ASID, and allows the TLB to store combined translations (from virtual address to machine-physical address) under the combined ASID.

The third commit introduces a new base register for paging, `hatp`, which makes the SPRV=1/SPV=1 mode much easier to use.

### DIFF
--- a/src/hypervisor.tex
+++ b/src/hypervisor.tex
@@ -206,11 +206,10 @@ Table~\ref{h-sprv} enumerates the cases.
 \label{h-sprv}
 \end{table*}
 
-When SPV=1, however, two-level translation swaps the usual meaning of
-{\tt satp} and {\tt bsatp}. {\tt bsatp} provides the virtual address to
-guest-physical address mapping, as well as the S-level ASID; {\tt satp}
-provides the guest-physical to machine-physical mapping and the HS-level
-ASID.
+When SPV=1, two-level translation uses the {\tt bsatp} register (instead
+of {\tt satp} to retrieve the virtual to guest-physical mapping, as
+well as the S-level ASID; as usual, {\tt hatp} provides the guest-physical
+to machine-physical mapping and the HS-level ASID.
 
 \begin{commentary}
 For simplicity, SPRV is in effect even when in S-mode or U-mode, but in normal
@@ -269,6 +268,71 @@ to U-mode by setting the corresponding bit in {\tt sedeleg} or {\tt sideleg}.
 When V=0 and the N extension for user-mode interrupts is implemented, any trap
 that has been delegated to HS-mode can be further delegated to U-mode by
 setting the corresponding bit in {\tt sedeleg} or {\tt sideleg}.
+
+\subsection{Hypervisor Address Translation and Protection Register ({\tt hatp})}
+
+The {\tt hatp} register is an XLEN-bit read/write register formatted as shown
+in Figure~\ref{rv32hatpreg} for RV32 and Figure~\ref{rv64hatpreg}  for RV64.
+The interpretation of the MODE, ASID, and PPN fields is the same as for
+{\tt satp}.
+
+\begin{figure}[h!]
+{\footnotesize
+\begin{center}
+\begin{tabular}{c@{}E@{}K}
+\instbit{31} &
+\instbitrange{30}{22} &
+\instbitrange{21}{0} \\
+\hline
+\multicolumn{1}{|c|}{{\tt MODE} (\warl)} &
+\multicolumn{1}{|c|}{{\tt ASID} (\warl)} &
+\multicolumn{1}{|c|}{{\tt PPN}  (\warl)} \\
+\hline
+1 & 9 & 22 \\
+\end{tabular}
+\end{center}
+}
+\vspace{-0.1in}
+\caption{RV32 hypervisor address translation and protection register {\tt hatp}.}
+\label{rv32hatpreg}
+\end{figure}
+
+\begin{figure*}[h!]
+{\footnotesize
+\begin{center}
+\begin{tabular}{@{}S@{}T@{}U}
+\instbitrange{63}{60} &
+\instbitrange{59}{44} &
+\instbitrange{43}{0} \\
+\hline
+\multicolumn{1}{|c|}{{\tt MODE} (\warl)} &
+\multicolumn{1}{|c|}{{\tt ASID} (\warl)} &
+\multicolumn{1}{|c|}{{\tt PPN}  (\warl)} \\
+\hline
+4 & 16 & 44 \\
+\end{tabular}
+\end{center}
+}
+\vspace{-0.1in}
+\caption{RV64 hypervisor address translation and protection register {\tt hatp}, for MODE
+values Bare, Sv39, and Sv48.}
+\label{rv64hatpreg}
+\end{figure*}
+
+The {\tt hatp} register is only active when second-level translation
+is in effect, namely when either V=1 or SPRV=1 and SPV=1.  The first
+level of translation comes from {\tt satp} when V=1, or from {\tt
+bsatp} when SPRV=1 and SPV=1.  Any physical address that is accessed
+by the hart to retrieve a page table entry, or that is specified in
+the leaf page table entry, is treated as a {\em guest physical address}
+and goes through a second stage of translation.  The second stage of
+translation uses paging structures described by {\tt hatp}.
+
+If MODE is 0 in {\tt hatp}, second-level translation uses an identity
+mapping from guest-physical to machine-physical addresses.  However,
+it is {\em not} entirely disabled.  In particular, when SPRV=1 and SPV=1,
+conversion of virtual addresses to guest physical addresses will use
+{\tt bsatp} rather than {\tt satp}.
 
 \subsection{Background Supervisor Status ({\tt bsstatus}) Register}
 
@@ -610,14 +674,10 @@ When V=0, the {\tt bsatp} register holds the S-mode guest's version of the
 V=1, or vice-versa), the implementation swaps the contents of {\tt bsatp} and
 {\tt satp}.
 
-{\tt bsatp} affects the behavior of the machine when two-level address
-translation is active.  When V=1, it controls the translation of guest
-physical addresses to machine physical addresses.  When V=0, but SPRV=1
-and SPV=1, it controls the translation of guest virtual addresses to
-guest physical addresses.
-
-The interpretation of the MODE, ASID, and PPN
-fields is the same as for {\tt satp}.
+{\tt bsatp} only affects the behavior of the machine when SPRV=1 and SPV=1.
+In that case, second-level translation uses {\tt bsatp} to translate
+guest virtual addresses to guest physical addresses, and {\tt hatp} to
+translate guest physical addresses to machine-physical addresses
 
 \begin{figure}[h!]
 {\footnotesize
@@ -845,19 +905,17 @@ and protection is in effect.  For any virtual memory access, the original
 virtual address is first converted
 by S-level address translation, as controlled by the {\tt satp} register, into
 a {\em guest physical address}.  The guest physical address is then
-converted by HS-level address translation, as controlled by the {\tt bsatp}
+converted by HS-level address translation, as controlled by the {\tt hatp}
 register, into a {\em machine physical address}.
 Although there is no option to disable two-level address translation when V=1,
 either level of translation can be effectively disabled by zeroing the
-corresponding {\tt satp} or {\tt bsatp} register.
+corresponding {\tt satp} or {\tt hatp} register.
 
 Two-level address translation and protection is also in effect if V=0 but
-SPRV=1 and SPV=1.  In that case, the process is the same but the roles
-of {\tt satp} and {\tt bsatp} are inverted.  The original virtual address
-is first converted by S-level address translation (now controlled by the
-{\tt bsatp} register) into a guest physical address.  The guest physical
-address is then converted by HS-level address translation (now controlled
-by the {\tt satp} register, into a {\em machine physical address}.
+SPRV=1 and SPV=1.  In that case, the process is the same except that
+S-level address translation is controlled by the {\tt bsatp} register.
+The guest physical address is then converted by HS-level address translation
+as usual, under control of of the {\tt hatp}.
 
 For the purposes of HS-level address translation and protection, all memory
 accesses made with V=1---including those made by the S-level
@@ -912,7 +970,8 @@ addresses and is in effect regardless of virtualization mode.
 
 When two-level address translation is in effect, there are separate
 HS-level and S-level ASIDs.  HS-level ASIDs share the ASID space with
-regular one-level page tables.
+regular one-level page tables.  The ASID field of {\tt hatp} provides the
+{\em active HS-level ASID}.
 
 \begin{commentary}
 	The processor may cache guest-physical to machine-physical
@@ -923,8 +982,7 @@ regular one-level page tables.
 	need to introduce an HFENCE.VMA instruction.  However, unless
 	there is a 1:1 correspondence between guest physical addresses
 	and host virtual addresses, the hypervisor must use different
-	ASIDs for virtualized guests and U-mode programs, even if
-	MODE is zero for the virtualized guest.
+	ASIDs in {\tt hatp} and {\tt satp}, even if MODE is zero.
 \end{commentary}
 
 While second-level translation is in effect, the ASID in {\tt satp}
@@ -933,7 +991,8 @@ or {\tt bsatp} (respectively when V=1, or when SPRV=1 and SPV=1) is a
 addresses to guest physical addresses.  S-level ASIDs are combined with
 the active HS-level ASID to form a {\em combined ASID}; combined ASIDs
 identify translations from guest virtual addresses to machine physical
-addresses.
+addresses.  As part of a combined ASID, the active HS-level ASID may be used
+to cache S-level translations even if MODE is zero in {\tt hatp}.
 
 \subsection{Memory-Management Fences}
 
--- a/src/hypervisor.tex
+++ b/src/hypervisor.tex
@@ -908,20 +908,79 @@ entry must not have its U bit set, unless overridden by {\tt sstatus}.SUM.
 Machine-level physical memory protection applies to machine physical
 addresses and is in effect regardless of virtualization mode.
 
+\subsection{Address-space IDs}
+
+When two-level address translation is in effect, there are separate
+HS-level and S-level ASIDs.  HS-level ASIDs share the ASID space with
+regular one-level page tables.
+
+\begin{commentary}
+	The processor may cache guest-physical to machine-physical
+	translations in the TLB under the active HS-level ASID.
+
+	Sharing the ASID space between one-level translations and
+	guest-physical to machine-physical translations removes the
+	need to introduce an HFENCE.VMA instruction.  However, unless
+	there is a 1:1 correspondence between guest physical addresses
+	and host virtual addresses, the hypervisor must use different
+	ASIDs for virtualized guests and U-mode programs, even if
+	MODE is zero for the virtualized guest.
+\end{commentary}
+
+While second-level translation is in effect, the ASID in {\tt satp}
+or {\tt bsatp} (respectively when V=1, or when SPRV=1 and SPV=1) is a
+{\em S-level ASID}.  S-level ASIDs identify translations from guest virtual
+addresses to guest physical addresses.  S-level ASIDs are combined with
+the active HS-level ASID to form a {\em combined ASID}; combined ASIDs
+identify translations from guest virtual addresses to machine physical
+addresses.
+
 \subsection{Memory-Management Fences}
 
 The behavior of the SFENCE.VMA instruction is affected by the current
 virtualization mode V.  When V=0, the virtual-address argument is an HS-level
 virtual address, and the ASID argument is an HS-level ASID.  If either argument
 is provided, the instruction orders stores only to HS-level address-translation
-structures with subsequent address translations.  If neither argument is
-provided, the instruction orders stores to all HS-level and S-level address-translation structures
-with subsequent address translations.
+structures with subsequent address translations; in addition, if the HS-level
+ASID is provided, SFENCE.VMA will only order reads and writes for the
+combined ASIDs formed by the provided HS-level ASID and any S-level ASID.
+If neither argument is provided, the instruction orders stores to all
+HS-level and S-level address-translation structures with subsequent
+address translations.
 
 When V=1, the virtual-address argument to SFENCE.VMA is a guest virtual
 address, and the ASID argument is an S-level ASID.  The instruction
 orders stores only to the S-level address-translation structures within the
-HS-level address-space specified by the {\tt bsatp} register.
+HS-level address-space specified by the {\tt bsatp} register.  More precisely,
+SFENCE.VMA will only order reads and writes for:
+\begin{itemize}
+	\item if {\em rs2}={\tt x0}, the combined ASIDs formed by
+		the active HS-level ASID and any S-level ASID;
+	\item if {\em rs2}$\neq${\tt x0}, the combined ASID formed by
+		the active HS-level ASID and the S-level ASID
+		specified in {\em rs2}.
+\end{itemize}
+
+\begin{commentary}
+	Simplified implementations are possible where the processor doesn't
+	store combined ASIDs in the TLB.  For example:
+	\begin{itemize}
+		\item The processor may cache guest-virtual to
+			machine-physical translations in the TLB under
+			the active HS-level ASID.  In this case,
+			every write to {\tt satp} while V=1 acts
+			as a fence for the active HS-level ASID, and
+			SFENCE.VMA while V=1 must behave as if {\em
+			rs2} specified the active HS-level ASID.
+		\item The processor may cache guest-virtual to
+			machine-physical translations in the TLB under
+			the active S-level ASID; in this case, the
+			implementation must flush the TLB completely
+			every time V changes.
+	\end{itemize}
+	In either case, the processor must not use cached translations
+	while SPRV=1 and SPV=1.
+\end{commentary}
 
 When V=0, attempts to execute SFENCE.VMA in U-mode or when {\tt mstatus}.TSR=1
 raise an illegal instruction exception.  When V=1, attempts to execute
--- a/src/hypervisor.tex
+++ b/src/hypervisor.tex
@@ -206,6 +206,12 @@ Table~\ref{h-sprv} enumerates the cases.
 \label{h-sprv}
 \end{table*}
 
+When SPV=1, however, two-level translation swaps the usual meaning of
+{\tt satp} and {\tt bsatp}. {\tt bsatp} provides the virtual address to
+guest-physical address mapping, as well as the S-level ASID; {\tt satp}
+provides the guest-physical to machine-physical mapping and the HS-level
+ASID.
+
 \begin{commentary}
 For simplicity, SPRV is in effect even when in S-mode or U-mode, but in normal
 use will only be enabled for short sequences in HS-mode.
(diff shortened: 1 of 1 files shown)


jhauser-us:
To justify adding `hatp`, you wrote:
> In general the HVA->MPA and GPA->MPA mappings will be different, ...

From our perspective, this needs more motivation.  Why should we assume the HVA->MPA and GPA->MPA mappings generally need to be different?

Due to the existence of M mode and possible physical memory protection, a RISC-V operating system shouldn't normally assume it has full access to the entirity of a machine's physical address space.  Rather, we require that a RISC-V OS cope with some amount of the PA space being inaccessible to it because the space is dedicated to M-mode functions.  When a RISC-V OS is run as a guest, we know that a comparable portion of the GPA space may be made off-limits to the guest and dedicated for use by the hypervisor.  For a 64-bit RISC-V, it seems to us that the hypervisor's portion of a combined HVA/GPA space need not be small.

This arrangement does force a hypervisor to set aside some fixed regions of its virtual address space to be the GPA space of a guest virtual machine, after which, whatever virtual address space the hypervisor wants for its own use will have to be allocated outside/around those fixed GPA regions.  There may be times when a hypervisor simply _must_ remap some of the dedicated GPA space for a different purpose.  However, assuming such forced remappings are sufficiently infrequent, the hypervisor already has the ability to accomplish this by swapping `satp`.

Can you explain why the solutions outllined above are inadequate?

bonzini:
There are several reasons for that. In order of importance, I can think of the following:

1) while the "preferred" solution for virtualizing RISC-V involves implementing the SBI in the hypervisor, that is not required. Hypervisor extensions can be used to emulate M-mode and virtualize a real machine, and constraining GPA->MPA mappings complicates this.

2) Hypervisor extensions are defined for 32-bit RISC-V as well. Assume an architecture like KVM, where a user-mode program has all guest memory mapped into its own address space. If GPA->MPA and HVA->MPA mappings are the same, migrating an active guest from a machine to another would only work if the address space is (almost) the same between the source and destination, and the chance of an address space collision would be high on a 32-bit host with ASLR. Guests with 1 GiB memory or even more are possible on a 32-bit host with 3GiB/1GiB address space separation between U-mode and S-mode.

3) Even if you can reserve part of the address space for the hypervisor, the assumption that GPA->MPA mapping is equal to the HVA->MPA mapping is wrong. In general, the GPA->MPA mapping is only a _subset_ of the HVA->MPA mapping; there may be other HU-mode data structures in the HVA->MPA mapping that must not be available to code running in VS-mode. If you use the same mapping, the guest has the possibility to peek at (or even write to) host HU-mode data structures just by accessing the "wrong" GPA.

bonzini:
To be clear, none of the above is a blocker. The only page that has to be the same between HVA and GPA is the one where hypervisor executes HRET; as long as that one page matches, the hypervisor can switch `satp` just before executing HRET and restore it afterwards. However, this would be slower and it would prevent usage of SPRV=1/SPV=1 to speed up hypervisor accesses to guest memory.

jhauser-us:
Thanks for the information, Paolo.  I'm afraid I have more questions about the reasons you've listed, but I'd like to consider first number 2, which I feel I understand the best of the three:
> 2. Hypervisor extensions are defined for 32-bit RISC-V as well.  [...]

Of course you are correct that the proposal doesn't exclude 32-bit RISC-V cores from supporting the H extension, but I wonder whether there is really a market for such a configuration.  Or will all _actual_ uses of the H extension have 64-bit HS mode?  I'm asking because I really don't know.  If we believe that 32-bit RISC-V hypervisors will be a real thing, then I'm prepared to accept that the 32-bit address space may be constrained enough to justify `hatp`.

bonzini:
That depends on hardware implementors, as I have no idea what the expected distribution is going to be for RV32 versus RV64...

Even for a simple 32-bit RTOS, however, it may make sense to use HS-mode to run a single instance of e.g. Linux, whenever no real-time process is running. Depending on the design however there might not be an HU-mode while Linux runs, and certainly no live migration, which mitigates my second concern.

Even if this were enough to convince you :-) I would like to hear your questions on points 1 and 3. 3 is the most concerning one to me and I would like to be sure that we are on the same page.

jcb62281:
The main goal that I want here is for a guest virtual machine to resemble a user process as much as possible.  The current draft of RVH specifies that guest physical to host physical translations are performed with user permissions in the host page tables.  As such, the guest physical address is translated the same way that a host user virtual address is translated.  It is translated like a user virtual address, so I maintain that we should call it exactly that:  a (host) user virtual address.  Switching into or out of a guest should be as similar to switching between user tasks in the host as possible.

I am currently working on a proposal for disjoint user/supervisor address spaces.  With disjoint addressing, the guest second level translation would use the same paging root CSR as any other user process in the host supervisor.

I understand that current architectures probably needlessly complicate this area, but RISC-V has a chance to do it right and keep virtualization simple.  For this reason, I oppose adding a new hatp CSR.

I also suggest rethinking the combined ASID concept to be more clear that this effectively partitions a single host-level ASID into some number of guest ASIDs.  I also note that I have been told several times on isa-dev that the RISC-V spec does not actually define a TLB, so either I am misinformed or the combined ASID design needs to be reworded.  Are combined ASIDs even needed in the spec, beyond a brief commentary that hardware is permitted to use them internally as an optimization?  How is a combined ASID visible to program code?

Generally, I see potential for lots of problems with SPRV and would like to see a less dangerous alternative.  If we can agree on RVH requiring disjoint addressing support, then the host suatp CSR can be the paging root for guest-physical -> host-physical translations.  I expect that guests should be mostly opaque and a hypercall interface could easily use guest-physical addresses.  Using physical addresses would also be helpful for a future SBI that monitors are expected to implement.  I suggest adding an HPROBE.VM instruction for translating guest-virtual addresses to guest-physical addresses (using the TLB if possible) if some fast means for hosts to use guest virtual addresses is desired.  (Since page tables store physical page numbers and guest physical addresses are translated equivalently to host user virtual addresses, walking the guest page tables in host software would be straightforward.)

aswaterman:
Re: point 1, efficient virtualization of M-mode was not a design goal.  It probably won't perform that well because it will incur many emulation traps, so not being able to use the SPRV mechanism isn't 

I'm not certain I'm understanding point 3 correctly, but it sounds similar to the problem that exists without virtualization, when the supervisor needs to access memory on behalf of the user.  Linux handles this by constraining user virtual addresses to the nonnegative addresses, so it can perform a simple bounds check to make sure the address doesn't belong to the kernel.  I don't see much of a downside of constraining the GPA space similarly, so performing a bounds check before going into SPRV mode should suffice to prevent access to HS-level or HU-level data structures.

To be clear, I'm not currently arguing against this new mechanism--I'm still trying to wrap my head around your motivations for proposing it.

sorear:
(3) isn't about SPRV sequences, it's about ensuring that a guest cannot access qemu-kvm's _user mode_ memory in its GPA space.

Tangent: Months ago when we were doing preparatory discussions for basic virtualization (the mechanism enabled by sstatus.TVM and sstatus.TSR), we observed that for partitioning use cases the GPA→HPA map could be a subset of an identity mapping, in which case no two-level mapping is needed and it was sufficient to swap PMPs on a world switch. Unclear if KVM would use such a feature or if it makes sense to include anything similar in extended virtualization.

bonzini:
First, a quick reply to Jacob's remark on combined ASIDs:

> the RISC-V spec does not actually define a TLB, so either I am misinformed or the combined ASID 
> design needs to be reworded

The TLB is mentioned only in commentary, similar to the supervisor spec.

> How is a combined ASID visible to program code?

It is not visible, but it is the simplest way I could think of to describe the interaction between HS-mode and S-mode executions of SFENCE.VMA. The current version of the spec uses the terms _HS-level ASID_ and _S-level ASID_ without ever defining the concepts.

> Re: point 1, efficient virtualization of M-mode was not a design goal. It probably won't
> perform that well because it will incur many emulation traps, so not being able to use
> the SPRV mechanism isn't

I disagree for two reasons. First, I don't see while it would perform badly. You certainly don't want to spend your time executing SBI, so it shouldn't matter much if the SBI is paravirtualized or uses actual M-mode. The hypervisor can easily avoid HS-mode traps for exceptions or interrupts that (according to `medeleg`/`mideleg`) are to be delivered in S-mode. M-mode features such as the PMP can also be handled pretty efficiently, just by merging their effect in the second-level page tables; they should be set up just once and should never fire during S-mode operation, so again they shouldn't affect performance.

Second, I think there is worth in allowing the same hypervisor to run both full-system emulation and paravirtual SBI. If paravirtual SBI lets you use SPRV=1/SPV=1, for full-system emulation you would require a second path that manually walks page tables to translate GPAs. "It wouldn't perform that well" (or at least it would perform worse) becomes a self-fulfilling prophecy.

Jacob's point that "a guest is the same as a U-mode process" is not true in general. With KVM, for example, a guest is _executed on behalf_ of a U-mode process. Hence, as explained by Stefan, there is plenty of other memory mapped in the HVA space that shouldn't be in the GPA space. It is possible to work around the issue:

* place HRET at a trampoline page, as explained [here](https://github.com/riscv/riscv-isa-manual/pull/125#issuecomment-356242735), and swap `satp` around HRET

* do a range-check around SPRV=1/SPV=1 accesses, as suggested [here](https://github.com/riscv/riscv-isa-manual/pull/125#issuecomment-356504973) by Andrew, or don't use them at all.

However, that incurs a performance cost; `hatp` is a pretty standard solution and I don't see it as "other architectures needlessly complicating things". I'm all for a simple design (that's why I'm not adding an HFENCE.VMA instruction, for example), but a wise man said things should not be simpler than necessary. :)

sorear:
Two small points of possibly no consequence:

1. The architecture as currently defined provides no S-mode timer and no S-mode IPI facility, so timer interrupts and IPIs are always delivered through SBI code.

2. There is a potential practical problem with requiring a real SBI implementation in qemu, which is that the implementation has to be supplied somehow.  Either the end user needs to compile bbl themselves, which adds steps, or qemu provides one, which adds a dependency on a risc-v cross-toolchain (a problem qemu already has on several architectures, but it'd be nice to avoid).

bonzini:
The SBI can be very simple so it can be filled out by qemu directly as binary code. I wouldn't be worried about that.

In fact, for both bare metal (M-mode) and virtualized (HS-mode) implementations the SBI page can be simply filled with "ECALL" instructions; the low 11 bits of the program counter can be used in the ECALL handler to determine which function is being invoked.

sorear:
The SBI page is no longer a thing due to difficulties specifying its execution environment.  ECALL instructions now appear directly in the guest kernel, and "SBI" is used to refer to the ECALL protocol.

You're right that it can be implemented in a hypervisor or emulator by trapping S-mode ECALL instructions but I thought you were proposing to emulate a complete machine including M-mode layers?

bonzini:
> The SBI page is no longer a thing due to difficulties specifying its execution environment. 

Then it's even simpler, the ECALL handler would be in the hypervisor and (in the case of KVM) can ask QEMU for help on some of the hypercalls.

> I thought you were proposing to emulate a complete machine including M-mode layers?

Optionally. When ECALL is invoked from S-mode (with V=1), the hypervisor's behavior would depend on whether M-mode is being emulated: if not, the ECALL would be handled in the hypervisor as in the previous paragraph; if it is, the hypervisor would switch the process to virtual M-mode, set up virtual M-mode CSRs and divert guest execution to `mtvec`. Virtual M-mode would talk to the hypervisor via CSRs and emulated MMIO devices.

jhauser-us:
Paolo wrote:
> 3. Even if you can reserve part of the address space for the hypervisor, the assumption that GPA->MPA mapping is equal to the HVA->MPA mapping is wrong. In general, the GPA->MPA mapping is only a subset of the HVA->MPA mapping; there may be other HU-mode data structures in the HVA->MPA mapping that must not be available to code running in VS-mode. If you use the same mapping, the guest has the possibility to peek at (or even write to) host HU-mode data structures just by accessing the "wrong" GPA.

And later,
> With KVM, for example, a guest is executed on behalf of a U-mode process. Hence, as explained by Stefan, there is plenty of other memory mapped in the HVA space that shouldn't be in the GPA space.

Since I'm not really an expert in the ways that hypervisors get used, the appearance of non-virtual U mode in the picture came as news to me.  It's still not obvious to me why it should necessarily be true, but I've decided to defer to those with greater knowledge about such things, if you're sure such arrangements are valuable.  That gives us at least one very good reason for splitting the HVA->MPA and GPA->MPA mappings.

With your forebearance, and for the benefit of myself and any others who may be following along, I will try to summarize some other justifications as I understand them, as follows:

For a given guest virtual machine, it's not unusual for the layout of the GPA space (the particular address ranges populated by memory or devices) to be dictated mostly or entirely by forces outside the control of the hypervisor, though the hypervisor is responsible for providing a virtual implementation of those resources.  In other words, in the GPA->MPA mapping, the allocated addresses on the GPA side may be fixed, not under the control of the hypervisor.  If the HVA->MPA and GPA->MPA mappings are combined, as originally proposed, a hypervisor must carve out usable HVA space from the addresses unused in the virtual machine's GPA layout.  Although this should be possible in theory, there are some complicating factors in practice:

1.  The allocated portion of a virtual machine's GPA space may consume most of the combined HVA + GPA addresses, leaving too little HVA space for the hypervisor's needs.

2.  A type-2 hypervisor lives within an operating system, which invariably comes with its own conventions for allocating virtual address space, and it's likely at least some of those conventions will conflict with the required allocation of the GPA space for a particular virtual machine.

Dealing with either points 1 or 2  can easily require swapping the address mapping (changing `satp` within HS mode), which has performance consequences.  Point number 2 might in theory be mitigated if the operating system designers plan far enough ahead, adopting conventions for address space allocation that will be compatible with a type-2 hypervisor.  However, that's probably expecting too much, especially in light of this:

3.  The required allocation of GPA space might not be the same for all virtual machines.  And the universe of all possible GPA layouts might not be knowable in advance.

Does that sound about right?

Unless somebody can successfully argue against it, I'm now on board with having a separate GPA->MPA page table.  But instead of `hatp`, may I suggest we call the new CSR `hgatp`, for "Hypervisor Guest Address Translation and Protection"?  Or something similar?

jcb62281:
Paolo Bonzini wrote:

  Jacob's point that "a guest is the same as a U-mode process" is
not true in general. With KVM, for example, a guest is executed on
behalf of a U-mode process. Hence, as explained by Stefan, there
is plenty of other memory mapped in the HVA space that shouldn't be in
the GPA space. It is possible to work around the issue:
  
    
      place HRET at a trampoline page, as explained here,
and swap satp around HRET
    
  


If you have an suatp CSR (used for guest-physical to host-physical and
user-virtual to supervisor-physical translations), there is no need for
this: simply load the guest paging root into suatp before executing
HRET.


  
    
      do a range-check around SPRV=1/SPV=1 accesses, as suggested here
by Andrew, or don't use them at all.
    
  


At this point, I would prefer that SPRV be removed in its current form
and replaced with something that has fewer sharp edges.


  However, that incurs a performance cost; hatp is a
pretty standard solution and I don't see it as "other architectures
needlessly complicating things". I'm all for a simple design (that's
why I'm not adding an HFENCE.VMA instruction, for example), but a wise
man said things should not be simpler than necessary. :)


Is an hatp CSR still needed if an suatp CSR (user-level paging root) is
provided?  To adapt the KVM example, a guest can be treated as a traced
process, with the U-mode KVM handler as the tracer.  Alternately, the
guest can be another "thread" (or group thereof with one host thread
per guest hart) with its own address space, but still notionally in
qemu-kvm's process.  If I understand the Linux process model correctly,
a guest is simply another task.  (Or should be simply another task on
RISC-V.)

sorear:
`suatp` is not going to happen, you can stop asking.

jcb62281:
Well, GitHub made a right mess of my previous comment sent in by email.

There are two major points I was trying to make:

1.  Disjoint user addressing allows to split the host/guest address spaces without adding an RVH-specific CSR.  Switching into or out of a guest always goes through the host supervisor, so reloading the user/guest paging root CSR is a non-issue on this path.

2.  I believe that SPRV has too many potential sharp edges and should be removed.  I note that guest-physical addresses are effectively host-virtual addresses and that RISC-V page tables use "physical" page numbers.  This makes walking guest page tables in host software easy.

If software guest PTE walks are not fast enough, I suggest an HPROBE.VM instruction that accesses the guest TLB.  HPROBE.VM would accept a guest virtual address in *rs1* and produce a guest physical address in *rd* or zero if a page fault was encountered.  Since physical page zero has been tacitly reserved for debugging support, excluding it from guests should be acceptable.  The host can still permit a guest to use physical page zero by performing a software walk in the rare case where zero from HPROBE.VM is ambiguous:  when the input address is page-aligned and "physical" page zero is available to the guest.

bonzini:
I agree with all the points made by John. Regarding the name, `hgatp` is okay if course.

Regarding `suatp`, I think disjoint address space support is in too early a stage. It's possible to unify the concept later, since the HS-mode is also relatively far from being ratified.

jhauser-us:
I wrote:
> Unless somebody can successfully argue against it, I'm now on board with having a separate GPA->MPA page table.

Assuming we do go this way, I don't see how we can avoid having a corresponding version of SFENCE.VMA for the new page table.  "HFENCE.GMA"?

bonzini:
Yes, I think I also convinced myself that it is bad to share ASIDs between hatp and HS-mode satp.

I will leave only the first patch in this pull request, and rework the rest. Thanks!

sorear:
Regarding `hgatp` there's one thing I'm still not entirely sold on.  Most 64-bit implementations are not going to provide a full 64-bit GPA space; combining `satp` and `hsatp` costs 1 bit (since the hypervisor will typically reserve the upper half for itself), and going from 37 to 36 bits (for instance) is qualitatively less impactful than going from 64 to 63, since either way parts of the architectually allowed GPA space are unusable.

bonzini:
The hypervisor need not reserve any part of the GPA space. I am not sure what you mean by `hsatp`.

GPA space is 39/48/57/64 bits depending on the paging mode that is chosen.

sorear:
`hsatp` was a typo, edited to `hgatp`.

Even with this proposal it is not possible to virtualize arbitrary RV32 memory maps, because the architectural physical address in RV32 is 34-bits (22-bit page frame numbers and a 10-bit offset) but no 34-bit VA paging mode is defined.

Regarding https://github.com/riscv/riscv-isa-manual/pull/125#issuecomment-356226006 , I understand and agree with the argument for KVM-like hypervisors to use different paging structures (potentially overlapping in the entire kernel area, not just a trampoline page) for HUVA→HPA and GPA→HPA, but I don't entirely follow the argument for having two separate registers for the roots.

Implementation advantages, since they will commonly feed separate TLBs?  Doubling GPA space from 31 to 32 or 36 to 37 bits more valuable than I realize?  Freedom to repurpose the UGX bits?

jhauser-us:
> Even with this proposal it is not possible to virtualize arbitrary RV32 memory maps, because the architectural physical address in RV32 is 34-bits (22-bit page frame numbers and a 10-bit offset) but no 34-bit VA paging mode is defined.

I was just thinking the same thing!

Somehow we missed this earlier.  To virtualize a 32-bit RISC-V, it seems the structure of the GPA->HPA page table tree will need to differ a little from a regular page table tree.  Here's one possibility:  While regular page tables are 4 kB at every level, perhaps the top-most table pointed to by `hgatp` can be made to be 16 kB.  I suspect this is the least disruptive tweak, though I'll be curious about other suggestions people may have.

For a certain kind of consistency, we could adopt the same rule for 64-bit addresses, that the top-most GPA->HPA page table is always 16 kB.   For a given number of levels in the page table hierarchy, that would always give us two extra bits of guest physical address relative to the supported virtual address width.

Thoughts?

aswaterman:
Sv32 supporting 34 bits of physical address is a hack to make RV32 applicable to the narrow class of systems with large address spaces and tiny processors.  We almost didn't even include this feature in Sv32.  We certainly shouldn't go out of our way to support it in the hypervisor extension.

I reject the possibility that the top RV64 page table be 16 KiB.  It's an unnecessary complication for RV64-only systems for only hypothetical benefit.

Mixed-XLEN systems already have to deal with complexity surrounding ```hgatp``` and friends, so if RV32's root page table is a different size than RV64, it is little incremental complexity.

sorear:
If the paging structure is going to diverge that much (yuck) then I withdraw my objection.

jhauser-us:
I'm surprised my suggestion for making the top-most GPA->MPA page table pointed to by `hgatp` be 16 kB drew such a strong negative reaction.  I suggested it because it has essentially zero cost for hardware, and nothing else about address translation is changed one iota.  The only software that would even be aware of the deviance would be a hypervisor, for which this would hardly count as a minor imposition compared to all its other worries.  I think you've rejected an elegant solution much too quickly, but I look forward to hearing your alternatives.

Unless somebody has a better idea, I think at a minimum we ought to do this for the 32-bit version of the GPA->MPA page tables.  Considering both hardware and software, the greatest cost I can see is that we might waste an extra 12 kB of memory for each active virtual machine.  The benefit is that there will be real RISC-V machines that can be virtualized that we wouldn't otherwise be able to.

Andrew wrote:
> I reject the possibility that the top RV64 page table be 16 KiB. It's an unnecessary complication for RV64-only systems for only hypothetical benefit.

Although I recognize it'll be a harder sell to get this also for 64-bit, the benefit is not just hypothetical.  For any RISC-V, the GPA->MPA page table determines the maximum physical address space that can be virtualized, at least without falling back on shadow page tables.  The major cost is the same as before:  potentially 12 kB of wasted memory for each active virtual machine, relative to the baseline assumption of a 4-kB top-most page table.  That's a cost I would gladly pay for the ability to support a 4-times-larger guest physical address space.

bonzini:
I am also not particularly fond of extending the top layer of page tables to 16 KiB.

In fact, what would prevent an RV32 implementation from supporting Sv39? For the hypervisor, but actually even for supervisor page tables.

jcb62281:
I disagree that limiting the guest physical address space is a problem.  Real hardware does not have arbitrary physical address maps and I would expect most RV32 systems to have only 32-bit physical addresses, since accessing memory beyond 4GiB is difficult for the monitor.

Limiting RV32-on-RV32H guests to 4GiB seems reasonable, especially since this entire issue does not apply to RV64H hosts.

For RV32-on-RV64H, arbitrary 34-bit RV32 physical address maps work fine, using only part of the host virtual address space, leaving the rest of the 39-bit or larger virtual space for the hypervisor.  Selecting RV32-on-RV64H is simple:  the host sstatus.UXL field sets the guest S-XLEN parameter, or, when V=1, S-XLEN is determined by bsstatus.UXL instead of mstatus.SXL.

Extending top-level page tables is only useful for RV32H.  On RV64, the physical address space is currently only 56-bit.  While it could expand to be analogous to RV32, I expect that RV128, with even larger physical address spaces, will be a reality before RV64 needs to worry about 66-bit physical addresses.  Like RV32, I expect that implementations using physical addresses larger than XLEN would be rare, due to the difficulties M-mode has with such addresses.

Lastly, RV32 cannot support Sv39 because the MODE field in RV32 satp is only one bit and selects between Bare and Sv32 modes.  RV64 takes advantage of a wider satp CSR to provide wider MODE and ASID fields at the cost that root page tables will always be required to be in the low 56-bit physical address space, even if PTE PPNs are expanded.

jhauser-us:
> In fact, what would prevent an RV32 implementation from supporting Sv39? For the hypervisor, but actually even for supervisor page tables.

Andrew can confirm, but I'm pretty sure the current expectation is that Sv39 will _never_ be standardized for 32-bit RISC-V.  Among other obstacles is the simple fact that the MODE field in a 32-bit `satp` is only 1 bit, and both possible values of 0 and 1 are already allocated, so there's no way even to encode the choice of Sv39 in the existing 32-bit `satp`.  We just heard Andrew say that he thought widening the Sv32 physical addresses to 34 bits was a stretch (so to speak), so I can't imagine he now wants to widen them to 39 bits or more.

Anyway, what's even more relevant to this discussion:  outfitting a 32-bit RISC-V to also support Sv39 is far, far more costly in hardware than just making the top-most page table of the GPA->MPA mapping be 16 kB.  It would make no sense to choose the costly solution because the cheaper solution is too expensive.

> I am also not particularly fond of extending the top layer of page tables to 16 KiB.

Okay, let's say you're not.  What's your alternative suggestion?  Abandon a large class of 32-bit RISC-V machines as simply not virtualizable on an H-extended 32-bit RISC-V?  That seems to be the choice everyone wants to make, but I think it's the wrong one.

jhauser-us:
Jacob wrote:
> I disagree that limiting the guest physical address space is a problem. Real hardware does not have arbitrary physical address maps and I would expect most RV32 systems to have only 32-bit physical addresses, since accessing memory beyond 4GiB is difficult for the monitor.

> Limiting RV32-on-RV32H guests to 4GiB seems reasonable, especially since this entire issue does not apply to RV64H hosts.

Well, maybe I'm just laboring under the wrong impression about the RISC-V Foundation's intentions concerning 34-bit physical addresses.  Even though the standard seems to endorse them, how do we really feel about machines that make use of physical addresses outside the 32-bit range?  Should I consider them orphans, not deserving of equal support or attention?

aswaterman:
I do agree that if we want to support 34-bit GPAs in RV32 guests, @jhauser-us's suggestion to make the GPA -> MPA root PT 16 KiB is the best way to do it.  I am warming to this idea for RV32.

Like @jcb62281, I'm not convinced that the extra two physical address bits for RV64 guests is important.  I'm not worried about 12 KiB of wasted memory, nor about hardware cost; I just consider it a weird feature.  It would also be the only aspect of the RV virtual-memory system that necessitates allocating adjacent physical-memory pages: having the page table size equal the page size is a feature.

jhauser-us:
I'm glad to see we're mostly in alignment on the technical details, so we can get down to the real business of arguing about assumptions and taste.  :^)

Andrew wrote:
> It would also be the only aspect of the RV virtual-memory system that necessitates allocating adjacent physical-memory pages

True....  Megapages and gigapages are optional, so the hardware doesn't _necessitate_ using them.

But you must know it's not totally rare for devices in the system to require the allocation of physical memory that's aligned and contiguous and spans multiple 4-kB pages.  It's common enough that operating systems normally support this among their internal memory allocation functions.  I know Linux does.  So is it really as strange as all that?

jcb62281:
> Abandon a large class of 32-bit RISC-V machines as simply not virtualizable on an H-extended 32-bit RISC-V?

I do not believe that this is a large class of RV32 machines that we are talking about.  I expect most RV32 systems to use 32-bit physical address spaces, since M-mode has difficulty accessing addresses beyond 2^32.  (Access to memory beyond 4GiB in M-mode on RV32 requires setting up page tables, setting MPP to S and using MPRV.)  I strongly doubt that anything other than (extra) main memory will ever appear in such address space.

34-bit physical addresses on RV32 are a hack much like Intel x86 PAE.  The difference is that PAE was introduced long before x86-64 existed, while RISC-V is starting out with RV64 available.  I doubt that there will be many RV32H machines in the first place and I suspect that RV32 machines that actually use the upper 3/4 of a 34-bit address space will be rarer still, with RV64 being in easy reach for applications that would need that.

Expanding the guest root page table will also break future unification of hypervisor guest address mapping and supervisor user mapping, requiring a CSR that is *only* used for the hypervisor instead of allowing the same (host supervisor) CSR to provide a user paging root and a hypervisor guest second-level paging root.  I believe that unifying hypervisor guest space and user space from the host supervisor's point of view will help to simplify supervisors and is worth more than full support for an edge case on RV32.  Remember that RV64H *is* able to virtualize all RV32 systems.  I expect that most RISC-V VM hosts will use RV64H or RV128H.

sorear:
I too agree that a 16 KiB root is the best way to handle a 34-bit GPA and a 32-bit HVA.  I agree that Sv39-on-RV32 is foolish, and my "yuck" was not intended to be a "strong negative reaction".

Writing off RV32H too soon would be a mistake; embedded hypervisors are becoming a thing, and I expect RV32 cores to outnumber RV64 for quite a while.

The GVA/HPA TLB and the GPA/HPA TLB are inevitably going to be different structures and having different ASID spaces for `hgatp` and `satp` seems to provide only benefits.

jcb62281:
sorear wrote:
>
> Writing off RV32H too soon would be a mistake; embedded hypervisors 
> are becoming a thing, and I expect RV32 cores to outnumber RV64 for 
> quite a while.
>

We are not writing off RV32H at all.  Embedded hypervisors are unlikely 
(in my estimation) to *have* more than 4GiB of RAM, which makes the 
issue irrelevant.  I expect that most RV32 implementations will not have 
even a 32-bit physical address space implemented, so RV32-on-RV32H 
guests being limited to 4GiB of address space will not be a problem.

Perhaps the spec should be clarified that smaller physical address 
spaces are allowed?

aswaterman:
@bonzini We've incorporated something very similar to this proposal in the most recent updates to the hypervisor chapter.

PR #126, opened 2018-01-10
Maintain symmetry between traps and MRET with respect to MPV

jcb62281:
If MPP is M, then V can only have been zero.  The outer frame must have
its own copy of the previous value of MPP somewhere, but that is no reason
to destroy MPV.

This change allows monitor trap handlers to recognize when the monitor was
initially entered from a guest virtual machine, even when handling traps
within the monitor.

### DIFF
--- a/src/hypervisor.tex
+++ b/src/hypervisor.tex
@@ -783,10 +783,11 @@ hypervisor extension is provided.
 \end{figure*}
 
 The MPV bit (Machine Previous Virtualization Mode) is written by the implementation
-whenever a trap is taken into M-mode.  Just as the MPP bit is set to the privilege
-mode at the time of the trap, the MPV bit is set to the value of the virtualization
-mode V at the time of the trap.  When an MRET instruction is executed, the
-virtualization mode V is set to MPV, unless MPP=3, in which case V remains 0.
+whenever a vertical trap is taken into M-mode from a less-privileged mode.  The MPV bit is
+set to the value of the virtualization mode V at the time the trap is taken, unless the
+trap is taken from M-mode.  MPV is unchanged when a horizontal trap is taken within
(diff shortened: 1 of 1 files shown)


aswaterman:
I'll let @jhauser-us weigh in, but I can't say that I see the utility in this change.  MPV was surely saved whenever MPP was saved, as there's no reason to save only part of mstatus.  Preserving MPV when traps are taken from M-mode adds a special case to the mstatus-update logic, which isn't costly in terms of gates, but is yet another thing to verify.

jcb62281:
The primary motivation is elegance.  The other side is to remove the special handling of MPV when MPP is M at MRET, since it is not possible for a horizontal monitor trap to set MPV if entering the monitor always clears V.

I would also be content with the "other side" of removing both special cases; the asymmetry between trap entry and exit just looks "off" to me.

jhauser-us:
The hypervisor extension's mstatus.MPV works the same way that the existing MPP bit does.  Even without the H extension, an M-mode trap handler that may be subject to nested traps must save and restore MPP at the handler's entry and exit.  This can be done by saving a copy of mstatus on entry and restoring mstatus on exit.  The H extension expands this requirement to two bits, MPP and MPV.  As it happens, a trap handler that saves and restores MPP by saving and restoring mstatus requires no change to also save and restore MPV.  (This is Andrew's observation as well, and I concur.)  I see no problem here that needs further solving.

jcb62281:
I have since realized that this cannot work and the asymmetry is an unfixable wart:  the monitor cannot be executed with V=1, which honoring MPV at MRET when MPP is M would attempt, and MPV *must* be written on horizontal monitor traps, to ensure that misaligned load/store emulation works correctly.

PR #128, opened 2018-01-23
Add disjoint addressing mode draft proposal

jcb62281:
This is a compromise between the proposals that Christoph Hellwig and I have previously advanced.

The design permitting hardware to always use disjoint addressing mode is the key part from my earlier proposal, while the use of SSU/SLU fields in the sstatus CSR instead of adding a new sats CSR is a significant contribution from Christoph Hellwig's proposal.

This compromise proposal also better balances the burdens between hardware and software for supervisors that support disjoint addressing mode as an optional feature than either of the original proposals.  Christoph Hellwig's earlier proposal put the entire complexity into hardware, defining disjoint addressing as a distinct hardware mode, while my earlier proposal radically changed all access to user memory when disjoint addressing is used.

This compromise allows supervisor user memory access routines to act as if the Sdas extension is always present, since the SSU/SLU fields must be hardwired to zero on systems that do not support disjoint addressing and writing them is harmless in that case.  The hardware can always use satp for supervisor translations and suatp for user translations, avoiding the introduction of a new mode switch in hardware, at the small cost of requiring dual-mode supervisors to choose which CSR to update during user task switch.  This is a small run-time patch (changing the CSR number in the supervisor text) to adapt to differing hardware support and should not be a significant concern.

I suggested this compromise in an off-list email to Christoph Hellwig over a week ago and have decided to proceed, since I have received no reply.

### DIFF
--- a/src/s-das.tex
+++ b/src/s-das.tex
@@ -1,0 +1,201 @@
+\chapter{``Sdas'' Standard Extension for Disjoint User/Supervisor
+  Address Spaces, Version 0.1}
+\label{s:das}
+
+This chapter describes the Sdas standard extension for disjoint user and
+supervisor address spaces.  Two operating modes are defined: {\em unified
+  addressing mode} is the traditional model, where the supervisor is mapped
+into the user address space and isolated only by page permissions, and {\em
+  disjoint addressing mode}, where the user and supervisor address spaces
+use independent page tables.
+
+All CSRs defined in this extension must be implemented if this extension is
+implemented.
+
+\section{Supervisor CSRs}
+
+This extension adds one CSR to the supervisor CSR set, modifies the
+interpretation of {\tt satp}, and adds two bits to {\tt sstatus}.
+
+\subsection{Supervisor Address Translation and Protection ({\tt satp}) Register}
+\label{s:das:satp}
+
+The {\tt satp} CSR controls address translation in S-mode and retains its
+format from the baseline supervisor support.
+
+For backwards compatibility, writes to {\tt satp} also write the same value
+to {\tt suatp}.  The CSR read-modify-write instructions may either read the
+value from {\tt satp}, modify it and write it to both {\tt satp} and {\tt
+  suatp} or independently modify both {\tt satp} and {\tt suatp}.
+Supervisors that use disjoint addressing mode must ensure that {\tt suatp}
+is reloaded with a known value after performing a read-modify-write on {\tt
+  satp}.  Supervisors that use unified addressing mode will observe no
+difference, since {\tt satp} and {\tt suatp} will at all times contain the
+same value if only {\tt satp} is written.
+
+\subsection{Supervisor-controlled User Address Translation and Protection
+  ({\tt suatp}) Register}
+\label{s:das:suatp}
+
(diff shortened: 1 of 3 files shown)


aswaterman:
Thanks.  To simplify our lives, I'd like to consider this proposal as part of v1.12.  Its interaction with the hypervisor spec needs to be carefully studied before it can be ratified, and adding that to the critical path of v1.11 finalization will cause us to miss our ratification deadline.

jcb62281:
Are you sure that delaying the merge is needed?  This proposal adds version 0.1, just as RVP is 0.1 in the user ISA spec and RVH is 0.1-draft.  Considerable revision on the road to 1.0/2.0 (and ratification) is expected even if the technical details do not change.
Will the hypervisor spec be out of draft stage for v1.11-final?  If not, why would RVH 0.1 block merging RVSdas 0.1?  We can harmonize them both for v1.12 even if their early versions are in conflict in v1.11.

aswaterman:
That's fair.

Can you also send this version of the proposal to the priv arch working group so it can be formally discussed (before merging)?

jcb62281:
I happen to know the address, so I will send an announcement, although that will depend on the list moderator approving it.

src/s-das.tex
```diff
@@ -0,0 +1,209 @@
...
+The SSU and SLU bits affect instructions using a specific major opcode:
+\begin{compactitem}
+\item SLU (Supervisor Load from User memory) affects LOAD instructions
+\item SSU (Supervisor Store to User memory) affects STORE instructions
+\end{compactitem}
+
+The SLU bit also affects LQ on RV128, even though LQ is encoded using the
+MISC-MEM major opcode because LOAD is full.
+
+The SSU and SLU bits are ignored if SPP is S.
```
jrtc27, line 110: This is pretty ugly and is going to end up causing problems. If you're doing a split address space you really need explicit translated load/store instructions, otherwise you're just asking to have something somewhere not save/restore sstatus appropriately and start accessing the wrong place.

allenjbaum:
This definition: 
   If the MODE field in {\tt suatp} contains an illegal value, all user translations will fault.
doesn't work, because MODE is a WARL field, and it is not possible to write an illegal value into it (because they are converted to some legal value)

allenjbaum:
Also, this says that the extension adds 2 bits to sstatus. 
Isn't it technically the case that it adds to bits to mstatus, but are visible (and writable) in sstatus aa a restricted view CSR? ( I believe it later says they are visible in mstatus, but I would say that is backwards from every other bit defined to visible in sstatus)

kersten1:
This PR started in 2018! If this info is still required, it will need to be converted to asciidoc before it can be merged. If that is the case, please work with @wmat and I to move it forward. If not, please close.

aswaterman:
@kersten1 I think that converting the PR to AsciiDoc would be putting the cart before the horse, since the proposal didn't get traction within RVIA.  To simplify the bookkeeping, I'm going to close the PR.  The discussion on the tech-privileged mailing list.  If that happens and it gains momentum, then we can convert it to AsciiDoc.

jcb62281:
I think we can declare this to be a failed proposal.

-- Jacob

PR #130, opened 2018-01-25
Clarify that canonical .{aq,rl,aqrl}->fence mappings must always be used

asb:
The current table caption might give the impression that the suggested
alternative mapping can be used at your discretion. However, it is
essential that the same mapping is used consistently for any code using
these primitives to access shared memory. Given that we can't guarantee
the same toolchain will be used for all such code, we must mandate the
canonical mapping is used.

To be more specific: if one translation unit used `fence rw, w;
s{b|h|w|d|c}` for stores while another used the alternative mapping and
so used `l{b|h|w|d|r}; fence r, rw` for loads, you would not end up with
the desired semantics.

This issue has been discussed on the RISC-V memory model working group
mailing list.

The patch currently deletes mention of the alternative mapping - let me know if there's somewhere else in the document you'd like that mapping to be preserved.

### DIFF
--- a/src/memory.tex
+++ b/src/memory.tex
@@ -1063,7 +1063,7 @@ The {\tt memory\_order\_release} mappings may use {\tt .rl} as an alternative.
     \tt amo<op>.aqrl             & \tt fence rw,rw; amo<op>; fence rw,rw \\
     \hline
   \end{tabular}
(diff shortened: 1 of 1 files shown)


aswaterman:
@daniellustig do you want this PR to be made against your branch instead, or should I merge this?

daniellustig:
For sure I'll want it on my branch as well, but I have no preference on whether you want to merge it here too in the meantime, or whether you just want to wait for the next big pull request from the memory model TG

aswaterman:
It'll simplify merging our branches if all the changes are made to yours.

daniellustig:
@asb, want to resubmit to https://github.com/daniellustig/riscv-isa-manual then?

asb:
Submitted as https://github.com/daniellustig/riscv-isa-manual/pull/8. Closing this PR.

PR #132, opened 2018-02-08
Fix typo in FENCE diagram

zakkak:
As discussed in
https://groups.google.com/a/groups.riscv.org/forum/#!topic/sw-dev/iiXF1PXvwrM

### DIFF
--- a/src/rv32.tex
+++ b/src/rv32.tex
@@ -1529,7 +1529,7 @@ Finally, they also imply that all AMO instructions are fully-fenced; nothing wil
 \instbitrange{11}{7} &
 \instbitrange{6}{0} \\
 \hline
-\multicolumn{1}{|c|}{0} &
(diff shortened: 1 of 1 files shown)


aswaterman:
This change makes the figure spill into the margins.  I'll just fix the comment to say "instruction bits 31:28" instead.

sorear:
Can we also update the "Instruction listing" to show the ignored fields of FENCE and FENCE.I as actually-ignored rather than must-be-zero?

aswaterman:
Any suggestions as to how that should be presented?

sorear:
Might be enough to include rs1/rd in the table (like in [1.0](https://www2.eecs.berkeley.edu/Pubs/TechRpts/2011/EECS-2011-62.pdf#page=31), oddly), and let people refer to the long description.

PR #139, opened 2018-02-22
Introduce IALIGN; propose misa.C semantics

aswaterman:
This is still being debated; I'm making the PR so the PrivArch group can discuss something concrete.


src/machine.tex
```diff
@@ -171,9 +171,11 @@ \subsection{Machine ISA Register {\tt misa}}
 interface without burdening hardware designs.
 \end{commentary}
 
-When clearing the ``C'' bit in {\tt misa}, software must ensure that the
-current {\tt pc} is 4-byte aligned and that all {\em x}{\tt epc} registers
-contain 4-byte-aligned values.
+Writing {\tt misa} may reduce IALIGN, e.g., by disabling the ``C''
```
sorear: s/reduce/increase/

aswaterman: thanks

PR #142, opened 2018-03-05
early intra-hart --> early inter-hart

bshanks:
perhaps this was a typo?


kasanovic:
Original looks correct to me.  The sentence is talking about store-load forwarding within a hart.

bshanks:
That's true -- i amended the pull request to remove the word 'other' also -- it seems to me that the first part of the sentence is saying that early intra-hart forwarding IS permitted, and the second part (should be) saying that early inter-hart forwarding IS NOT permitted. This makes sense to me because the property of multi-copy atomicity is about inter-hart behavior, i think. But i could be wrong -- sorry if i am incorrect.

daniellustig:
Thanks @bshanks, you're right, it was meant to say "inter".  I think your latest proposed change (also removing "other") does a good job clarifying the text.

PR #148, opened 2018-03-21
Make it explicit that the page-fault exceptions from address translation correspond to the original access type, as is done for access exceptions.

pmundkur:
It is implicit, but being more explicit seemed better.

### DIFF
--- a/src/supervisor.tex
+++ b/src/supervisor.tex
@@ -1212,8 +1212,10 @@ must be cleared by software for forward compatibility.
 \subsection{Virtual Address Translation Process}
 \label{sv32algorithm}
 
-A virtual address $va$ is translated into a physical address $pa$ as
-follows:
+A virtual address $va$ used for a memory access is translated into a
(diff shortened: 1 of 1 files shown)


aswaterman:
What do you think about making this change for the first sentence only?  It is pretty verbose.

pmundkur:
On 15:39 Wed 21 Mar, Andrew Waterman wrote:
> What do you think about making this change for the first sentence only?  It is pretty verbose.

Sure.  It would be good to make the treatment consistent for access
exceptions and page-faults.  Perhaps a single line about both before
the enumerated steps, and omit the qualifications inside the
enumerated steps?

aswaterman:
Yeah, let’s see how that looks. Thanks!

pmundkur:
@aswaterman  any opinion on how it looks?

aswaterman:
After comparing them, I prefer your first version :) I've merged it.  Thanks.

PR #156, opened 2018-04-13
Inkscape figures

### LINKED ISSUE #86: build figures from source, use non-proprietary source format
The `src/figs/*.pdf` files were created from `src/graffles/*.graffle` but the build system in `build/Makefile` does not rebuild the PDF figures when the OmniGraffle files change.

I would recommend to:

* remove the prebuilt PDF figures from the git repository
* switch from the proprietary OmniGraffle file format as source for your figures to an open format like SVG that can be converted to PDF at build time.

It looks like OmniGraffle can export to SVG, which can then be edited in vector graphics editors like Inkscape and then Inkscape can convert to PDF for import into the TeX document:

https://en.wikipedia.org/wiki/OmniGraffle
https://tex.stackexchange.com/questions/2099/how-to-include-svg-diagrams-in-latex

terpstra:
Inkscape seems more than capable enough to maintain these images going forward; Fixes #86.


aswaterman:
Thanks.  Software libre!

PR #157, opened 2018-04-17
Misaligned superpage exception occurs if pte.ppn[i−1:0]!=0

premjithav:
Misaligned superpage exception occurs if **pte.ppn[i−1:0]!=0**. In the current specification instead of pte.ppn it given as pa.ppn.


aswaterman:
Thanks

PR #158, opened 2018-04-18
Clarify interrupt delegation semantics

aswaterman:


gameboo:
Hello.
This is how I understand the trap delegation mechanism for interrupts:
example

About the new bit of text:

> For example, if the supervisor timer interrupt (STI) is _delegated to S-mode_ by setting _mideleg[5]_, STIs _will not be __taken__ when executing in M-mode_. By contrast, if _mideleg[5]_ is clear, STIs can be taken in any mode and regardless of current mode will transfer control to M-mode.

Do you mean not taken at all, or taken but with a mode switch to S-mode? I was under the impression that the _x_**ideleg** registers were simply used to control "how far an interrupt bubbles up" whereas the _x_**ie** registers where used to block them from being taken. In other words, is the blue path in my example drawing correct or not? (Perhaps it is worth adding a similar diagram in the document?)

Thanks

aswaterman:
The blue line is indeed impossible. An interrupt can only be taken when the current privilege mode is less than or equal to the privilege mode to which it has been delegated.

Diagrams would definitely help explain this unusual architectural feature (but fair warning, Krste will go several rounds of revisions on them :))

I am happy to review pull requests that improve the clarity of the text.

gameboo:
Interesting, I definitely didn't understand that. So, when running in machine mode an receiving interrupt *m*, if mie[*m*] is 1 and mideleg[*m*] is 0, then the interrupt would be handled in machine mode, and if mie[*m*] is 1 and mideleg[*m*] is 1, the interrupt that is not masked is simply not handled?

aswaterman:
In the latter case, the interrupt would be effectively masked until returning to a less-privileged mode.

gameboo:
I see. Thanks for helping me out ;)

aswaterman:
It's also helpful for me to know which parts of the document are confusing... this being one of them.

gameboo:
To be fair, the expected behaviour is the one that was actually described. It's a bit of an involved process to get your head around every detail is all.

gameboo:
I am having an attempt at a couple improvements, following up on the suggestion. I was wondering: can bits 2, 6 and 10 of *mideleg* be considered *wiri*?

aswaterman:
I’d like @kasanovic to review it

PR #161, opened 2018-04-28
Forward-reference NaN boxing section when describing FMV.W.X

aswaterman:


aswaterman:
I changed my mind.  This is actually more confusing than before, because it suggests

Issue #163, opened 2018-05-01
URET from U-mode without N extension


gameboo:
Hello.
I think the spec is not clear in [3.2.2](https://github.com/riscv/riscv-isa-manual/blob/ba5cfa4d1eb0aeb9e694dbaf6ccb05312cca02b0/src/machine.tex#L1970) about what should happen when a URET is executed from U-mode when the N extension is not present:

> URET is only provided if user-mode traps are supported

I presume that implicitly means "raise an illegal instruction exception if URET is executed from U-mode when the N extension is not present"?

benjaminselfridge:
Yep, that's right. "Not supported" generally means "is considered an illegal instruction."

gameboo:
OK cool. It would be worth being explicit here, no?
Thanks again.

benjaminselfridge:
If you like, go ahead and submit a pull request; it has a decent chance of being integrated, I'd bet.

PR #164, opened 2018-05-02
Clarify URET with no user-mode traps support

gameboo:
Following up on [issue 163](https://github.com/riscv/riscv-isa-manual/issues/163)...


gameboo:
Thank you for pointing this out! It made me realise that I actually got my SRET instruction encoding wrong! I very much agree with your observation. New commit coming up.

aswaterman:
Thanks!

PR #167, opened 2018-05-03
Add graphviz sources of figures from the "RVWMO Explanatory Material" appendix.

maranget:
Those sources (.dot) are to be processed by neato.


aswaterman:
Thanks!

Issue #144, opened 2018-03-15
LA pseudo instruction implementation


andrewt0301:
The load address pseudo instruction (`la rd, symbol`) is described in the reference manual as follows:

`auipc rd, symbol[31:12]`
`addi rd, rd, symbol[11:0]`

In some situations, this implementation seem to cause incorrect results.
The `addi` instruction sign-extends the immediate value (`symbol[11:0]`). So, when `symbol[11]` equals 1, the immediate value is considered negative.

For example, when symbol=0x0000_0FFF `symbol[11:0]` is turned into 0xFFFF_FFFF. 

How to deal with such situations?
Do I get anything wrong?

aswaterman:
You are right, of course.  The syntax in the manual is shorthand, and deliberately neglects both PC relativity and sign extension.

Regardless, we are moving the assembly-programming stuff to a separate document entirely: https://github.com/riscv/riscv-asm-manual

andrewt0301:
Thanks. I have read this document too.

Eventually, I found out the trick that the RISC-V assembler does to deal with sign-extension of symbol[11:0]. When symbol[11] equals 1, it adds 1 to symbol[31:12]. This resets the high bits that were set to 1 as a result of sign-extension:

`auipc rd, symbol[31:12] + 1`
`addi rd, rd, symbol[11:0]`

Probably, this is worth mentioning in the documentation.

aswaterman:
I'll leave that to the folks writing https://github.com/riscv/riscv-asm-manual

PR #177, opened 2018-05-26
Clarify pseudoinstruction sign extension neglected

diekmann:
As discussed in issue 144, the table "RISC-V pseudoinstructions" in the
placeholder "RISC-V Assembly Programmer's Handbook" chapter neglects
sign extension for brevity. This commit adds this assumption explicitly
to the table.

I was lying on the beach (no Internet, no laptop), reading a printout of the RISC V manual. Nice read, thanks! But I spent a long time staring at the pseudoinstruction table and reading the parts about sign extension again and again, because I thought my understanding of sign extension was wrong. I was so desperate that left the beach to find a computer. Fortunately, @andrewt0301 already raised the same question in issue #144. Though the whole assembly programming chapter will be replaced, please accept this little commit to make the risc v manual a better reading material for the upcoming holiday season :smile:

### DIFF
--- a/src/assembly.tex
+++ b/src/assembly.tex
@@ -50,16 +50,17 @@ RISC-V pseudoinstructions.
 \begin{tabular}{l l l}
 pseudoinstruction & Base Instruction(s) & Meaning \\ \hline
 
-\multirow{2}{*}{\tt la rd, symbol} & {\tt auipc rd, symbol[31:12]} & \multirow{2}{*}{Load address} \\
-                                   & {\tt addi rd, rd, symbol[11:0]} \\
-\multirow{2}{*}{\tt l\{b|h|w|d\} rd, symbol} & {\tt auipc rd, symbol[31:12]} & \multirow{2}{*}{Load global} \\
-                                           & {\tt l\{b|h|w|d\} rd, symbol[11:0](rd)} \\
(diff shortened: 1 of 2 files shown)


diekmann:
Ping @aswaterman

aswaterman:
What do you think about actually fixing the table, rather than adding the caveat?  It's just a matter of adding bit 11 into the high part.

For example, for la,

    _Let delta = symbol - pc_
    auipc rd, delta[31:12] + delta[11]
    addi rd, rd, delta[11:0]

diekmann:
Sure! To avoid further round trips, fell free to make this change yourself. Otherwise, I will modify this pull request *soon*™.

aswaterman:
I can wait ;)

diekmann:
Updated (via force push) as discussed. I split content-related changes and typesetting-only into separate commits.

Did I get `call offset` and `tail offset` right?

This is how the improved typesetting looks like: imgur

aswaterman:
OK, I took this with slight tweaks in 3f05dd6

diekmann:
:+1: :+1: :+1: 

Yep, your wording is nicer.

Nit-picking: Your explanation uses passive voice. It is not specified which magical entity actually subtracts and adds the bits. Using active voice instead, we could clarify that the compiler does this.

But the current version is absolutely fine and everything is understandable from the context.

aswaterman:
I chose to use the passive voice to be deliberately vague (because it's not the compiler, it's the linker).  If you want to change it to active voice with linker as the subject, that's fine with me, too.

PR #178, opened 2018-05-26
Nit: no hyphen in instruction-set

diekmann:
Unify the writing of "instruction set".

### DIFF
--- a/src/d.tex
+++ b/src/d.tex
@@ -2,7 +2,7 @@
 Version 2.0}
 
 This chapter describes the standard double-precision floating-point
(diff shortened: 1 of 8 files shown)


aswaterman:
It's more correct with the hyphen.  Nevertheless, it's clearly inconsistent.

I'm leaving this to @kasanovic

aswaterman:
OK, I've made the style consistent.  "instruction set" is now always hyphenated when it's part of a greater noun phrase, and never hyphenated when it stands alone as a compound noun.

Thanks for pointing this out, even though our resolutions diverged!

PR #179, opened 2018-05-30
Specify operand order for div instructions

gameboo:
Same as [#172](https://github.com/riscv/riscv-isa-manual/pull/172), bu for `div` (is it `rs1/rs2` or `rs2/rs1`?). `rem` seems to be defined in terms of `div` in the text, so that one is no longer ambiguous either. I wonder whether there are more noncommutative operations that would have also a similar ambiguity...


aswaterman:
Thanks. Can you check FDIV and FSQRT, too?

PR #180, opened 2018-05-30
FDIV.S and FSUB.S disambiguation

gameboo:
Following up on [#179](https://github.com/riscv/riscv-isa-manual/pull/179). `FDIV` and `FSQRT` are actually defined in terms of the single precision instructions. `FSQRT.S` looked fine to me. I just rephrased the paragraph to be explicit about `FSUB.S` and `FDIV.S`.


aswaterman:
Thanks!

PR #181, opened 2018-05-31
Disambiguation of MULHSU + explicit rd as a destination for DIV,DIVU, REM,REMU

gameboo:
Same idea as [#180](https://github.com/riscv/riscv-isa-manual/pull/180).
This time `MULHSU` as I understand it is non-commutative, but I am not sure I fully understand this instruction. I guess more than a pull request, this is me asking for someone to help me understand this bit:

> *MULHSU is used in multi-word signed multiplication to multiply the most-significant word of the multiplier (which contains the sign bit) with the less-significant words of the multiplicand (which are unsigned).*

Should I understand "word" in the text above as being XLEN-bit chuncks? I think it makes sence if I do that... Could someone give me an example to make sure I got it? 

Also there is a more trivial change to explicitly mention a destination register `rd` for `DIV` & co.

### DIFF
--- a/src/m.tex
+++ b/src/m.tex
@@ -40,11 +40,13 @@ MULDIV & multiplier & multiplicand & MULW           & dest & OP-32 \\
 \end{center}
 
 MUL performs an XLEN-bit$\times$XLEN-bit multiplication and places the
-lower XLEN bits in the destination register.  MULH, MULHU, and MULHSU
-perform the same multiplication but return the upper XLEN bits of the
-full 2$\times$XLEN-bit product, for signed$\times$signed,
-unsigned$\times$unsigned, and signed$\times$unsigned multiplication
-respectively.  If both the high and low bits of the same product are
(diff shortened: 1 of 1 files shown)


aswaterman:
Yes, "word" = "xlen-bit chunks" in t he commentary.

Your PR is fine, but I chose to phrase it differently in 41cab885f15d5bd2f4e915c087bbc1438029c1cd -- let me know if you have any comments on my version.

BTW, I added you to the list of contributors.  Thanks for helping  clarify these ambiguities in the spec.

aswaterman:
Re: division: For the "W" instructions, the table is accurate if you substitute "32" for "XLEN."  In other words, DIVW overflows on RV64 when dividing -2^31 by -1; the result is -2^31.

I would prefer to clarify this issue with English, rather than adding a new table (or, worse, more columns in the existing table).  But feel free to use your judgment and propose something in another PR.

PR #182, opened 2018-06-01
Specify division by zero for DIVW, DIVUW, REMW, REMUW

gameboo:
Follow up on #181.
I lifted the paragraph defining the `W`-version of the instructions above the table, and tried to make the table generic in that the 32-bit values for the "all ones" and "most negative int" are different for the `W`-version (for example, XLEN can be 64, but the most negative 32-bit int is not -2^{XLEN-1} but rather -2^{32-1}).
This PR simplifies the text a bit, but does make the table harder to read. Another way to specify it without adding an extra table would be to simply add a note in the `W`-instruction paragraph as...

### DIFF
--- a/src/m.tex
+++ b/src/m.tex
@@ -101,25 +101,33 @@ cannot be the same as {\em rs1} or {\em rs2}).  Microarchitectures can
 then fuse these into a single divide operation instead of performing
 two separate divides.
 
-The semantics for division by zero and division overflow are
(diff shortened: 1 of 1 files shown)


aswaterman:
Looks fine.  I tweaked it to use L instead of y, and moved the definition of L into the caption.  See 2cc659bb2e0f86ca9e7a3d99100c1bb7fd3affcc

PR #183, opened 2018-06-01
Added language that permits sharing mtimecmp CSRs between harts at an…

allenjbaum:
…y granularity

### DIFF
--- a/src/m.tex
+++ b/src/m.tex
@@ -160,3 +160,36 @@ unsigned divider implementations.  Signed division is often
 implemented using an unsigned division circuit and specifying the same
 overflow result simplifies the hardware.
 \end{commentary}
+
+\section{ Zmmul Subextension}
+
+The Zmmul extension implements the multiply operations subset of the M-extension.
+It defines instructions comprising M-extension unprivileged spec section 8.1 above
+\\*    MUL, MULH, MUHU and MULHSU 
+\\*and (for RV64/RV128) 
+\\*    MULW 
+\\*without section 8.2 of the M-extension: DIV* or REM* instructions.
+No other architectural state is changed.
+The encodings are identical to corresponding M-extension.  
+The ISA string is \textbf{Zmmul}.
+
+This extension implicitly allows the DIV* and REM* opcodes to cause
+illegal-instruction exceptions (e.g. if a platform requires DIV*/REM*
+emulation) or be used for (non-conforming) custom instructions. It
+therefore cannot co-exist with the M-extension. It has no dependencies
+on other extensions and can be used in any combination that the
+M-extension could be used.
+
+\begin{commentary}
+The rationale for this extension is that, particularly in the
+microcontroller market where cost reduction is critical, divide
+operations are infrequent and divider hardware is expensive, while
+multiplication operations are more useful and both easier and cheaper to
+implement. A specific example is that some FPGAs come with built-in
+multipliers but not built-in dividers, so that it makes the core
+implementation much more expensive.
+
+In short: this enables very low end microcontroller implementations that
+need a multiplier, but without the requirement of having to also
+implement divide or remainder instructions.
+\end{commentary}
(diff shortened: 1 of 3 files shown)


gameboo:
Hello. I was looking at the `mtime` and `mtimecmp` section and got confused with the definition as it currently is in the spec, and then saw this pull request.

First, am I correct in understanding that the expected behaviour of `mtimecmp` is not defined as part of the riscv architecture, but rather should be specified on a per-platform basis?

Second, `mtimecmp` and `mtime` are listed as "Machine level CSRs" (under subsection 3.1). This is quite misleading/confusing for me:

- They are claimed to be memory mapped
- `mtime` has a CSR address but `mtimecmp` does not (unless I've missed something)

My questions here are:

- Why isn't `mtime` exclusively memory mapped or exclusively mapped as a CSR?
- Why are both memory mapped in the first place?
- If it is indeed desirable for a CSR to be exclusively memory mapped, than it's not really a CSR... So `mtimecmp` should probably not be described as a CSR.

Can anybody comment on those concerns?
Thanks.

ilg-ul:
> On 16 Jul 2018, at 19:27, allenjbaum <notifications@github.com> wrote:
> 
> ... the timer has to keep running at a
> constant rate even if the core is clock gated, powered down, or its
> frequency changes.
> The CSR block is inside the core, so if mtime is implemented as a CSR, it
> will stop if the core is clock gated or powered down, and its count rate
> will change if its frequency changes

even for an ignorant software guy like me, this explanation does not stand.

probably the intent was to save some transistors in adapters between different voltage domains. at the expense of severely complicating the software. remember, people are more expensive than transistors.


regards,

Liviu

allenjbaum:
That;s why there are experienced HW guys around.

gameboo:
>  it is in some thread somewhere than can be found

Does someone have a pointer to this thread?

> the timer has to keep running at a constant rate  even if the core is clock gated, powered down, or its
frequency changes. The CSR block is inside the core, so if mtime is implemented as a CSR, it will stop if the core is clock gated or powered down, and its count rate will change if its frequency changes

Thank you, that is useful information to help me understand the kind of reasons that motivate this choice. I hadn't considered this issue.

> An implementation could choose to translate the CSR read into a MMIO read

I was thinking of something along those lines. This would allow for `mtime` and `mtimecmp` to be actual CSRs, and would address the clock boundary challenge that was pointed out. Naively (i.e. without having gone through the thread that was mentioned), I would tend to prefer this approach to having those memory mapped. As far as I understand, memory mapped registers could mean bits of memory space private to certain cores and the likes, leading to some amount of pain from a memory consistency perspective? What was the main argument for memory mapped over CSR mapped?

> I would say that  behavior of mtimecmp is defined; how that behavior is implemented is not.

I guess what I should have asked is something more precise: one could argue that, from an interrupt perspective, whether `mtimecmp` is to trigger an interrupt to the current, or to some arbitrary combination of harts is be part of the behaviour itself rather than simply the implementation of that behaviour. I do however see that it is convenient to define this on a more case by case basis as platform specific. The pull request is quite helpful for that I think.

> That doesn't answer why mtimecmp is not also listed as a CSR

I guess what I would suggest is that, if a so called "CSR" is memory mapped exclusively, the use of the term "CSR" be avoided as it is confusing. It is also confusing to have `mtime` CSR mapped and Memory Mapped (Not sure what the reason is to have two ways to access this). Should both `mtime` and `mtimecmp` end up memory mapped, then it would probably make sense to not list them under the Machine CSR section at all.

Thanks a lot for the explanation!

jnk0le:
Zmmul probably shouldn't go there.

allenjbaum:
Ah - I added to a branch that has been sitting around for a looong time, thinking it was one that I had just cloned... or something. I hadn't realized that old pull request was still there.

allenjbaum:
this was a combination of 3 different extensions (mistakenly). Two are now obsolete, as zmmul as been added to the isa manual separately, as has a different counter overflow  extension. I'll reintroduce the remaining one (sharing mtimecmp between harts) if there is agreement that it should be.

Issue #147, opened 2018-03-21
Allow FS to be imprecise


palmer-dabbelt:
I can't find anything in the spec that explicitly allows FS to be imprecise.  This table lists the possible FS state transitions:

https://github.com/riscv/riscv-isa-manual/blob/5f8294b56e0447775c444c5c641713d8e67c4b88/src/machine.tex#L824

There's some vague "When the status is Dirty, the corresponding state has potentially been modified since the last context save.", but the state transition table makes it appear that it'l not legal to set the dirty state without executing an instruction.

aswaterman:
@palmer-dabbelt i still agree this is a problem, but i'm going to punt it back to you to propose a PR.

PR #185, opened 2018-06-05
 Be more explicit about allow less precise FS and XS transitions

palmer-dabbelt:
The goal here is to explicitly allow for implementations that use two
states to track FS.  The previous wording was ambiguous here, as it
seemed to allow that in the text but then specify a set of state
transitions.  This new wording makes it explicit that the listed state
transitions are just the most precise possible, and defines how
implementations are allowed to be less precise.

There's also one commit that has no output changes, it just re-aligns a table.

Fixes: https://github.com/riscv/riscv-isa-manual/issues/147

### DIFF
--- a/src/machine.tex
+++ b/src/machine.tex
@@ -831,10 +831,16 @@ register state.  In particular, setting FS=Off does not destroy the state, nor
 does setting FS=Initial clear the contents.  Other extensions might not
 preserve state when set to Off.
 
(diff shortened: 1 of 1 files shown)


aswaterman:
I agree with the semantics you proposed, but I'm somewhat inclined to add an explicit exception for FS-dirty-vs-off, rather than generalizing it to everything else.  I'll make a dueling PR tmrw, and we can choose which one works best.

Issue #186, opened 2018-06-07
fence.tso and ignored fence fields


asb:
The current draft specification introduces fence.tso, which makes use of the previously unused bits in instr[31:28]. Since at least version 2.0, implementers were told to ignore these fields (although this was made much more clear in version 2.1):

> The unused fields in the FENCE instruction, imm[11:8], rs1, and rd, are reserved for finer-grain
> fences in future extensions. For forward compatibility, base implementations shall ignore these
> fields, and standard software shall zero these fields.

The above statement was modified in the current draft. I'd suggest that in the new revision, the description of alternate values for imm[31:28] as "invalid" might muddy the waters a bit. I would expect trapping to be a valid behaviour for "invalid" field values and the specification doesn't seem to indicate otherwise. But surely we want implementers to continue to ignore unrecognised fm field values to allow future extensibility?

I would suggest:
1) Changing use of 'Invalid' in table 2.2 to 'Ignored'. This isn't an ideal description as the table tries to simultaneously describe mnemonics as well as encodings, and of course unrecognised mnemonics should result in errors in software.
2) Clarify the intended interpretation of "Likewise, all combinations of fm and predecessor/successor
set not listed in Table 2.2 are also reserved for future use." Again, I assume the intent is that these encodings are ignored by compliant cores. "Reserved for future use" doesn't seem sufficiently clear about whether these encodings should trap or not.

Or is it explicitly intended that unrecognised field values now trap? If so, this probably deserves a mention in the list of major changes given that all core implementers would probably want to revise their RV32I/RV64I instruction decoders.

A related comment:
`fence.tso` was added to the table of pseudo-instructions and expands to `fence.tso rw, rw`. Other pred/succ specifications aren't valid (as explained in section 2.9). Given that, is there any point in accepting the `fence.tso rw, rw` form? Wouldn't it be simpler to define `fence.tso` as an instruction and then later specify it as a pseudoinstruction in the case that alternate pred/succ arguments are accepted.

aswaterman:
I agree `fence.tso` should take no operands, and this is how I implemented it in binutils.  If fence.tso is later extended to accept things other than rw, rw, then at that time it can be extended to accept operands.

aswaterman:
IMO, changing it to "Ignored" isn't quite right, either: it implies to me that software is allowed to put stuff there.  "Ignored" correctly describes what hardware should do, but it remains "invalid" for software to use those encodings.

@daniellustig want to take a stab at clarifying this?

[...]

PR #187, opened 2018-06-07
Don't describe fence.tso as a pseudoinstruction

asb:
As discussed in #186, it makes no sense for RISC-V assemblers to accept
`fence.tso rw, rw`. Removing this expansion from the pseudoinstruction
table (23.2) seems consistent with the description of fence.tso
elsewhere in the document.

If different pred/succ for fence.tso are made legal in a future spec, at
that point it would make sense to list fence.tso as a pseudoinstruction.


aswaterman:
Thanks, I did the same thing in 3943720d13ce0b10c25507aeb085eac70fe6e507.

PR #188, opened 2018-06-08
Disambiguation of accesses to SSIP/USIP bits through mip/sip/uip

gameboo:
The bold section in the initial text

> Each lower privilege level has a separate software interrupt-pending bit (SSIP, USIP), which can
be both read and written by CSR accesses from code running on the local hart **at the associated
or any higher privilege level**.

suggests that `mip` could be accessed from other privilege modes than `M`. According to the CSR Address Mapping Conventions chapter, if I understand correctly, any access to `mip` from another mode than `M` should raise an illegal instruction exception. I presume that what is really being refered to here is the access to those bits through the `sip`/`uip` views of the `mip` register, so I am suggesting to make this explicit...

### DIFF
--- a/src/machine.tex
+++ b/src/machine.tex
@@ -1268,9 +1268,10 @@ UTIE for M-mode, S-mode, and U-mode timer interrupts respectively.
 
 Each lower privilege level has a separate software interrupt-pending
 bit (SSIP, USIP), which can be both read and written by CSR accesses
-from code running on the local hart at the associated or any higher
-privilege level. The machine-level MSIP bits are written by accesses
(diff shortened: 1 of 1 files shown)


aswaterman:
I don't see how it suggests `mip` can be accessed from less-privileged modes.

The accessibility of these fields through `sip`/`uip` is made explicit a few paragraphs earlier: https://github.com/riscv/riscv-isa-manual/blob/master/src/machine.tex#L1165

gameboo:
Yes... I guess when I first read through, I was under the impression that this was talking specifically about `mip`. I think I am happy dropping this pull request :) sorry about that!

PR #192, opened 2018-06-11
Explicitly mention that FS may be imprecise

### LINKED ISSUE #147: Allow FS to be imprecise

aswaterman:
Closes #147

### DIFF
--- a/src/machine.tex
+++ b/src/machine.tex
@@ -831,6 +831,19 @@ register state.  In particular, setting FS=Off does not destroy the state, nor
 does setting FS=Initial clear the contents.  Other extensions might not
 preserve state when set to Off.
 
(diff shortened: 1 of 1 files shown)


palmer-dabbelt:
Also closes https://github.com/riscv/riscv-isa-manual/pull/185

PR #194, opened 2018-06-14
Help the reader by pointing at TVM, TW and TSR in the relevant sections

gameboo:
Sections that are possibly distant in the document from the `mstatus` section are often affected by bits in the `mstatus` register.
I suggest helping the reader by mentioning that specific instructions/csrs are influenced by `mstatus` where those instructions/csrs are defined as opposed to exclusively in the `mstatus` section.

EDIT: Added a mention of TSR when describing SRET as well.

Additionally, when describing the `mstatus` TW field, shoud U-mode be mentioned in the case that user interrupts are enabled?

### DIFF
--- a/src/machine.tex
+++ b/src/machine.tex
@@ -2011,9 +2011,10 @@ MRET/SRET/URET & 0 & PRIV & 0 & SYSTEM \\
 To return after handling a trap, there are separate trap return
 instructions per privilege level: MRET, SRET, and URET.  MRET is
 always provided. SRET must be provided if supervisor mode is
-supported, and should raise an illegal instruction otherwise.
-URET is only provided if user-mode traps are supported, and
-should raise an illegal instruction otherwise.
+supported, and should raise an illegal instruction exception otherwise. SRET
+should also raise an illegal instruction exception when TSR=1 in {\tt mstatus},
+as described in Section~\ref{virt-control}. URET is only provided if user-mode
+traps are supported, and should raise an illegal instruction otherwise.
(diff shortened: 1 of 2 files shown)


aswaterman:
I don't actually think TVM is germane to the supervisor-ISA section.  It seems like an abstraction violation to describe it here.

Supervisor-mode software shouldn't be able to tell whether TVM is set, and a hypothetical pure-supervisor (no-M-mode) hardware implementation wouldn't have TVM.

gameboo:
I see... When coming from the angle of the implementer, it is still somewhat frustrating to think that you are "doing everything that the instruction should do" and realise after chasing various different things that your answer lies in that other section you should have remembered... Would simply mentioning the `mstatus` section without naming the field be acceptable ?

aswaterman:
Re: edit 2: yes, these are just compliant w.r.t. different platform specifications.  For example, qemu-user is meant to be a compliant implementation of the RV user-level ISA, but of course it isn't a compliant RV S-mode implementation, since it doesn't implement S-mode.

I'll read and comment on your revised commit soon.

aswaterman:
btw, editing posts doesn't cause github to send more email, so it's better to write a new post (or delete then write a new post)

gameboo:
> editing posts doesn't cause github to send more email

Good to know! thanks

gameboo:
I added a new commit. Not sure if those trigger notifications...

aswaterman:
Yep, new commits send email notifications to people who have already interacted with the PR.

aswaterman:
Looks good now, thanks!

PR #198, opened 2018-06-18
Strengthen guidance on the need to clear a reservation using SC

asb:
Commit 170f3c5 clarified that reservations can be cleared with an SC to
a dummy memory location. As discussed
<https://github.com/riscv/riscv-isa-manual/commit/170f3c52bd134ac90c3467f77925fdaa68e9b8f6#commitcomment-29386537>,
this patch makes it clear that the reservation "should" be cleared in
this way during a preemptive context switch.

Anyone writing preemptive context switch code should be forcibly
clearing load reservations. Although other mechanisms might be used,
this is the standard way of doing it (hence "should" rather than
"must").

### DIFF
--- a/src/a.tex
+++ b/src/a.tex
@@ -215,8 +215,9 @@ any address) between the LR and the SC in program order.  In other
 words, multiple outstanding reservations are not permitted.
 
 \begin{commentary}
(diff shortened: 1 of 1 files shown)


aswaterman:
thanks

PR #201, opened 2018-06-20
Clarify that AMOs are always stores, and clarify fence.tso configs

daniellustig:
Two small updates to the memory model text

1) Clarify that AMOs are always semantically considered stores
2) Attempt to clarify #186


aswaterman:
@asb this looks to me to address #186, so I’m going to merge.

Thanks @daniellustig

PR #203, opened 2018-06-26
Clarification of NOP description

wdc-pnl:
Addition of instret to state changed by NOP.


aswaterman:
Merged manually in f94493344f5c8351b277f28c3a200b3e5e5b63d6

PR #205, opened 2018-06-28
Clarifications on the virtual memory translation mechanism

gameboo:
This is based on a confusion I had that memory translation should always take place when enabled, that is even for memory accesses in M-mode. After chatting with a colleague, I understand now that it should only be the case for S-mode and U-mode (although memory translation with an effective current privilege of M-mode can take place when writing S-mode or U-mode to MPP and 1 to MPRV).

Is this correct?
If so, this set of changes tries to make it more explicit that M-mode virtual translation are not a thing.

### DIFF
--- a/src/machine.tex
+++ b/src/machine.tex
@@ -638,15 +638,20 @@ widest supported width not wider than the new MXLEN.
 
 \subsection{Memory Privilege in {\tt mstatus} Register}
 
-The MPRV (Modify PRiVilege) bit modifies the privilege level at which
-loads and stores execute in all privilege modes.  When MPRV=0,
-translation and protection behave as normal, that is, M-mode accesses only go
-through the PMP mechanism (Section~\ref{sec:pmp}), and S-mode and U-mode
-accesses may also go through the address translation mechanism
-(Section~\ref{sv32algorithm}).  When MPRV=1, load and store memory addresses
-are translated and protected as though the current privilege mode were set to
-MPP.  Instruction address-translation and protection are unaffected.  MPRV is
-hardwired to 0 if U-mode is not supported.
+The MPRV (Modify PRiVilege) bit modifies the privilege level at which loads
+and stores execute in all privilege modes. When MPRV=1, load and store memory
+addresses are translated and protected as though the current privilege mode
+were set to MPP. When MPRV=0, loads and stores are unaffected. Instruction
+address-translation and protection are unaffected. MPRV is hardwired to 0 if
+U-mode is not supported.
+
+\begin{commentary}
+MPRV enables M-mode software to use the translation and protection mechanisms
+active in S-mode and U-mode, typically the one described in
+Section~\ref{sv32algorithm}. When MPRV=0, M-mode accesses do not go through
+these mechanisms, but still undergo checks active in M-mode, such as the PMP
+mechanism described in Section~\ref{sec:pmp}.
+\end{commentary}
 
 The MXR (Make eXecutable Readable) bit modifies the privilege with which loads
 access virtual memory.  When MXR=0, only loads from pages marked readable (R=1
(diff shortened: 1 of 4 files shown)


src/priv-intro.tex
```diff
@@ -146,9 +146,9 @@ \section{Privilege Levels}
 
 Each privilege level has a core set of privileged ISA extensions with
 optional extensions and variants.  For example, machine-mode supports
-several optional standard variants for address translation and memory
-protection.  Also, supervisor-mode can be extended to support Type-2
-hypervisor execution as described in Chapter~\ref{hypervisor}.
+an optional standard variant for memory protection.  Also, supervisor-mode
```
aswaterman: I'd say "extension" rather than "variant" here

src/supervisor.tex
```diff
@@ -805,6 +805,10 @@ \subsection{Supervisor Address Translation and Protection ({\tt satp}) Register}
 and if {\tt satp} is written with an unsupported MODE, the entire write has
 no effect; no fields in {\tt satp} are modified.
 
+When memory translation is active, instruction fetches and memory loads and stores
+from S-mode and U-mode go through the translation mechanism described in
+Section~\ref{sv32algorithm}.
+
```
aswaterman: This is only true for currently defined MODEs.  I would remove this paragraph and instead augment the table of MODEs to have pointers to the various SvXX sections (e.g. Page-based 32-bit virtual addressing (see Section~\ref{sec:sv32}))

src/machine.tex
```diff
@@ -627,11 +627,13 @@ \subsection{Memory Privilege in {\tt mstatus} Register}
...
 loads and stores execute in all privilege modes.  When MPRV=0,
-translation and protection behave as normal.  When MPRV=1, load and
-store memory addresses are translated and protected as though the
-current privilege mode were set to MPP.  Instruction
-address-translation and protection are unaffected.  MPRV is hardwired
-to 0 if U-mode is not supported.
+translation and protection behave as normal, that is, M-mode accesses only go
+through the PMP mechanism (Section~\ref{sec:pmp}), and S-mode and U-mode
+accesses may also go through the address translation mechanism
+(Section~\ref{sv32algorithm}).  When MPRV=1, load and store memory addresses
```
aswaterman: I don't think the MPRV description should explicitly mention any mechanisms like PMP or VM.  The main reason for this is composability: the MPRV text shouldn't depend on what protection/translation schemes exist.

Suppose someone adds a new custom extension for translation/protection.  MPRV should still work in the context of that mechanism, because all it does is change the effective privilege mode.

src/priv-intro.tex
gameboo: sure thing

src/machine.tex
gameboo: I think I agree with you... As I was writing that, I was wondering about this.
However, I would like to have a way to easily know what "normal" find somewhere in the document an easy way to know what

> behave as normal

means. Doing this would break composability with newer mechanisms, but maybe giving an example would do the trick:
say a setup with MUS modes, no pmp, virtual memory and say that "behave as normal" in this case means translation happens for U / S and not M. This would clearly state that M-mode accesses don't go through mem translation... What do you think / where would this be appropriate in the document?

aswaterman: Agreed...  "behaves as normal" is pretty vague, but was the best I could come up without abandoning generality.

Rather than clarifying this in the normative text, it might be best to explain the intended use of MPRV in a commentary block, and therein provide an example, which makes its effect more clear.  You want to give that a stab, or should I?

aswaterman:
OK, thanks, I've incorporated most of these suggestions into c7a81b2be5e210da2134b29aeb8e0fda1893cedd

gameboo:
great

PR #206, opened 2018-07-01
Change hypervisor virtual mode names to VS-mode and VU-mode

jhauser-us:
For the hypervisor extension, change the names of S-mode and U-mode when V = 1 to VS-mode and VU-mode, as agreed earlier.


aswaterman:
Looks good, thanks.

PR #208, opened 2018-07-05
Memory model clarifications

daniellustig:
A few more small textual changes:

1) Address some feedback from Ken Dockser (thanks Ken!)
2) Reorganize Ztso a bit, to clarify what is commentary and what has actual force


src/rvwmo.tex
```diff
@@ -37,10 +37,14 @@ \subsection*{Memory Model Primitives}
 A memory operation can be either a {\em load operation}, a {\em store operation}, or both simultaneously.
 All memory operations are single-copy atomic: they can never be observed in a partially-complete state.
 
-Each aligned memory instruction other than an unsuccessful SC instruction gives rise to exactly one memory operation.
+Among instructions in RV32GC and RV64GC, each aligned memory instruction other than an unsuccessful SC instruction gives rise to exactly one memory operation.
```
aswaterman: I believe this is true for RV32IMAFC and RV64GC, but not RV32D: https://github.com/riscv/riscv-isa-manual/blob/master/src/d.tex#L150

daniellustig: Thanks for pointing that out.  Let me make a quick correction commit then.

daniellustig: If XLEN==32, are FLD and FST intended to at least be atomic at 32b granularity?  Or can they be performed even at byte granularity, as for misaligned loads and stores?

aswaterman: Good question.  I can rationalize either choice, but the latter is compatible with the spec as written, whereas the former would add a new constraint.

The latter choice should also simplify the memory model text: loads/stores of >XLEN bits have the same properties as misaligned loads/stores.

daniellustig:
OK, I made the correction regarding FLD and FST.

Note that I also added one more unrelated memory model change as well: I versioned the appendices and noted that they should be treated as commentary.

aswaterman:
Good call on the appendices.

aswaterman:
Merged in, along with a commit fixing FST to FSD (the 'D' in 'FLD' is for 'double precision', not part of a contraction of 'LoaD').

Issue #145, opened 2018-03-15
Preferred form of jalr


asb:
The ISA manual consistently uses `jalr $rd, $rs1, $imm` for the jalr instruction. You might reasonably assume this is the preferred form, but it seems the current binutils prefers to disassemble to an alternative format:`jalr $rd, $imm($rs1)`. Is this a spec bug or a binutils bug?

PR #209, opened 2018-07-06
Make JALR assembly format consistent with binutils

aswaterman:
Addresses #145


aswaterman:
@asb did I miss anything?

PR #215, opened 2018-08-01
Un-mix trap and exception in {m,s}epc

palmer-dabbelt:
This might be a bit pedantic, but the current wording in the spec mixes
up traps and execptions here in a way that makes it sound like
interrupts don't write {m,s}epc if you're reading quickly.

I don't really like the wording I've proposed, so this is more of a bug
report than a fix.

### DIFF
--- a/src/machine.tex
+++ b/src/machine.tex
@@ -1723,8 +1723,9 @@ addresses.  Implementations may convert some invalid address patterns into
 other invalid addresses prior to writing them to {\tt mepc}.
 
 When a trap is taken into M-mode, {\tt mepc} is written with the virtual
-address of the instruction that encountered the exception.  Otherwise,
-{\tt mepc} is never written by the implementation, though it may be
+address of the instruction that encountered the exception or the virtual
+address of the next instruction to be executed in the interrupted context.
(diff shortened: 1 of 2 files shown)


src/machine.tex
```diff
@@ -1723,8 +1723,9 @@ \subsection{Machine Exception Program Counter ({\tt mepc})}
 other invalid addresses prior to writing them to {\tt mepc}.
 
 When a trap is taken into M-mode, {\tt mepc} is written with the virtual
-address of the instruction that encountered the exception.  Otherwise,
-{\tt mepc} is never written by the implementation, though it may be
+address of the instruction that encountered the exception or the virtual
+address of the next instruction to be executed in the interrupted context.
```
aswaterman: @kasanovic should verify, but I think with the new proposed terminology, it would suffice to use the existing text, but replace "exception" with "trap."

kasanovic:
The intro text was changed to make clear that exception or interrupt can cause a trap.  Text regarding mepc and sepc was changed to indicate both are written with PC of trapping instruction regardless of cause of trap.

PR #223, opened 2018-08-30
Remove text stating C.ADDI4SPN is RV32C/RV64C-only

T-J-Teru:
The paragraph of text describing C.ADDI4SPN states the instruction is
RV32C/RV64C-only, but the opcode map indicates the instruction is
available for RV128C too.  This change updates the descriptive text,
assuming that the opcode map is correct.


aswaterman:
Yes, this looks like an editing error.  (Note RV128C isn't yet frozen, so anything could happen, but for the time being C.ADDI4SPN is part of it.)

PR #244, opened 2018-10-19
adding counter overflow interrupt facility

allenjbaum:
This adds a counter overflow facility for all counters. The timer is grandfathered in as it effectively already has an overflow interrupt. This piggybacks on that interrupt cause, with new CSR that identifies which counter(s) overflowed. It is completely backwards compatible with existing implementations.

### DIFF
--- a/src/machine.tex
+++ b/src/machine.tex
@@ -1681,6 +1681,91 @@ the memory-mapped {\tt mtime} register, or hard-wire the TM bit in
 and emulate this functionality in M-mode software.
 \end{commentary}
 
+\subsection{Counter Interrupts ({\tt mtip})}
+\label{sec:mcip}
+
+\begin{figure*}[h!]
+{\footnotesize
+\begin{center}
+\setlength{\tabcolsep}{4pt}
+\begin{tabular}{cccMcccccc}
+\instbit{31} &
+\instbit{30} &
+\instbit{29} &
+\instbitrange{28}{6} &
+\instbit{5} &
+\instbit{4} &
+\instbit{3} &
+\instbit{2} &
+\instbit{1} &
+\instbit{0} \\
+\hline
+\multicolumn{1}{|c|}{HPM31} &
+\multicolumn{1}{c|}{HPM30} &
+\multicolumn{1}{c|}{HPM29} &
+\multicolumn{1}{c|}{...} &
+\multicolumn{1}{c|}{HPM5} &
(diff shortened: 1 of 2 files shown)


allenjbaum:
There is some rush here however, as we intend to ship something that may
have a conflicting definition... so if you have a different idea, please
lay it out (much) sooner rather than later.

aswaterman:
If you’re planning to ship it soon, I strongly advise making it an M-mode-only feature. It’s virtually certain to be at least slightly different than what gets ratified, but if that nonconformance is isolated to M-mode, then standard software is a lot less likely to notice/care.

kersten1:
@allenjbaum Not sure what needs to happen with this one. Obviously, it can't go forward as is because we are using adoc now. There is an issue attached to this PR that is at least from 2021. Can we port the basic info to this issue and close this PR?

gfavor:
What is the issue that remains valid and  given the counter overflow extensions that have been done since way back (both Sscofpmf and the more recent extensoin by Beeman which I off-hand forget the name of)?

allenjbaum:
It's ratified now. 
Any concerns we had are either resolved or moot.

kersten1:
So maybe we just close this PR?

sorear:
The only part of this that can't currently be done in a standardized way is mhpmevent access to instret and cycle, c.f. https://github.com/riscv/riscv-smcntrpmf/issues/1#issuecomment-1637036334

gfavor:
It's unclear what is the remaining issue preventing closure of this PR?  The instret and cycle counters don't have associated event CSRs since they are fixed event counters, Smcntrpmf  provides mode-based filtering for these two counters, and there has not been a justified intent to try and add overflow interrupts to these counters for a few reasons (that I don't remember off-hand), as well as the programmable counters can be used for this purpose if needed.

ved-rivos:
> and there has not been a justified intent to try and add overflow interrupts to these counters for a few reasons (that I don't remember off-hand)

Use of overflow interrupts typically involves setting these counters close to overflow such that the overflow interrupt can be used as a sampling event. With instret and cycles there were uses that sampled these counters periodically for purposes such as IPC estimation and hence these could not be used in this mode for sampling. The equivalent events can be counted on the programmable counters for the sample-on-overflow uses. I believe this was the reason they were excluded.

kersten1:
Based on comments, I'm closing this one. If there is something still relevant, please reopen and I'll get the info moved ASAP.

PR #245, opened 2018-10-22
Clarification on the meaning of the annotations

kdockser:
As discussed in the weekly memory model meeting, this text clarifies that it is not necessary (and perhaps not desirable) to research the common definitions of these memory ordering annotation terms - in the context of RISC-V their semantics are  fully defined in this chapter.

### DIFF
--- a/src/rvwmo.tex
+++ b/src/rvwmo.tex
@@ -72,7 +72,7 @@ An ``RCsc annotation'' refers to an acquire-RCsc annotation or a release-RCsc an
 \begin{commentary}
   In the memory model literature, the term ``RCpc'' stands for release consistency with processor-consistent synchronization operations, and the term ``RCsc'' stands for release consistency with sequentially-consistent synchronization operations~\cite{Gharachorloo90memoryconsistency}.
   
-While there are many different informal defintions of these annotations, in the context of the RISC-V RVWMO (and by extension RVTSO) these terms are concisely and completely defined axiomatixally by Preserved Program Order rules 4, 5 and 6 on page ##.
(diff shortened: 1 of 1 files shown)


src/rvwmo.tex
```diff
@@ -71,6 +71,8 @@ \subsection*{Memory Model Primitives}
 
 \begin{commentary}
   In the memory model literature, the term ``RCpc'' stands for release consistency with processor-consistent synchronization operations, and the term ``RCsc'' stands for release consistency with sequentially-consistent synchronization operations~\cite{Gharachorloo90memoryconsistency}.
+  
+While there are many different informal defintions of these annotations, in the context of the RISC-V RVWMO (and by extension RVTSO) these terms are concisely and completely defined axiomatixally by Preserved Program Order rules 4, 5 and 6 on page ##.
```
aswaterman: Can you add a section link or something instead of “page ##”?

tommythorn: "defintions" should probably be definitions

daniellustig:
I think it's a helpful note to add in general, but I'd suggest the following subtle tweaks:

While there are many different definitions for acquire and release annotations in the literature, in the context of RVWMO these terms are concisely and completely defined by Preserved Program Order rules \ref{ppo:acquire}--\ref{ppo:rcsc}.

- delete "informal": formal definitions differ too
- "these" -> "acquire and release" (if it's going to be a separate paragraph)
- "the RISC-V RVWMO" -> "RVWMO"
- let the RVTSO chapter handle RVTSO.  RVTSO is not otherwise defined in this RVWMO chapter
- fix typo in "axiomatixally" by just deleting the word

aswaterman:
Thanks, merged (with slight tweak) via 8e7b4b43ea05d99c1b56b17f7f5fdbba5c610f04

PR #253, opened 2018-11-07
Update marchid.md for VectorBlox ORCA

vbx-glemieux:


aswaterman:
The Syntacore folks already claimed that ID.  Can you update the PR to choose another?

vbx-glemieux:
I filed this change request before Syntacore. Shouldn't it be accepted first?

aswaterman:
Sorry I didn't drain my inbox in order.  I didn't think this would be a political issue.

Perhaps we can agree that 8 is luckier than 7?

aswaterman:
The resolution was that we could not agree that 8 is luckier than 7

PR #262, opened 2018-11-08
Change funct to funct2 in the CA format

luismarques:
In the RVC instruction formats figure and the respective subsection, the CA format row is changed such that the funct field is renamed to funct2, for consistency with the base ISA (which distinguishes between funct7 and funct3).

In the RVC instruction listings, the instructions with the CA format have the table formatting updated to match the new CA format, which was introduced in the commit 00557c314bd288c6eb511a201ce90bbdd4450d19.

### DIFF
--- a/src/c.tex
+++ b/src/c.tex
@@ -309,7 +309,7 @@ CS & Store &
 CA & Arithmetic &
 \multicolumn{6}{|c|}{funct6} &
 \multicolumn{3}{c|}{rd$'$/rs1$'$} &
-\multicolumn{2}{c|}{funct} &
+\multicolumn{2}{c|}{funct2} &
 \multicolumn{3}{c|}{rs2$'$} &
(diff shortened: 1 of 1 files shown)


aswaterman:
IMO, the encoding table is easier to read when the funct6 field remains split in two, even though it's inconsistent with the CA format diagram.  Note we made the same tradeoff for the CR format (e.g. C.MV).

luismarques:
@aswaterman So I suppose you're OK with the part of the pull request that changes `funct` to `funct2`? Also, you say "remains split in two", but in the original table the `funct6` was split into 3 subfields. Are you proposing to leave `funct6` as is, split into 3 fields, or changing my commit to split it into two fields (3+3 bits)?

In any case, I have to wonder: what's the point of defining a new instruction format if it's then not actually used in the tables? It's not like the hardware cares either way.

PR #278, opened 2018-11-21
Fix colliding latex labels.

pmundkur:
needed to fix only one label reference to mcsrnames.

### DIFF
--- a/src/instr-table.tex
+++ b/src/instr-table.tex
@@ -462,7 +462,7 @@
 \end{center}
 \end{small}
 
-\label{instr-table}
+\label{instr-table0}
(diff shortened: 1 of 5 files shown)


aswaterman:
the various instr-table.tex files are auto-generated.  But I guess I'm confused why we are even labeling them, because they are never referenced :) So I'll change the script to not label them.

pmundkur:
Ah, ok :)  You might want to retain the fix to the sec:bits label in v.tex though.

aswaterman:
Yep, I kept the non-auto-generated set of your changes.

PR #286, opened 2018-11-29
Extend mstatus.TW to U-mode for M/U systems

aswaterman:


zhonghochen:
Hi,

Is the Rocket-Chip implementation compatible with the spec? 
val allow_wfi = Bool(!usingVM) || reg_mstatus.prv > PRV.S || !reg_mstatus.tw

When S-Mode is implemented in Rocket-Chip and TW=0, executing WFI in U-mode does cause an illegal instruction exception.

Thanks,
Zhong-Ho

aswaterman:
Rocket doesn’t implement this new feature yet. We might or might not add the feature in the future.

We were careful to write the spec so that existing implementations, which ban WFI in U-mode, are still valid. The spec says that when TW=0, WFI is valid in U-mode unless prevented for some other reason. In Rocket’s case, the “some other reason” is that WFI is never allowed in U-mode.

aswaterman:
The spec is compatible with v1.10 in both cases.

When S-mode is not implemented, WFI in U-mode *must* raise an exception if mstatus.TW=1 and if the WFI does not complete in a bounded amount of time.  If mstatus.TW=0, WFI in U-mode *might* raise an exception if it does not complete in a bounded amount of time. 
 (This vague specification maintains compatibility with v1.10.)

When S-mode is implemented, WFI in U-mode *must* raise an exception if it does not complete within a bounded amount of time.  (This is the same as v1.10.)

Note that, in all of these cases, we never say "WFI *must not* raise an exception."  It's always valid for instructions to be unimplemented and emulated in M-mode.

PR #301, opened 2018-12-10
Double-Precision Floating-Point Move Instructions proposal for XLEN=32.

dp-sc:
We would like to propose to define the FMVH.X.W instruction (binary alias of the FMV.X.D) and FMV.D.X instructions for RV32IMFD architectures

### DIFF
--- a/src/d.tex
+++ b/src/d.tex
@@ -354,27 +354,79 @@ FMV.D.X & D & 0    & src  & 000  & dest & OP-FP  \\
 \end{tabular}
 \end{center}
 
-\begin{commentary}
-  Early versions of the RISC-V ISA had additional instructions to
-  allow RV32 systems to transfer between the upper and lower portions
-  of a 64-bit floating-point register and an integer register.
-  However, these would be the only instructions with partial register
-  writes and would add complexity in implementations with recoded
-  floating-point or register renaming, requiring a pipeline read-modify-write
-  sequence.  Scaling up to handling quad-precision for RV32 and RV64
-  would also require additional instructions if they were to follow
-  this pattern.  The ISA was defined to reduce the number of explicit
-  int-float register moves, by having conversions and comparisons
-  write results to the appropriate register file, so we expect the
-  benefit of these instructions to be lower than for other ISAs.
+For XLEN$=$32 there are the following instructions to move bit patterns
+between the floating-point and integer registers.
 
-  We note that for systems that implement a 64-bit floating-point unit
-  including fused multiply-add support and 64-bit floating-point loads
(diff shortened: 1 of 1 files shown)


aswaterman:
If these instructions are ultimately added, they should be considered part of new extensions (Zdmv for double on RV32, Zqmv for quad on RV64), since the base D/Q extensions are already frozen.  I'm including the "Base ISA Ratification" tag for issue-tracking purposes.

dp-sc:
Hi Andrew, thank you for the response, but we respectfully disagree this should necessary belong to the new extension, because:

1.	First, we were under impression D and Q extensions are being ratified now as part of the Base ISA
2.	The nature of the proposal is purely additive, - it defines useful behavior in the currently undefined space
3.	We can see other cases when significant, in our opinion, additive changes are being applied as part of this ratification process:
(i.e. https://github.com/riscv/riscv-isa-manual/commit/54498519cd8bcc28cbc79654eb49137594a0e943)

So, if you maintain the proposal cannot be applied this time, could you please clarify, what are criteria to choose to right place for the changes, i.e. next revision of the existing spec (2.3 or 2.4) vs new extension. 

Thank you, Dmitri

michael-roe:
"Me too" --- I would like to have an FMVH instruction.

Is there a draft online of the proposed floating point extensions?

Is there a recommended opcode from the implenentation-specific opcoide block that we could use for FMVH in the meantime?

jiangshuaili:
As I mentioned at https://github.com/riscv/riscv-openocd/issues/546,  FMVH is really needed by a rv32 with 64 bit float debug. The sooner this Instruction is defined, the better. Is it advancing?

jiangshuaili:
FMVH may not be as important in floating-point programming, but it is necessary in debugging interface, so that the user cannot use the memory resources on the development board because of debugging. (If the RV32 CPU does not support FMVH, then it can only read and write with instructions that interact with MEM, so that the debugger needs to use a memory address.)

OpenOCD requires users to specify an address for debugging in order to read the floating-point register. Although the scheme is feasible, it increases the burden of debugging developers and also brings the scenario where the memory on the development board is not fully available for the program developers.

If user B debugs user A's development board and programs, and user A does not reserve an unused memory for user B, then RV32's CPU reading the 32-bits high value of the double-precision floating-point register will cause more tedious operation for user B.User B may need to figure out a available memory and then manually manipulate the DM to complete the high 32-bits read and write of a double-precision floating-point number.

kersten1:
Did this work happen in a different PR? I can convert changes to asciidoc and put in another PR if needed to advance this one.

aswaterman:
@kersten1 This work already happened as part of the Zfa extension; we forgot to close this PR.

PR #303, opened 2018-12-11
added passage explaining conditional branches and misaligned accesses

benjaminselfridge:
There is currently no explanation, within the Conditional Branches section, of how branches to misaligned addresses work in the absence of C. My understanding is that if the branch is taken, a misaligned exception should be raised, but if it is not taken, then an exception is not raised.

I have added two brief paragraphs explaining this; it could be merged now if it is deemed correct. If it is not correct, can someone correct it and add the correct explanation?

### DIFF
--- a/src/rv32.tex
+++ b/src/rv32.tex
@@ -946,6 +946,19 @@ allowing the branch and following instructions to be executed
 speculatively and out-of-order with respect to other code~\cite{ibmpower7}.
 \end{commentary}
 
+All conditional branch instructions will generate an
+instruction-address-misaligned exception if the target address is not
+aligned to a four-byte boundary, whenever the result of executing the
+instruction would be to take the jump (i.e. if the branch condition is
+true). If the branch condition is false, a misaligned address
+exception will not be raised.
+
+\begin{commentary}
+Instruction-address-misaligned exceptions are not possible on machines
+that support extensions with 16-bit aligned instructions, such as the
+compressed instruction-set extension, C.
+\end{commentary}
+
 \section{Load and Store Instructions}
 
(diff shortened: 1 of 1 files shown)


benjaminselfridge:
I have no idea why the CI build failed, it works fine on my system and all I did was add a couple paragraphs.

aswaterman:
There is already some text to this effect:

    \section{Base Instruction Formats}
    
    In the base RV32I ISA, there are four core instruction formats
    (R/I/S/U), as shown in Figure~\ref{fig:baseinstformats}.  All are a
    fixed 32 bits in length and must be aligned on a four-byte boundary in
    memory.  An instruction-address-misaligned exception is generated on a
    taken branch or unconditional jump if the target address is not
    four-byte aligned.  This exception is reported on the branch or jump
    instruction, not on the target instruction.  No
    instruction-address-misaligned exception is generated for a
    conditional branch that is not taken.

benjaminselfridge:
Yes, which is how I discovered it; I am proposing to add this text somewhere in the Conditional Branch section itself, as it is currently included in the unconditional jump section, but not in the conditional branch section (which caused me, initially, to believe that conditional branches never raised misaligned exceptions).

aswaterman:
Ah, I didn't realize we explicitly said this for unconditional jumps but not conditional branches.  That asymmetry is confusing.

benjaminselfridge:
Yeah; I suppose you could either add some text, or delete the text in the unconditional branch section.

aswaterman:
I'm going to err on the side of redundancy in this case, and merge your PR with slightly tweaked wording.

PR #309, opened 2018-12-20
Suggest gp as the base for a millicode library

brucehoult:
This is likely to be available on more systems. For example HiFive1 :-)

On Linux you could potentially put an execute-only page at address 0, but I think it's better to put an execute-only page with millicode routines (.stext?) at the TOP of the page (the -msave-restore routines are currently 96 bytes), followed by R/W page(s) with .sdata, then .data and point GP at 2048 bytes above the start of the millicode.

### DIFF
--- a/src/rv32.tex
+++ b/src/rv32.tex
@@ -769,7 +769,9 @@ address will usually quickly raise an exception.
 When used with a base {\em rs1}$=${\tt x0}, JALR can be used to implement
 a single instruction subroutine call to the lowest \wunits{2}{KiB} or highest
 \wunits{2}{KiB} address region from anywhere in the address space, which could
-be used to implement fast calls to a small runtime library.
+be used to implement fast calls to a small runtime library. If the ABI in use
+has a Global Pointer register ({\em rs1}$=${\tt x3} in the standard calling
+convention) this could also be used.
 \end{commentary}
 
 The JAL and JALR instructions will generate an


aswaterman:
I concur there are circumstances where this makes sense, but I think it's actually too boutique to mention here, for a few reasons:

- In the current ABIs, gp points to the middle of a writable data section.  In systems with memory protection, you wouldn't want to intermix code and data.  This can be ameliorated with more ABI complexity, but...

- ...I don't want to give security-conscious readers the impression we don't know what we're doing.  To dispel that concern would require more descriptive complexity.

- In many situations, the millicode library can be within reach of JAL.  Other things being equal, it's better to favor JAL over JALR, since JAL is easier to accelerate in hardware.

- It's not clear that reducing the amount of gp displacement available for data addressing is a win.

brucehoult:
It's great if JAL will reach, but binaries over 1 MB are not uncommon and it's a real tragedy if you get an AUIPC/JALR instead, in almost every function. You could try to sprinkle multiple copies of these 96 bytes every 2 MB through the program -- that's trivial expansion compared to the several percent savings from using them. But it seems like a big job to get the tools to do this.

I covered how to handle the X^W issue in my initial comment here. That could be moved into the actual document to show we *so* know what we're doing.

If it's only the save-restore millicode then gp data addressing is reduced to 97.65% of what it is now. I would be *very* surprised if bytes 4000-4095 of .data are statically or dynamically directly addressed more often than the save-restore code.

By the way, I've found that on HiFive1 CoreMark with -msave-restore runs 0.4% *FASTER* as well as being 0.35% smaller.

aswaterman:
I'll reword in a way that avoids most of my concerns.

brucehoult:
Sadly, using c.jal and sprinkling copies every 4 KB in a binary costs 2.3% overhead .. which is probably as much as the space savings in the functions. And only works in 32 bit.

Sprinkling plain J instructions to four save and four restore functions would be 32 bytes instead of 96. A saving, but at even more time overhead.

brucehoult:
Cheers.

PR #310, opened 2018-12-20
Update commentary to reflect MIPS r6 conditional branches

brucehoult:


aswaterman:
Good suggestion.  Merging with a slight tweak: I'm going to keep saying "MIPS" instead of "MIPS pre-R6", because that's the convention in the rest of the document and I don't want to add clutter.

aswaterman:
Merged manually via 87e4ae880bb3169a1128ad9fa2bcdac91a1219e4

PR #312, opened 2018-12-20
Add commentary on sufficiency of rv128

brucehoult:
A little bit whimsical :-)

Could add comment that even quite sparse usage will not exhaust it quickly.

Calculation (Python):

    ((2**128 # bytes
    *8 # bits
    /6.022e23 # moles
    *12 # grams
    /3.51 # CCs
    /(4.0/3*3.1416*0.5**3) # volume of a sphere inscribed in a unit cube
    )**(1.0/3) # cm
    /100 # m
    /1609.344) #miles

### DIFF
--- a/src/rv128.tex
+++ b/src/rv128.tex
@@ -33,6 +33,10 @@ segmentation, 96-bit address spaces, and software workarounds, until,
 finally, flat 128-bit address spaces will be adopted as the simplest
 and best solution.
 
+If you could somehow store one bit of information in every atom of a
+diamond, storing $2^{128}$ bytes would require a diamond almost exactly
+two miles in diameter --- 128 bits ought to be enough for anybody.
+
 We have not frozen the RV128 spec at this time, as there might be need
 to evolve the design based on actual usage of 128-bit address spaces.
 \end{commentary}


jim-wilson:
2 miles?  I'd prefer 3.2km in a technical document.

brucehoult:
That would certainly be more natural for me, but I'm not American :-) Or "over 3000m"

aswaterman:
It is a little too glib of a remark.   Because one application of the extra bits is tagging, rather than addressing, it's not abundantly clear that it will be enough for any purpose.

brucehoult:
If you use 32 bits (!!) for tags or sparse address space, 96 bits still needs a diamond 1.9m in diameter under the same assumptions.

OK, so it could be a bit less whimsical and/or glib (if you mean the reference to Bill Gates' 640 KB remark), but I think there should be some kind of words about whether and why 128 bits is probably enough for a long long time.

256 bits would need a diamond 2.2 light years in diameter, .

kersten1:
Why do I suddenly want to use a Twinkie? (Ghostbusters reference)

In all seriousness, what can we do here? Could go with a bit/maximum integer table, but that doesn't really express the size, which is what I think you are getting at here?

Don't want to link to Wikipedia, but is there an idea in [here](https://en.wikipedia.org/wiki/128-bit_computing) that we can grab? (nice call out to RISC-V as well).

Or we could do a "wanna learn more about 128 bit, see .... )  But I'm not sure what is a good, "official" location to link to. 

Else we can just close this PR.

@brucehoult thoughts? I'm trying to clear out zombie PRs.


| Bit | Maximum Integer |
| -- | -- |
| 16-bit | 65,535 |
| 32-bit | 4,294,967,295 |
| 64-bit | 18,446,744,073,709,551,615 |
| 128-bit | 340,282,366,920,938,463,463,374,607,431,768,211,455 |

aswaterman:
The way things are trending, I think we'll need to significantly rework the RV128 proposal.  In the meantime, it is causing some confusion and clutter in the specs.  So I think we might end up excising it altogether, then reviving it later.  With that in mind, we can probably close the RV128-centric PRs, as fun as they are.

kersten1:
Closing due to other rv128 work being discussed.

PR #315, opened 2018-12-20
Clarify use of C instructions between LR and SC

brucehoult:
Is it true? If not then substitute "but not" for "or"


aswaterman:
Yes, the C extension chapter covers it, but I think it’s better compartmentalized how it is now. (Just my opinion.)

PR #334, opened 2019-01-29
Fix a couple of typos and inconsistencies.

pmundkur:
Perhaps its time to use \newcommand to avoid this fragility.


aswaterman:
Thanks.  I'd happily merge a PR that macro-ized some of this stuff, but don't have the time to do the heavy lifting myself.

PR #336, opened 2019-02-01
Make the mcause table easier to find.

pmundkur:
It is forward referenced a couple of times around 10 pages before it actually appears.  Added a pageref for specifically this table, and add figures and tables to ToC for other cases.


aswaterman:
The list of tables/figures looks kind of bad because we didn't caption things appropriately for that.  But the point about the mcause table is well taken so I'll pluck that part of this PR.  Thanks!

aswaterman:
Closed by a4d773f29aa5e623f712d41f9068e11270d51d80

PR #343, opened 2019-02-24
Reword FNMADD.S and FNMSUB.S to be clearer

brucehoult:


brucehoult:
OK, if we can't change the equations, can we at least keep the paragraphing and grouping fnmadd with fmadd instead of last?

Maybe the final "in other words" can say "almost the negation of X, except for the effects of signed zeros and directed rounding"

aswaterman:
(Posting this for posterity:) While this description is certainly clearer, it's unfortunately the case that -x*y+z isn't equal to -(x*y-z) in general.  For example,

    printf("%g\n", fma(-0., 0., 0.));
    printf("%g\n", -fma(0., 0., -0.));

will print 

    0
    -0

brucehoult:
I see that the ARMv8 manual describes everything as using A * B + C with A and C conditionally pre-negated:
neither for FMADD i.e. A * B + C
both for FNMADD i.e. -A * B + -C
C for FMSUB i.e. A * B + -C
A for FNMSUB. i.e. -A * B + C

Presumably this formulation gives the correct behaviour with signed zeroes and directed rounding.

aswaterman:
Yeah, our definitions match theirs, except for the naming of FNMADD and FNMSUB being swapped.

I agree splitting into paragraphs might help; I'll take care of that.

aswaterman:
https://github.com/riscv/riscv-isa-manual/commit/cbc2663ef7df1b2bcda627740d94f3e95adbc141

PR #361, opened 2019-04-05
Privileged Spec: Add dscratch0/1 to CSR listing

imphil:
dscratch1 was missing from the listing, dscratch0 was named only dscratch.


aswaterman:
Thanks.

PR #370, opened 2019-04-17
Finesse ligatures to work with Adobe Acrobat Reader search and cut-and-paste

pdonahue-ventana:
Adobe Acrobat Reader DC 19.010.20099 for Windows doesn't work well with the ligatures found in the RISC-V specs.  Searching for words like "defined" doesn't work due to the "fi" ligature.  Copying "defined" to the clipboard will drop the "fi" and only copy "dened" (or there's possibly an unprintable character there).

Other PDF viewers seem to have better behavior than the Adobe product (which is odd since Adobe invented PDF).  However, enough people use Acrobat that it ought to be supported if possible.

I found a fix for this situation on:
https://superuser.com/questions/165073/is-it-possible-to-remove-ligatures-from-copied-text
Ligatures appear to be rendered the same after the fix, so there's no visual difference.  I tried viewing and searching using evince on Linux and it works as before.  But searching and clipboard operations on Adobe Acrobat Reader now work as expected with this change.

### DIFF
--- a/src/riscv-privileged.tex
+++ b/src/riscv-privileged.tex
@@ -4,6 +4,9 @@
 
 \documentclass[twoside,11pt]{book}
 
+\input{glyphtounicode.tex}
+\pdfgentounicode=1 %
+
 \input{preamble}
 
 \newcommand{\privrev}{\mbox{20190405-Priv-MSU-Ratification}}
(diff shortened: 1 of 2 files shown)


aswaterman:
What a curious bug.  We use Apple Preview so haven't stumbled across this.  Thanks for the fix.

aswaterman:
This was merged in commit 3392205530510f02a2d8197a9969a525b6359f7d, but I guess since I modified the patch to include a commit, GitHub didn't automatically close this PR.

PR #374, opened 2019-04-22
Add coherency bit in PTE attributes

guoren83:
The bit designates a coherent memory model for this page mapping.
Software set the bit to tell the hardware that the region of the
page's memory area must be coherency with IOs and dma devices in
SOC system by PMA settings.

Signed-off-by: Guo Ren <ren_guo@c-sky.com>

### DIFF
--- a/src/supervisor.tex
+++ b/src/supervisor.tex
@@ -1153,10 +1153,11 @@ before being directly converted to machine-level physical addresses.
 \begin{figure*}[h!]
 {\footnotesize
 \begin{center}
-\begin{tabular}{@{}E@{}O@{}Fcccccccc}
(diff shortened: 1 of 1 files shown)


aswaterman:
In RISC-V, these properties are controlled by physical-memory attributes, not by virtual-memory properties, which are used only for translation and protection purposes.

guoren83:
See:
https://lore.kernel.org/linux-riscv/1555947870-23014-1-git-send-email-guoren@kernel.org/T/#u

Our goal is to support non-coherency memory model in riscv linux, this pull request just a part. I also Cc your email, let's discuss in that mail thread.

PR #375, opened 2019-04-24
Two separate changes

vmurali:
1) Specifiying the sign of the remainder in signed division
2) Explaining the effect of reducing MXLEN on integer register values, PC and CSRs.

### DIFF
--- a/src/m.tex
+++ b/src/m.tex
@@ -99,7 +99,20 @@ MULDIV & divisor & dividend & DIV[U]W/REM[U]W & dest & OP-32 \\
 DIV and DIVU perform an XLEN bits by XLEN bits signed and unsigned integer
 division of {\em rs1} by {\em rs2}, rounding towards zero.
 REM and REMU provide the remainder of the
-corresponding division operation.  If both the quotient and remainder
+corresponding division operation.
+For signed division, the sign of the remainder and the value of the quotient
+follow the usual conventions: the sign of the remainder is the sign of the
+dividend and the absolute value of the quotient is the same as the quotient for
+the unsigned division of the absolute value of dividend by the absolute value
+of the divisor.
+
+\begin{commentary}
+This ensures that $\textrm{dividend} = \textrm{divisor} \times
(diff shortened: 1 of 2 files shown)


src/m.tex
```diff
@@ -99,7 +99,20 @@ \section{Division Operations}
 DIV and DIVU perform an XLEN bits by XLEN bits signed and unsigned integer
 division of {\em rs1} by {\em rs2}, rounding towards zero.
 REM and REMU provide the remainder of the
-corresponding division operation.  If both the quotient and remainder
+corresponding division operation.
+For signed division, the sign of the remainder and the value of the quotient
+follow the usual conventions: the sign of the remainder is the sign of the
+dividend and the absolute value of the quotient is the same as the quotient for
```
aswaterman: I believe the remark about the absolute value of the quotient is redundant with the previous sentence ("rounding towards zero").  Agreed?

aswaterman: Could move it to the commentary you added below (e.g. "Rounding towards zero implies that the abs val of the quotient is the same as [...].  In combination with the sign of the remainder, this ensures that [...]")

src/machine.tex
```diff
@@ -88,6 +88,19 @@ \subsection{Machine ISA Register {\tt misa}}
 zero after two shifts, then the machine is RV64, else RV128.
 \end{commentary}
 
+\begin{commentary}
```
aswaterman: Did you determine that a commentary block is sufficient for this?  i.e. do we not need normative text explaining MXLEN's effect on the PC?

src/m.tex
aswaterman: I convinced myself that the quotient definition follows from the rounding scheme.  I'm committing a change that includes the sign-of-REM clarification and the commentary below.

src/machine.tex
aswaterman: I think we need new normative text about the PC.  Working out the language with JohnH.

vmurali:
Sorry 😂 I will check the commit when I get a chance this weekend

Issue #377, opened 2019-04-29
Proposal to explicitly declare things as unspecified


pdonahue-ventana:
The recent discussion of FENCE.I behavior [1] brings up a larger concern I have.  There are several cases where it is clear what is expected of software but unclear how much latitude implementations have.  Software should avoid FENCE.I unless it's supported but there was confusion about hardware requirements if a FENCE.I is attempted by software. Atomicity of instruction fetches[2] is also not considered in the base architecture.

It is not obvious if a lack of clarity in any particular case is an attempt to unconstrain hardware or if it is an oversight.  It would be beneficial to all involved if the architecture strived to specify behavior in all cases, even if the specification allows complete freedom.  For instance, other architectures have terms such as:
- MIPS has "UNPREDICTABLE" results and operations.  It has "UNDEFINED" operations and behavior. It has "UNSTABLE" results and values.
- Alpha has "UNPREDICTABLE" results and occurrences and "UNDEFINED" operations.
- ARM has "UNPREDICTABLE" and "IMPLEMENTATION DEFINED" behavior.  It has "UNKNOWN" and "Architecturally UNKNOWN" values.
- Power has “boundedly undefined" behaviors and "undefined value" values.

As a straw man, I propose:
- The term "UNSPECIFIED" should be defined in the introduction to apply to a behavior or value that the architecture intentionally leaves open to extensions, platform standards, or implementations.
- Extensions or platform standards may provide normative content to further constrain "UNSPECIFIED" cases or they may not.  That decision is beyond the scope of the architecture.
- All behavior and values must be defined either concretely or using the term "UNSPECIFIED."  This is the key to providing clarity.
- This term is available to and should be used by extensions for behavior that they want to delegate to other extensions or to platform standards or implementations.  This is the key to providing clarity in extension documentation.

For instance:
- To address the FENCE.I issue, the behavior when decoding a reserved instruction is UNSPECIFIED.
- The atomicity of instruction fetches is UNSPECIFIED.
- On reset, "All other hart state is unspecified" becomes "UNSPECIFIED" to explicitly allow extensions, platform standards, or implementations to specify it (which was likely the goal anyway).
- "The number of supervisor physical address bits is implementation-defined" becomes "UNSPECIFIED" to allow requirements to be introduced by extensions or platform standards, not just implementations.
[...]

PR #380, opened 2019-05-06
Proposal to explicitly declare things as UNSPECIFIED

pdonahue-ventana:
Here are my proposed changes in support of https://github.com/riscv/riscv-isa-manual/issues/377.

Note that many more things would presumably fall into this UNSPECIFIED category but this pull request contains the basic definition and a few concrete examples.  Documenting other behaviors and values as UNSPECIFIED can be done later as necessary.

### DIFF
--- a/src/intro.tex
+++ b/src/intro.tex
@@ -712,21 +712,21 @@ Invisible traps are, by their nature, out of scope for this document.
 Instruction encodings that are not defined here and not defined by
 some other means may cause a fatal trap.
 
-\section{Unspecified Behavior}
+\section{UNSPECIFIED Behaviors and Values}
 
 The architecture fully describes what implementations must do and any
 constraints on what they may do.  In cases where the architecture
-intentionally does not constrain implementations, the term UNSPECIFIED
+intentionally does not constrain implementations, the term \unspecified\ 
 is explicitly used.
 
-The term UNSPECIFIED refers to a behavior or value that is
+The term \unspecified\ refers to a behavior or value that is
 intentionally unconstrained.  The definition of these behaviors or
 values is open to extensions, platform standards, or implementations.
-Extensions or platform standards may provide normative content to
-further constrain cases that the base architecture defines as
-UNSPECIFIED.
+Extensions, platform standards, or implementation documentation may
+provide normative content to further constrain cases that the base
+architecture defines as \unspecified.
 
 Like the base architecture, extensions should fully describe allowable
-behavior and values and use the term UNSPECIFIED for cases that are
+behavior and values and use the term \unspecified\ for cases that are
 intentionally unconstrained.  These cases may be constrained or defined
 by other extensions, platform standards, or implementations.
--- a/src/intro.tex
+++ b/src/intro.tex
@@ -711,3 +711,22 @@ defined solely to cause requested traps are documented here.
 Invisible traps are, by their nature, out of scope for this document.
 Instruction encodings that are not defined here and not defined by
 some other means may cause a fatal trap.
+
+\section{Unspecified Behavior}
+
+The architecture fully describes what implementations must do and any
+constraints on what they may do.  In cases where the architecture
+intentionally does not constrain implementations, the term UNSPECIFIED
+is explicitly used.
+
+The term UNSPECIFIED refers to a behavior or value that is
+intentionally unconstrained.  The definition of these behaviors or
+values is open to extensions, platform standards, or implementations.
+Extensions or platform standards may provide normative content to
+further constrain cases that the base architecture defines as
+UNSPECIFIED.
+
+Like the base architecture, extensions should fully describe allowable
+behavior and values and use the term UNSPECIFIED for cases that are
+intentionally unconstrained.  These cases may be constrained or defined
+by other extensions, platform standards, or implementations.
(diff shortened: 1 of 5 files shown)


pdonahue-ventana:
Any thoughts or feedback on this?  Thanks.

aswaterman:
I agree with the sentiment, but I think a finer categorization might be appropriate.  For example, the Alpha spec distinguishes UNPREDICTABLE from UNDEFINED.  Their UNDEFINED is similar in some respects to your UNSPECIFIED, whereas their UNPREDICTABLE is very constrained (namely, only registers the instruction could ordinarily write are written, and memory isn't accessible to the current protection domain isn't accessed).   ARMv8 uses the term UNPREDICTABLE in a similar way as Alpha.

The remark about reserved opcodes is addressed to some extent by this paragraph: https://github.com/riscv/riscv-isa-manual/blob/master/src/intro.tex#L706

pdonahue-ventana:
> I agree with the sentiment, but I think a finer categorization might be appropriate. For example, the Alpha spec distinguishes UNPREDICTABLE from UNDEFINED. Their UNDEFINED is similar in some respects to your UNSPECIFIED, whereas their UNPREDICTABLE is very constrained (namely, only registers the instruction could ordinarily write are written, and memory isn't accessible to the current protection domain isn't accessed). ARMv8 uses the term UNPREDICTABLE in a similar way as Alpha.

I would be very happy with a finer categorization and my original inclination was to go with an Alpha-like approach (having worked on Alpha long ago).  By the time I wrote this proposal, I had decided to simplify to more closely approximate what I believe to be the current state of RISC-V where there aren't really any constraints.  In doing so, I hoped that it would be less controversial.

I believe that there ideally ought to be constraints that prevent security holes.  My current thinking is that those constraints belong in the privileged architecture (as part of the further constraint that UNSPECIFIED allows extensions to do) rather than in the unprivileged architecture since I think that an implementation that doesn't include the privileged architecture has no security and so can't have any security holes.  Perhaps hangs should be mentioned in the unprivileged architecture, though.

While UNSPECIFIED can more closely mirror UNPREDICTABLE (whether in the presence of the privileged architecture or not), I'm not sure there's a need for something like the UNDEFINED behavior of Alpha and MIPS.  ARM and Power don't seem to have such a notion, at least not explicitly.

> The remark about reserved opcodes is addressed to some extent by this paragraph: https://github.com/riscv/riscv-isa-manual/blob/master/src/intro.tex#L706 -- I guess I'm also worried about people reading "undefined opcodes are have unspecified behavior" as "these architects have no idea what they're doing," when in actuality, platform specifications are expected to nail down this behavior precisely.

In reality, undefined opcodes do have unspecified and completely unbounded behavior in the base architecture.  I think that it's better to say this explicitly rather than having half of the people assume (as I did initially) that reserved opcodes must trap since that's how "everyone else" does it and the other half (who come from more of a microcontroller background) maybe assuming something else.  Perhaps your concern would be addressed with additional non-normative text that says that platforms are expected to describe the behavior.  Overall, I think that having the rigor of UNSPECIFIED makes it less likely that people believe that the architects don't know what they're doing and non-normative text can further help bolster confidence in the architects.

The few uses of UNSPECIFIED I included were just some examples for illustrative purposes.  My main goal is to advocate for everything being fully described either precisely or using UNSPECIFIED.  If my reserved instruction verbiage is blocking things then I'd rather pull that out for a separate discussion at a later time and focus now on whether UNSPECIFIED is a good idea in general and what constraints it should have (such as security holes and hangs).

aswaterman:
Thanks for the additional insight, Paul.

I agree that what concretely constitutes a security hole belongs in the privileged architecture, but feel that the unprivileged architecture should still state that UNSPECIFIED behavior mustn't open security holes, in the abstract sense.

My current thinking is that everything that's UNSPECIFIED in the unprivileged architecture could instead be said to be PLATFORM-SPECIFIED.  The definition of the two terms might not be any different at this level, as ultimately, a given platform may leave the behavior relatively unconstrained.  But I think the alternate terminology better indicates the intent that the behavior be specified--just specified elsewhere.  Any thoughts on that?

pdonahue-ventana:
My understanding of the current state of affairs is that there is this hierarchy:
1. base architecture
2. extensions
3. platform standard*
4. implementation

Each level can constrain anything that is not specified by the levels above.  That's generally similar to what you seem to be saying but you have fewer levels.

I tried to capture this in the pull request:
> Extensions, platform standards, or implementation documentation may
> provide normative content to further constrain cases that the base
> architecture defines as \unspecified.

and:
> Like the base architecture, extensions should fully describe allowable
> behavior and values and use the term \unspecified\ for cases that are
> intentionally unconstrained.  These cases may be constrained or defined
> by other extensions, platform standards, or implementations.

However, the term PLATFORM-SPECIFIED implies:
1. extensions cannot constrain the behavior of things not specified in the base architecture
2. platforms such as the Unix platform can be expected to constrain all such behaviors

For those reasons, I believe that this term isn't quite right.  I'm open to other terms but cannot think of one that is as descriptive as UNSPECIFIED.

As for prohibiting security holes in all cases, I'm completely in favor of such restrictions.

** I believe that there may be some confusion about the dual use of the word "platform" in RISC-V to mean both:
1. a standard, such as the Unix platform, with which implementations may comply and
2. any such hardware implementations themselves, as defined in section 1.1 of the base architecture.

Issue #376, opened 2019-04-26
LR.D and SC.D aren’t mentioned in extension A


Columbus240:
Only an issue with the presentation, not with the actual content.
When the instructions LR and SC are introduced in chapter 8 about extension A, there is no mention made, that there are two forms for each. LR.W, LR.D and SC.W, SC.D. The text talks about LR and SC as if they were LR.W respectively SC.W. It uses the word “word (of memory)”, which got defined as 32 bits in section 1.4, slightly misleading the reader that there is only one instruction each, loading/storing only 32b values.
In Figure 8.1 the `.w` variants are used explicitly, and both variants of the instructions are included in the two instruction listings towards the end of the spec. They are also mentioned in Figure A.4. So it is possible to learn about there being two variants from the spec, just a little difficult.

For AMOs, the relevant text is at the end of the first paragraph of section 8.3. This might be copied and put in an appropriate place in section 8.2, and the names in the instruction format in section 8.2 should be changed from “LR” to “LR.W/D” and from “SC” to “SC.W/D” to clarify that two variants of each exist.
An idea how to mention both words and doublewords of memory in the text is, to first talk (explicitly) only about LR.W and SC.W and then say that the same applies analogously to LR.D and SC.D but with doublewords instead of words.

aswaterman:
Thanks again; resolved https://github.com/riscv/riscv-isa-manual/commit/169da704eb3eec1c729f98d11532a9b10b7fd158

PR #381, opened 2019-05-07
Clarify the behaviour of LR.W/D and SC.W/D

Columbus240:
Concerning issue #376.

I didn’t rewrap the text yet, so the diff is easier to read.


aswaterman:
Thank you for these clarifications.  I've manually merged the commit with a minor tweak.

Columbus240:
That’s how I expected you’d do it. Thanks for merging and rewording.

PR #382, opened 2019-05-09
Clarify reserved/HINT encodings in C chapter text

aswaterman:
This PR essentially copies the information from the encoding table at the
end of the chapter into the mainline text.  The intent is to remove any
doubt about what happens when an instruction's operand constraints are
not met.

### DIFF
--- a/src/c.tex
+++ b/src/c.tex
@@ -32,8 +32,6 @@ on any 16-bit boundary, i.e., IALIGN=16.  With the addition of the C
 extension, no instructions can raise instruction-address-misaligned
 exceptions.
 
-\pagebreak
-
 \begin{commentary}
 Removing the 32-bit alignment constraint on the original 32-bit
 instructions allows significantly greater code density.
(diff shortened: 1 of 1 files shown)


src/c.tex
```diff
@@ -761,15 +767,21 @@ \subsection*{Integer Constant-Generation Instructions}
 \end{tabular}
 \end{center}
 C.LI loads the sign-extended 6-bit immediate, {\em imm}, into
-register {\em rd}.  C.LI is only valid when {\em rd}$\neq${\tt x0}.
+register {\em rd}.
 C.LI expands into {\tt addi rd, x0, imm[5:0]}.
+C.LI is only valid when {\em rd}$\neq${\tt x0};
+the code points with {\em rd}={\tt x0} encode HINTs.
```
palmer-dabbelt: I guess I haven't actually read the ISA manual since hints were introduced, but I wouldn't say `C.LI is only valid when {\em rd}$\neq${\tt x0}` -- it's still valid, it's just a hint.  From reading the rest of this it looks like hints are being described as different instructions, so if that's the case then the wording is OK.

src/c.tex
```diff
@@ -871,12 +889,16 @@ \subsection*{Integer Register-Immediate Operations}
...
-  shamt[5:0]}, except for RV128C with {\tt shamt=0}, which expands to
-{\tt slli rd, rd, 64}.
+The shift amount is encoded in the {\em shamt} field.
+For RV128C, a shift amount of zero is used to encode a shift of 64.
+C.SLLI expands into {\tt slli rd, rd, shamt[5:0]}, except for
+RV128C with {\tt shamt=0}, which expands to {\tt slli rd, rd, 64}.
+
+For RV32C, {\em shamt[5]} must be zero; the code points with {\em shamt[5]}=1
+are reserved for custom extensions.  For RV32C and RV64C, the shift
+amount must be non-zero; the code points with {\em shamt}=0 are HINTs.  For
```
palmer-dabbelt: Here's another one.

aswaterman:
I vacillated on this matter, but I think it's best to describe HINTs as different instructions entirely, rather than as subcases of e.g. C.LI.  It's not that the instruction is invalid, it's just not a valid C.LI.

palmer-dabbelt:
> I vacillated on this matter, but I think it's best to describe HINTs as different instructions entirely, rather than as subcases of e.g. C.LI. It's not that the instruction is invalid, it's just not a valid C.LI.

I actually think that's the right way to go, it's just not how I remember these things being described.  Assuming the rest of the ISA manual matches then it's fine, but if I understand correctly that would extend past the C chapter.

aswaterman:
I think it will take some work to make the base-ISA chapters line up, but we can handle that separately.

PR #383, opened 2019-05-14
Fixed typo in 'nzimm' field of CSRR*I instructions

tovine:
In the RV32/64G Instruction Set Listings the immediate field of the `CSRRWI`, `CSRRSI` and `CSRRCI` instructions was called `zimm`, there was a missing `n` there - enforcing all-zero immediates wouldn't have made any sense... 🙃

### DIFF
--- a/src/instr-table.tex
+++ b/src/instr-table.tex
@@ -721,7 +721,7 @@
 
 &
 \multicolumn{6}{|c|}{csr} &
-\multicolumn{1}{c|}{zimm} &
+\multicolumn{1}{c|}{nzimm} &
 \multicolumn{1}{c|}{101} &
 \multicolumn{1}{c|}{rd} &
 \multicolumn{1}{c|}{1110011} & CSRRWI \\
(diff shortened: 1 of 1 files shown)


aswaterman:
We use `zimm` there as shorthand for `zero-extended immediate`, not `zero immediate`.  `nzimm` isn't quite right, as there's no constraint that the immediate must be nonzero.

tovine:
Huh... 🤔 
That was not what I expected, might I suggest using an abbreviation that is slightly less similar to nzimm or dropping a footnote to explain it?

aswaterman:
There are some things that match this pattern in the vector spec, so down the road we can see what they do, and consider changing this abbreviation to match

tovine:
But what about just using `uimm` there, like it's done in the chapter text (or `uimm[4:0]`, to follow the style of other instructions in the same table)?
I think that if something is listed as "unsigned immediate", then it is a given default that it will be zero-extended to the required width, don't you agree?

aswaterman:
I had forgotten the chapter text used uimm; that is preferable.  I will fix it.

aswaterman:
ba63c32418dcb3b72859957f83373b8a5644e169

PR #384, opened 2019-05-15
Correct iteration in LR/SC CAS example.

David-Horner:
Address argument was clobbered.

Signed-off-by: David Horner <ds2horner@gmail.com>


aswaterman:
oops, thanks.

PR #395, opened 2019-06-20
Add endianness control proposal to priv spec

aswaterman:
Contributed by @jhauser-us

### DIFF
--- a/src/supervisor.tex
+++ b/src/supervisor.tex
@@ -1,4 +1,4 @@
-\chapter{Supervisor-Level ISA, Version 1.11}
+\chapter{Supervisor-Level ISA, Version 1.12}
 \label{supervisor}
 
 This chapter describes the RISC-V supervisor-level architecture, which
@@ -46,7 +46,8 @@ register keeps track of the processor's current operating state.
 {\footnotesize
 \begin{center}
 \setlength{\tabcolsep}{4pt}
-\begin{tabular}{cWcccccWcWccWcc}
+\scalebox{0.95}{
+\begin{tabular}{cWcccccWcccccWcc}
 \\
 \instbit{31} &
 \instbitrange{30}{20} &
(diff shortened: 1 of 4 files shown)


src/supervisor.tex
```diff
@@ -242,6 +247,42 @@ \subsection{Memory Privilege in {\tt sstatus} Register}
 alternate mapping.
 \end{commentary}
 
+\subsection{Endianness Control in {\tt sstatus} Register}
+
+The UBE bit is a \warl\ field that controls the endianness of explicit
+memory accesses for U-mode, which may differ from the endianness of
+memory accesses in S-mode.
+An implementation may hardware UBE to specify always the same endianness
```
marceg: hardware => hardwire

aswaterman: thanks

sorear:
The feature whereby the endianness of a load or store is not known until the TLB lookup returns is pretty unusual and I think it could benefit from a rationale.  (Among the bi-endian architecture manuals I have handy, only SPARC supports different endianness depending on virtual addresses, and the manual calls out the feature as only intended for I/O use and likely to interfere with cacheability; POWER, ARM, MIPS, and IA64 select endianness based on processor mode bits only, and Alpha doesn't support runtime endian switches at all.)

jhauser-us:
Stef O'Rear wrote:
> The feature whereby the endianness of a load or store is not known until the TLB lookup returns is pretty unusual and I think it could benefit from a rationale.

Logically, whatever endianness that U mode is configured for, that's how values will be stored in user-space memory.  So, if S mode accesses values in user space using its own, opposite endianness, it will usually have to swap the bytes.  The proposal is to have the hardware do the swapping automatically when page-based address translation is in use.

Do you have a concern that this may impact an implementation's performance (clock speed)?  The issue can arise only when UBE is the opposite from SBE and also sstatus.SUM = 1; otherwise, the correct endianness is always known before the TLB lookup.  SUM is expected to be enabled only rarely.  One option might be for a bi-endian implementation to execute loads/stores with more delay (if necessary) when both UBE != SBE and SUM = 1.

A core that doesn't care about supporting UBE != SBE can simply hardwire UBE = SBE, even if SBE is variable.  I would not be surprised if many implementations went that route.

PR #398, opened 2019-06-24
Add PAUSE hint

### LINKED ISSUE #43: Guidance requested: instruction for spin-wait loops
Linux is presently using `div a5,a5,zero` for `cpu_relax` (actually the register is variable but it winds up being `a5` in 80% of sites); glibc and Go currently generate no instructions for `atomic_spin_nop` and `runtime.procyield`.

aswaterman:
Closes #43

### DIFF
--- a/src/preface.tex
+++ b/src/preface.tex
@@ -110,6 +110,7 @@ The changes in this version of the document include:
   December 2019.
 \item Defined big-endian ISA variant.
 \item Moved N extension for user-mode interrupts into Volume II.
+\item Defined PAUSE hint instruction.
 \end{itemize}
 
 \section*{Preface to Document Version 20190608-Base-Ratified}
--- a/src/riscv-spec.tex
+++ b/src/riscv-spec.tex
@@ -78,6 +78,7 @@ Andrew Waterman and Krste Asanovi\'{c}, RISC-V Foundation, \specmonthyear.
 \input{intro}
 \input{rv32}
 \input{zifencei}
+\input{zihintpause}
 \input{rv32e}
 \input{rv64}
 \input{rv128}
--- a/src/rv32.tex
+++ b/src/rv32.tex
@@ -1384,15 +1384,19 @@ supervisor-level operating system or debugger.
 
 RV32I reserves a large encoding space for HINT instructions, which are
 usually used to communicate performance hints to the
-microarchitecture.  HINTs are encoded as integer computational
+microarchitecture.  Most HINTs are encoded as integer computational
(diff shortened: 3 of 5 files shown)


aswaterman:
I’m probably just rehashing the discussion we already had, but: I don’t think there’s much benefit from distinguishing the two, as the processor can just unstall when either event (IO or memory) occurs, and there’s likely no loss in efficiency in practice. In fact, given the likelihood of misuse, it might perform better this way around ;)

I would view the predecessor field as just another opcode field when the successor field is zero. It doesn’t really matter if the bits line up with their usual purpose, as they are feeding into different control logic (or so I claim without proof).

I’m not outright opposed to the change, but I need a stronger argument to change my mind.

palmer-dabbelt:
Can we just talk in person when I'm back next week?  I'm running out of
email writing energy... 😁

aswaterman:
Sure.

aswaterman:
@gfavor this is rebased and ready for merging whenever we deem appropriate.

knightsifive:
Minor suggestion: perhaps the enhancements to the HINT discussion could be separated from the addition of the PAUSE discussion. (I suggest this only because there has been some contention about PAUSE on the mailing lists.)

David-Horner:
Perhaps you saw Krste response on google ISA list

https://groups.google.com/a/groups.riscv.org/g/isa-dev/c/0VUZAEl6o0g/m/KCYUUMZTCQAJ.

This revision to HINT is distinct from an enhanced version of PAUSE.

There was also extensive discussion on [RISC-V] [tech] Start of 45-day 
public review period for "PAUSE Hint instruction" extension.

The enhancement discussion will most likely always occur.

What the process has demonstrated this time is that clarifications and 
cleanups of proposals are incorporated,
even if design enhancements with fundamentally substantial changes are not.



On Wed, Dec 16, 2020, 16:38 Nick Knight, <notifications@github.com 
<mailto:notifications@github.com>> wrote:

    Minor suggestion: perhaps the enhancements to the HINT discussion
    could be separated from the addition of the PAUSE discussion. (I
    suggest this only because there has been some contention about PAUSE
    on the mailing lists.)

    —
    You are receiving this because you are subscribed to this thread.
    Reply to this email directly, view it on GitHub
    <https://github.com/riscv/riscv-isa-manual/pull/398#issuecomment-747056570>,
    or unsubscribe

aswaterman:
@knightsifive If the PAUSE discussion drags on, I'll separate these into PRs.  I kept them together because redoing LaTeX these tables is time-consuming, and the changes are coupled.

PR #400, opened 2019-06-25
Allow mcycle to be shared between harts

palmer-dabbelt:
The commentary says this is allowed, but the normative text doesn't.  I don't like this behavior, but the manual should at least be self-consistent.


aswaterman:
Spell check is your friend, Mr. `mcountintihib`

PR #404, opened 2019-07-06
revise table of hypervisor states to include misa[7] status

David-Horner:
Adding misa[7] state to Table 5.1: Operating modes with the hypervisor extension :

1) emphasizes the recommendation to make hypervisor extension selectable.
2) clarifies U-mode is unchanged by adding the extension
3) emphasizes that M-mode,  in addition to HS-Mode, is also enhanced by the extension.
(even though we do not designate it HM-mode).

Signed-off-by: ds2horner <ds2horner@gmail.com>

### DIFF
--- a/src/hypervisor.tex
+++ b/src/hypervisor.tex
@@ -64,18 +64,23 @@ possible operating modes of a RISC-V hart with the hypervisor extension.
 
 \begin{table*}[h!]
 \begin{center}
-\begin{tabular}{|c|c||l|l|l|}
+\begin{tabular}{|c|c|c||l|l|l|}
   \hline
-   Virtualization & Privilege & \multirow{2}{*}{Abbreviation} & \multirow{2}{*}{Name} & Two-Level \\
-   Mode (V)       & Encoding  &                               &                       & Translation \\ \hline
-   0              & 0         & U-mode  & User mode & Off \\
-   0              & 1         & HS-mode & Hypervisor-extended supervisor mode & Off \\
-   0              & 3         & M-mode  & Machine mode & Off \\
+  misa[7]& Virtualization & Privilege & \multirow{2}{*}{Abbreviation} & \multirow{2}{*}{Name} & Two-Level \\
+    (H) & Mode (V)       & Encoding  &                               &                       & Translation \\ \hline
+  0 or 1 &   0*   & 0         & U-mode  & User mode & Off \\
+\hline
+  0 &  0*              & 1         & S-mode & Supervisor mode & Off \\
+  0 &  0*              & 3         & M-mode  & Machine mode & Off \\
   \hline
-   1              & 0         & VU-mode & Virtual user mode & On \\
-   1              & 1         & VS-mode & Virtual supervisor mode & On \\
+  1 &  0              & 1         & HS-mode & Hypervisor-extended supervisor  & Off \\
+  1 &  0              & 3         & M-mode  & Hypervisor-extended Machine  & Off \\
+  \hline
+  1 &  1              & 0         & VU-mode & Virtual user mode & On \\
+ 1 &   1              & 1         & VS-mode & Virtual supervisor mode & On \\
   \hline
  \end{tabular}
+* Note: Virtualization Mode defaults to 0 when H = 0.
 \end{center}
 \caption{Operating modes with the hypervisor extension.}
 \label{h-operating-modes}


aswaterman:
cc @jhauser-us

aswaterman:
I appreciate the concept and the relevance to misa.H.  I'm not sold on distinguishing H=0/mode=M from H=1/mode=M, though.  By that rationale, we should also distinguish M-when-S-is-present from M-when-S-is-not-present, since adding S-mode also adds features to M-mode.  It's not an irrational scheme; it's just at odds with the current approach.

One simple alternative is to revert to the original figure, but change the caption to indicate the table applies to the misa.H=1 case.

I'd like to see what JohnH thinks

jhauser-us:
Andrew wrote:
> I'm not sold on distinguishing H=0/mode=M from H=1/mode=M, though. By that rationale, we should also distinguish M-when-S-is-present from M-when-S-is-not-present, since adding S-mode also adds features to M-mode.

Yes, I had basically the same reaction.  We could also try to account for user mode (determines whether `mstatus`.UXL meaningfully exists), the F extension ( whether `mstatus`.FS exists), and the N extension (whether `mstatus`.UPIE and UIE exist).  Should we have "hypervisor-extended, supervisor-extended, user-extended, F-extended, N-extended machine mode"?  And various subset combinations, too?

Andrew knows my personal philosophy, that in a document like this, excessive pedantry risks readability. In practice, the authors end up making a great many judgement calls for clarity, over which reasonable people may disagree.

My own opinion is that the proposed edit doesn't add anything to comprehension.  However, I may not be an untainted judge, able to see the text with fresh eyes.

> One simple alternative is to revert to the original figure, but change the caption to indicate the table applies to the misa.H=1 case.

That would be my preference.

marceg:
By the same token, what about dropping "HS" in favor of simply "S" ?  The hypervisor extension simply adds functionality to S.  VS and VU are really the only new modes, correlating exactly to V=1.

jhauser-us:
Marc Gauthier wrote:
> By the same token, what about dropping "HS" in favor of simply "S" ?

That's one of those judgement calls that reasonable people can disagree about.  One justification was that the hypervisor extension opens up to supervisor mode a whole category of additional hypervisor-level CSRs and (not so many) instructions, enough perhaps to be worthy of a new label as an enhanced mode.  Another reason was due to HS mode being where type-1 *hypervisor* code executes, which in those contexts is often distinguished from *supervisor-level* code (in VS mode).  A third was to make a better connection to the letter *H* in the names of CSRs and instructions.

I wasn't solely responsible for the name "HS mode", but I happen to believe it's helpful.  I'm not surprised if there are other opinions.  There's no one right answer here.

We actually went through a long process to settle on the current names.  Believe it or not, there was a long time when VU mode was called just "U mode", and I think that was supportable too.  But I'm happy to have the name "VU mode", if only because people find it more intuitive that way.

jhauser-us:
Also, in support of the name "HS mode", I should add that it makes it easy for us to say that "VS mode supports everything that S mode supports".

marceg:
Ah, that's an easy one:  "VS mode supports everything that S mode supports without the hypervisor extension".

Opposite reactions to HM and HS are a bit surprising.  Yes the hypervisor adds many more CSRs.

I'll note that numerous places where I've been documenting or referring to privilege modes, I end up saying "S or HS" or "S/HS" or something similar.  It's extra weight that doesn't add to the discussion.

VU is meaningfully different from U, due to effects of V=1 vs V=0.  Also because although U and VU are identical from the point of view of the user code, U is managed by S whereas VU is managed by VS.

aswaterman:
We’ve gone through a few iterations of naming schemes for this extension,
including, IIRC, what you’re proposing, and the current one appears to
result in the least confusion. Other defensible positions exist, but have
their own downsides (e.g. in some contexts, it’s more convenient or clearer
that “S-mode” means “S-mode, sans HS”—the complement of your observation
that in other contexts it’s more convenient that “S” is shorter than
writing “S or HS”).

David-Horner:
On 2019-07-06 8:17 p.m., John Hauser wrote:
>
> Andrew wrote:
>
>     I appreciate the concept and the relevance to misa.H.
>
Thank you both for your thoughtful consideration.
>
>     I'm not sold on distinguishing H=0/mode=M from H=1/mode=M, though.
>     By that rationale, we should also distinguish M-when-S-is-present
>     from M-when-S-is-not-present, since adding S-mode also adds
>     features to M-mode.
>
In effect the documentation does exactly that. Chapter 3, Machine-Level 
ISA, has numerous occurrences detailing changes when S is present (and 
to a lesser degree when U is present). And not only features added but 
behaviour changes; notably medeleg and mideleg (that apply to S when 
present, to U-mode when N present)
>
>     It's not an irrational scheme; it's just at odds with the current
>     approach.
>
I assume you mean current approach for Machine/Supervisor-Level ISA 
documentation.

However, Hypervisor Extension documentation does not take the same 
approach. V state is considered an add-on and not an integral part of 
the MSU-modes design.

Conversely, Chapter 3 defines M-mode in the context of all three 
privilege modes (It even has a bias towards the originally conceived 
four mode hypervisor support). And also in the context of N extension, F 
extension and provisions for extended ISA state (mstatus.FS and 
mstatus.XS fields). Therefor, by definition, there is no extension of 
M-mode with F or N extensions, nor for U and S mode implementation.


> Yes, I had basically the same reaction. We could also try to account 
> for user mode (determines whether |mstatus|.UXL meaningfully exists), 
> the F extension ( whether |mstatus|.FS exists), and the N extension 
> (whether |mstatus|.UPIE and UIE exist). Should we have 
> "hypervisor-extended, supervisor-extended, user-extended, F-extended, 
> N-extended machine mode"? And various subset combinations, too?
>
Conceptually we do have such modes and this is how I perceive the 
overall ISA. By excellent design they are substantially orthogonal and 
so can be considered independently as well as collectively. The various 
subset combinations, too!
>
> Andrew knows my personal philosophy, that in a document like this, 
> excessive pedantry risks readability.
>
I agree, it is better for documentation to provide clarity through 
precise concepts that once understood cancel the need for 
explanation/clarification in exhaustive contexts.  It is imperative, 
however, that those concepts are inherently present in the underlying 
subject matter and are sufficiently presented as to be readily understood.

It is the concept that setting H-mode fundamentally changes M-mode that 
I hoped to highlight with the table.

I hope by my arguments above that you both now agree.

> In practice, the authors end up making a great many judgement calls 
> for clarity, over which reasonable people may disagree.
>
I am of the opinion that once reasonable people agree on the intent, 
objectives and priorities such disagreements disappear or are limited to 
minor wordsmithing. However, I have not had your experience with 
"reasonable people" nor is agreement on those fundamentals generally 
easy to achieve.

> My own opinion is that the proposed edit doesn't add anything to 
> comprehension. However, I may not be an untainted judge, able to see 
> the text with fresh eyes.W
>
So all that raises many questions for me, not the least: What 
understanding we are intending to convey.

I agree that it is not essential that the reader envision a 
multiplicative plethora of distinct states in the ISA with each feature 
or facility.

However, with such a set of fresh eyes, improved means of conveying 
fundamental concepts and explaining inherent characteristics may emerge.

I realize now that augmenting the table is insufficient. The inferences 
are not sufficiently apparent that highlighting them will cause an "aha" 
moment. Another approach will need to convey that M-mode behaviour has 
significantly changed; that select-able H-mode is inherently different 
from how the "original" U and S-Modes are handled;


>     One simple alternative is to revert to the original figure, but
>     change the caption to indicate the table applies to the misa.H=1 case.
>
> That would be my preference.
>
I considered that but decided that it would be superfluous. My reasoning 
was that the whole chapter assumes misa.H set (excepting the one 
sentence that explicitly deals with disabling it). If the table 
augmented as proposed did not elicit awareness of ramifications then 
stating the obvious even less so. Not as an alternative, but to have a 
visual reinforcement, it may be warranted.


*Are the requisite underlying concepts present for sufficiently complete 
documentation ?*

*Upward compatible code.*

M-mode is not completely orthogonal to the other states. Code written 
for some features run completely unchanged in M-mode, for others M-mode 
code must change its behaviour to have the previous effect.

M-mode only code will run unchanged with the introduction of U-mode.

Similarly for N extension. It is effectively disabled by default with 
mstatus.NS zero. I'd prefer N functionality disabled if sedeleg and 
sideleg are both zero (their default values) rather than extend 
mstatus/sstatus with an NS field. But this is probably why this 
extension is still in draft. (and mandating N be writable does not help as

However, M-mode code supporting U+N is perturbed by the introduction of 
S-mode. The default sedeleg and sideleg are zero and therefor medeleg 
and mideleg are not propagated through to N. If they defaulted to all 
ones then M-mode code would be invariant. It is perhaps not too late to 
impose this hardware initialization requirement on sedeleg and sideleg 
when N is enabled, however, this introduces its own anomaly within S-mode.

If I were to suggest and alternative it would be, N extension be defined 
with M-mode CSR control through sedeleg and sideleg. It would then be S 
agnostic. I gather this may still be possible to change.

I now believe any deviation from the concept (or meme) of default upward 
code compatibility should be explicitly highlighted. I previously 
thought it was sufficient to explicitly state the behavioral deviation . 
As in this example, "medeleg applies to U-mode when S is absent and to S 
mode when it is present". However, I now believe the inference can 
easily be missed; that opportunities to improve the design might be 
missed. I presumed that there were mitigating technical reasons. I now 
believe if this is indeed true, than they should be a part of the 
explicit highlighting documentation.

*Facilities by default should be minimally pervasive.*

Ideally the default condition should allow for ignoring the facility 
with little or no adverse consequences. Provisions for approximating 
running of code as if the facility were not present.

The F extension is a good example. Unless it is explicitly activated by 
initializing status.FS it is as if it were absent. This not only allows 
for upward code compatibility, but further provides an environment that 
can validate the software will run on a machine without the facility 
(eg. soft-float routine validation, that no F instructions were 
inadvertently included).

As with the previous "concept" I believe deviations from it should be 
avoided, and if present explicitly highlighted.

I believe the current hypervisor design fails this concept.

But before I ramble on more I'd like to pause  to get your feedback on 
the importance of these two concepts and if I misjudged the "S-mode 
disrupting U+N" scenario.

**And related other:**

Do we present a consistent model of the various privilege modes 
(including V state)?

Should Hypervisor V-mode be an extension or implemented/not-implemented 
mode just as S and U?

What are the tradeoffs of making/documenting Hypervisor as a first class 
state? *
*

*
*

*
*





*
*

jhauser-us:
David Horner wrote:
> that select-able H-mode is inherently different from how the "original" U and S-Modes are handled

Not as much as you seem to believe.  See below.

> the whole chapter assumes misa.H set (excepting the one sentence that explicitly deals with disabling it).

Naturally, because having `misa`.H set means the hypervisor extension is enabled (thus "active", "supported", whatever).  A RISC-V implementation is allowed to have other bits in `misa` be writable too.  If`misa`.S is writable, setting `misa`.S = 1 means supervisor mode is enabled ("supported"); setting `misa`.S = 0 means supervisor mode is not enabled (not supported).  Likewise for `misa`.U and U mode.  That the same relationship would be true for `misa`.H and the hypervisor features doesn't feel to me like a new concept by the time we get to the hypervisor chapter.

> Similarly for N extension. It is effectively disabled by default with mstatus.NS zero.

Field `mstatus`.FS is explicitly left unspecified after reset, and there's no reason to expect NS to be any different.  So I'm not quite sure what you mean by "disabled by default".  M-mode software must know enough to zero those fields to disable the features.

> I now believe any deviation from the concept (or meme) of default upward code compatibility should be explicitly highlighted.  [...]

I hear you that you wish that M-level software could always simply ignore "extra" features it doesn't care about.  To repeat one of your own examples, if an implementation has M, S, and U modes, and supports the N extension, you'd like it to be possible for M-level software to be oblivious of S mode, such that manipulating `medeleg` and `mideleg` alone, never touching `sedeleg` or `sideleg`, would be sufficient to control trap propagation to U mode.  And you want this same principle of allowed obviousness to work for all RISC-V features, including those involving hypervisor support.

While this might (or might not) have been a good idea for RISC-V, it's clear there's been no attempt to satisfy such a principle in the RISC-V design so far.  Instead, as you've observed, M-mode software can only sometimes be oblivious in this way, and then probably only by accident.  I have no gauge for how valuable such a principle might be within the breadth of the RISC-V community.  I'll leave it to Andrew to speculate on the prospects of adopting this principle for RISC-V at this date and retrofitting it into the existing ISA specifications.

> Should Hypervisor V-mode be an extension or implemented/not-implemented mode just as S and U?

I think this makes no practical difference to the hardware or software.  The technical details would all be the same if we scrubbed the word *extension* from the hypervisor chapter.  Only we might show the hypervisor-specific fields of `mstatus` (MPV and MTL) earlier, in the machine-level chapter.

jhauser-us:
Marc Gauthier wrote:
> I'll note that numerous places where I've been documenting or referring to privilege modes, I end up saying "S or HS" or "S/HS" or something similar.

I would just say "S mode" most of the time.  Since HS mode is a superset of S mode, anything that works for S mode ought to be good for HS and VS modes too.

If necessary, I'd say up front:  "The term *S mode* is meant to include HS mode too, unless stated otherwise."

PR #406, opened 2019-07-10
Clarify that mtvec is WARL

asb:
Other WARL registers seem to be explicit (e.g. "The misa CSR is a WARL
read-write register..."). This patch adds a similar indication for
mtvec. This consistency is important, as otherwise the reader will spend
time trying to determine if the behaviour is different. You can
determine it's WARL by reading the field layout diagram, but I think a
little redundancy in favour of easing readability makes sense.

At least one simulator started off trapping on invalid field
modifications
<https://lists.gnu.org/archive/html/qemu-devel/2018-04/msg04510.html>.


aswaterman:
Yeah.

PR #416, opened 2019-07-24
Updates to LR/SC proposal

jscheid-ventana:
Updates to the base specification.
This is an attempt to capture the results of the recent rounds of discussion on the new proposal in the memory model group list.
It also includes some re-phrasing for clarity.
Additional related work to introduce a Reservability PMA in the privileged architecture remains.

### DIFF
--- a/src/a.tex
+++ b/src/a.tex
@@ -124,6 +124,13 @@ synchronization libraries and other software that assumes DW-CAS is
 the basic machine primitive.  A possible mitigating factor is the
 recent addition of transactional memory instructions to x86, which
 might cause a move away from DW-CAS.
+
+More generally, a multi-word atomic primitive is desirable, but there is
+still considerable debate about what form this should take, and
+guaranteeing forward progress adds complexity to a system.  Our
+current thoughts are to include a small limited-capacity transactional
+memory buffer along the lines of the original transactional memory
+proposals as an optional standard extension ``T''.
 \end{commentary}
 
 The failure code with value 1 is reserved to encode an unspecified
@@ -189,10 +196,27 @@ during a preemptive context switch to forcibly yield any existing load
 reservation.
 \end{commentary}
 
-LR/SC can be used to construct lock-free data structures.  An example
-using LR/SC to implement a compare-and-swap function is shown in
-Figure~\ref{cas}.  If inlined, compare-and-swap functionality need
-only take four instructions.
+An SC instruction can never be observed by another RISC-V hart
+before the immediately preceding LR.
+The LR/SC
+sequence can be given acquire semantics by setting the {\em aq} bit on
+the LR instruction.  The LR/SC sequence can be given release semantics
+by setting the {\em rl} bit on the SC instruction.  Setting the {\em
+  aq} bit on the LR instruction, and setting both the {\em aq} and the {\em
+  rl} bit on the SC instruction makes the LR/SC sequence sequentially
+consistent, meaning that it cannot be reordered with earlier or
+later memory operations from the same hart.
+
+If neither bit is set on both LR and SC, the LR/SC sequence can be
+observed to occur before or after surrounding memory operations from
+the same RISC-V hart.  This can be appropriate when the LR/SC
+sequence is used to implement a parallel reduction operation.
+
+Software should not set the {\em rl} bit on an LR instruction unless the {\em
+aq} bit is also set, nor should software set the {\em aq} bit on an SC
+instruction unless the {\em rl} bit is also set.  LR.{\em rl} and SC.{\em aq}
+instructions are not guaranteed to provide any stronger ordering than those
+with both bits clear, but may result in lower performance.
 
 \begin{figure}[h!]
 \begin{center}
@@ -217,36 +241,10 @@ only take four instructions.
 \label{cas}
 \end{figure}
 
-An SC instruction can never be observed by another RISC-V hart
-before the immediately preceding LR.
-The LR/SC
-sequence can be given acquire semantics by setting the {\em aq} bit on
-the LR instruction.  The LR/SC sequence can be given release semantics
-by setting the {\em rl} bit on the SC instruction.  Setting the {\em
-  aq} bit on the LR instruction, and setting both the {\em aq} and the {\em
-  rl} bit on the SC instruction makes the LR/SC sequence sequentially
-consistent, meaning that it cannot be reordered with earlier or
-later memory operations from the same hart.
-
-If neither bit is set on both LR and SC, the LR/SC sequence can be
-observed to occur before or after surrounding memory operations from
-the same RISC-V hart.  This can be appropriate when the LR/SC
-sequence is used to implement a parallel reduction operation.
-
-Software should not set the {\em rl} bit on an LR instruction unless the {\em
-aq} bit is also set, nor should software set the {\em aq} bit on an SC
-instruction unless the {\em rl} bit is also set.  LR.{\em rl} and SC.{\em aq}
-instructions are not guaranteed to provide any stronger ordering than those
-with both bits clear, but may result in lower performance.
-
-\begin{commentary}
-In general, a multi-word atomic primitive is desirable but there is
-still considerable debate about what form this should take, and
-guaranteeing forward progress adds complexity to a system.  Our
-current thoughts are to include a small limited-capacity transactional
-memory buffer along the lines of the original transactional memory
-proposals as an optional standard extension ``T''.
-\end{commentary}
+LR/SC can be used to construct lock-free data structures.  An example
+using LR/SC to implement a compare-and-swap function is shown in
+Figure~\ref{cas}.  If inlined, compare-and-swap functionality need
+only take four instructions.
 
 \section{Eventual Success of Store-Conditional Instructions}
 \label{sec:lrscseq}
--- a/src/a.tex
+++ b/src/a.tex
@@ -267,8 +267,10 @@ the following properties:
 \item The code to retry a failing LR/SC sequence can contain backwards jumps
   and/or branches to repeat the LR/SC sequence, but otherwise has the same
   constraint as the code between the LR and SC.
-\item The LR address must lie either within a main memory region or within some
-  other memory region specified by the execution environment.
+\item The LR address must lie within a memory region with the {\em LR/SC
+  eventuality} property.  Main memory regions have this property.  The
+  execution environment may specify additional memory regions that have
+  this property.
 \item The SC must be to the same virtual address and of the same data size as
   the latest LR executed by the same hart.
 \end{itemize}
@@ -308,8 +310,7 @@ must guarantee that one of the following events eventually occurs:
   the reservation set of the LR instruction in {\em H}'s constrained LR/SC
   loop, or some other device in the system writes to that reservation set.
 \item {\em H} executes a branch or jump that exits the constrained LR/SC loop.
-\item {\em H} executes an instruction that raises a synchronous exception.
-\item {\em H} takes an interrupt.
+\item {\em H} takes a trap.
 \end{itemize}
 
 \begin{commentary}
--- a/src/a.tex
+++ b/src/a.tex
@@ -241,35 +241,36 @@ proposals as an optional standard extension ``T''.
 \section{Eventual Success of Store-Conditional Instructions}
 \label{sec:lrscseq}
 
-The standard A extension defines {\em constrained LR/SC sequences}, which have
+The standard A extension defines {\em constrained LR/SC loops}, which have
 the following properties:
 \vspace{-0.2in}
 \begin{itemize}
 \parskip 0pt
 \itemsep 1pt
-\item The static code for the LR/SC sequence, plus the code to retry the
-  sequence in the case of failure, must comprise at most 16 instructions
-  placed sequentially in memory.
-\item The dynamic code executed between the LR and SC instructions can only
-  contain instructions from the base ``I'' instruction set, excluding loads,
-  stores, backward jumps or taken backward branches, JALR, FENCE, and SYSTEM
-  instructions.  If the ``C'' extension is supported, then compressed forms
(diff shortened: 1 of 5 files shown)


jscheid-ventana:
My current development environment doesn't have the capability to build the PDF via LaTeX, so apologies in advance for markup-related issues.

jscheid-ventana:
cc/ @aswaterman @dkruckemyer-ventana

src/a.tex
```diff
@@ -157,29 +159,49 @@ \section{Load-Reserved/Store-Conditional Instructions}
...
+
+The following events are guaranteed to invalidate the reservation on hart
+{\em H}:
+\vspace{-0.2in}
+\begin{itemize}
+\parskip 0pt
+\itemsep 1pt
+\item {\em H} executes a successful SC.
+\item {\em H} executes an SC that addresses bytes that are not included in the
+reservation set.
```
aswaterman: The reservation should be invalidated if any SC instruction is executed, and it's not clear to me these two bullet points are collectively exhaustive of SC instructions, e.g. because SCs can fail for no particular reason from time to time.  I would merge them into a single item: "{\em H} executes an SC instruction (irrespective of the success of the SC instruction)."

src/a.tex
```diff
@@ -157,29 +159,49 @@ \section{Load-Reserved/Store-Conditional Instructions}
...
+Architectural specifications that extend the base specification, e.g.
+extensions, privileged architecture, and platform specifications, may identify
+additional conditions that cause the reservation to be invalidated.
+
+In general, an SC in an LR/SC sequence may succeed only if certain atomicity
+requirements, as defined by the Atomicity Axiom in Section 14.1, are satisfied.
+If these requirements cannot be satisfied, the SC in an LR/SC sequence must
+fail. In particular, if two harts do not agree on the memory order of accesses
+to a reservation set, the atomicity of accesses from those two harts to that
+reservation set cannot be maintained, and so the SC must fail.
```
aswaterman: @daniellustig your thoughts on this paragraph?

daniellustig: This text seems to suggest that after checking if the reservation is still valid, there is some kind of additional step to see if the Atomicity Axiom has also been satisfied.  I'd say its largely the opposite: the Atomicity Axiom just formalizes the atomicity semantics that the reservation already enforces for successful SCs.

I'd also quibble about the wording about harts agreeing on memory order, but before we go there...what would you all say to just deleting this paragraph altogether?

src/a.tex
dkruckemyer-ventana: We were trying to list the things that were guaranteed to invalidate the reservation across all implementations of the architecture. I intentionally omitted a failed SC because the reservation would already be invalid in that case.

marceg: The two bullets look correct.  The single bullet is simpler and seems clearer -- it's okay to invalidate a reservation that's already invalid.  The distinction of the 2 bullets is perhaps most relevant in showing that a constrained LR/SC sequence never itself invalidates the reservation except by succeeding -- material for non-normative comments?

@aswaterman, as defined so far, SC cannot fail for no particular reason from time to time, unless saying that the reservation can be cleared for no particular reason from time to time (which would be another bullet in this list).

src/a.tex
dkruckemyer-ventana: There are a few things we were trying to do here.  The first was to expand on the existing reference to the Atomicity Axiom:

"The precise statement of the atomicity requirements for successful LR/SC sequences is defined by the Atomicity Axiom in Section 14.1."

The second was to point out that the SC can only succeed if the Atomicity Axiom holds; otherwise, the SC must fail. Upon reflection, I see this is a clumsy attempt to tie this back to the memory model, especially since a failed SC won't appear in the global memory order, rendering the axiom moot. So yeah, I'd say take it out.

The third was to introduce a notion of "big-C Coherence" that was floating around in the discussion groups. Again, I was attempting to tie this back to the memory model, and I was trying to create an architectural "litmus test" for events that must cause an SC to fail that could apply once more than the base spec was considered. For example, differences in or modifications to the PTEs and PMAs could break big-C Coherence between harts, because each hart accesses the reservation set in a different manner, so there should be no reasonable expectation of success. (Spoiler alert: I'm anticipating the necessity to describe the behavior of accesses with mixed attributes at some point in the future, so I'm trying to lay the groundwork now....)

Independent of this particular text, I do think there needs to be a formal definition of the coherence order, which is related to the global memory order, and which should be a function of PTEs/PMAs/etc. once the various extended architecture specs are considered, and it seems clear that if two harts can't observe the same coherence order for memory locations in the reservation set, they can't use LR/SC between them. Maybe that's a topic for a different day, but I'd like not to kick the can down the road too much longer....

aswaterman:
Marc, in the previous formulation, it was perfectly fine for the SC to fail
for no particular reason, as long as it eventually would succeed. If this
new formulation disallows that, it’s a bug in the new formulation.

src/a.tex
```diff
@@ -298,19 +320,27 @@ \section{Eventual Success of Store-Conditional Instructions}
...
+\parskip 0pt
+\itemsep 1pt
+\item A hart or a device performs a memory write to a byte included in the
+  reservation set.
+\item Any additional conditions specified by any architectural specifications
+  that extend the base specification.
+\end{itemize}
+
+An SC in a constrained LR/SC loop, executed by an individual hart {\em H}, may
+not succeed if any of the following events occurs:
```
aswaterman: I’m unsure whether this change (splitting the eventuality concerns into two sections) is an improvement. I find it harder to reason about than the previous monolithic formulation, and it’s not clear to me whether the two are equivalent.

src/a.tex
daniellustig: I don't think it's helpful to sneak in a new concept of coherence just anticipating that it might later be used somehow.  Can you give an example of where you think the existing global memory order formulation is incomplete?

src/a.tex
dkruckemyer-ventana: Admittedly, I don't believe they are exactly equivalent, but it's also not clear that, if one hart traps or branches out of its loop, the eventual success guarantee should be scrapped.

What we were trying to call out is that 1. there is activity by agents not performing constrained LR/SC loops that can impact those performing said loops, and 2. there are some self-inflicted ways that cause SC failure on one hart, but shouldn't affect the overall guarantee. One could then presume that if 2 happened on all harts executing the loops, then there's no guarantee that one SC would succeed.

In some ways, it's the difference between expressing what guarantees success of one SC (our wording) and what is legal "forward progress" in the EE (original wording, which may include SC failures).

dkruckemyer-ventana: @aswaterman Re: "it was perfectly fine for the SC to fail
for no particular reason, as long as it eventually would succeed. If this
new formulation disallows that, it’s a bug in the new formulation."
That wasn't clear to me from the original text, but I understand the intent.
I feel the wording we added specifies what *must* make the SC fail (it's a restatement of the "must fail" sentences in the original text, correct?); however, maybe an explicit statement that an individual SC execution can fail for any reason as long as 8.3 is satisfied is warranted.

marceg:
In present and prior formulations, including both the currently published spec and the update you proposed on June 28th, the only way to randomly cause SC to fail is to invalidate the reservation.  Only this new formulation makes explicit what may invalidate a reservation (which IMO is a valuable way of uncovering such issues).  It currently leaves the door open to specifications but not implementations.

What about adding:

> Implementations may occasionally invalidate a reservation for any reason, conditional upon never impeding eventual success of constrained LR/SC sequences by some hart for each reservation set.

The condition is to avoid diluting the entire forward progress guarantee.  For example, allowing cache eviction to invalidate the reservation leaves "Joe-User-Level-Programmer" with no reasonable means to ensure forward progress (likely worth mentioning in non-normative comments).

src/a.tex
dkruckemyer-ventana: Fair point. Perhaps the overall discussion should be reserved (so to speak....) for another time and forum.

To answer your question, I think there's an assumption that all harts have the same architectural state with respect to how they access memory (e.g. memory attributes) and that assumption ensures that the global memory order is observed by all harts. There are two (or more) things that undermine that assumption: mismatched architectural state (e.g. programming memory attributes differently) and changing architectural state (e.g. modifying memory attributes at runtime, which is kinda mismatched architectural state, but in a dynamic sense). In my experience, the abstraction that all agents access memory in the same big-C Coherent manner is not practical (e.g. due to different capabilities of agents) or even desirable (e.g. enforcing cache coherence may negatively impact system cost or performance). In addition, changing the architectural state that affects how memory is accessed is never atomic across all agents, so during those transition periods, different agents can in principle observe different memory orders. (One can carefully craft software sequences to avoid this; I'm aware of these sequences in other architectures, but I'm uncertain whether they exist in RV yet.)

(Note that I don't believe I'm the first to raise these issues, and my understanding is that some of these would be covered in MM2.0, no?)

For the purposes of LR/SC, I was trying to express the notion that LR/SC can't be atomic if harts don't observe the same memory order, at a minimum, on the subset of the global memory order pertaining to the addresses in the reservation set. (LR/SC may still be atomic if harts observed different memory orders on addresses not in the reservation set, however.) But perhaps introducing the tip of the iceberg in the LR/SC section is the wrong move.

Long story short, I think the memory model description and the behavior of LR/SC is consistent in the absence these issues. So for now, I'm OK removing the text Andrew highlighted in his comment here.

aswaterman:
@marceg I claim that previous formulations _implicitly_ permitted SC to fail for no particular reason, as long as it didn’t impede eventuality. I’m totally fine with making that explicit using language like what you’ve proposed.

src/a.tex
strikerdw: OK. (in my best NPR car talk voice) though Andrew cringes every time he hears me say it, I agree with him on this one. 

SC should clear a reservation no matter what (including it already being cleared). 

First off, it's legal to fail an SC even if the address and size match and the reservation is still standing if the sequence is non-conforming ("LR/SC sequences that do not lie withing constrained LR/SC loops are unconstrained. Unconstrained LR/SC sequences might succeed on some attempts on some implementations, but might never succeed on other implementations"). Hence you cannot assume the reservation is absent if the SC fails.

Second, even if it were the case that you know the reservation was gone on a failing SC, I would still argue that it's better to be explicit that the reservation is always gone after an SC. Don't require the reader to make the leap of though that ... oh, a failing SC, so the reservation must be down, so that's why were skpping it here in the architecture text. It more robust and less prone to mental failure to be explicit.

Finally, it does no harm to reset an already reset reservation.

I would stick with SC always kill the reservation (even dead ones).

@marceg, I think SC can fail for no particular reason... at least in non-normative text (that if not supported by normative text should be made normative text). To wit, the non-normative text says "These definitions admit the possibility that SC instructions may spuriously fail for for [someone should remove the second for :-)] implementation reasons, provided progress is eventually made."

This was a change I asked for specifically to ensure that an implementation that had a spurious rarely occurring glitch or simply a rarely occurring case where they had to fail an SC that would otherwise pass, they could without being held up as "broken". You don't want to be absolute here on the must pass when spurious fails that don't impede progress do no harm and really don't have  to be fixed.

dkruckemyer-ventana: OK. My meaning was to specify what events cause the reservation to transition from valid to invalid, but I think what folks are asking for is more along the lines of "after these events, the reservation is guaranteed to be invalid."

So here's my proposal for the bullets and surrounding text:

```
The reservation is guaranteed to be invalid after the following events occur:
- H executes an SC instruction
- A memory write (e.g., an unconditional store, an AMO, a successful SC, or a device write) performed by another hart or device to a byte included in the reservation set is observed between the LR and the SC executed by H

The reservation may be invalid after implementation-defined events occur, provided the invalidation does not impede the eventual success guarantee in Section 8.3. In addition, architectural specifications that extend the base specification, e.g. extensions, privileged architecture, and platform specifications, may identify additional events that result in an invalid reservation; however, the specification of these events may change the conditions under which the guarantee holds.
```

Or something like that...

src/a.tex
strikerdw: OK... I'm going to refrain from the long speech I would have put in here, other than to say that I think forward referencing to 14 here is a bad idea (so yes, Dan cringes every time I say it -- I'm with him here).

For the record, I believe the axiomatic model in 14 does include the notion of Big-C coherence. You take the global memory order and then restrict it to writes to the same location. Out pops what other axiomatic memory models call the "co" or coherence order which is the total order over all writes to a given location and is equivalent to the notion I called "big-C" coherence. 

But I think the rules for LR/SC succeeding operationally should ALL be clearly in the A extension text without needing 14 (i.e. we don't want an implementer to have to swallow and understand, or really even read, 14's much more abstract mathematical worldview in order to build the right hardware for LR/SC).

Now, I hear you about the PMAs moving about and in particular the coherence PMA mismatching. And this is most definitely a MM2.0 discussion, but at the point you drop coherence, I don't think you can reasonably define an architecture to LR/SC. You can define a lot of different implementations, but I don't see how you make a one-size fits all architecture for LR/SC work.

Derek

dkruckemyer-ventana: @strikerdw I agree with you, but I have two additional comments. Number 1: It might be useful to explicitly define coherence order as you described it (I agree with that definition). Number 2: There has to be a consistent memory order on the entire reservation set which is more expansive, I think, than the coherence order on the individual memory locations within the set.

Maybe 2 is not worth the virtual ink because, at the end of the day, the eventual success guarantee requires the LR and SC accesses to be to the same memory location (Coherence!), but I think there is a subtle difference in the "pure" architectural definition of LR/SC.

marceg:
> @marceg I claim that previous formulations _implicitly_ permitted SC to fail for no particular reason, as long as it didn’t impede eventuality. I’m totally fine with making that explicit using language like what you’ve proposed.

Agreed.  Only the newest formulation had broken that in the explicit list of what invalidates a reservation -- sorry, my awkward wording totally did not convey this clearly.

aswaterman:
Closing, as this has been partially incorporated in f7c9203d9d22523f3049bd7942ec15d80e77cec4

PR #417, opened 2019-07-24
Define EE responsibility for hart forward progress

marceg:
This adds a generic forward progress definition into the normative text of Section 1.2.

One question came up.  That section appears to freely alternate between the terms "execution environment" and "execution environment interface", without establishing anywhere that they are equivalent or otherwise what distinguishes them.  Are they in fact, equivalent terms?

### DIFF
--- a/src/intro.tex
+++ b/src/intro.tex
@@ -170,7 +170,10 @@ forward progress of each of its harts --- until their possible
 termination and while not exercising a mechanism that explicitly
 waits for an event, such as the WFI instruction waiting for interrupts.
 Forward progress includes:
+\vspace{-0.2in}
 \begin{itemize}
+\parskip 0pt
+\itemsep 1pt
 \item Completing the execution of an instruction.
 \item Other events defined by extensions, such as traps.
 \end{itemize}
--- a/src/intro.tex
+++ b/src/intro.tex
@@ -166,12 +166,13 @@ Some EEIs support the creation and destruction of additional harts,
 for example, via environment calls to fork new harts.
 
 The execution environment is responsible for ensuring eventual
-forward progress of each of its harts.  Forward progress includes:
+forward progress of each of its harts --- until their possible
+termination and while not exercising a mechanism that explicitly
+waits for an event, such as the WFI instruction waiting for interrupts.
+Forward progress includes:
 \begin{itemize}
 \item Completing the execution of an instruction.
 \item Other events defined by extensions, such as traps.
-\item Mechanisms that explicitly wait for an event, such as the WFI
-  instruction waiting for interrupts, or various environment calls.
 \end{itemize}
 
 \begin{commentary}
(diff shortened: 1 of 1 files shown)


src/intro.tex
```diff
@@ -165,6 +165,15 @@ \section{RISC-V Software Execution Environments and Harts}
 Some EEIs support the creation and destruction of additional harts,
 for example, via environment calls to fork new harts.
 
+The execution environment is responsible for ensuring eventual
+forward progress of each of its harts.  Forward progress includes:
+\begin{itemize}
+\item Completing the execution of an instruction.
+\item Other events defined by extensions, such as traps.
+\item Mechanisms that explicitly wait for an event, such as the WFI
+  instruction waiting for interrupts, or various environment calls.
```
pdonahue-ventana: I like the first two.  I parse the third as "Mechanisms that explicitly wait for ... various environment calls."  What environment calls do you envision here?  Wouldn't ECALL instructions fall under the first bullet?

Are mechanisms that wait for an event (which may never come) really considered to be making forward progress (as stated here) or are we waiving the forward progress guarantee in this case?  My intuition is the latter so it would be more like:
Unless a mechanism that explicitly waits for an event, such as the WFI instruction waiting for interrupts, is being employed, the execution environment is responsible for ensuring eventual forward progress of each of its harts.  Forward progress includes:
- Completing the execution of an instruction.
- Other events defined by extensions, such as traps.

src/intro.tex
```diff
@@ -165,6 +165,15 @@ \section{RISC-V Software Execution Environments and Harts}
...
+\end{itemize}
```
jscheid-ventana: I think reference to or accommodation of the trap taxonomy in Section 1.6 is useful here.

Fatal traps and some requested traps halt the hart.

src/intro.tex
marceg: Good points.  I did ponder a bit about whether to make "waiting" part of forward progress or an exception to it.  Your alternative looks good.

src/intro.tex
marceg: What about rewording, "forward progress of each of its harts" to "forward progress of each of its harts until they terminate" ?

jscheid-ventana: I think we should leverage existing formal terms.  Terminate isn't formally defined, but is used in the description of "Fatal Trap".

aswaterman:
EEI and EE aren't equivalent terms---to make an analogy, EEI is to ABI as EE is to OS kernel---but I think the terminology is abused in some cases.

src/intro.tex
marceg: It's a good point.  I looked and found that Section 1.6 specifically calls out that Requested Traps may request termination (footnote 1 to Table 1.1).  So neither "Fatal Trap" nor "Requested Trap" exactly fit the bill, and both use the termination terminology which seems pretty clear.  Saying "until their termination by Fatal Traps or Requested Traps" rather than simply "until their termination" seems to weigh things down without adding clarity.  ??

marceg: (Also, a hart or EE can terminate without needing a trap.)

aswaterman: Agreed @marceg

aswaterman:
I've committed this with some stylistic adjustments; LMK if there are any more quibbles. 9a39b75241914a210d1f4f7a342d62743cd1dcb7

PR #418, opened 2019-07-26
 LR/SC Updates to Sections 8.2 and 8.3

marceg:
Here's another tentative wording update, as comprehensive as I could make it in one pass for these two sections.

Changes made (slightly longer here than in commit message):

- Shamelessly copy selected wording changes from jscheid.

- Make a clear distinction between reservations and reservation sets,
  the first potentially covering a subset of the second.
  Many references to reservation sets are now to the reservation instead.
  This allows optimizations that distinguish non-overlapping reservations
  within a reservation set.

- Note that misaligned LR/SC raises the possibility of accessing multiple
  reservation sets.

- As done originally, leave implementation defined what happens across
  virtual to physical mappings, now in terms of reservations and reservation sets.
  And add the case of mappings changing between LR and SC.

- For compatibility with Amba, avoid requiring that SC to non-reservable regions
  raise an exception, but highly recommend it in non-normative text.

- Add that LR also invalidates the reservation (and establishes a new one).

- Add PTE updates as possible writes.  Thus, writes by "any hart" not "other harts".

- Add that an implementation can randomly invalidate a reservation as long
  as it doesn't impede forward progress.

- Add that invalidating reservations to separate bytes but within the same
  reservation set is implementation defined.

- Add that the dummy SC is needed not only on context-switch, but also
  on certain virtual to physical remappings (such as page migration).

- Note that one-reservation-per-hart is beyond the Atomicity Axiom,
  in order to ensure software works with expected implementations.
  (Actually, that can probably be said of several constraints in these
  sections that are beyond the simple "must be atomic" statement
  of the Atomicity Axiom.)
  And a hint of what an extension might do to lift that restriction.

- Note cache operations that can act as writes to invalidate reservations.

- Clarify in notes that simple implementations can only track reservations
  in the cache for non-shared caches.

- Instead of repeating the list of writes etc, refer back to the list of
  situations that invalidate reservations, to keep the text simple and
  avoid ambiguity.

- Use the phrases "establish a reservation" rather than "register a reservation"
  (perhaps pure personal preference; it sounds clearer to me),
  and "invalidate a reservation" rather than "yield a reservation".
  ("Yield" can mean "let go" as was meant here, or "give" which was not.)

Also includes a PDF, not to be pulled.  (If that complicates things, I'll remove it.)


gfavor:
What is the motivation behind the distinction between reservation and
reservation sets (and 'sets' both singular and plural)?  Especially given
that a reservation can already be a superset of the bytes accessed by an
lr/sc?

marceg:
> What is the motivation behind the distinction between reservation and reservation sets (and 'sets' both singular and plural)? Especially given that a reservation can already be a superset of the bytes accessed by an lr/sc?

Paul Loewenstein brought this up here:  [https://lists.riscv.org/g/tech-memory-model/message/1880](url)   It allows for example multiple harts in the same core to successfully do LR/SC to non-overlapping areas of the same reservation set, as they can directly see each others' reservations -- even though that might not work across multiple cores in the same system (due for example to the line granularity of cache snoops used to communicate invalidations).

Looking at this again now, I'm curious whether the memory model was really the motivation for this.  The memory model simply says whether an LR/SC sequence is atomic, without any constraints to derive a practical implementation.  I'll followup there.

I'm not sure where you see singular vs plural as an issue.  Both are used depending on context.

gfavor:
Sorry for a belaboring a little more: What's the point or significance of
having multiple reservation sets (given that one can have multiple
reservations within one reservation set)?

marceg:
By all means.
The idea is it’s implementation specific whether you can have multiple reservations in a reservation set.  So portable software must assume only one per set.
I see this can be a bit confusing, and given the direction of discussions this approach seems likely to get dropped in favor of simply always invalidating the entire reservation set.

src/a.tex
```diff
@@ -285,7 +348,7 @@ \section{Eventual Success of Store-Conditional Instructions}
 cache and TLB size and associativity.
 Similarly, we disallowed other loads and stores within the loops to avoid
 restrictions on data-cache associativity in simple implementations that track
-the reservation within the cache.
+the reservation within a non-shared cache.
```
aswaterman: It's not impossible to do this with a shared cache, just more difficult.

src/a.tex
```diff
@@ -327,10 +389,12 @@ \section{Eventual Success of Store-Conditional Instructions}
 We note this constraint implies, among other things, that loads and
 load-reserved instructions executed by other harts (possibly within the same
 core) cannot impede LR/SC progress indefinitely.
-For example, conflict misses caused by another hart sharing the cache cannot
+For example, capacity misses caused by another hart sharing the cache cannot
```
aswaterman: Both are true.  I'll rephrase this to be more neutral.

src/a.tex
marceg: Okay.  Then it's no longer a simple implementation, right?  The concern is about accidentally leading implementors in a non-productive direction.  These are subtle issues that can easily not be caught until after release, or still cause waste of effort if caught in time.

aswaterman: That's a fair point.

src/a.tex
```diff
@@ -306,19 +369,18 @@ \section{Eventual Success of Store-Conditional Instructions}
...
 \begin{commentary}
 As a consequence of this requirement, if some harts in an execution
-environment are executing constrained LR/SC loops, and no other harts or
-devices in the execution environment execute an unconditional store or AMO to
-that reservation set, then at least one hart will eventually exit its
-constrained LR/SC loop.
+environment are executing constrained LR/SC loops, and no harts or
+devices in the execution environment write to the reservation set
+(as described in causes of reservation invalidation),
+then at least one hart will eventually exit its constrained LR/SC loop.
```
aswaterman: This isn't quite right, because a successful SC is itself a write, so some hart does write the reservation set.

marceg: Good point.

marceg:
This pull request is superceded by pull request #420 and others to follow (single topic per PR as suggested by Andrew).

PR #420, opened 2019-07-30
Refactor LR/SC reservation set terminology

marceg:
Define invalidations in terms of the reservation itself,
defining the invalidation set simply as a mechanism that
expands the size of reservations. This keeps definitions
independent of whether a reservation might span multiple
reservation sets, as is unavoidable with unaligned LR/SC.
A minimum sized reservation set allows exact reservation.

Reservations being separate from the bytes they refer to,
introduce the term "reserved byte" to refer to the bytes.

### DIFF
--- a/src/a.tex
+++ b/src/a.tex
@@ -197,7 +197,8 @@ contiguous and no greater than the virtual memory page size.
 The standard A extension requires aligned LR/SC:  in the common case, each
 reservation set covers a whole number of the largest supported LR/SC,
 and each reservation covers exactly one reservation set.
-Otherwise, a reservation covers one or more (whole) reservation sets.
+Otherwise and in all cases, a reservation covers
+one or more (whole) reservation sets.
 \end{commentary}
 
 \begin{commentary}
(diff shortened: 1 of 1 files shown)


src/a.tex
```diff
@@ -195,6 +193,11 @@ \section{Load-Reserved/Store-Conditional Instructions}
 A platform specification may constrain the size and shape of the reservation
 set.  For example, the Unix platform is expected to require the set be
 contiguous and no greater than the virtual memory page size.
+
+In the common case where only aligned LR/SC is supported and each
+reservation set covers one or multiple of the largest supported LR/SC,
+each reservation covers exactly one reservation set.  Otherwise,
+a reservation covers one or more (whole) reservation sets.
```
aswaterman: I would suggest strengthening the premise.  Something like, "In the standard A extension, only aligned LR/SC is supported, and so each reservation covers exactly one reservation set."

marceg: The standard extension wording works.  It also makes sense for reservation sets to have the same alignment restrictions as LR/SC, I'll add that.  Independently, reservation sets might still be smaller than 8 bytes -- a doubleword LR/SC might cover two.  So the statement isn't so straightforward.  Perhaps:

> In the standard A extension, only aligned LR/SC is supported:  assuming each reservation set covers the largest LR/SC, each reservation covers exactly one reservation set.  Otherwise, a reservation covers one or more (whole) reservation sets.

aswaterman:
Without misalignment, I don't think it makes sense for reservation sets to ever be plural--i.e., there should never be more than one reservation set active at a time.

marceg:
Superceded by #444

PR #421, opened 2019-07-30
Non-normative details for LR/SC reservation sets

marceg:
Add further details on motivation and expectations with respect to the definition of reservation sets, as defined in PR #420.   (As in #420, "address space" is more accurately "reservable space", which is not yet defined.)

### DIFF
--- a/src/a.tex
+++ b/src/a.tex
@@ -124,6 +124,13 @@ synchronization libraries and other software that assumes DW-CAS is
 the basic machine primitive.  A possible mitigating factor is the
 recent addition of transactional memory instructions to x86, which
 might cause a move away from DW-CAS.
+
+More generally, a multi-word atomic primitive is desirable, but there is
+still considerable debate about what form this should take, and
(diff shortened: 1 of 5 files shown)


src/a.tex
```diff
@@ -195,6 +195,23 @@ \section{Load-Reserved/Store-Conditional Instructions}
...
+harmonizing reservation invalidation with the cache coherence protocol.
+
+It is software's joint responsibility to ensure forward progress, for
+example by avoiding the allocation of data subject to progress-impeding
+writes in the same reservation set(s) as the LR/SC word.
+
+Where this is not a concern, an implementation might define a
+single reservation set covering the entire address space.
+At the other extreme, a suitably constrained system might define
+byte-sized reservation sets, relieving software of the need for
```
aswaterman: should be "word-sized" instead of "byte-sized", given that misaligned LR/SC is not supported and the minimum LR/SC size is one word.

aswaterman:
Closing, as this has been superseded by other changes.

PR #422, opened 2019-07-30
Use consistent terms for exception types.

jscheid-ventana:
Consistently use "access fault".

Consistently use "page fault".

Consistently use "instruction" instead of "instruction-fetch" or "fetch"
when qualifying exception types.

Consistently order list of "instruction, load, or store" with respect to
exception types.

No normative change is intended.

### DIFF
--- a/src/a.tex
+++ b/src/a.tex
@@ -138,7 +138,7 @@ For LR and SC, the A extension requires that the address held in {\em
   rs1} be naturally aligned to the size of the operand (i.e.,
 eight-byte aligned for 64-bit words and four-byte aligned for 32-bit
 words).  If the address is not naturally aligned, a misaligned address
-exception or an access exception will be generated.  The access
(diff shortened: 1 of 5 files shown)


aswaterman:
Thanks for pointing out these inconsistencies.  But I feel the need to revise the hyphenation decision ("page faults" should cause "page-fault exceptions").  I'll make those changes atop yours and merge it in.

jscheid-ventana:
That improves English usage at the expense of consistent terms for searches.  Unfortunately, not all PDF viewers find both when searching for "page fault".

aswaterman:
Oh, that's unfortunate.  The pendant in me still favors the more grammatical approach.  I think we should look into making a reverse index, so that you can click on terms like "access-fault exception" and get linked to an index that lists all uses of that term and its cousins, like "access fault."

Issue #298, opened 2018-12-06
Hypervisor guest time delta


aswaterman:
It has been requested that we add htimedelta[h] CSRs so that hosts can lie to guests about the current time, without requiring trapping and emulating.  There is no intent to add equivalent CSRs for cycle/instret.

https://groups.google.com/a/groups.riscv.org/forum/#!searchin/isa-dev/Time$20offset%7Csort:date/isa-dev/9F_-XcUd2kU/bhxHgb62BQAJ

cc @jhauser-us

bonzini:
Please do include the same CSRs for `cycle`, since the SBI set timer callback has absolute cycles as the argument. I agree that instret and performance counters can be handled via trap and emulate.

aswaterman:
sbi_set_timer is specified in terms of the time counter, not the cycle counter.  We're covered, as far as timer interrupts go.

PR #423, opened 2019-07-31
hypervisor: add performance counter delta registers

bonzini:
It has been requested that we add htimedelta[h] CSRs so that hosts can
lie to guests about the current time, without requiring trapping and
emulating. cycle is also included, since the SBI set timer callback has
absolute cycles as the argument.  There is no intent to add equivalent
CSRs for instret and performance counters.

Fixes: #298

### DIFF
--- a/src/hypervisor.tex
+++ b/src/hypervisor.tex
@@ -404,6 +404,51 @@ indicating reads to the corresponding counter will cause an exception
 when V=1.
 Hence, they are effectively \warl\ fields.
 
+\subsection{Hypervisor time delta registers ({\tt htimedelta}, {\tt htimedeltah})}
+
+The {\tt htimedelta} CSR is a VSXLEN-bit read/write register
+that contains a delta between the hardware value of the {\tt time} CSR
+and the value returned in VS/VU-mode.  That is, reading the {\tt time}
+or {\tt timeh} CSRs, or executing the {\tt RDTIME} instruction returns the
+value of the hardware counter plus the contents of {\tt htimedelta}.
+
+\begin{figure*}[h!]
+{\footnotesize
+\begin{center}
+\begin{tabular}{@{}J}
+\instbitrange{VSXLEN-1}{0} \\
+\hline
+\multicolumn{1}{|c|}{\tt htimedelta} \\
+\hline
+VSXLEN \\
+\end{tabular}
+\end{center}
+}
+\vspace{-0.1in}
+\caption{Hypervisor time delta register.}
+\label{hdeltareg}
+\end{figure*}
+
+On RV32 only, the time delta is split between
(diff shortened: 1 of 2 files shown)


aswaterman:
For the reason I mentioned in #298, I would prefer to add only the time-delta CSR.

bonzini:
Sorry, I was confused by Linux code that does rdtime and calls it get_cycles...

I would like to have cycles as well but I will remove it so that we can progress on this missing feature.

aswaterman:
Ah, yes, that API name mismatch is unfortunate.  Thank you.

src/hypervisor.tex
```diff
@@ -404,6 +404,51 @@ \subsection{Hypervisor Counter-Enable Register ({\tt hcounteren})}
 when V=1.
 Hence, they are effectively \warl\ fields.
 
+\subsection{Hypervisor time delta registers ({\tt htimedelta}, {\tt htimedeltah})}
+
+The {\tt htimedelta} CSR is a VSXLEN-bit read/write register
+that contains a delta between the hardware value of the {\tt time} CSR
+and the value returned in VS/VU-mode.  That is, reading the {\tt time}
+or {\tt timeh} CSRs, or executing the {\tt RDTIME} instruction returns the
+value of the hardware counter plus the contents of {\tt htimedelta}.
```
avpatel: should this not be "returns the value of the hardware counter minus the contents of htimedelta" ??

Otherwise looks good to me.

aswaterman: It's arbitrary, and we can define it either way, it's just a matter of whether software negates the value before writing it to htimedelta.

aswaterman: I looked up how the delta registers were defined in the older versions of the specs, and we also defined it like Paolo did: the delta is added, not subtracted.

avpatel: I am fine with both approaches but we should have clear documentation of HTIMEDELTA CSR.

Maybe Replace
".... contains a delta between the hardware value of the {\tt time} CSR and the value returned in VS/VU-mode."
With
".... contains difference between the {\tt time} CSR value read in VS/VU-mode and the hardware value of the {\tt time} CSR."

aswaterman:
One detail is this register should be HSXLEN bits wide, not VSXLEN bits wide.  If the host is RV64 and the guest is RV32, the host shouldn't have to use two CSRs to maintain the delta.  I will fix this, and a couple other stylistic things, and merge this in.

aswaterman:
Merged manually via 4bfae7862e436ce67c1aebaf2d95f0d35c658c51.

Note we've also changed the CSR assignment to keep the h* registers in the 0x6xx range, and to reflect that we will not add delta registers for other counters.  (Time is special because it is global. The other counters are per-hart, so it is practical to migrate them on VM migration even without delta registers.)

PR #432, opened 2019-08-05
define CSRs reserved for trapping to HS-mode

bonzini:
This patch defines a range of CSRs that always trap to hypervisors, namely 0x400-0x5FF. Future releases of the standard should not define `vs*` CSRs that map to this range. These could be used to enable paravirtualization capabilities for the HEE, including:

* reporting of time when no host CPU was available to run the guest

* reporting host wallclock time at VM boot

* requesting the host to flush the TLB on the next guest entry (which avoids IPIs on overcommitted host)

* enabling asynchronous page faults (which could be delivered through scause>=16)

Standard CSRs in these ranges can be used to define paravirtualized interfaces that are accessible from multiple hypervisors.

Because U-mode CSRs in those range are all defined as standard CSRs rather than custom, the patch also redefines the range 0x4C0 to 0x4FF to custom read/write.

### DIFF
--- a/src/hypervisor.tex
+++ b/src/hypervisor.tex
@@ -118,6 +118,13 @@ for HS-mode and U-mode.
 Hypervisor software is expected to manually swap the contents of these
 registers as needed.
 
+Finally, CSRs in the range from 0x400 to 0x5FF are guaranteed to cause
+an illegal instruction when accessed from VS-mode or VU-mode, even in
+future versions of the spec\footnote{The exception could be delivered to
+  VS-mode if bit 2 is set in {\tt hedeleg}.}.  Hypervisors can use the
+custom CSR ranges 0x4C0--0x4FF and 0x5C0--0x5FF to implement
+platform-specific functionality.
+
 In this section, we use the term {\em HSXLEN} to refer to the effective XLEN
 when executing in HS-mode, and {\em VSXLEN} to refer to the effective
 XLEN when executing in VS-mode.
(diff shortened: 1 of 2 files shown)


aswaterman:
I'm having trouble understanding the rationale here.  SBI calls seem more appropriate for a paravirtualization interface.  This is an enormous fraction of the CSR space to allocate for this purpose, and by contrast, the SBI-call space is effectively unlimited.

Separately, I do not see justification in allocating more custom CSR space at this time.

bonzini:
>  This is an enormous fraction of the CSR space to allocate for this purpose

Note that this patch does *not* allocate CSR space for paravirtualization. It just says that there are no `vs*` registers allocated in this range, which actually matches how the current specification allocates hypervisor-level CSR addresses (compare 0x200-0x2FF to 0x600-0x6FF for `h*` in Table 2.4). The idea is to guide the choice of CSR addresses in future extensions to the privileged specification.

Paravirtualization (whether standard or hypervisor-specific) is one possible use, but the important part is that CSRs trap to HS-mode.

aswaterman:
It _is_ an allocation, as this part of the address space can no longer be used for hardware CSRs in those modes.

Why not employ a level of indirection (e.g. ECALL with an x-register calling convention)? In addition to the larger address space, this approach will perform better, since software won’t have to decode instructions to determine the trap cause.

bonzini:
Fair enough...

PR #433, opened 2019-08-06
Describe the Reservability PMA

jscheid-ventana:
I'm open to other terms for the enumerated PMA values.


aswaterman:
Merged manually.

PR #434, opened 2019-08-07
Fix transposed args in LR/SC CAS example

hegemanjwh2:
This commit fixes transposed arguments in the LR/SC CAS example of Chapter 8, Section 8.2.

According to the spec, "`sc.w` writes a word in `rs2` to the address in `rs1`...".  In the example, `a0` contains the address and `a2` contains the value to be written.  Therefore it appears that the instruction should be written as
```
        sc.w t0, (a0), a2
```

### DIFF
--- a/src/a.tex
+++ b/src/a.tex
@@ -232,7 +232,7 @@ only take four instructions.
     cas:
         lr.w t0, (a0)        # Load original value.
         bne t0, a1, fail     # Doesn't match, so fail.
-        sc.w t0, a2, (a0)    # Try to update.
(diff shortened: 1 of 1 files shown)


aswaterman:
It's actually correct as-is.  This way it matches the other store instructions, where the store data comes before the address.  (I realize this is not obvious from the spec, because the spec doesn't contain a definition of the assembly language syntax.)

hegemanjwh2:
Thanks; confirmed also that this is consistent with riscv-gnu-toolchain.

PR #439, opened 2019-08-21
Define Upper-Half User-Level Timer CSRs for RV32I

Columbus240:
Up to now the behavior of `cycleh`, `instreth`, `hpmcounterN` and `timeh` wasn’t defined in the privileged spec, other than in the initial table.


aswaterman:
Can you fix the build error you introduced?

Columbus240:
Sorry for being too lazy to try building it.

aswaterman:
That's what Travis CI is for :)

Issue #440, opened 2019-08-21
Usage of “RV32I” in the CSR listing


Columbus240:
In the CSR listing, the commentary to some CSRs like `timeh` or `htimedeltah` says “RV32I only”. That would exclude the RV32E base ISA, which would mean that RV32E has less CSRs than RV32I, namely it isn’t able to access the upper halves of these registers.

Is this intentional or should the spec say “RV32 only” as it does for `mstatush`?

aswaterman:
You can make a pull request to replace RV32I with RV32 if you’d like.

Columbus240:
Thank you.

PR #441, opened 2019-08-21
Allow RV32E to access the same CSRs as RV32I

Columbus240:
A fix for issue #440.
Includes fixes for trailing whitespace.


aswaterman:
BTW, @Columbus240, I did some archaeology and this idea dates back to an older description of RV32E that indicated the perf counters were excluded.  Now that perf counters are a separate extension, that point is moot.

Columbus240:
I thought something like that might come up. I only checked git blame a little.

PR #444, opened 2019-08-29
Changes nominally for AMBA

marceg:
There are two commits.  The first removes a fair bit of redundancy between the two paragraphs at lines 167 through 192 of a.tex.  The second adds a simplified version of wording suggested earlier.


aswaterman:
Closing, as a variant of this has been incorporated in 26492e36885acbac1eeeef54ad27589898f0def5

PR #447, opened 2019-09-05
Add files via upload

wdc-pnl:
Change \, to \/ so that PDF string searching works.


scottj97:
This PR seems pretty broken...the PR title (and commit message header) "Add files via upload" doesn't make sense. The PR description suggests it's a tiny change but the diff is huge. I suspect that the file was copied from a different branch.

aswaterman:
Looks like the intent was to change e.g. `{\em x}\,IE` to `{\em x}\/IE` to fix searching the PDF.  I'll manually make the change.

PR #453, opened 2019-10-19
Consistently claim that U-immediate is 32 bits, not 20 bits

### LINKED ISSUE #452: Clarify descriptions of LUI and AUIPC immediate values
On pages 18-19 section 2.4 and 36-37 section 5.2 and 105 section 16.5, in the descriptions of LUI and AUIPC, I don't think it's clear what the "U-immediate value" and "nzimm" are.

    LUI (load upper immediate) is used to build 32-bit constants and uses the
    U-type format. LUI places the U-immediate value in the top 20 bits of the
    destination register rd, filling in the lowest 12 bits with zeros.

It places the "U-immediate value" in the top 20 bits. What is the "U-immediate value"? In the figure above the quoted text I can see:

```
 31                12 11   7 6        0
+--------------------+------+----------+
|    imm[31:12]      |  rd  |  opcode  |
+--------------------+------+----------+
          20             5        7
  U-immediate[31:12]   dest      LUI
  U-immediate[31:12]   dest     AUIPC
```

So it looks like the upper bits (31:12) of the "U-immediate value" are defined as the bits from the instruction encoding. The lower bits are still undefined.

Now it is not clear what the "U-immediate value" is.

* Is it a 32 bit value with the lower 12 bits undefined but bits 31:12 defined, as seen in the figure? That would mean that the value constructed in `rd` will be a 44 bit value (20 defined bits (imm[31:12]) followed by 12 undefined bits (imm[11:0]) followed by 12 zeros).
* Is it a 20 bit value that has no lower bits? But then how is it possible to have a value with no lower bits? That just doesn't make sense. The bit range 31:12 suggests that there should be bits 11:0 too.

---

I can see a few ways to make it clearer:

1. Always refer to the upper bits and never to the whole "U-immediate value". Example: change "LUI places the U-immediate value" to "LUI places the upper 20 bits of the U-immediate value" and "AUIPC forms a 32-bit offset from the 20-bit U-immediate" (on the next page) to "AUIPC forms a 32-bit offset from the 20 upper bits of the U-immediate". **This solution creates a problem mentioned at the bottom.**
2. Define the "U-immediate value" as a 32 bit value and define the lower bits of it as 0. Example: Change the whole sentence starting with "LUI places the U-immediate value" to "LUI places the U-immediate value, with the lower 12 bits set to 0, in the destination register rd", and a similar change for AUIPC.  **This solution creates a problem mentioned at the bottom.**
3. Define the "U-immediate value" as a 20 bit value (with bits 19:0) and change the figure to refer to "imm[19:0]" and "U-immediate[19:0]". That way it doesn't look like there are more than 20 bits in this value.
4. Define it as a 20 bit value but make it clearer somehow that 31:12 really refer to bits 19:0 of that value.

I think the intention is that it is a 20 bit value but the bits are numbered 31:12 instead of starting from 0, like all other numbers do, to show that this number will be used for upper bits. I think this just makes it more confusing though. This adds confusion when writing assembly code too because this bit numbering makes it look like you have to give really large values as arguments to lui and auipc (values that are >= 0x1000) to make them count as more than 0 since it looks like bits 11:0 will be ignored.

In chapter 25 when `auipc` is written in assembly code, it is not clear how the value given as argument to `auipc` (`delta[31:12]+delta[11]`) should be interpreted. It is given as a 20 bit number and that means that the "U-immediate" should probably not be interpreted as a 32-bit number because then these 20 bits would be bits 19:0 and there would be no way to reach all of the bits 31:12 of the "U-immediate".
 
Same in section 16.5 and the instruction `lui rd, nzimm[17:12]`.

aswaterman:
The text vacillates between describing the U-immediate as a 20-bit
quantity, and as a 32-bit quantity whose lower 12 bits are zero.
Standardize on the latter.  Note the discrepancy in the ASM syntax.

Resolves #452

### DIFF
--- a/src/rv32.tex
+++ b/src/rv32.tex
@@ -562,16 +562,19 @@ U-immediate[31:12] & dest & AUIPC
 \end{center}
 
 LUI (load upper immediate) is used to build 32-bit constants and uses
-the U-type format.  LUI places the U-immediate value in the top 20
-bits of the destination register {\em rd}, filling in the lowest 12
+the U-type format.  LUI places the 32-bit U-immediate value into
+the destination register {\em rd}, filling in the lowest 12
 bits with zeros.
 
 AUIPC (add upper immediate to {\tt pc}) is used to build {\tt pc}-relative
 addresses and uses the U-type format.  AUIPC forms a 32-bit offset from the
-20-bit U-immediate, filling in the lowest 12 bits with zeros, adds this offset
+U-immediate, filling in the lowest 12 bits with zeros, adds this offset
 to the address of the AUIPC instruction, then places the result in register {\em rd}.
 
 \begin{commentary}
+The assembly syntax for {\tt lui} and {\tt auipc} does not represent the lower
+12 bits of the U-immediate, which are always zero.
+
 The AUIPC instruction supports two-instruction sequences to access
 arbitrary offsets from the PC for both control-flow transfers and data
 accesses.  The combination of an AUIPC and the 12-bit immediate in a
--- a/src/rv64.tex
+++ b/src/rv64.tex
@@ -134,15 +134,16 @@ U-immediate[31:12] & dest & AUIPC
 \end{center}
 
 LUI (load upper immediate) uses the same opcode as RV32I.  LUI places
-the 20-bit U-immediate into bits 31--12 of register {\em rd} and
-places zero in the lowest 12 bits.  The 32-bit result is
-sign-extended to 64 bits.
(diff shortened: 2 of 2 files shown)


aswaterman:
As mentioned in the referenced issue, the other obvious solution is to stop claiming the U-immediate is 32 bits, and claim it's 20 bits.  That would require reworking Figure 2.4 in a way that is less helpful to hardware implementers.

robalni:
I think this PR solves the problem I was describing.

PR #460, opened 2019-11-23
Refinements to UNSPECIFIED

pdonahue-ventana:
- Software should avoid it
- Prohibit security holes

### DIFF
--- a/src/intro.tex
+++ b/src/intro.tex
@@ -763,3 +763,6 @@ Like the base architecture, extensions should fully describe allowable
 behavior and values and use the term \unspecified\ for cases that are
 intentionally unconstrained.  These cases may be constrained or defined
 by other extensions, platform standards, or implementations.
+
+Software should avoid \unspecified\ behavior and should not rely on
+\unspecified\ values.
--- a/src/priv-intro.tex
+++ b/src/priv-intro.tex
@@ -199,6 +199,29 @@ Horizontal traps can be implemented as vertical traps that
 return control to a horizontal trap handler in the less-privileged mode.
 \end{commentary}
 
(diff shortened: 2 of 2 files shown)


aswaterman:
I agree with the sentiment, but "Implementations must not allow \unspecified\ behaviors to be dependent on state that is not accessible to the hart in its current privilege level" is too restrictive, I think.  Consider a custom ISA extension that can be dynamically enabled or disabled based upon a custom M-mode CSR.  When the extension is enabled, executing a certain reserved opcode will perform some custom function; when disabled, it will raise an illegal-instruction exception.  Since we write that the behavior on executing a reserved opcode is unspecified, this is an example of an unspecified behavior that (legitimately) depends on state not accessible to the hart at its current privilege level.

I'm not sure how best to craft a general rule that permits more-privileged state from affecting less-privileged unspecified behavior when we want it to, but disallows it when it could open a security hole.

pdonahue-ventana:
> Consider a custom ISA extension that can be dynamically enabled or disabled based upon a custom M-mode CSR. When the extension is enabled, executing a certain reserved opcode will perform some custom function; when disabled, it will raise an illegal-instruction exception. Since we write that the behavior on executing a reserved opcode is unspecified, this is an example of an unspecified behavior that (legitimately) depends on state not accessible to the hart at its current privilege level.

That's a very good point.

I did a survey of several other architectures.  For instance, Alpha says: "UNPREDICTABLE results must not depend upon, or be a function of, the contents of memory locations or registers that are inaccessible to the current process in the current access mode."  But illegal instructions are not UNPREDICTABLE on Alpha so your scenario doesn't exist there.

Let me ponder this.

PR #477, opened 2020-01-20
ignore write to "controlled" SBE and UBE.

David-Horner:
Behaviour when endian fields are hard-coupled is not well defined.
If both fields are written with opposite endianness which takes precedence,
 or are the fields left unchanged?
Both approaches are currently used in CSRs.
Further complicating the situation on RV32 systems is the separation
of UBE from MBE/SBE into two CSRs.
Whereas precedence and conflict cannot be determined for two discrete CSR
 updates I propose instead that controlled fields not be directly writable.

Signed-off-by: David Horner <ds2horner@gmail.com>

[ I would have preferred to introduce the concept of "controlling fields" which would be generally applicable to the other situations where this coupling occurs and allows for soft coupling in addition to  "hard-wiring". More possibilities, including transition setting by multiple controlling fields is possible. Future evolution of the standard may benefit from the concept but I expect there is currently little desire for greater flexibility in setting endian fields. ]

### DIFF
--- a/src/machine.tex
+++ b/src/machine.tex
@@ -746,9 +746,9 @@ ordinarily be required for a world switch.
 \end{commentary}
 
 If S-mode is supported, an implementation may hardwire SBE so that
-SBE=MBE.
+SBE=MBE and writes to SBE are ignored.
 If U-mode is supported, an implementation may hardwire UBE so that
-UBE=MBE or UBE=SBE.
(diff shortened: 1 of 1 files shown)


aswaterman:
In my mind, "hardwired" implies the intended behavior, but I can see how it might be interpreted differently.  I'll merge this, but will bring it to @jhauser-us' attention since it's his prose.

PR #483, opened 2020-02-12
The RVWMO is version 2.0

palmer-dabbelt:
The preface says the RVWMO is version 2.0 and has been ratified, but the text
of the standard still says it's version 0.1.

Signed-off-by: Palmer Dabbelt <palmerdabbelt@google.com>


aswaterman:
Whoops, thanks.

PR #484, opened 2020-02-13
Repurpose misa.T to represent Ztso

aswaterman:


paul-walmsley-sifive:
The more we can make autodiscoverable at runtime, the better.  Looks good to me

src/machine.tex
```diff
@@ -165,6 +165,10 @@ \subsection{Machine ISA Register {\tt misa}}
 bit always reads as the complement of the ``I'' bit.  An implementation that
 supports both RV32E and RV32I can select RV32E by clearing the ``I'' bit.
 
+The ``T'' bit is read-only, and all harts within the execution environment
+must hardwire the ``T'' bit to the same value, indicating either that they
+all implement RVWMO (0) or all implement RVTSO (1).
```
jrtc27, line 170: This is a bit of an unusual way to put it; an RVTSO implementation is an RVWMO implementation that just doesn't exploit the flexibility it could. It could also be confusing for people who don't realise that but are faced with, say, a big.LITTLE implementation where one set of the harts are RVTSO and the other set is RVWMO; they might not know what to do in that case as it'd be seemingly a third state. Similarly people might start filing non-compliance bug reports about implementations that happen to be RVTSO but don't set misa.T; whilst setting the bit would be permitted and helpful for some software, it's by no means required. I'd instead rephrase it more along the lines of:

```latex
The ``T'' bit is read-only, and all harts within the execution environment
must hardwire the ``T'' bit to the same value. If all harts within the
execution environment implement RVTSO then this value may be 1, otherwise
it must be 0.
```

sorear: I agree with the statement that TSO is a special case of WMO and the requirement should be rephrased to reflect that, but I am unsure about "all harts within the execution environment".  A system may include harts with very different capabilities which share some or all physical memory but are not intended to ever migrate threads between them (thinking about CPU and GPU cores, or embedded controllers of various types) and it would be odd to constrain mixing TSO and non-TSO in that way.

jrtc27: I think the real issue is that "execution environment" is only really implicitly-defined in the spec when talking about SEE/HEE/AEE/etc (at least as far as I can tell) and there are several different notions you want in different places. In some cases you want, say, your RV64IMAC management hart to be regarded as in the same execution environment as your RV64GC harts, e.g. when requiring `mhartid` uniqueness, but in other cases like here you want to regard them as separate-but-related environments. I don't think a CPU and a GPU would ever be considered the same execution environment though even if they were both RISC-V-based on the same SoC.

PR #496, opened 2020-03-28
Draft of Zfh extension for IEEE 754 binary16 support

aswaterman:
This proposal follows the template of the existing FP extensions.  Future extensions might also provide widening ops, but I think that's out of scope for the baseline.


aswaterman:
See https://github.com/riscv/riscv-v-spec/issues/349

kito-cheng:
Any plan for merge this? I think it's just 0.1 that mean we still have chance to change anything? otherwise it's like a hidden/secret extension in branch.

When I said there is already have a spec draft for zfh/fp16 and then people told me they don't seem the spec, that happened several times, and then I need to point out this is not exist in master branch yet, you need to find it out in `zfh` branch :P

aswaterman:
I would love to merge it, but IIRC the plan was to wait until @kasanovic wrote an email proposing it.

kito-cheng:
Oh, OK, thanks for clarification :)

vowstar:
I really hope this PR could merge and freeze soon, this is a very useful extension.
I wonder if there is a corresponding extension for bfp16?

src/zfh.tex
```diff
@@ -348,3 +348,46 @@ \section{Half-Precision Floating-Point Classify Instruction}
...
+FSGNJ.S is used to move half-precision values between floating-point registers.
+\end{commentary}
+
+\begin{commentary}
+Half-precision addition, subtraction, multiplication, division, and
+square-root operations can be faithfully emulated by converting the
+half-precision operands to single-precision, performing the operation
+using single-precision arithmetic, then converting back to half-precision.
+Performing half-precision fused multiply-addition using this method incurs
+a 1-ulp error on some inputs.
```
knightsifive, line 384: Suggest clarifying that the "error" is only on RNE and RMM (double-rounding is innocuous for directed rounding).

aswaterman: @knightsifive Yeah, good point.

PR #503, opened 2020-04-06
Clarify that FENCE + remote FENCE.I is insufficient for local hart

jrtc27:
This sequence only synchronises the remote harts; the local hart still needs to perform a normal FENCE.I.


aswaterman:
Good point.

PR #518, opened 2020-05-05
Add virtual instruction exceptions to the hypervisor extension

jhauser-us:
For the hypervisor extension, add another exception type, "virtual instruction exception", that substitutes for the illegal instruction exception in cases where a hypervisor is expected to emulate the instruction.  This saves time because this exception type can be delegated directly to HS mode, whereas illegal instructions will typically be processed first in M mode and delegated to HS mode only by software.


alistair23:
You use both "virtual instruction exception" and "virtual instruction trap" I think it would be easier to search if you only use one.

PR #530, opened 2020-06-26
Not all agents may be coherent

pdonahue-ventana:
For instance, instruction fetches and page table accesses may involve non-coherent caches and TLBs so there is no eventuality guarantee for such non-coherent agents (except via the relevant fence instructions which are described elsewhere).

(There's a risk that this is a circular definition so maybe "coherent agent" should be defined?)

### DIFF
--- a/src/machine.tex
+++ b/src/machine.tex
@@ -2887,7 +2887,7 @@ the same interconnect path and the path does not reorder requests.
 
 Coherence is a property defined for a single physical address, and
 indicates that writes to that address by one agent will eventually be
-made visible to other agents in the system.  Coherence is not to be
+made visible to other coherent agents in the system.  Coherence is not to be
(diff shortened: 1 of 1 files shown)


allenjbaum:
IF only that definition were that simple. You could also have clusters that
are coherent, but not coherent with others, and a different cache levels.
So, some cluster would be coherent at L1, and would be coherent with a
supercluster at L2, and be coherent with the entiere system at L3, or
something like that.

pdonahue-ventana:
That's interesting and presumably legal as long as there's some "other mechanism" to meet the RVWMO Progress Axiom in A.3.4.

kersten1:
Closing in favor of this PR. https://github.com/riscv/riscv-isa-manual/pull/1281

PR #531, opened 2020-06-29
Clarifications to the perceived ordering of CSR accesses and their effects

jhauser-us:
More clearly distinguished the ordering rules that apply to the same hart from those that involve other agents (harts and devices).  Made CSR side effects completely synchronous to the same hart by default.  (Deviances are still allowed, when documented.)  Added a comment that rules about the reordering of CSR accesses in the global memory order should probably be moved to the RVWMO chapter.

### DIFF
--- a/src/csr.tex
+++ b/src/csr.tex
@@ -157,19 +157,23 @@ bits in the CSR when the old value is not required: CSRS/CSRC {\em
 
 \subsection*{CSR Access Ordering}
 
-On a given hart, explicit and implicit CSR access are performed in program
-order with respect to those instructions whose execution behavior is affected
-by the state of the accessed CSR. In particular, a CSR access is performed
+Each RISC-V hart normally observes its own CSR accesses, including its
+implicit CSR accesses, as performed in program order.
+In particular, unless specified otherwise, a CSR access is performed
 after the execution of any prior instructions in program order whose behavior
 modifies or is modified by the CSR state and before the execution of any
 subsequent instructions in program order whose behavior modifies or is modified
-by the CSR state. Furthermore, a CSR read access instruction returns the
-accessed CSR state before the execution of the instruction, while a CSR write
-access instruction updates the accessed CSR state after the execution of the
+by the CSR state.
+Any side effects from a CSR access are likewise observed to occur
+synchronously in program order, unless documented otherwise.
+Furthermore, an explicit CSR read returns the
+CSR state before the execution of the instruction, while an
+explict CSR write updates the CSR state after the execution of the
 instruction.
 
-Where the above program order does not hold, CSR accesses are weakly ordered,
-and the local hart or other harts may observe the CSR accesses in an order
+For the RVWMO memory consistency model (Chapter~\ref{ch:memorymodel}),
+CSR accesses are weakly ordered by default,
+so other harts or devices may observe CSR accesses in an order
 different from program order. In addition, CSR accesses are not ordered with
 respect to explicit memory accesses, unless a CSR access modifies the execution
 behavior of the instruction that performs the explicit memory access or unless
@@ -194,16 +198,19 @@ CSR.  With the exception of the {\tt time}, {\tt cycle}, and {\tt mcycle} CSRs,
 the CSRs defined thus far in Volumes I and II of this specification are not
 directly accessible to other harts or devices and cause no side effects visible
 to other harts or devices.  Thus, accesses to CSRs other than the
-aforementioned three can be freely reordered with respect to FENCE instructions
+aforementioned three can be freely reordered in the global memory order
+with respect to FENCE instructions
 without violating this specification.
 \end{commentary}
-
-For CSR accesses that cause side effects, the above ordering constraints apply
-to the order of the initiation of those side effects but does not necessarily
-apply to the order of the completion of those side effects.
 
 The hardware platform may define that accesses to certain CSRs are
 strongly ordered, as defined by the Memory-Ordering PMAs section in Volume II
 of this manual. Accesses to strongly ordered CSRs have stronger ordering
 constraints with respect to accesses to both weakly ordered CSRs and accesses
 to memory-mapped I/O regions.
+
+\begin{commentary}
+The rules for the reordering of CSR accesses in the global memory order
+should probably be moved to Chapter~\ref{ch:memorymodel} concerning the
+RVWMO memory consistency model.
+\end{commentary}
--- a/src/csr.tex
+++ b/src/csr.tex
@@ -164,12 +164,18 @@ after the execution of any prior instructions in program order whose behavior
 modifies or is modified by the CSR state and before the execution of any
 subsequent instructions in program order whose behavior modifies or is modified
 by the CSR state.
-Any side effects from a CSR access are likewise observed to occur
-synchronously in program order, unless documented otherwise.
 Furthermore, an explicit CSR read returns the
 CSR state before the execution of the instruction, while an
-explict CSR write updates the CSR state after the execution of the
-instruction.
+explict CSR write suppresses and overrides any implicit writes or
+modifications to the same CSR by the same instruction.
+
+Likewise, any side effects from an explicit CSR access are normally
+observed to occur synchronously in program order.
+Unless specified otherwise, the full consequences of any such side
+effects are observable by the very next instruction, and no consequences
+may be observed out-of-order by preceding instructions.
+(Note the distinction made earlier between side effects and indirect
+effects of CSR writes.)
 
 For the RVWMO memory consistency model (Chapter~\ref{ch:memorymodel}),
 CSR accesses are weakly ordered by default,


src/csr.tex
```diff
@@ -157,19 +157,23 @@ \section{CSR Instructions}
...
 subsequent instructions in program order whose behavior modifies or is modified
-by the CSR state. Furthermore, a CSR read access instruction returns the
-accessed CSR state before the execution of the instruction, while a CSR write
-access instruction updates the accessed CSR state after the execution of the
+by the CSR state.
+Any side effects from a CSR access are likewise observed to occur
+synchronously in program order, unless documented otherwise.
+Furthermore, an explicit CSR read returns the
+CSR state before the execution of the instruction, while an
+explict CSR write updates the CSR state after the execution of the
```
David-Horner, line 171: > updates the CSR state after the execution

shouldn't that rather be 
>updates the CSR state at the end of the execution

Or words to that effect (the state is after all changed _by_ the instruction execution.

jhauser-us: Perhaps it would be clearer to say that an explicit CSR write suppresses and overrides any implicit writes or modifications to the same CSR by the same instruction.

David-Horner: yes. I agree but perhaps:
> by the same instruction.

to be more inclusive of any concurrent change including side effects of the instruction, with wording:
> during the same instruction,  ...  

?

... , unless documented otherwise.

;-)  

This qualifier could, perhaps, be a prefix to the full text, but I do think it helps to be explicit at each point of potential exception.

In effect these becomes the "unless otherwise specified" exception within the "this is the default behaviour" prefix.  

So, should we document the "default behaviour" behaviour more explicitly than "normally"?

Although we are providing a precise description, and I like it, does the general reader understand the context? Do we need to provide some background in the"default behaviour" prefix?

Something like "CSRs are unique within the RV model as their values can change independent of currently executing code or can have side-effects related to execution of that code., including execution of instructions that explicitly access/modify CSRs"

src/csr.tex
```diff
@@ -194,16 +198,19 @@ \subsection*{CSR Access Ordering}
...
-aforementioned three can be freely reordered with respect to FENCE instructions
+aforementioned three can be freely reordered in the global memory order
+with respect to FENCE instructions
 without violating this specification.
 \end{commentary}
 
-For CSR accesses that cause side effects, the above ordering constraints apply
-to the order of the initiation of those side effects but does not necessarily
-apply to the order of the completion of those side effects.
-
```
gfavor, line 204: Should this be left in place until the "rules for the reordering of CSR accesses in the global memory order" are added to the RVWMO chapter.  Maybe with a note that this statement is part of what should be moved.

src/csr.tex
```diff
@@ -157,19 +157,23 @@ \section{CSR Instructions}
...
+In particular, unless specified otherwise, a CSR access is performed
 after the execution of any prior instructions in program order whose behavior
 modifies or is modified by the CSR state and before the execution of any
 subsequent instructions in program order whose behavior modifies or is modified
-by the CSR state. Furthermore, a CSR read access instruction returns the
-accessed CSR state before the execution of the instruction, while a CSR write
-access instruction updates the accessed CSR state after the execution of the
+by the CSR state.
+Any side effects from a CSR access are likewise observed to occur
+synchronously in program order, unless documented otherwise.
```
dkruckemyer-ventana, line 168: What does "synchronously" mean in this context? Does this imply "atomically" as well (since the side-effect statement was removed below)?

For example, assume there is an interrupt pending and that interrupt is disabled. If a CSR write enables that interrupt, does the architecture require the interrupt to be taken on the next instruction? Or does the architecture imply that is the behavior unless the implementer states otherwise?

src/csr.tex
```diff
@@ -194,16 +198,19 @@ \subsection*{CSR Access Ordering}
...
 strongly ordered, as defined by the Memory-Ordering PMAs section in Volume II
 of this manual. Accesses to strongly ordered CSRs have stronger ordering
 constraints with respect to accesses to both weakly ordered CSRs and accesses
 to memory-mapped I/O regions.
+
+\begin{commentary}
+The rules for the reordering of CSR accesses in the global memory order
+should probably be moved to Chapter~\ref{ch:memorymodel} concerning the
+RVWMO memory consistency model.
+\end{commentary}
```
dkruckemyer-ventana, line 216: With regards to the interaction between the memory ordering model and the CSR access ordering model, it seems appropriate to discuss those behaviors in the RVWMO section.

Generally speaking, though, I think this section belongs in a broader "instruction execution" model, that ties together all accesses and modifications of architectural state, including implicit memory accesses (e.g. instruction fetches and table walks), explicit memory accesses, CSR accesses, traps, etc. (In general, RV seems to require a pretty strict "sequential execution model," but there are exceptions; otherwise, there wouldn't be FENCE.I or SFENCE.VMA....)

RVWMO only captures a portion of the interesting events, so perhaps that would be expanded or incorporated into a broader conceptual model some day. (Or we just wait for the formal model....?)

src/csr.tex
jhauser-us: David Horner wrote:
> yes. I agree but perhaps:
> > by the same instruction.
>
> to be more inclusive of any concurrent change including side effects of the instruction, with wording:
> > during the same instruction, ...

The ISA has no requirements for what must happen "during an instruction" except what is caused by the instruction itself.  For that reason, "during the same instruction" wouldn't restrict anything more than the words "by the same instruction" do.

When executing a CSR instruction, the latter wording gives the hardware the freedom to make independent, concurrent reads/writes to the same CSR appear to occur entirely before or after the CSR instruction, or to be suppressed (because the independent accesses occur after the instruction reads the CSR but before the instruction writes the CSR), or some mixture.  Again, because the ISA never says what independent events must occur "during an instruction", all of these possible outcomes seem legitimate to me, at least until there are other rules with more to say (presumably about specific CSRs and specific independent reads/writes).  But as those other rules wouldn't contravene what is written here about suppressing implicit CSR accesses by the same instruction, we also don't need to add here the usual fudge, "unless specified otherwise".

src/csr.tex
jhauser-us: "Synchronously" is intended to mean that, by default, the full consequences of any side effects are expected to be observable by the very next instruction (to the extent they are observable).  I think the word *atomic* has implications with respect to other, independent observers that I don't intend.

> For example, assume there is an interrupt pending and that interrupt is disabled. If a CSR write enables that interrupt, does the architecture require the interrupt to be taken on the next instruction?

If the interrupt is pending in CSR mip, I've always believed that was an existing requirement.  I would be surprised to learn it's not.

> Or does the architecture imply that is the behavior unless the implementer states otherwise?

That raises a good point.  By "documented" I meant not by an implementor but by the standard(s), assuming we're talking about a standard CSR.

Is "unless specified otherwise" sufficient to plug that hole?  That same crutch already appears in a few places in the documents, so if it's also ambiguous, we have a larger problem.

src/csr.tex
jhauser-us: I agree there is still much to be done to standardize and document these matters, just as you say.  The comment I've added is merely a temporary patch, obviously.

src/csr.tex
dkruckemyer-ventana: > 
> 
> "Synchronously" is intended to mean that, by default, the full consequences of any side effects are expected to be observable by the very next instruction (to the extent they are observable). I think the word _atomic_ has implications with respect to other, independent observers that I don't intend.

OK. Makes sense.

> 
> > For example, assume there is an interrupt pending and that interrupt is disabled. If a CSR write enables that interrupt, does the architecture require the interrupt to be taken on the next instruction?
> 
> If the interrupt is pending in CSR mip, I've always believed that was an existing requirement. I would be surprised to learn it's not.

Hmmm.... Originally I was thinking that, in general, we didn't want to mix the notions of "synchronous execution" and "side-effects" because doing so would imply that a CSR access may have high synchronization or hazarding costs. So I had been separating the idea of a side-effect, which I considered to be asynchronous, from the simple change of architectural state due to the CSR access that could affect the next instruction in program order. Along those lines, I had been considering that recognizing an interrupt was one such asynchronous side-effect, so that taking a trap wasn't required immediately after enabling a pending interrupt.

Regardless of what further arguments I could make, I understand the new text implies that all side-effects of a CSR access are synchronous unless otherwise allowed by the specification. That seems like a tighter description and clearly leads to your conclusion regarding interrupts in all cases.

> 
> > Or does the architecture imply that is the behavior unless the implementer states otherwise?
> 
> That raises a good point. By "documented" I meant not by an implementor but by the standard(s), assuming we're talking about a standard CSR.
> 
> Is "unless specified otherwise" sufficient to plug that hole? That same crutch already appears in a few places in the documents, so if it's also ambiguous, we have a larger problem.

To me, at least, "specified" concerns itself with the architectural specifications, while "documented" seems broader (but that's probably because I've spent a bit to much time pouring over "product documentation"). My vote would be for "specified."

David-Horner:
My vote is also for "specified" .
Of course it implies "documented" as the specification is in the 
documentation, specifically that for the architectural specification.

gfavor:
I also vote for "specified".

gfavor:
On Mon, Jun 29, 2020 at 11:43 AM John Hauser <notifications@github.com>
wrote:

> For example, assume there is an interrupt pending and that interrupt is
> disabled. If a CSR write enables that interrupt, does the architecture
> require the interrupt to be taken on the next instruction?
>
> If the interrupt is pending in CSR mip, I've always believed that was an
> existing requirement. I would be surprised to learn it's not.
>
I want to expand a little bit on this example.  Whether one uses the word
"synchronous" or some similar word, there can still be some
interpretational ambiguity by some readers.  From an architectural/software
perspective, the only way to judge that an interrupt is "pending" is to
observe that it is pending via mip.  Until that point of observation, the
delay from some device out in the system asserting an interrupt request to
the relevant "interrupt request pending" signal inside a hart being
asserted such that hart logic can recognize it as "pending", is
unknown/uncertain.

The point being that the only testable version of the "synchronous"
architectural requirement is that an enabled interrupt that would be
observable as pending in mip must be recognized and taken before that CSR
instruction (and of course after the last CSR write instruction that
enabled the interrupt).

The extreme case of this would be the scenario of back-to-back CSR write
instructions that enable and then disable the interrupt.  The architectural
requirement implies that an interrupt that is observable as pending in mip
must be recognized and taken in between these two instructions.  (Btw, this
is exactly what x86 is like.)

Greg

allenjbaum:
Thanks for giving the compliance TG some guidance for testing interrupt
handling :-).

Would you expect that, if there were 2 IP bits set when the
enabling/disabling pair of instructions was executed, that both would be
taken before disabling? (So, executing an xRET would immediately cause the
second interrupt to be taken before the CSR write that disabled interrupts
could take effect - assuming the second IP bit wasn't cleared).
Ditto, but if only a single IP bit were set when the interrupt is taken,
but another appears before the xRET.
It gets hairier when delegation is involved, of course, but that should
mostly affect which CSR bits must be written to enable the interrupt.

These are actually tests that I would have included, but would just let the
formal model decide what would happen instead of knowing the answer, which
begs the question: what does the formal model do in these cases?

It's hard to test interrupt handling without explicitly setting a bit in
MIP with a CSRRS.

gfavor:
Comments below:

On Tue, Jun 30, 2020 at 12:09 PM Allen Baum <notifications@github.com>
wrote:

> Thanks for giving the compliance TG some guidance for testing interrupt
> handling :-).
>
> Would you expect that, if there were 2 IP bits set when the
> enabling/disabling pair of instructions was executed, that both would be
> taken before disabling? (So, executing an xRET would immediately cause the
> second interrupt to be taken before the CSR write that disabled interrupts
> could take effect - assuming the second IP bit wasn't cleared).
>

Yes.  That would seem to be the unavoidable implication or extension of
what I described.  (And this is what would happen on x86 as I remember.)


> Ditto, but if only a single IP bit were set when the interrupt is taken,
> but another appears before the xRET.
>

As long as that new interrupt is observable as being pending in mip, then
Yes.

Greg


> It gets hairier when delegation is involved, of course, but that should
> mostly affect which CSR bits must be written to enable the interrupt.
>
> These are actually tests that I would have included, but would just let the
> formal model decide what would happen instead of knowing the answer, which
> begs the question: what does the formal model do in these cases?
>
> It's hard to test interrupt handling without explicitly setting a bit in
> MIP with a CSRRS.
>
>

jhauser-us:
Note also #546.

PR #539, opened 2020-07-13
Add an "lga" pseudoinstruction for GOT loads even in non-PIC

jrtc27:
This is useful for cases where compilers and/or assembly writers want to force the use of a GOT. For example, this could be used to address [1], but more generally it's a useful thing to have, and is a strange hole in the set of pseudoinstructions provided.

Whilst here, alter the description of "la" to not mention absolute. All addresses are absolute, but highlighting it for "la" specifically is confusing as it potentially implies that it's using an absolute lui/addi sequence rather than a PC-relative sequence.

[1] https://github.com/riscv/riscv-elf-psabi-doc/issues/126

I have not yet created patches against riscv-asm-manual and riscv-elf-psabi-doc. However, those are already outdated and only mention `la` without `lla`. Defining `lga` will also make it far less confusing to fix that; then, `lla` and `lga` can be documented individually, with `la` simply saying it's equivalent to the relevant one depending on -fPIC.

### DIFF
--- a/src/assembly.tex
+++ b/src/assembly.tex
@@ -59,12 +59,14 @@ RISC-V pseudoinstructions.
 \begin{tabular}{l l l}
 pseudoinstruction & Base Instruction(s) & Meaning \\ \hline
 
-\tt la rd, symbol (\emph{non-PIC}) & {\tt auipc rd, ${\tt delta[31:12]} + {\tt delta[11]}$} & Load absolute address, \\
+\tt la rd, symbol (\emph{non-PIC}) & {\tt auipc rd, ${\tt delta[31:12]} + {\tt delta[11]}$} & Load address, \\
                   & {\tt addi rd, rd, delta[11:0]}                         & where ${\tt delta} = {\tt symbol} - {\tt pc}$ \\[1ex]
-\tt la rd, symbol (\emph{PIC})& {\tt auipc rd, ${\tt delta[31:12]} + {\tt delta[11]}$} & Load absolute address, \\
+\tt la rd, symbol (\emph{PIC})& {\tt auipc rd, ${\tt delta[31:12]} + {\tt delta[11]}$} & Load address, \\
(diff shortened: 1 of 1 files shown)


jrtc27:
Yeah, this stuff feels like it doesn't belong here, but that's where it is at the moment... 🤷‍♀️

jim-wilson:
lga may not work when compiling without -fpic, but if so we can give an error for that in the assembler or linker.

FYI In GNU as, both la and lld are equivalent to li when the symbol is a constant address.  E.g.
	.equ foo, 0x12345678
	la a0, foo
	lla a1, foo
expands to two lui/addi sequences.  But that is a detail that doesn't need to be documented here.

jrtc27:
It should work just fine. In that case, the linker will just put the absolute address in the GOT.

jrtc27:
(and your constant address case is more of a fact that `l(l)a` are relatively permissive and let you pass them constant values like `la a0, 1`; `.equ` is more like `#define` than an actual symbol)

aswaterman:
I think we should move this to https://github.com/riscv/riscv-asm-manual

jrtc27:
To be honest the whole chapter should just be removed and superseded by that, no?

aswaterman:
Yes. The ISA spec does depend on the ASM spec because of code examples, but it would suffice to cite it and reference it early on.

kito-cheng:
Maybe @Nelson1225 could implement this on binutils.

Nelson1225:
> Maybe @Nelson1225 could implement this on binutils side.

OK I can do this.

Nelson1225:
> It should work just fine. In that case, the linker will just put the absolute address in the GOT.

Yeah, if it is a static link, then the linker behavior should be the same as @jrtc27 said.

jrtc27:
(and for dynamic linking without PIC it’s just another R_RISCV_32/64 relocation for the run-time linker)

aswaterman:
I reposted this issue to the riscv-asm-manual repo.  https://github.com/riscv/riscv-asm-manual/issues/50

Issue #544, opened 2020-07-21
Is it appropriate to specify an exception vs interrupt priority?


GregAC:
The current privileged spec specifies:

> Synchronous exceptions are of lower priority than all interrupts

I feel this may be better off being left as implementation defined or the meaning of priority in this instance should have further clarification.

Consider the following case:

1. Core fetches a load instruction and begins its execution
2. Core issues a load request on its external memory interface (i.e. it's not being satisfied by a cache if one exists)
3. Before a response to the load request is received an interrupt is raised
4. When the response to the load request does appear it signals a bus error which results in a synchronous exception (either interrupt and bus error response are seen on the same cycle or bus error response occurs after interrupt)

The obvious interpretation of the line I quote from the spec would be that in this scenario we should take the interrupt (whether we see the bus error response on the same cycle or not). However what do we then do with our pending load/pending exception on bus error?

If the response has yet to be received for the load then we need to keep some state around so we're aware there is a pending load response along with some information about the instruction and similarly if we saw the error response the same cycle as the interrupt we need some state to indicate a 'pending' exception with logic to handle the special pending load -> pending exception transition when the response returns. For larger cores this may be perfectly fine and indeed just 'fall out' of the existing micro-architectural design of a super-scalar out of order core. At the other end of the spectrum with tiny 2 or even 1 stage in-order pipelines this adds extra area and complexity.

Then there's also the question of how a 'pending exception' should be dealt with. One answer would be to immediately take the exception upon returning from an interrupt with an mret that results in a jump back to the code that saw the exception though I don't think the architecture strictly specifies an interrupt handler must end with mret (though it is probably inadvisible for software to use a different mechanism). You also have to deal with nested cases where you have a pending exception and end up with another interrupt and exception co-incident to get a second pending exception. For further complexity you could end having the same PCs both times. Clearly a big hardware stack of pending exceptions is not something you want to architecturally mandate.

In some cases on taking the interrupt we could set the mepc to the currently executing load and ignore any response received for it, so we simply repeat the load on return from the interrupt. However whether this is permissible is highly system/implementation dependent. For example repeating the load may fine when loading from some 'ordinary' memory but would not be fine when the load targets some device register where a read triggers some action (such as popping from a FIFO) and clearly wouldn't be fine for stores.

[...]

PR #545, opened 2020-07-22
Pmp wording fix

sorear:
First patch is editorial and tangentially related to #544; the second patch is a functional clarification since the sentence is ambiguous and interpreted differently by [spike](https://github.com/riscv/riscv-isa-sim/blob/67b7edd0270d5d074136eb45474477d15aeee725/riscv/processor.cc#L809) and [ibex](https://github.com/lowRISC/ibex/blob/d7c13aa9045838e7b9751e87b6b4630e7893db39/rtl/ibex_cs_registers.sv#L828).  The spike interpretation makes more sense to me.


aswaterman:
Thanks.  Please avoid unnecessary whitespace changes in future.

PR #546, opened 2020-07-23
Clarify what is a side effect of a CSR access

jhauser-us:


aswaterman:
Thanks for following through with this.

PR #559, opened 2020-08-10
Add scontext, hcontext, and mcontext CSRs for Debug

ernie-sifive:
These registers may be written with process ID information by operating systems running in their respective privilege levels.  A debugger may use tdata3 (aka textra) to  qualify breakpoints based on a comparison of context register(s) with corresponding trigger field(s).  These registers and comparison fields are described in the RISC-V Debug Specification.

### DIFF
--- a/src/priv-csrs.tex
+++ b/src/priv-csrs.tex
@@ -225,6 +225,10 @@ Number    & Privilege & Name & Description \\
 \hline
 \tt 0x180 & SRW  &\tt satp       & Supervisor address translation and protection. \\
 \hline
+\multicolumn{4}{|c|}{Debug/Trace Registers} \\
+\hline
+\tt 0x5A8 & SRW &\tt scontext & Supervisor-mode context register. \\
+\hline
 \end{tabular}
 \end{center}
 \caption{Currently allocated RISC-V supervisor-level CSR addresses.}
@@ -258,6 +262,10 @@ Number    & Privilege & Name & Description \\
 \multicolumn{4}{|c|}{Hypervisor Protection and Translation} \\
 \hline
 \tt 0x680 & HRW  &\tt hgatp      & Hypervisor guest address translation and protection. \\
+\hline
+\multicolumn{4}{|c|}{Debug/Trace Registers} \\
+\hline
+\tt 0x6A8 & HRW &\tt hcontext & Hypervisor-mode context register. \\
 \hline
 \multicolumn{4}{|c|}{Hypervisor Counter/Timer Virtualization Registers} \\
 \hline
@@ -378,8 +386,6 @@ Number    & Privilege & Name & Description \\
 \tt 0x7A1 & MRW &\tt tdata1 & First Debug/Trace trigger data register. \\
 \tt 0x7A2 & MRW &\tt tdata2 & Second Debug/Trace trigger data register. \\
 \tt 0x7A3 & MRW &\tt tdata3 & Third Debug/Trace trigger data register. \\
-\tt 0x5A8 & SRW &\tt scontext & Supervisor-mode context register. \\
-\tt 0x6A8 & HRW &\tt hcontext & Hypervisor-mode context register. \\
 \tt 0x7A8 & MRW &\tt mcontext & Machine-mode context register. \\
 \hline
 \multicolumn{4}{|c|}{Debug Mode Registers } \\
--- a/src/priv-csrs.tex
+++ b/src/priv-csrs.tex
@@ -124,9 +124,6 @@ for use by the debug system.  Of these CSRs, {\tt 0x7A0}--{\tt 0x7AF} are
 accessible to machine mode, whereas {\tt 0x7B0}--{\tt 0x7BF} are only visible
 to debug mode.  Implementations should raise illegal instruction exceptions on
 machine-mode access to the latter set of registers.
-In addition, supervisor-mode standard read-write CSRs {\tt 0x5A0}--{\tt 0x5AF}
-and hypervisor-mode standard read-write CSRs {\tt 0x6A0}--{\tt 0x6AF} are
-reserved for use by the debug system.
 
 \begin{commentary}
 Effective virtualization requires that as many instructions run natively as
--- a/src/priv-csrs.tex
+++ b/src/priv-csrs.tex
@@ -124,6 +124,9 @@ for use by the debug system.  Of these CSRs, {\tt 0x7A0}--{\tt 0x7AF} are
 accessible to machine mode, whereas {\tt 0x7B0}--{\tt 0x7BF} are only visible
 to debug mode.  Implementations should raise illegal instruction exceptions on
 machine-mode access to the latter set of registers.
+In addition, supervisor-mode standard read-write CSRs {\tt 0x5A0}--{\tt 0x5AF}
+and hypervisor-mode standard read-write CSRs {\tt 0x6A0}--{\tt 0x6AF} are
+reserved for use by the debug system.
 
 \begin{commentary}
 Effective virtualization requires that as many instructions run natively as
@@ -375,6 +378,9 @@ Number    & Privilege & Name & Description \\
 \tt 0x7A1 & MRW &\tt tdata1 & First Debug/Trace trigger data register. \\
 \tt 0x7A2 & MRW &\tt tdata2 & Second Debug/Trace trigger data register. \\
 \tt 0x7A3 & MRW &\tt tdata3 & Third Debug/Trace trigger data register. \\
+\tt 0x5A8 & SRW &\tt scontext & Supervisor-mode context register. \\
+\tt 0x6A8 & HRW &\tt hcontext & Hypervisor-mode context register. \\
+\tt 0x7A8 & MRW &\tt mcontext & Machine-mode context register. \\
 \hline
 \multicolumn{4}{|c|}{Debug Mode Registers } \\
 \hline


src/priv-csrs.tex
```diff
@@ -375,6 +378,9 @@ \section{CSR Listing}
 \tt 0x7A1 & MRW &\tt tdata1 & First Debug/Trace trigger data register. \\
 \tt 0x7A2 & MRW &\tt tdata2 & Second Debug/Trace trigger data register. \\
 \tt 0x7A3 & MRW &\tt tdata3 & Third Debug/Trace trigger data register. \\
+\tt 0x5A8 & SRW &\tt scontext & Supervisor-mode context register. \\
+\tt 0x6A8 & HRW &\tt hcontext & Hypervisor-mode context register. \\
```
jhauser-us, line 382: The `scontext` and `hcontext` CSRs need to go into the respective tables for supervisor CSRs and hypervisor CSRs.

src/priv-csrs.tex
```diff
@@ -124,6 +124,9 @@ \section{CSR Address Mapping Conventions}
 accessible to machine mode, whereas {\tt 0x7B0}--{\tt 0x7BF} are only visible
 to debug mode.  Implementations should raise illegal instruction exceptions on
 machine-mode access to the latter set of registers.
+In addition, supervisor-mode standard read-write CSRs {\tt 0x5A0}--{\tt 0x5AF}
+and hypervisor-mode standard read-write CSRs {\tt 0x6A0}--{\tt 0x6AF} are
+reserved for use by the debug system.
```
jhauser-us, line 129: I'm concerned that this makes it sound like the debug spec can allocate registers in these ranges at will, which is not what I hoped we would say.  I might be in the minority on this, but I think I'd prefer not to add this sentence.

src/priv-csrs.tex
ernie-sifive: done

src/priv-csrs.tex
ernie-sifive: Works for me.  Change is made.

ernie-sifive:
@jhauser-us Hi John, is there anything further I need to do here? What else has to happen to get this approved and merged?

sorear:
meta: you're adding new S-mode state which is enabled by default & impossible for M-mode to disable, which means a hypervisor needs to know this extension exists in order to avoid creating a storage channel.  How does a hypervisor know that and what happens with (draft) hypervisors that exist already?

ernie-sifive:
All debug features are optional.  If a hypervisor doesn't know about or doesn't support swapping scontext, that's ok.  It just means that hardware breakpoints effectively can't be used.  This is the current state of the world.

An "enlightened" hypervisor will check whether hcontext and scontext exist and handle setting hcontext and swapping scontext.  The debugger is above all of that and will present hardware breakpoints as an option if and only if all the components (hypervisor, OS, and debug hardware) in the system support context qualification.

sorear:
If a hypervisor doesn't know about or support swapping scontext, it is insecure to use on hardware that implements scontext.  This has nothing to do with hardware breakpoints per se; any new register that can be read and written introduces a covert storage channel if it is enabled by default.

ernie-sifive:
@sorear So you are concerned about one VS-mode OS writing something to scontext and it being seen by another VS-mode OS because the hypervisor didn't know to swap the scontext state.

To be clear, this PR isn't changing the behavior of scontext.  It is only changing the CSR address of scontext to make it compatible with the recommendations in the Priv Arch spec regarding accessibility.  Previously, the ratified Debug Spec defined scontext at CSR address 0x7aa (MRW range) but with the descriptive text "_This optional register is only writable in S mode, M mode and Debug Mode.  Note that the register number does not follow the read and write accessibility of the CSRs according to privilege level as defined in the Privileged Spec._"

That said, I understand your concern.  I think you are suggesting that scontext and possibly hcontext should have enable bits so that they remain disabled unless the components making up a particular system support this feature.  Do you have any suggestion where the enable bit(s) should be?  I think the enable needs to be writable from at least Debug Mode and M-mode, and maybe scontext-Enable should be writable from HS-mode.

sorear:
You correctly understand my concern.  I don't have a suggestion for where the enable bit should live.

There also needs to be some thought given to how to prevent breakpoints intended for the guest from firing in the host trap code before the context registers have been changed, but I suspect I'm just not finding the relevant discussion.

ernie-sifive:
@sorear I have updated my Debug Spec PR related to this feature to add enable bits for `scontext` and `hcontext`.  The enables default to 0 and must be written to 1 in M-mode or Debug mode before the `scontext` and `hcontext` registers can be written.  See section 5.5.6 of https://github.com/riscv/riscv-debug-spec/blob/scontext2-csr/riscv-debug-draft.pdf.

Triggers meant for a guest would be qualified to trigger in VU mode only.  There are bits in the control register of every trigger type to configure which modes the trigger can occur in.  Hypervisor-related changes to the Debug Spec, including adding mode enables for VU and VS modes, are mostly being prepared in a separate PR.

sorear:
@ernie-sifive Thanks, LGTM so far.

For the other `hwhatever` CSRs, access to them is defined to trap in VS-mode, so they they can be emulated to support nested virtualization — I think the PR does that, but it could be said more explicitly.

Looking forward to seeing the "separate PR" when it's ready.  If it helps, I think you can share a single bit between U and VU since there are no direct privilege transitions between them, and VS will have an opportunity to change hcontext between them.

jhauser-us:
Stefan O'Rear wrote:
> you're adding new S-mode state which is enabled by default & impossible for M-mode to disable, which means a hypervisor needs to know this extension exists in order to avoid creating a storage channel. How does a hypervisor know that and what happens with (draft) hypervisors that exist already?

As long as the hypervisor extension remains officially in draft, we are not guaranteeing not to add new CSRs that a hypervisor must swap.

There's an inherent conflict between the desires of implementors who want the draft to be fixed in stone *right now* and the need for sufficient experience and feedback before cutting off all changes.  Since the hypervisor draft hasn't been elevated to *frozen* status yet, spec changes that force a corresponding change to software are most certainly not ruled out.

sorear:
> As long as the hypervisor extension remains officially in draft, we are not guaranteeing not to add new CSRs that a hypervisor must swap.

Adding new CSRs which are guaranteed to exist in all implementations, and which a hypervisor can swap unconditionally after the draft is updated, would be fine.  But this is proposing a new optional CSR.  Since S-mode code cannot detect features by probing, it cannot know whether it needs to swap the register, even if it implements the final spec!

aswaterman:
> Since S-mode code cannot detect features by probing

Standard platforms that implement S-mode will raise a precise illegal-instruction trap that's visible to S-mode:

```
  la t0, 1f
  csrw stvec, t0
  csrr x0, scontext
  # feature exists
.align 2
1:
  # feature doesn't exist
```

sorear:
@aswaterman I'm having trouble finding text in the as-ratified privileged spec that forbids a non-standard CSR from decoding number 0x5A8?

aswaterman:
Hence my use of the word "platform", which can mandate that custom extensions not occupy standard encoding space.

Obviously, if there is a custom CSR at that address, the same side channel can possibly exist, and we aren't attempting to solve that problem.

avpatel:
We don't need hcontext CSR because Hypervisor is also S-mode software so Hypervisor itself can use scontext of HS-mode.

What we need is vscontext CSR to allow HS-mode access VS-mode context.

Regards,
Anup

aswaterman:
@avpatel I think Ernie's scheme and your scheme can both work.  As for Ernie's scheme, `hcontext` describes the guest OS (if any), whereas `scontext` describes the context of the currently executing U-mode or VU-mode program.  The debugger knows which register to examine based on the interrupted privilege mode.  If the mode is U, then `mcontext` and `scontext` matter.  If the mode is S or HS, then only `mcontext` matters.  If the mode is VS, then `mcontext` and `hcontext` matter.  If the mode is VU, then all three registers matter.  Did I get that right, @ernie-sifive?

I think many systems will choose to omit mcontext, since it's implicit that there's only one hypervisor or supervisor (agreed, @ernie-sifive?)

sorear:
If M-mode runs a multithreaded RTOS, even if there is no lower privilege level implemented, mcontext is useful to distinguish threads.  Likewise scontext can be used to distinguish kernel threads in S-mode.  For hcontext it may be useful to have hcontext set while running hypervisor code on behalf of a single VM (possibly less interesting for breakpoints, but more so for tracing); this probably requires the hypervisor to reserve a specific hcontext value (probably 0) for "not on behalf of any VM".

gfavor:
It should be noted that there was discussion (which I had raised) as to
whether there should also be a vscontext CSR.

Having the additional vscontext CSR would be ideal (mirroring what the
H-extension does for most but not all CSR's).  But is the extra hardware
justified?  Would the alternative and compromise be acceptable, i.e. of a
hypervisor directly context switching scontext between VMs, and between
them and its own use of scontext?  And that the switching of the active
scontext value wouldn't change at exactly the moment of entering/exiting a
VM.  The answer was 'yes' those would be acceptable compromises and the
preferred (lower cost) option for debug support purposes.  (Note that there
is also support in the Debug Spec for direct ASID and VMID matching that
will, of course, switch in and out precisely on VM entry/exits.)

Similarly, regarding the mcontext CSR (which exists in the ratified spec)
and the new hcontext CSR, the preferred choice was to make these the same
single physical register - with the legacy mcontext alias and the new
hcontext alias referring to the same physical register.

Greg

gfavor:
As hopefully is apparent from the other emails, scontext is intended to
hold OS process-related context info, while hcontext holds VM-related
context info.  Then the addition of vscontext would allow the clean
separation of HS/U-mode scontext and VS/VU-mode scontext.  But it was felt
by Ernie and others in the debug group that the extra hardware to support
that ideal behavior was unnecessary (i.e. of limited value).

Greg

jhauser-us:
The hypervisor chapter says we create VS CSRs only when we need to.  It appears to me that this is not such a case.

Stefan O'Rear wrote:
>  this probably requires the hypervisor to reserve a specific hcontext value (probably 0) for "not on behalf of any VM".

Yes.

avpatel:
@gfavor, we have to save/restore CSR in low-level world-switch of KVM if we don't corresponding  VS-CSR . For example, save/restore of scounteren will end-up in KVM low-level world-switch.

The KVM world-switch is in hot-path for every trap and interrupt.

Having VS-CSR help us lazily save/restore CSR outside low-level KVM world-switch.

I suggest we should also have vsscounteren CSR

Regards,
Anup

gfavor:
Exactly.  That was the thought process by Ernie and others in the debug
group.

ernie-sifive:
@aswaterman and @gfavor are correct.  The underlying idea is for the context registers to identify the specific application in effect at any time so that hardware breakpoints intended for one application do not falsely trigger in another.   scontext is the U or VU application identification.  If there is a hypervisor, then hcontext identifies the VM.   The hypervisor would assign a unique hcontext ID for applications of its own (i.e. U mode) -- zero, maybe.  The debugger does have to know the proper hcontext and scontext to program into the hardware breakpoint qualification.

avpatel:
@jhauser-us @gfavor Even if a CSR is not frequently accessed by S-mode software still absence of VS-CSR forces us to save/restore it in low-level KVM world switch.

Regards,
Anup

gfavor:
As John just noted "The hypervisor chapter says we create VS CSRs only when
we need to."

The chapter also says "A few standard supervisor CSRs (scounteren) have no
matching VS CSR. These supervisor CSRs continue to have their usual
function and accessibility even when V=1, except with VS-mode and VU-mode
substituting for HS-mode and U-mode. Hypervisor software is expected to
manually swap the contents of these registers as needed"

gfavor:
I have to leave this to John - as architect - to respond to.

Greg

avpatel:
@ernie-sifive, the vsscontext CSR is for identifying process/application running inside Guest/VM without actively context switching scontext CSR it in low-level KVM world-switch

It's for better KVM performance and in-general any Type-2 Hypervisor. The Type-1 Hypervisor don-t face  performance issue for missing VS-CSR

Regards,
Anup

avpatel:
I suggest having four CSRs mcontext, hcontext, vsscontext, and scontext for good KVM world-switch performance.

On similar lines we should also have vsscounteren CSR

Regards,
Anup

gfavor:
One question I have - to all concerned - is:  What value would a standard
hypervisor put into hcontext (and similarly what would a standard OS put
into scontext), and how would a debugger come to know the value assigned to
a specific process or VM?

Or would these CSRs be used *only* in more "custom" embedded systems where
the hypervisor/OS code would be instrumented to use these CSRs and put
whatever useful values into them for debug purposes.  Certainly for
"standard software" debug purposes VMID and ASID based trigger filtering
can be used (which doesn't involve these *context CSRs).

Greg

aswaterman:
> One question I have - to all concerned - is: What value would a standard hypervisor put into hcontext (and similarly what would a standard OS put into scontext), and how would a debugger come to know the value assigned to a specific process or VM?

@gfavor See discussion here: https://github.com/riscv/riscv-debug-spec/pull/535 -- TL;DR is that whatever opinions you hold of ARMv8's CONTEXTIDR_EL2 and CONTEXTIDR_EL1 would apply here.  The OS would put the PID into scontext.

The ASID approach has been suggested (including by yours truly), but it was consciously rejected, because the mapping from ASID back to PID isn't available to the debugger, may change over time, and may even be nonexistent, depending on the ASID-management scheme.

gfavor:
Thanks.  Also note that the debug spec says (for scontext): " It’s
recommended to implement no more than 16 bits on RV32, and 34 on RV64."
(This reflects a hard constraint elsewhere in the spec on what the compare
value can be.)  I'll assume a PID fits in 34 bits (?), but probably not 16
bits.

For similar reasons, the constraint on the size of context values is 7 bits on
RV32, and 14 bits on RV64.

Greg

P.S. This all goes to whether "standard" hypervisor software (and OS
software) will support these *context CSRs with meaningful values given the
size constraints?  Or, practically speaking, are *context CSRs only usable
by a suitably instrumented embedded OS/hypervisor?

gfavor:
That is "hcontext values" below.

aswaterman:
I'm not sufficiently familiar with the debug spec to understand the hard caps of 16 and 34, but I know 32-bit Linux supports a 15-bit PID space, and 64-bit Linux supports 15-22 bits.  So the scontext restrictions are sufficiently generous to handle current Linux, at least.

gfavor:
Thanks.  A few observations:

- 16 bit ASIDs and 15-22 bits aren't radically different sizes.

- External debug is more interesting for embedded systems - that maybe have
PID sizes towards the lower end of that 15-22 bit range?  (Not that ASIDs
would all of a sudden be exact proxies for PIDs.)

- In a self-hosted debug situation for higher-end systems, the debugger is
attached (loosely speaking) to a process to be debugged and so would the
ASID be more readily knowable to that debugger as part of it being set up
to debug that process??

Note that I am far from knowledgeable in this area, but am questioning what
- in actual practice - happens wrt scontext usage (and eventually also
hcontext usage).  Irrespective of what can happen in theory, does RV Linux
currently maintain scontext as it switches between processes?  Or is this
left as a some form of custom config of Linux?  In general, when and how is
scontext actually being supported in RV systems today?

Greg

aswaterman:
> Note that I am far from knowledgeable in this area, but am questioning what - in actual practice - happens wrt scontext usage (and eventually also hcontext usage). Irrespective of what can happen in theory, does RV Linux currently maintain scontext as it switches between processes? Or is this left as a some form of custom config of Linux? In general, when and how is scontext actually being supported in RV systems today?

@gfavor I hope our more knowledgeable colleagues can answer your other questions, but I think I can answer this one.  scontext is not currently populated by the Linux kernel.  Presumably, we would follow the same path as some other ISAs' kernel ports: we would provide a Kconfig option to bake the scontext write into the context-switch path.  By default, this code would be statically excluded, and so scontext would be an optional feature.  (As a further extension, there could be a dynamic check (or even self-modification), but that might prove to be over-engineered.)

sorear:
If scontext is enabled at runtime after consent by software, as proposed in [the current PR](https://github.com/riscv/riscv-debug-spec/pull/535/files#diff-079b8e5f673e0f18ecddf999b26a2f8bR118), then hypervisors can also treat scontext as an optional feature, not touching it at all in some configurations.

If scontext is always enabled but must be probed as described above, then hypervisors always need to probe it and switch it in the context switch path.

Requiring scontext to be enabled allows hypervisors to treat it as optionally as supervisors can.

avpatel:
For Linux RISCV, the preferred approach will be to detect scontext CSR at boot-time and at runtime we do things differently in context-switch based on availability of scontext CSR.

The rationale behind above approach is to support single kernel Image that works on existing and future RISC-V platforms.

Regards,
Anup

aswaterman:
@avpatel I think part of the reason it's a Kconfig option is that it's only necessary in a relatively rare use case.  Most users don't even have the capability to attach an external debugger.

I agree with your point in spirit, but I think the ARM Linux folks got this one right by disabling it by default.

avpatel:
@awaterman, For hypervisors, we can't assume what will be running inside Guest/VM so for KVM-RISCV we might endup keeping the kconfig option always enabled.

I think the context CSR can be used for other purposes as well. Maybe tag cache lines with context id to achieve cache level isolation? OR Maybe do cache coloring based in context id?

Are context CSR limited to debugger only?

Regards,
Anup

aswaterman:
@avpatel If your guest assumes the existence of this CSR, you can simply give it an illegal-instruction exception when it attempts to access it.  VMMs aren't emulators; they don't need to support all hypothetical RISC-V configurations.

avpatel:
@awaterman, the scontext CSR won't trap when accessed from VS-mode. Right?

FYI, the KVM ARM64 unconditionally save/restore CONTEXTIDR_EL1 msr for each Guest VCPU.
(Refer, https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/arch/arm64/kvm/hyp/include/hyp/sysreg-sr.h?h=v5.9-rc2)

Regards,
Anup

aswaterman:
@avpatel These are all optional registers, so they might trap on any implementation.  But yes, if scontext is implemented, it won't trap unless disabled (see https://github.com/riscv/riscv-debug-spec/pull/535/files#diff-079b8e5f673e0f18ecddf999b26a2f8bR118).

avpatel:
This means KVM-RISCV will have to detect scontext/vscontext CSR at boot-time and only save/restore if CSR is present.

I think it will be much cleaner and faster for KVM if we have vsscontext for HS-mode on platform having both H-Extension and scontext CSR.

Regards,
Anup

aswaterman:
vsscontext would also be optional, so I don't see how that affects the decision.  Either way, you'd need to conditionalize (statically or dynamically) the management of those registers.

I'll state again, this is not a big deal.  It's a boutique feature.  It's OK if we treat it as a compile-time, opt-in feature.  Then, OSes and hypervisors can just check at boot time that the feature _doesn't_ exist, and assert-fail if it does.

I really think we're spending too much time on this one...

fintelia:
To clarify, the _scontext_ register would only be used for triggering breakpoints in U and VU modes? And _hcontext_ only in VS and VU modes?

ernie-sifive:
@fintelia  scontext would be used for U-mode breakpoints.  Both hcontext and scontext would be used for VU-mode breakpoints.  hcontext alone would be used for VS-mode breakpoints.

gfavor:
>
> I think the context CSR can be used for other purposes as well. Maybe tag
> cache lines with context id to achieve cache level isolation? OR Maybe do
> cache coloring based in context id?
>
> Are context CSR limited to debugger only?
>
This CSR is defined as part of the External Debug spec and it's only
current usage is as part of trigger conditioning.  Any new arch extension,
such as for things you mention, should not be reusing the CSR from an
unrelated arch spec.

gfavor:
I'm in support of this.  It handles the matter in a simple and sufficient
way.

Greg

pdonahue-ventana:
I've seen a couple of comments mentioning that this topic is related to external debug.  Note that these registers can also be used for triggers that are set up to fire for trap-based/self-hosted debug.  I would expect that to be more a more common debug method for implementations that have hypervisor support.

sorear:
"Assert at startup that the extension is not present" is fine by me.  A little bit concerned about the effect on TVM-based hypervisors, but those require some M-mode code and we generally agree that M-mode must be aware of extensions anyway.

ernie-sifive:
Any other objections or discussion?

aswaterman:
I'm pulling the trigger; if there are related but new issues, let's continue the discussion on a new issue or on the tech-privileged list.

PR #565, opened 2020-08-14
Change "reserved for custom" to "available for custom"

jhauser-us:
In Volume I, section 1.3, "RISC-V ISA Overview", the words *reserved* and *custom* are defined as disjoint categories.  I believe the phrase "reserved for custom use" undermines the meaning we usually want to convey with the word *reserved*.


aswaterman:
Do you think "available" sufficiently connotes that "standard extensions will not use this space"?  "Reserved" seems to me to get that point across more clearly.

kasanovic:
I agree "available" sounds too weak.  I'd suggest "designated" or "allocated" instead.

aswaterman:
Superseded by #566.

PR #566, opened 2020-08-15
Change "reserved for custom" to "designated for custom"

jhauser-us:
In Volume I, section 1.3, "RISC-V ISA Overview", the words *reserved* and *custom* are defined as disjoint categories.  I believe the phrase "reserved for custom use" undermines the meaning we usually want to convey with the word *reserved*.


aswaterman:
Thanks.

PR #571, opened 2020-08-25
Change "hardwired to other field" to "read-only field"

jhauser-us:
Instead of describing CSR fields for XLEN (SXL, VSXL, UXL) and endianness (SBE, VSBE, UBE) as possibly being "hardwired" to another field, describe them as possibly being read-only fields with matching values.


aswaterman:
Yeah, this is an improvement.

PR #574, opened 2020-08-28
Implicit reads

dkruckemyer-ventana:
Proposal for clarifying the intended behavior of implicit reads

### DIFF
--- a/src/intro.tex
+++ b/src/intro.tex
@@ -431,12 +431,13 @@ implementation could attempt to read all of main memory at the earliest
 opportunity, cache as many fetchable (executable) bytes as possible for later
 instruction fetches, and avoid reading main memory for instruction fetches ever
 again. An instruction fetch accessing I/O devices, however, may only read the
-bytes within an implementation-defined naturally-aligned power-of-2 region,
-provided that the region was reached non-speculatively. Such a region is reached
-non-speculatively if a simple instruction-by-instruction execution performs a
-non-speculative instruction fetch in that region via sequential code, including
-fetching an instruction that spans adjacent regions, or via a branch, jump, or
-trap. By implication, an implementation may always cache any data read by an
+bytes within an implementation-defined, naturally-aligned power-of-2 region
+containing the address of the instruction fetch, provided that the region was
+reached non-speculatively. Such a region is reached non-speculatively if a
+simple instruction-by-instruction execution performs a non-speculative
+instruction fetch in that region via sequential code, including fetching an
+instruction that spans adjacent regions, or via a branch, jump, or trap. By
+implication, an implementation may always cache any data read by an
 instruction fetch.
 
 In addition, except when specified otherwise, any additional implicit read that
@@ -444,11 +445,12 @@ results from an instruction fetched as described above and that does not raise
 an exception, e.g. a read of an address translation data structure, may be
 performed arbitrarily early and speculatively. An implicit read accessing main
 memory may read any number of bytes, whereas an implicit read accessing I/O
-devices may only read the bytes within an implementation-defined naturally-aligned
-power-of-2 region. This region does need not be the same region as the related
-instruction fetch region, and the size of this region does not need to be the
-same as the size of the related instruction fetch region. By implication, an
-implementation may cache any data read by an implicit read.
+devices may only read the bytes within an implementation-defined,
+naturally-aligned power-of-2 region containing the address of the implicit read.
+This region does need not be the same region as the related instruction fetch
+region, and the size of this region does not need to be the same as the size of
+the related instruction fetch region. By implication, an implementation may
+always cache any data read by an implicit read.
 
 To ensure that certain implicit reads are ordered only after writes to
 the same memory locations, software must execute specific fence or cache-control
--- a/src/intro.tex
+++ b/src/intro.tex
@@ -440,6 +440,13 @@ instruction that spans adjacent regions) or via a branch, jump, or trap. By
 implication, an implementation may always cache any data read by an
 instruction fetch.
 
+(Fetch a bounded number of NAPOT regions, and continue executing within those
+regions)
+
+(The practical restriction is that an implementation that limits speculation 
+can fetch from any memory without PMAs, but if an implementation knows where
+executable memory is, it can fetch more.)
+
 In addition, except when specified otherwise, other implicit reads, e.g.
 reads of an address translation data structure, that do not raise an exception
 may be performed arbitrarily early and speculatively. An implicit read
--- a/src/intro.tex
+++ b/src/intro.tex
@@ -422,32 +422,30 @@ Ordinarily, if an instruction attempts to access memory at an inaccessible
 address, an exception is raised for the instruction.
 Vacant locations in the address space are never accessible.
 
-Except when specified otherwise, an implicit read for an instruction fetch that
-does not raise an exception may
-be performed arbitrarily early and speculatively, i.e. before the instruction
-fetch is required by a simple instruction-by-instruction execution of a program.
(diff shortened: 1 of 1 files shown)


dkruckemyer-ventana:
Note that 3a81a0a implies that implicit reads other than instruction fetches occur only if a corresponding instruction fetch occurs. The result is that other implicit reads, e.g. for translation, cannot run ahead of the instruction fetch, even if those other implicit reads access main memory regions. I'm assuming that's not what we want, so I'll push another edit.

src/intro.tex
```diff
@@ -422,14 +422,38 @@ \section{Memory}
...
+bytes within an implementation-defined, naturally-aligned power-of-2 region
+containing the address of the instruction fetch, provided that the region was
+reached non-speculatively. Such a region is reached non-speculatively if a
+simple instruction-by-instruction execution performs a non-speculative
+instruction fetch in that region via sequential code (including fetching an
+instruction that spans adjacent regions) or via a branch, jump, or trap. By
+implication, an implementation may always cache any data read by an
+instruction fetch.
+
+In addition, except when specified otherwise, additional implicit reads, e.g.
```
aswaterman, line 443: Additional => other?

src/intro.tex
```diff
@@ -422,14 +422,38 @@ \section{Memory}
...
+raise an exception may be performed arbitrarily early and speculatively, i.e.
+before the instruction fetch is required by a simple instruction-by-instruction
+execution of a program. An instruction fetch accessing main memory may read any
+number of bytes and may be performed at any time. For instance, a valid
 implementation could attempt to read all of main memory at the earliest
 opportunity, cache as many fetchable (executable) bytes as possible for later
 instruction fetches, and avoid reading main memory for instruction fetches ever
-again.  To ensure that certain implicit reads are ordered only after writes to
+again. An instruction fetch accessing I/O devices, however, may only read the
+bytes within an implementation-defined, naturally-aligned power-of-2 region
```
aswaterman, line 434: No hyphen after naturally (sorry for being the grammar police!)

src/intro.tex
```diff
@@ -422,14 +422,38 @@ \section{Memory}
...
+instruction that spans adjacent regions) or via a branch, jump, or trap. By
+implication, an implementation may always cache any data read by an
+instruction fetch.
+
+In addition, except when specified otherwise, additional implicit reads, e.g.
+reads of an address translation data structure, that do not raise an exception
+may be performed arbitrarily early and speculatively. An implicit read
+accessing main memory may read any number of bytes and may be performed at any
+time. An implicit read accessing I/O devices, however, may only read the bytes
+within an implementation-defined, naturally-aligned power-of-2 region containing
```
aswaterman, line 448: No hyphen after naturally

aswaterman:
> I'm assuming that's not what we want

Agreed.

aswaterman:
Thanks - I think this is what we want.  We've only discussed this between ourselves and Ken, right?  Perhaps you should send a note to tech-privileged (or tech-base-isa?) about this proposed change.

Can you also add a one-sentence changelog entry in preface.tex?

dkruckemyer-ventana:
Will do. Would you consider this to be a compatible change, i.e. the second set of bullets?

sorear:
This PR has the effect of prohibiting simple implementations of instruction prefetch unless the prefetch unit is aware of PMAs and I suspect it is incompatible with many existing implementations.  For instance cv32e40p (formerly known as R1SCY):

https://github.com/openhwgroup/core-v-docs/blob/master/cores/cv32e40p/user_manual/source/instruction_fetch.rst

> As a result of this (speculative) prefetch, CV32E40P can fetch up to four words outside of the code region and care should therefore be taken that no unwanted read side effects occur for such prefetches outside of the actual code region.

dkruckemyer-ventana:
The primary goal is to allow speculative instruction fetches (including prefetches) from anywhere in main memory and bounded speculative instruction fetches (including prefetches) from I/O devices, so I believe this implementation meets the goal without needing PMAs. (In fact, this section cannot be defined in terms of PMAs, since this pull request is part of the user spec. Despite the similarity to the PMA description, the terms "main memory" and "I/O devices" are actually introduced near the beginning of the "Memory" section.)

Where I think the conflict occurs is the shape of the bounds. Would you agree @sorear ? If that's all, I can easily accommodate different boundaries for instruction fetches from I/O devices. However, if the prefetcher can read four arbitrary words from anywhere in the address space, that would be against the spirit of this pull request (and probably would also cause system issues).

aswaterman:
If that design has a constraint that the only devices that can be fetched from are memory-like, it’s already compliant. I’m not sure it _should_ be considered compliant if I/O regions can be fetched from in the same manner. But if we decide it should be compliant, we could relax this definition from aligned power-of-2 to some maximum fixed distance before and after.

jrtc27:
Prefetching from I/O space is a terrible idea regardless of whether or not the letter of the spec permits it. We have run into issues when combining a highly out-of-order core with I/O memory and had wedges due to misbehaving AXI slaves when speculatively sent unexpected instruction fetch requests (it was speculatively doing an indirect branch to a nonsense register value due to mis-speculation on an earlier branch), though I could easily have trained the predictor to do silly things like speculatively fetch instructions from the UART with some clever page tables, and even on very simple pipelined in-order cores. It 100% should be forbidden in the spec because otherwise people don't think of these things and end up with buggy cores.

jrtc27:
And I think fetching from non-memory-like I/O regions is something that implementations should be allowed to permit, because whilst it's probably not a useful thing to do, if people really want to do it in some niche case there's nothing inherently wrong with allowing it so long as it's done non-speculatively if side-effecting.

sorear:
My comment was about simple cores that don't know whether any given address is I/O space because they aren't integrated into a larger framework that tracks that information.  Such cores are obviously limited in their ability to speculatively perform memory references (although that can change in the context of a PTE attribute extension) but with the right definitions it's possible to allow _some_ pipelining.

jrtc27:
I would regard cv32e40p as being buggy though as it could be used in an SoC with side-effecting I/O memory directly after DRAM.

dkruckemyer-ventana:
Here's what I think the goals are for this edit:

- Allow unbounded speculation for explicit and implicit accesses to main memory (which _must not_ have side-effects)
- Allow implementation-defined bounded speculation for implicit accesses to I/O devices (which _may_ have side-effects)

An implementation is not required to distinguish main memory from I/O devices, in which case the more restrictive requirement applies.

An open issue is how to define "implementation-defined bounded" and whether those bounds need to be standardized across all implementations.

Thoughts or comments?

jscheid-ventana:
Instead of bounding the I/O access behavior directly, SW can be implored to use some protection mechanism (e.g., PMP) to restrict execute access to such areas if it wants defined behavior.

jrtc27:
> Instead of bounding the I/O access behavior directly, SW can be implored to use some protection mechanism (e.g., PMP) to restrict execute access to such areas if it wants defined behavior.

That has a chicken-and-egg problem: reliably stopping speculative side-effects for I/O memory requires configuring the PMP, but configuring the PMP requires executing code, which still needs to guarantee no speculative side-effects for I/O memory otherwise it would be legal to, say, speculatively write to a GPIO reset and reset the hart before any instruction has even committed.

aswaterman:
Folks, we are _way_ off topic here.  Where did the idea that speculative _writes_ are legal come from, regardless of the memory region?  Furthermore, no one is suggesting that PMP is the recommended solution to handling memory attributes.  Everyone participating in this thread understands why that's an unworkable solution.

@gfavor and I will make a refined proposal re: the I-fetch side, but I think we got the other half right.

PR #579, opened 2020-09-12
add architecture ID for NEORV32 core

stnolting:
According to [marchid.md](https://github.com/riscv/riscv-isa-manual/blob/master/marchid.md)

> Open-source project maintainers may make pull requests against this repository to request the allocation of an architecture ID.

I would like to request an architecture ID assignment (architecture ID 19) for the [NEORV32 processor](https://github.com/stnolting/neorv32).
:)


stnolting:
I definitely don't want to jostle and maybe this belongs into a new issue, but what are the specific requirements (if there are any) a core/project/you-name-it has to met in order to obtain an official (open-source) architecture ID? 🤔 😉

aswaterman:
@stnolting This fell through the cracks; jostling was the right move in this case.

stnolting:
> This fell through the cracks

Anyway, thank you very much! :)

PR #581, opened 2020-09-14
When disabling and then reenabling extensions, their state becomes UNSPECIFIED

pdonahue-ventana:
As in https://lists.riscv.org/g/tech-privileged/topic/76739961#324


pdonahue-ventana:
This sentence was apparently added via some other PR so this PR is no longer necessary.

PR #582, opened 2020-09-14
Require unimplemented instructions to trap

### LINKED ISSUE #576: Trapping hypervisor instructions when misa.H=0
The hypervisor chapter says:
> When misa[7] is clear, the hart behaves as though this extension were not implemented,
and attempts to use hypervisor CSRs or instructions raise an illegal instruction exception.

It makes sense that accesses to hypervisor CSRs must trap since chapter 2 says:
> Attempts to access a non-existent CSR raise an illegal instruction exception.

However, there is no analogous requirement in the privileged architecture that reserved or unimplemented instructions must raise an illegal instruction exception.  The unprivileged architecture is very explicit that the behavior is UNSPECIFIED and, other than specific cases like accesses to reserved CSRs or trapping particular instructions with TVM/TSR/etc., the privileged architecture doesn't add a blanket requirement.

Is it expected that HLV/HLVX/HSV must trap only on implementations that implement H but where H is disabled?  Or is this a broader requirement for all implementations, even those that hard-wire misa.H=0?  Or should this requirement to trap hypervisor instructions be removed?  Or should the privileged architecture be modified to require that illegal instructions must raise illegal instruction exceptions (regardless of support for H)?

/cc @jhauser-us

pdonahue-ventana:
Fixes #576

### DIFF
--- a/src/priv-intro.tex
+++ b/src/priv-intro.tex
@@ -208,3 +208,14 @@ M-mode. The separate debug specification proposal describes operation
 of a RISC-V hart in debug mode.  Debug mode reserves a few CSR
 addresses that are only accessible in D-mode, and may also reserve
 some portions of the physical address space on a platform.
+
+\section{Unimplemented Instructions}
+
+Implementations must raise an illegal instruction exception when decoding
+an instruction that is unimplemented.  This includes instructions defined by
+an extension which is disabled.
+
+\begin{commentary}
+Requiring unimplemented instructions to trap allows execution environments
+to use software emulation to implement functionality not provided in hardware.
+\end{commentary}


aswaterman:
Thanks, @pdonahue-ventana

David-Horner:
Although I favour this approach, I understood that this would be a profile
stipulation if required.
I can envision a class of minimalist implementations for embedded, say iot,
that would not need this emulation functionality and not want to pay for it
in gates/capacitance/area/routing or power use/low but would still want
certification.

allenjbaum:
This isn’t backwards compatible- just saying.

-Allen

David-Horner:
On Tue, Sep 15, 2020, 00:08 Allen Baum, <notifications@github.com> wrote:

> This isn’t backwards compatible- just saying.
>

Which isn't backwards compatible?
This recent push? If so, I agree that it is a change of policy/requirements
that unnecessarily constrain embedded products and possibly others.
Decisions have been made in the Vector TG with the tactical understanding
that platform profiles will determine this and other behaviours and that we
will need to rely on such profiles.

However, your comment could be read that: without traps on unimplement
instructions (by which I believe opcodes is meant) forward capabilities can
not be implemented and one cannot backwardly emulate newer standards.
This is also true, but a platform may not care that it has this
functionality if other competing requirements are deemed more important.

Further, a product that does not implement C-extension (unimplimented) for
instance should still be able to certify its RVI compliance even if the two
lower bit decode 00, 01 and 10 into non-standard custom instructions. This
is the implicit promise in chapter 26 Extending RISC-V.

-Allen

kasanovic:
This is not a compatible change, at least not as written.  It adds additional constraints that were not present when the spec was ratified.  Please revert.

The sentence about disabled extensions is particularly problematic as implies that otherwise compliant implementations can not have non-conforming use of extension space when the extension is turned off.

Trapping illegal instructions to support forward-compatible emulation is a non-goal in the ISA specs.
Platform profiles which can include execution environment definitions are the right place to put a given user community's desires to constrain implementation behaviors, but we cannot impose one community's world views on all users of the spec.

Requiring that all encodings do something "specified" might be a goal, but this is hard to define in a useful way at the level of ISA specs. Perhaps we can clarify that non-conforming instructions are also "implemented" instructions, so really is only saying that implementations must do something "defined" for any encoding - but I'm not sure that is a meaningful increment over saying nothing (some people might want a "halt and catch fire" instruction).

allenjbaum:
Sorry, replies crossed in the ether.

Requiring that all unimplemented opcodes trap is not backwards compatible
(unless all the implementations in existence trap on all unimplemented
opcode  - which include those that are associated with extensions that are
defined, and those that don't, e.g. have no standard defined assembler
mnemonic). Current implementations are free to halt and catch fire,  nop,
alias to some other architecturally defined op, do some weird custom op --
or trap. Some weird custom op is technically "implemented", and I suppose
you could argue that aliasing, no-oping and halt&catch fire are all
technically an "implemented" op. If so... what exactly is the point of this
change?

IS there a more specific definition of "unimplemented" I should use
instead?
I am not unhappy with this.. but this change really should have been made
2+ years ago.

If we do make it, it really should be discussed in a wider setting. I mean,
I'd vote for it, but there may be some very unhappy people out there that
have implementations that will break because of this.

kasanovic:
Requiring trap on illegal opcodes should not be in the ISA specs, as it breaks modularity.

Not having it in ISA specs allows a profile to add that constraint if desired by a sub-community.
Adding it to ISA specs would mean ISA specs become useless for many other use cases.

We made correct call, and I will argue very strongly against changing this.

aswaterman:
Reverted

pdonahue-ventana:
> The sentence about disabled extensions is particularly problematic as implies that otherwise compliant implementations can not have non-conforming use of extension space when the extension is turned off.

The sentence talks about extensions, not standard extensions.  In the situation you pose, there is a non-conforming extension that is enabled and the instruction in question is implemented as part of that non-conforming extension so no trap is required.  Furthermore, opcode space devoted to custom extensions can never be required to trap because whatever behavior happens can be considered the behavior of some custom extension.

This seems to be the exact same as "Attempts to access a non-existent CSR raise an illegal instruction exception" in section 2.1.  If you have a non-conforming extension then a standard CSR address can be reused (making it existent) and there is no requirement to trap.

Perhaps we could say something to the effect of: "If a standard instruction encoding or CSR address that is not implemented does not raise an illegal instruction exception, that instruction encoding or CSR is considered to be part of a non-conforming extension."  I think that is logically equivalent to this PR but perhaps stated in a more palatable way?  It makes it very clear that it requires no particular behavior, just clarifies definitions in the unprivileged architecture.

That doesn't really solve my original concern in #576, though.  I think we'd have to remove the requirement that HLV/HLVX/HSV trap if misa.H is 0 or somehow clarify it to deal with backward-compatibility, non-conforming extensions, etc.

allenjbaum:
The problem is that *anything* could be considered a non-conforming
extension (even if unintended) - so no trap is ever required.
e.g the universal noop extension - where everything not declared in MISA or
is in the base ISA will noop.

pdonahue-ventana:
> Trapping illegal instructions to support forward-compatible emulation is a non-goal in the ISA specs.

How do these ISA spec requirements comport with that statement:
1. Priv spec section 2.1: "Attempts to access a non-existent CSR raise an illegal instruction exception."
1. F extension section 11.2: "If frm is set to an invalid value (101–111), any subsequent attempt to execute a floating-point operation with a dynamic rounding mode will raise an illegal instruction exception."

PR #590, opened 2020-09-23
Improve description of interrupt traps

jhauser-us:
Simplify the description of when interrupt traps are taken, in part by abandoning the concept that interrupt sources are globally enabled selectively, depending on delegation registers.  (Global enablement on a per-interrupt basis had always been in conflict with other text in the document that says that interrupts are globally enabled for M-mode or S-mode simply when field MIE or SIE in mstatus is set.)

Also make clear that pending-and-enabled interrupts must trap immediately, before another instruction is executed.  This requirement is sufficient to prove (within the bounds of what is enforceable) that interrupts must have higher priority than synchronous exception traps.

### DIFF
--- a/src/hypervisor.tex
+++ b/src/hypervisor.tex
@@ -566,12 +566,6 @@ privilege than HS-mode;
 {\tt hip} and {\tt hie}; and
 (c)~bit~\textit{i} is not set in {\tt hideleg}.
 
-After execution of each instruction in program order, if the conditions
-for an interrupt trap to HS-mode are true, the trap occurs immediately,
-before another instruction executes.
-Interrupts to HS-mode take priority over any interrupts to operating
-modes of lower privilege.
-
 If bit~\textit{i} of {\tt sie} is hardwired to zero, the same bit in
 register {\tt hip} may be writable or may be read-only.
 When bit~\textit{i} in {\tt hip} is writable, a pending interrupt
--- a/src/hypervisor.tex
+++ b/src/hypervisor.tex
@@ -510,9 +510,6 @@ Registers {\tt hip} and {\tt hie} are HSXLEN-bit read/write registers
 that supplement HS-level's {\tt sip} and {\tt sie} respectively.
 The {\tt hip} register indicates pending VS-level and hypervisor-specific
 interrupts, while {\tt hie} contains enable bits for the same interrupts.
-As with {\tt sip} and {\tt sie}, an interrupt \textit{i} will be taken in
-HS-mode if bit~\textit{i} is set in both {\tt hip} and {\tt hie}, and if
-supervisor-level interrupts are globally enabled.
 
 \begin{figure}[h!]
 {\footnotesize
@@ -559,6 +556,21 @@ The active bits of {\tt hip} and {\tt hie} cannot be placed in HS-level's
 software to emulate the hypervisor extension on platforms that do not
 implement it in hardware.
 \end{commentary}
+
+An interrupt~\textit{i} will trap to HS-mode whenever all of the
+following are true:
+(a)~either the current operating mode is HS-mode and the SIE bit in the
+{\tt sstatus} register is set, or the current operating mode has less
+privilege than HS-mode;
+(b)~bit~\textit{i} is set in both {\tt sip} and {\tt sie}, or in both
+{\tt hip} and {\tt hie}; and
+(c)~bit~\textit{i} is not set in {\tt hideleg}.
+
+After execution of each instruction in program order, if the conditions
+for an interrupt trap to HS-mode are true, the trap occurs immediately,
+before another instruction executes.
+Interrupts to HS-mode take priority over any interrupts to operating
+modes of lower privilege.
 
 If bit~\textit{i} of {\tt sie} is hardwired to zero, the same bit in
 register {\tt hip} may be writable or may be read-only.
--- a/src/machine.tex
+++ b/src/machine.tex
@@ -1365,7 +1365,7 @@ privilege than M-mode;
 {\tt mideleg}.
 
 After execution of each instruction in program order, if the conditions
-for an interrupt trap to M-mode are true, the trap occurs immediately,
+for an interrupt trap are true, the trap occurs immediately,
 before another instruction executes.
 Interrupts to M-mode take priority over any interrupts to lower privilege
 modes.
--- a/src/machine.tex
+++ b/src/machine.tex
@@ -1355,17 +1355,20 @@ MXLEN \\
 \label{miereg}
 \end{figure}
 
-An interrupt \textit{i} will be taken if bit \textit{i} is set in both
-{\tt mip} and {\tt mie}, and if interrupts are globally enabled.  By
-default, M-mode interrupts are globally enabled if the hart's current
-privilege mode is less than M, or if the current privilege mode is M
-and the MIE bit in the {\tt mstatus} register is set.  If bit \textit{i}
-in {\tt mideleg} is set, however, interrupts are considered to be
-globally enabled if the hart's current privilege mode equals the
-delegated privilege mode and that mode's interrupt enable
-bit (\textit{x}\/IE in {\tt mstatus} for mode~\textit{x}) is set,
-or if the current
-privilege mode is less than the delegated privilege mode.
+An interrupt~\textit{i} will trap to M-mode (causing the privilege mode
+to change to M-mode) whenever all of the following are true:
+(a)~either the current privilege mode is M and the MIE bit in the
+{\tt mstatus} register is set, or the current privilege mode has less
+privilege than M-mode;
+(b)~bit~\textit{i} is set in both {\tt mip} and {\tt mie}; and
+(c)~if register {\tt mideleg} exists, bit~\textit{i} is not set in
+{\tt mideleg}.
+
+After execution of each instruction in program order, if the conditions
+for an interrupt trap to M-mode are true, the trap occurs immediately,
+before another instruction executes.
+Interrupts to M-mode take priority over any interrupts to lower privilege
+modes.
 
 Each individual bit in register {\tt mip} may be writable or may be
 read-only.
@@ -1555,17 +1558,15 @@ using uncached I/O writes to memory-mapped control registers depending
 on the platform specification.
 \end{commentary}
 
-Multiple simultaneous interrupts destined for different privilege modes are
-handled in decreasing order of destined privilege mode.  Multiple simultaneous
-interrupts destined for the same privilege mode are handled in the following
+Multiple simultaneous
+interrupts destined for M-mode are handled in the following
 decreasing priority order: MEI, MSI, MTI, SEI, SSI, STI.
-Synchronous exceptions are of lower priority than all interrupts.
 
 \begin{commentary}
   The machine-level interrupt fixed-priority ordering rules were developed
   with the following rationale.
   
-  Interrupts for higher privilege modes must be serviced before
+  Interrupts for M-mode must be serviced before
   interrupts for lower privilege modes to support preemption.
 
   The platform-specific machine-level interrupt sources in bits 16 and above
@@ -1586,9 +1587,6 @@ Synchronous exceptions are of lower priority than all interrupts.
   Software interrupts are located in the lowest four bits of {\tt mip}
   as these are often written by software, and this position allows the
   use of a single CSR instruction with a five-bit immediate.
-
-  Synchronous exceptions are given the lowest priority to minimize
-  worst-case interrupt latency.
 \end{commentary}
 
 Restricted views of the {\tt mip} and {\tt mie} registers appear as
--- a/src/supervisor.tex
+++ b/src/supervisor.tex
@@ -374,7 +374,7 @@ privilege than S-mode; and
 (b)~bit~\textit{i} is set in both {\tt sip} and {\tt sie}.
 
 After execution of each instruction in program order, if the conditions
-for an interrupt trap to S-mode are true, the trap occurs immediately,
+for an interrupt trap are true, the trap occurs immediately,
 before another instruction executes.
 Interrupts to S-mode take priority over any interrupts to lower privilege
 modes.
--- a/src/supervisor.tex
+++ b/src/supervisor.tex
@@ -366,12 +366,18 @@ SXLEN \\
 \label{siereg}
 \end{figure}
 
-An interrupt \textit{i} will be taken if bit \textit{i} is set in both
-{\tt sip} and {\tt sie}, and if supervisor-level interrupts are globally
-enabled.
-Supervisor-level interrupts are globally enabled if the hart's current
-privilege mode is less than S, or if the current privilege mode is S
-and the SIE bit in the {\tt sstatus} register is set.
+An interrupt~\textit{i} will trap to S-mode whenever both of the
+following are true:
+(a)~either the current privilege mode is S and the SIE bit in the
+{\tt sstatus} register is set, or the current privilege mode has less
+privilege than S-mode; and
+(b)~bit~\textit{i} is set in both {\tt sip} and {\tt sie}.
+
+After execution of each instruction in program order, if the conditions
+for an interrupt trap to S-mode are true, the trap occurs immediately,
+before another instruction executes.
+Interrupts to S-mode take priority over any interrupts to lower privilege
+modes.
 
 Each individual bit in register {\tt sip} may be writable or may be
 read-only.
@@ -499,7 +505,6 @@ they are shown as hardwired to 0 in Figures~\ref{sipreg-standard} and
 Multiple simultaneous
 interrupts destined for supervisor mode are handled in the following
 decreasing priority order: SEI, SSI, STI.
-Synchronous exceptions are of lower priority than all interrupts.
 
 \subsection{Supervisor Timers and Performance Counters}
 


allenjbaum:
The phrase 
      " and (c)if register mideleg exists, bit{i} is not set in mideleg"
seems ambiguous, since it doesn't describe what happens if mideleg doesn't exist.
Unambiguous would be:
   " and (c) register mideleg does not exist or bit{i} is not set in mideleg". or
   " and (c) register mideleg does not exist or bit{i} is clear in mideleg". or

David-Horner:
> Also make clear that pending-and-enabled interrupts must trap immediately, **before another instruction is executed.** This requirement is sufficient to prove **(within the bounds of what is enforceable)** that interrupts must have higher priority than synchronous exception traps.

Is this verifiable.
By definition interrupts are asynchronous events that can be in flux, appearing and disappearing.; asserted and released.,

When exactly, related to another instruction's execution, is the interrupt advanced sufficiently in Interrupt-Control-Circuitry to trap?
In the extreme, any transient at the "control" inputs would suffice to stop execution.

"(within the bounds of what is enforceable)": Doesn't this, itself, negate the requirement in the general case?

At best, this could be platform stipulated before software can rely on this precedence.
Where does it critically matter? Perhaps that scenario can be made bounded? 
If there is a place it critically matters, then is virtual HS doomed?

jhauser-us:
David Horner wrote:
> "(within the bounds of what is enforceable)": Doesn't this, itself, negate the requirement in the general case?

That's the right sort of question to ask, but in fact software can set up situations where the trap must occur, even given that many forms of interrupts may be arbitrarily delayed in arriving.  A simple example makes use of the fact that bit SSIP in sip is writable.  If we start with SSIP = 0, sie.SSIE = 1, and sstatus.SIE = 1, and then in S-mode execute a CSRS instruction to set SSIP = 1, an interrupt trap must occur after that CSRS instruction; no delay or fudging is allowed.  It's exactly for this type of example that the new text is proposed.

> When exactly, related to another instruction's execution, is the interrupt advanced sufficiently in Interrupt-Control-Circuitry to trap?

When it shows up in mip.

David-Horner:
This is not the general case.
Software Interrupt Enable is a very special case.
There is specific commentary for it:
>We allow a hart to directly write only its own SSIP bit, not those of other harts, as other harts
might be virtualized and possibly descheduled by higher privilege levels.

That is not present, nor can it be, for all other interrupts.

Consider STIP, which is writable by M-mode.
As currently defined, software cannot rely on setting of STIP to signal a trap to S-mode.
Perhaps STIP is under specified and needs special qualification  à la SSIP, but there is no reason that an external event or even prior setting (that is propagating) could not reset STIP during or immediately after a CSRS of STIP=1, and thus the trap legitimately does not occur.

Other interrupts are even more tenuous as signals to S-mode as their real time events are even less predictable than a level triggered comparison counter.
  
>>    When exactly, related to another instruction's execution, is the interrupt advanced sufficiently in Interrupt-Control-Circuitry to trap?

>When it shows up in mip.

Not precisely, as there is "virtual" mip until it is read by a CSRR.
Traps could be tripped even if no mip is ever actually set, certainly not sufficiently persistently to be  read.

jhauser-us:
The example I gave for SSIP isn't the only one that exists.  Here's another, involving an external interrupt from a device:

Let's say I have a machine with one hart and with a PLIC for external interrupts, and say I have a device interrupt line connected to the PLIC that, once raised, will not lower on its own without software action.  (Because the device's documentation tells me so.)  Let's start with the assumptions that this device interrupt has been configured at the PLIC to go to machine level, and that mie.MEIE = 1 but mstatus.MIE = 0.  I now execute code on my hart that does the following:

 1. Initiate something (anything) at the device that will eventually cause the device's interrupt to be raised.

 2. Execute a loop that repeats (doing nothing else) until mip.MEIP = 1.

 3. At this point, some interrupt has arrived in mip.MEIP, possibly from our device.  Now execute a CSRS instruction to set mstatus.MIE = 1.

After the CSRS instruction of step 3 has completed, the interrupt trap must be taken immediately.  It cannot be taken earlier, or delayed later.

> >  When it shows up in mip.
>
> Not precisely, as there is "virtual" mip until it is read by a CSRR.
> Traps could be tripped even if no mip is ever actually set, certainly not sufficiently persistently to be read.

You are correct, but that is entirely a matter for an implementation.  Except where allowed otherwise (the memory model comes to mind), RISC-V instructions are specified in a model where they execute in sequence.  Implementations are free to execute them in a different order, or use other tricks, but only so far as this isn't visible to software.

The proposed text for interrupt traps dictates when a trap must occur *in program order*, meaning what appears to happen from the perspective of software.  In the actual hardware, you are free to have your virtual mip register up until the moment the software reads mip.  Nothing proposed here changes the basic rules governing an implementor's freedom.

allenjbaum:
This is a consequence of the synchronous nature of CSRs, more than having
anything to do with the synchronous or asynchronous nature of interrupts.
Since the interrupt is dependent on both synchronous and asynchronous
events, whether it appears in program order is true only when the
consequence of causing the synchronous event is the event that cause the
interrupt.

Thanks for this discussion, I now have a handle on how to build an
architectural test for interrupts.

David-Horner:
>with a PLIC for external interrupts, ...  once raised, will not lower on its own without software action

This is where you have transitioned from ISA to platform.
I agree that it is very reasonable to stipulate characteristics according to such requirements.
However, it is equally important to distinguish between what is ISA and what is "platform".

STIP appears to require similar constraint for your stipulation that trap is immediately after conditions occur. However, I don't see a possibility to ensure the trap occurs immediately after enabling, even if there appeared to be a previous pending as a transient interrupt could be reasserted an instruction execution after the enabling. Hence, "delayed later" from a certain software perspective .

H is all about features for acceleration and optimization, Presumably portions can be implemented independently. Perhaps this "trap .... in program order" is of such an elective nature, it can give a benefit, but even without the remainder works well.
Fundamentally, I disagree with the notion that H must impose this constraint on hardware.
I don't see the rationale for it nor the use case that requires it to be so.

>The proposed text for interrupt traps dictates when a trap must occur in program order, meaning what appears to happen from the perspective of software. In the actual hardware, you are free to have your virtual mip register up until the moment the software reads mip. **Nothing proposed here changes the basic rules governing an implementor's freedom.**

Technically I agree with the last sentence. However, the statement itself is meaningless. No ISA can "changes the basic rules governing an implementor's freedom." What an ISA can do is stipulate a requirement that eliminates implementation options. That is what this requirement of "trap ... in program order" does.

Reconsider STIP: what constraints will be required on it to meet this new requirement?
minimally this "ratified" text will need to change for S-mode timer to be further constrained:

>4.1.4 Supervisor Timers and Performance Counters
...
The implementation must provide a facility for scheduling timer interrupts [to STIP] in terms of the real-time
counter, time.

And possibly this, as it gives guidance to the analogous STIP:
>A spurious timer interrupt might occur if an interrupt handler increments mtimecmp then im-
mediately returns, because MTIP might not yet have fallen in the interim. All software should be
written to assume this event is possible, but most software should assume this event is extremely
unlikely. It is almost always more performant to incur an occasional spurious timer interrupt
than to poll MTIP until it falls.

It may be that H extension must stipulate a specific environment that is conducive to this "trap ... in program order" decree.
But I am hopeful that it need not and that such impositions on the physical design can be eliminated.

@allenjbaum I hope this does not nullify your plan.

David-Horner:
I think what we are looking for is a guarantee, similar to the LR/SC forward progress: that an interrupt that is allowed to trigger, because all the conditions are met, will, if the in pending interrupt persists, trigger before software disables the interrupt.
The constraint then becomes a potential stall at the csr write that disables the interrupt, rather than a stall/roll back at the csr write that enables the interrupt. 
To me this allows much more flexibility to the implementation, which is free to stall and/or roll back at the csr write enabling, or free to complete all in flight instructions up to the csr write to disable, or anything in between.

I think this is a reasonable expectation for all implementations, not only those supporting H extension.
However, I don't believe it need be enforced on all implementations. Some platforms may not require such a guarantee.
Thus I suggest it is a (potential) platform guarantee that is stipulated for H (if we determine that this is needful or at least conducive to simplifying H or making it generally more useful/performant/reliable.)

Preemption and higher priority interrupt arriving before the original still pending interrupt can complicate this guarantee (e.g. that only one of them is required to trigger), but that is a complication that we can deal with later, once we have an agreement in principle that the basic guarantee is truly what is "needed" for H.

jhauser-us:
David Horner wrote:
> However, I don't see a possibility to ensure the trap occurs immediately after enabling, even if there appeared to be a previous pending as a transient interrupt could be reasserted an instruction execution after the enabling.

I had trouble parsing this sentence.  As best I can make out, you're saying that we can't guarantee that a specific interrupt trap is taken at that time because a different interrupt trap may be taken instead.  The proposed text says (for M mode, in this case):

> After execution of each instruction in program order, if the conditions for an interrupt trap to M-mode are true, the trap occurs immediately, before another instruction executes.

It doesn't say *which* interrupt, just *an* interrupt.  This could be your transient interrupt, for all I care.  I assumed it was already clear from elsewhere, combined with common sense, that there might be multiple interrupts whose conditions are true, and, if so, the interrupt with highest priority is the one that is taken.  But we could add more text if that's a source of confusion.

I'm afraid I don't understand why you keep speaking of the H extension, as though these are rules added only when the H extension is implemented.  Note that, in its list of files changed, GitHub shows "hypervisor.tex" first because it shows the files in alphabetical order.

> I think what we are looking for is a guarantee, similar to the LR/SC forward progress: that an interrupt that is allowed to trigger, because all the conditions are met, will, if the in pending interrupt persists, trigger before software disables the interrupt.

Clearly "trigger before software disables the interrupt" would be a much laxer requirement than "trigger immediately after software enables the interrupt", which is what the proposed text requires.  (Or clarifies what was always supposed to be required, if you believe as I do.)  I believe that "trigger before software disables the interrupt" is much too weak.  In fact, I believe under no circumstances should software be able to see evidence that an interrupt is pending-and-enabled, ready to be triggered, yet for some reason *no* interrupt trap has been taken.  The "trigger immediately" rule essentially follows from that.

David-Horner:
On Sat, Sep 26, 2020, 12:26 John Hauser, <notifications@github.com 
<mailto:notifications@github.com>> wrote:

    David Horner wrote:

        However, I don't see a possibility to ensure the trap occurs
        immediately after enabling.

        even if there appeared to be a previous pending.

        The transient interrupt could be **reasserted** an instruction
        execution after the enabling.

    I had trouble parsing this sentence.

Hopefully the added punctuation will help.

    As best I can make out, you're saying that we can't guarantee that a
    specific interrupt trap is taken at that time because a different
    interrupt trap may be taken instead.

Yes and no. Yes, that is one relevant case. No:
What I state is a reasserting of the same interrupt. That is, as far as 
software can tell. Software cannot,  with the mechanisms so far 
specified in the ISA, directly differentiate between these two.

    The proposed text says (for M mode, in this case):

        After execution of each instruction in program order, if the
        conditions for an interrupt trap to M-mode are true, the trap
        occurs immediately, before another instruction executes.

    It doesn't say /which/ interrupt, just /an/ interrupt. This could be
    your transient interrupt, for all I care. I assumed it was already
    clear from elsewhere, combined with common sense, that there might
    be multiple interrupts whose conditions are true, and, if so, the
    interrupt with highest priority is the one that is taken. But we
    could add more text if that's a source of confusion.

No that scenario is not confusing.
Rather, what is confusing for software is: the apparent same pending 
interrupt not immediately trapping after the enabling csr write.
Ironically it is the added text that is the source of the confusion; I 
think adding more will not help.

    I'm afraid I don't understand why you keep speaking of the H
    extension, as though these are rules added only when the H extension
    is implemented. Note that, in its list of files changed, GitHub
    shows "hypervisor.tex" first because it shows the files in
    alphabetical order.

I noticed that the text change was applied in multiple places (including 
M and S mode ).

I speak of H extension because

     1) the M and S mode texts had already been ratified and it could 
only be in the context of a further extension that the existing 
behaviour could reasonably be modified, and

     2) H extension is the elephant in the room wrt included text of a 
new extension.

On its face,  the changes for S and M modes presented on their own fail 
the backward compatibility test.



        I think what we are looking for is a guarantee, similar to the
        LR/SC forward progress: that an interrupt that is allowed to
        trigger, because all the conditions are met, will, if the in
        pending interrupt persists, trigger before software disables the
        interrupt.

    Clearly "trigger before software disables the interrupt" would be a
    much laxer requirement than "trigger immediately after software
    enables the interrupt", which is what the proposed text requires.

    (Or clarifies what was always supposed to be required, if you
    believe as I do.)

No, I don't believe as you do, nor I suspect does established silicon. 
This is not just a clarification but a new requirement.

    I believe that "trigger before software disables the interrupt" is
    much too weak.

    In fact, I believe under no circumstances should software be able to
    see evidence that an interrupt is pending-and-enabled, ready to be
    triggered, yet for some reason /no/ interrupt trap has been taken.

I disagree, there are many platforms that "for all they care" are 
indifferent to this quirk.

Further, there is no atomic test for pending and enabled and (even more 
difficult to ascertain) that no trap was taken.

Certainly not in S,U,VS,VU or HS modes. Maybe we can assure such a case 
for M-mde, maybe. But I see no reason to ensure it.

There are some legitimate reasons that any mode other than M may "see 
evidence that an interrupt is pending-and-enabled, ready to be 
triggered, yet for some reason /no/ interrupt trap has been taken".

Notably a preempting higher level interrupt that removes the condition. 
The lower levels need not know what the reason is, only that they need 
to accept it.

Now, a hypervisor may wish to ensure a higher level of 
consistency/simplicity but it would provide an artificial virtual 
environment. The real world has such quirks.

> The "trigger immediately" rule essentially follows from that. 

Both are too stringent requirements. However, to affect the hart 
observance that you want it is sufficient that a subsequent csr read 
that reveals a pending interrupt triggers a trap.

But, really, to what end? Is there a real need?



    —
    You are receiving this because you commented.
    Reply to this email directly, view it on GitHub
    <https://github.com/riscv/riscv-isa-manual/pull/590#issuecomment-699517058>,
    or unsubscribe

aswaterman:
The intent of the spec was always that, if some event causes a currently pending interrupt to become enabled (e.g., an MRET or a write to mstatus or mie), the interrupt must be taken before another instruction is executed.  This text clarifies that point.

This doesn't entirely contradict David's remark that "to affect the hart observance that you want it is sufficient that a subsequent csr read that reveals a pending interrupt triggers a trap".  But we can model this more cleanly, and consistently with John's framework, by saying that the interrupt simply hasn't shown up in mip yet.

allenjbaum:
I had the same comment about level/edge interrupt signals, and Andrew
confirmed that the assumption was that interrupts were levels.
Incoming interrupt signals could pulses - but (just as the PLIC requires)
need to be "converted" to level, else they disappear.

Non-posted writes wait until the write completion is received before
allowing further writes.
Using those to clear interrupts would fix this, and I thought PCIE required
that behavior, but I'm not quite sure how Risc-V does this: Fence followed
by a dummy read?
Handling spurious interrupts sounds messy (more corner cases) but easier .
But for compliance tests, I really want that non-posted behavior to be able
to test things properly.

David-Horner:
> The intent of the spec was always that, if some event causes a currently pending interrupt to become enabled (e.g., an MRET or a write to mstatus or mie), the interrupt must be taken before another instruction is executed. This text clarifies that point.

Like laws, these ratified text have to consider the letter and not just the intent.
Many times we have stated that omissions are as significant as explicit text.
We relied on this to interpret and justify leaving the text as initially presented,
 not changed to what we may latter recall was intended

So what do contemporaneous documents say?
The revisions from 2.2 of the RISC-V Instruction Set Manual
Volume I: Unprivileged ISA to ratified included:
>Added clearer and more precise definitions of execution environments, harts, traps, and memory accesses

So there was a significant review of the wording before the ratification (It was not a long forgotten passage).
Section 1.6 Exceptions, Traps, and Interrupts  describes extensively the trap and interrupt natures.
It is carefully worded to afford the latitude that is in the original text of Volume II.   
The section is explicit that:
>The EEI defines for each trap whether it is handled precisely, though the recommendation is to
maintain preciseness where possible.

The plain text thus says imprecise interrupts are allowed (deferred trap with intervening instructions is not forbidden). 
Although the expectation may have been the recommendation would be followed, this text challenges any notion that the intent was that it would be mandatory.
Further, the above text and this:
>The manner in which interrupts are generated, routed to, and enabled by a hart depends on the EEI.

makes clear that EEI are allowed to have different policies and characteristics appropriate to their purpose.
We have recently  called these platform requirements, and it is here that we could stipulate additional/specific EEI requirements. Not, I argue, in revisionist Volume II text.

In Volume II:
PLIC was removed from the specification prior to ratification.
This signalled that how and what PLIC mandated was not a part of the specification.
That PLIC ensured level based interrupts could no longer be assumed to apply.

Nor is CLINT mentioned.
The inferred intent is that Volume II be agnostic to EEI approaches.
e.g. Transient interrupts are not prohibited.

Although it may now be categorized as an oversight STIP is minimally specified wrt persistence of level and stability.
However: From 3.1.10 Machine Timer Registers (mtime and mtimecmp)
>Writes to mtime and mtimecmp are guaranteed to be reflected in MTIP eventually, but not necessarily immediately.

explicitly makes “errant” STIP status possible, and the explanatory note goes on to advise how to handle this situation. (It does not recommend a precise trap stipulation).


>This doesn't entirely contradict David's remark that "to affect the hart observance that you want it is sufficient that a subsequent csr read that reveals a pending interrupt triggers a trap".

My remark wasn’t a recommendation, but instead an observation that the concern that inconsistencies may arise can be addressed in other ways than “immediate trap”.
Indeed, this is a preferred way (but not the best) in that “inexplicable delays” no longer are puzzling. Instead, they are consistent with the spec.

>But we can model this more cleanly, and consistently with John's framework, by saying that the interrupt simply hasn't shown up in mip yet.

I abhor such word play intentionally entrenched in specifications.
Let our yes be yes, and our no be no. 
Not a “depends on how you want to interpret or contrive the definitions”.

In addition to my revulsion at word play I consider dishonest and deceptive, there are a number of practical reasons:
1) sometimes such ambiguities are necessary. We have to live with that (for a time at least until the science catches up with the art). To the extent possible we must reduce those occurrences, otherwise the impression/message will be that specs are suggestions.
2) such resolutions are not usually formalized in the documentation. Either it becomes a secret sauce known only to a few, or it is lost in the ether. Confusion if not acrimony occurs. “They say they are compliant, but this .... is definitely contrary to the specs.
3) a preferred solution is not pursued, an opportunity is lost, erroneous decisions are made and the quality of the ecosystem suffers.

If we formalize “ hasn't shown up in mip yet” we have gutted John’s change.

An implementation can just hold off posting to mip as long as it wants, without any guarantee, and certainly without any apparent consistency. The decision could be based solely on micro-architecture expediency related to its current state; with no regard to even the letter of the text let alone its intent. We have judge, jury and executioner in one design team if definitions are malleable.

That is not to say this will not occur in physical designs anyway, but hopefully to a lesser degree if we shape the narrative, especially by formulating  platform guarantees that are testable. Producers will either stipulate to them or not. If stipulated they can be “held accountable”, loose that certification, or a portion of the stipulation through erratas, etc.  The alternative of “We could do whatever we wanted” is otherwise a legitimate excuse.

I did not raise this concern previously, but I see problematic the phrase:
>After execution of each instruction in **program order**

First of all what does it mean?
Does it mean that during a sequence of code that could potentially be interrupted that a long running instruction, say a divide will stall from completing (retiring) other shorter duration instructions so that **program order** can prevail?
If the interrupt occurs during the afore mentioned divide and concurrent instruction execution, does it mean that the interrupt must occur on the divide (as it has precedence in program order).

I cannot believe that the intent is to nullify existing implementations that currently allow divides to complete over an interrupt and only stall when the handler accesses the result register. (e.g. SiFive E31)

But that is the natural consequence of the suggested “clarifying” text.
I suspect correcting this text will not be “clean” and simple. I suspect the framework is fundamentally flawed. Once **program order** is weakened, the significance of the **immediately, before another instruction executes** is completely lost; it becomes meaningless.

My recommendation remains to have a "platform" guarantee : interrupts that persist since being software enabled trap prior to being software disabled.
Not all platforms may subscribe to this, and a more stringent requirement could be devised, but this will suffice for the scenarios John has so far identified.
The significance of Interrupt priority over synchronous trap has not been explained, nor a scenario in which it is relevant, so I cannot attest to the "guarantee" being adequate for that.

jhauser-us:
David Horner wrote:
> If we formalize “hasn't shown up in mip yet” we have gutted John’s change.

No, that's exactly how I (and we) want it to work.  In fact, that's *why* the document says for timer interrupts:

> Writes to mtime and mtimecmp are guaranteed to be reflected in MTIP eventually, but not necessarily immediately.

David:
> An implementation can just hold off posting to mip as long as it wants, without any guarantee, and certainly without any apparent consistency.

I see for MTIP one guarantee:  it shows up eventually.  Aside from that, you are correct.  Platform specs could always tighten the upper and lower bounds on the delay, if that's needed.

> Does it mean that during a sequence of code that could potentially be interrupted that a long running instruction, say a divide will stall from completing (retiring) other shorter duration instructions so that program order can prevail?

Only that it must appear that way to software.

> I cannot believe that the intent is to nullify existing implementations that currently allow divides to complete over an interrupt and only stall when the handler accesses the result register. (e.g. SiFive E31)

That's neither the intent nor a consequence of the proposed text.  This falls under the heading of "inherent implementor freedom" to do whatever he/she wants, so long as software doesn't see the rules being violated.

David-Horner:
> David Horner wrote:
>
>     If we formalize “hasn't shown up in mip yet” we have gutted John’s change.
>
> No, that's exactly how I (and we) want it to work.

I will take you at your word that you (currently) want this behaviour.
But who exactly is "we"? Is it the collective "we"; you and Andrew, or Allan; or a silent partner?
>
> In fact, that's why the document says for timer interrupts:
>
>     Writes to mtime and mtimecmp are guaranteed to be reflected in MTIP eventually, but not necessarily immediately.
I cannot fathom your reasoning.

1) There are operational, physical propagation, other engineering design reasons and importantly micro-architecture impartiality reasons to specify MTIP eventuality this way.

    These are reasons that neither envision nor necessitate the revised wording.

2) If MTIP eventuality guarantee was predicated on the "intended" meaning, why was that intended meaning not incorporated at the time?

     I have detailed contemporaneous documents that support the contrary conclusion that the original wording was intentional;

      The original wording was, at the time of review and revision, consistent with and potentially motivate by the uncertainly inherent in the MTIP timing delay.
3) Does it make any sense to argue the MTIP eventuality guarantee exists to substantiate  "hasn't shown up in MIP yet" so as to  gut the meaning of the "clarification"?

Perhaps you misstated? perhaps a typo? Perhaps you can rephrase if I have grossly misconstrued your intent.

>David:

>     An implementation can just hold off posting to mip as long as it wants, without any guarantee, and certainly without any apparent consistency.
>
> I see for MTIP one guarantee: it shows up eventually.

It is worth noting that MTIP eventuality guarantee applies to both assertions and negation. and (thus) does not guarantee their observance by software.

> Aside from that, you are correct.

However, this is not about MTIP alone. With the pervasive implicit caveat, any interrupt could be delayed indefinitely by the micro-architecture.

"Immediately" has no meaning in this context. Why revise the text with a thus meaningless requirement.

Additionally, if the action is neutered, why stipulate a check for the trigger of that [in]action, and why prescribe a specific time and order for the check [after each instruction in program order]?


> Platform specs could always tighten the upper and lower bounds on the delay, if that's needed.

I believe it is necessary to establish a mechanism for such bounds.
This proposed revised text with the pervasive implicit caveat does nothing to assist in defining or establishing any bounds.
Rather, in my opinion, it muddies the waters.

I again recommend my proposal for a "platform" guarantee, which CAN be adjusted on upper and lower bounds.

>     Does it mean that during a sequence of code that could potentially be interrupted that a long running instruction, say a divide, will stall from completing (retiring) other shorter duration instructions so that program order can prevail?
>
> Only that it must appear that way to software.

So , two situations here

   1) software that is oblivious to the situation: it has no access to resources that could detect that this is happening
          In which case the in "program order" and "immediacy" requirements are moot.

   2) software that has access to reliable instrumentation (such as detailed and comprehensive instruction cycle counters )
          That could detect the case that a divide continued execution into a handler, but that the interrupt occurred on a later instruction.

          However, with this wording change, this common optimization could not be enabled (at least while the counters are active).

If you say 2 is not what you mean by "it appearing that way to software"; that such instrumentation, although part of the ISA, is cheating/over-reach.

If we are left with only 1, then the proposed change has no observable ISA impact and by definition cannot be a requirement of the ISA.


>     I cannot believe that the intent is to nullify existing implementations that currently allow divides to complete over an interrupt and only stall when the handler accesses the result register. (e.g. SiFive E31)
>
> That's neither the intent nor a consequence of the proposed text. This falls under the heading of "inherent implementor freedom" to do whatever he/she wants, so long as software doesn't see the rules being violated.

The above argument controls here as well.

If the revised text provides no meaningful qualifications on the ISA, then including it is superfluous at best, probably misleading/confusing and potentially harmful.

jhauser-us:
David Horner wrote:
> I will take you at your word that you (currently) want this behaviour.
> But who exactly is "we"? Is it the collective "we"; you and Andrew, or Allan; or a silent partner?

By "we" I meant Andrew and myself, at a minimum.  I believe Krste is on board too, but it's possible I'm mistaken about that.  It would be best to confirm with him.

> It is worth noting that MTIP eventuality guarantee applies to both assertions and negation. and (thus) does not guarantee their observance by software.

Once MTIP becomes asserted, it remains asserted for as long as mtimecmp < mtime; hence it remains asserted at least until software writes again to mtimecmp or mtime---or until mtime overflows, which with today's technology shouldn't happen in anything less than 50 years of continuous operation.  Assuming mtime overflow isn't a realistic concern, this guarantees that, if software is waiting for MTIP to become true, it will see it eventually.

If you believe that MTIP automatically returns to zero after a brief pulse, you are mistaken.

> If you say 2 is not what you mean by "it appearing that way to software"; that such instrumentation, although part of the ISA, is cheating/over-reach.

It's fair to say that your point 2 is not how I meant it, yes.

> If we are left with only 1, then the proposed change has no observable ISA impact and by definition cannot be a requirement of the ISA.

You mean you believe the proposed added requirement would be vacuous.  I'm sure you're wrong, but nevertheless I suggest we leave this pull request open for comments for a while and see what others may say.  Despite the fact that they haven't commented, I know there are other processor implementors who regularly scan the open issues and pull requests for the RISC-V ISA.

David-Horner:
On 2020-09-28 4:27 a.m., John Hauser wrote:
>
> David Horner wrote:
>
>     I will take you at your word that you (currently) want this behaviour.
>     But who exactly is "we"? Is it the collective "we"; you and Andrew, or Allan; or a silent partner?
>
> By "we" I meant Andrew and myself, at a minimum. I believe Krste is on board too, but it's possible I'm mistaken about that. It would be best to confirm with him.
>
>     It is worth noting that MTIP eventuality guarantee applies to both assertions and negation. and (thus) does not guarantee their observance by software.
>
> Once MTIP becomes asserted, it remains asserted for as long as mtimecmp < mtime; hence it remains asserted at least until software writes again to mtimecmp or mtime---or until mtime overflows, which with today's technology shouldn't happen in anything less than 50 years of continuous operation. Assuming mtime overflow isn't a realistic concern, this guarantees that, if software is waiting for MTIP to become true, it will see it eventually.

This is why I thought it worth noting:

If mtime is updated greater than mtimecmp, and mtimecmp is then updated to be greater than mtime, the assertion and negation are allowed to race with the net result that the software, even if it were polling continuously, need not see the assertion.

> If you believe that MTIP automatically returns to zero after a brief pulse, you are mistaken. 

Well that is one less thing about which I am mistaken (I hadn't even [consciously] considered an automatic return to zero (other than wrap around).)

(I'm puzzled that you thought I might)

However, I had not assumed that mtime was defined to increase only by the LSB as your 50 years infers.

I considered that it could advance by any constant as long as the frequency with which it was updated was constant.

For example I had considered that an implementation (a la 360 architecture) could define a specific bit to represent an epoch, say a microsecond, and the incremental advance INCR  would be defined by INCR*FREQ = epoch.

The increment would then be some obscure number such as 110111110B.

So, if that is definitely NOT allowed, then I can envision seeing the "clarification" that the counter must advance by 1 LSB at a time.
>
>     If you say 2 is not what you mean by "it appearing that way to software"; that such instrumentation, although part of the ISA, is cheating/over-reach.
>
> It's fair to say that your point 2 is not how I meant it, yes.
>
>     If we are left with only 1, then the proposed change has no observable ISA impact and by definition cannot be a requirement of the ISA.
>
> You mean you believe the proposed added requirement would be vacuous. 

I looked it up and in the archaic sense, "empty", yes.

But certainly not in the vernacular ( not my vernacular mind you).
I believe the considerations were thoughtful and useful intent contemplated. 
> I'm sure you're wrong, 

I'd appreciate if you can clarify precisely about what you are sure that I am wrong.
I have not perceived a specific rebuttal to my reasoning, nothing pinpointing where I've gone astray.
I have tried to do that in my responses.

As in previous correspondence, I  am of the opinion that once reasonable people agree on the intent,
objectives and priorities such disagreements disappear or are limited to
minor wordsmithing.

> but nevertheless I suggest we leave this pull request open for comments for a while and see what others may say.

Of course. That is indeed your prerogative.
> Despite the fact that they haven't commented, I know there are other processor implementors who regularly scan the open issues and pull requests for the RISC-V ISA.

The concern can also be raised on relevant TGs and when all else fails, isa-dev google group is fertile field for opinions.

jhauser-us:
David Horner wrote:
> If mtime is updated greater than mtimecmp, and mtimecmp is then updated to be greater than mtime, the assertion and negation are allowed to race with the net result that the software, even if it were polling continuously, need not see the assertion.

That's true enough as it stands, but software that actually cares about seeing the timer interrupt will refrain from this behavior.  When I spoke of the inevitability of software seeing an incoming MTIP signal, I was assuming the software would act rationally and not actively sabotage its ability to see the interrupt.  I guess I neglected to say that.

I wrote:
> If you believe that MTIP automatically returns to zero after a brief pulse, you are mistaken.

David:
> Well that is one less thing about which I am mistaken (I hadn't even [consciously] considered an automatic return to zero (other than wrap around).)
>
> (I'm puzzled that you thought I might)

Because I was grasping for an explanation for why you thought MTIP would get de-asserted before software would see it asserted.  Again:  Can we not agree that if software wants to see a timer interrupt, it will ordinarily refrain from causing the interrupt to be de-asserted until *after* the interrupt has been observed asserted?

> However, I had not assumed that mtime was defined to increase only by the LSB as your 50 years infers.

Searching through the documents just now, I could not find exactly where an increment step larger than 1 is prohibited.  Volume II says:
> `mtime` must increment at constant frequency.

And concerning the user-level 64-bit time value (same as mtime), Volume I, section 10.1 says:
> The real-time clocks of all harts in a single user application should be synchronized to within one tick of the [master] real-time clock.

One might interpret "increment" and "tick" as implying a step size of 1.  Or not.  Anyway, Volume I also says:
> The underlying 64-bit counter should never overflow in practice.

I would certainly read that as ruling out step sizes large enough to vitiate this claim of never overflowing in practice (i.e., within any reasonable amount of time).

David wrote:
> So, if that is definitely NOT allowed, then I can envision seeing the "clarification" that the counter must advance by 1 LSB at a time.

I'll leave that to others.

> I have not perceived a specific rebuttal to my reasoning, nothing pinpointing where I've gone astray.

I'm aware that you feel that way.  I have tried.

David-Horner:
I note you annotate authors.

Given how mangled github can be with email, that seems a good idea.

On 2020-09-28 5:27 p.m., John Hauser wrote:
>
> David Horner wrote:
>
>     If mtime is updated greater than mtimecmp, and mtimecmp is then
>     updated to be greater than mtime, the assertion and negation are
>     allowed to race with the net result that the software, even if it
>     were polling continuously, need not see the assertion.
>
> That's true enough as it stands, but software that actually cares 
> about seeing the timer interrupt will refrain from this behavior. When 
> I spoke of the inevitability of software seeing an incoming MTIP 
> signal, I was assuming the software would act rationally and not 
> actively sabotage its ability to see the interrupt. I guess I 
> neglected to say that.
>
David writes:

The point isn't that software shouldn't behave a certain way. But rather 
the ISA does not provide this guarantee of visibility.

It is this lack of guarantee that is worth mentioning.

Because, if a platform does want to detect each instance when mtimecmp 
is less than mtime, for instance a watchdog in which there are multiple 
harts/elements adjusting the RCT, then it will have to be explicitly 
supported at the platform level. [Not an impossibility given the ISA's 
lax stance on timeliness.]  The eventuality guarantee does not cover 
this case.

Further, I was using mtimecmp as a proxy for other interrupts that 
aren't even as well defined.

I'm sorry if you thought I was being oppositional or argumentative (or a 
jerk)..

> I wrote:
>
>     If you believe that MTIP automatically returns to zero after a
>     brief pulse, you are mistaken.
>
> David:
>
>     Well that is one less thing about which I am mistaken (I hadn't
>     even [consciously] considered an automatic return to zero (other
>     than wrap around).)
>
>     (I'm puzzled that you thought I might)
>
> Because I was grasping for an explanation for why you thought MTIP 
> would get de-asserted before software would see it asserted. Again: 
> Can we not agree that if software wants to see a timer interrupt, it 
> will ordinarily refrain from causing the interrupt to be de-asserted 
> until /after/ the interrupt has been observed asserted?
>
>     However, I had not assumed that mtime was defined to increase only
>     by the LSB as your 50 years infers.
>
> Searching through the documents just now, I could not find exactly 
> where an increment step larger than 1 is prohibited. Volume II says:
>
>     |mtime| must increment at constant frequency.
>
> And concerning the user-level 64-bit time value (same as mtime), 
> Volume I, section 10.1 says:
>
>     The real-time clocks of all harts in a single user application
>     should be synchronized to within one tick of the [master]
>     real-time clock.
>
> One might interpret "increment" and "tick" as implying a step size of 
> 1. Or not. Anyway, Volume I also says:
>
>     The underlying 64-bit counter should never overflow in practice.
>
> I would certainly read that as ruling out step sizes large enough to 
> vitiate this claim of never overflowing in practice (i.e., within any 
> reasonable amount of time).
>
David writes:

This may be the fundamental difference in our perspective.

I believe you are saying, "should never overflow in practice" is 
sufficient for software to rely on it never overflowing.

I see the statement as guidance that the ISA is likely to meet your 
requirements for the reasons stated, but YMMV. There are caveats.


The difference in perspective may also be a result of use cases.

I'm thinking (frequently) of bare metal, minimalist systems, just as 
many bits in each required csr as needed for the task.

If I put on my linux  hat, I expect the OS to make very reasonable 
efforts to do the right thing, just as you expected above, so that the 
user environment and other coprocesses/tasklets/system-threads will not 
see time glitches.

> David wrote:
>
>     So, if that is definitely NOT allowed, then I can envision seeing
>     the "clarification" that the counter must advance by 1 LSB at a time.
>
> I'll leave that to others.
>
>     I have not perceived a specific rebuttal to my reasoning, nothing
>     pinpointing where I've gone astray.
>
> I'm aware that you feel that way.
>

> I have tried.
>
David writes:

Thank you.

David-Horner:
@jhauser-us 
 I apologize.
It was disingenuous of me to ascribe that interpretation of the phrase to you.

Unjustified, as I should have asked if it was possibly yours.
But further, more subtle distinctions in interpretation, and more consistent with the text, is sufficient to crystallize what might be our differing perspectives.

Multiple ways of perceiving “The underlying 64-bit counter should never overflow in practice.”

One is [software sufficiency centric]:
 that 64-bits should be sufficient for software to maintain the counter such that it will not roll-over unexpectedly [or at all] with corresponding glitches. Software should be able to manage all practical situations.

Another is [ISA tradeoff analysis]:
that 64-bits is sufficiently large that the ISA can allow software to avoid overflow ill-effects. The ISA need not  increase the number of bits, nor provide any other exception mechanism. It is sufficient for practical purposes. 

There are even more perceptions and inferrences that can be drawn from the statement.
(Ranging from an imperative “Do not let overflow occur because software needn’t allow it to happen” to “In [usual] practice you should not expect overflow to happen” to ideas that  deviate even more from the original text)

I won’t attribute any of these interpretations to you.
As I said it was wrong for me to do so in the last post.
Suffice it to say that I read the text as a tradeoff determination for the ISA.
A different reading may come to different conclusions about expectations of software vs expectations of hardware.

allenjbaum:
quick rule of thumb: there are pi*10^7 seconds in a year.. 64 bit counter
that incremented by one at a 1Ghz rate would (if initialized to zero at
powerup)  overflow in 128 years. At a 100Mhz rate, that's 1280 years. Just
saying.

A platform requirement that mtime MSB be cleared at poweron should cover
all practical concerns (unless someone deliberately writes it with a small
negative value)

jrtc27:
This somewhat-needlessly constraints out-of-order implementations. Taking an interrupt is expensive, so you might as well finish executing all the instructions you already have in the reorder buffer before taking it, which can easily be done by checking for interrupts as instructions enter the reorder buffer (turning them into special exception-like entries instead), but in theory one of those could be a CSR read of MIP that would allow you to observe an interrupt *not* being taken immediately after it. Does it really matter enough that you want to force implementations to abide by this additional restriction, either checking for interrupts at the commit stage and abandoning everything immediately, or having special handling of MIP reads to ensure that it reads as 0 except for possibly the final instruction before the interrupt? What about if the (initially-speculative) fetch of the next instruction faults; do we really need to mandate that interrupts have higher priority?

It seems to me that this is unnecessarily strict and that instead you just need to require that interrupts be taken within a bounded number of cycles of observing that one is pending?

jrtc27:
> quick rule of thumb: there are pi*10^7 seconds in a year.. 64 bit counter that incremented by one at a 1Ghz rate would (if initialized to zero at powerup) overflow in 128 years. At a 100Mhz rate, that's 1280 years. Just saying. A platform requirement that mtime MSB be cleared at poweron should cover all practical concerns (unless someone deliberately writes it with a small negative value)
> […](#)

I don't think your maths quite checks out; I get about 584 years for your calculation (`2^64 / (1*10^9*365*24*60*60)`). But even then, what if I have a counter that's counting the number of instruction bytes fetched from memory on a high-performance 4 GHz core that achieves an average IPC of 2? That gives a maximum uptime of between 18 and 36 years before an interrupt will fire depending on the distribution of 16-bit and 32-bit instructions, which is a long time but it's still in the feasible range.

allenjbaum:
You are correct: pi*10^7 ~= 2^25, 1G ~= 2^30, so that's 2^55 ns/year, so it
would roll over every 2^(64-55)  ~= 29 ~= 512 years.  I'm not sure where
128 ~= 2^7 came from.
Initializing just the MSB in HW would guarantee 256 years for the first
rollover.
I don't know anyone that is implementing an RTC that is counting at a 1Ghz
(as opposed to something like 10-100Mhz).

So, I'm missing some context here: what exactly is constraining OOO
implementations? There was a lot of discussion above and it isn't clear
which one you're referring to.

jrtc27:
> You are correct: pi*10^7 ~= 2^25, 1G ~= 2^30, so that's 2^55 ns/year, so it would roll over every 2^(64-55) ~= 29 ~= 512 years. I'm not sure where 128 ~= 2^7 came from. Initializing just the MSB in HW would guarantee 256 years for the first rollover. I don't know anyone that is implementing an RTC that is counting at a 1Ghz (as opposed to something like 10-100Mhz).

Sure, but timers aren't the only thing that are counting up. I gave an example of a meaningful counter one might wish to implement that could feasibly overflow during the lifetime of a device. Not adding a mechanism to completely disable interrupts is short-sighted, especially given how easy it is to just add another bitmask register; AArch32 and AArch64 both have one, for example.

>So, I'm missing some context here: what exactly is constraining OOO implementations? There was a lot of discussion above and it isn't clear which one you're referring to.

My separate OOO comment was regarding the original pull request's stricter requirements for when interrupts should be taken.

jhauser-us:
Jessica Clarke,

I do not believe that reads of CSR `mip` are the obstacle you suggest.  There are actually a couple of simple ways for a complex processor core to handle this case, the obvious one being simply for reads of `mip` to serialize execution, i.e., flush the execution pipeline.  This is not at all an extreme solution, for a combination of two reasons:  First, it is quite common for complex cores, including out-of-order ones, to flush the pipeline on certain CSR accesses, as a way to obviate what would otherwise be a great deal of complexity over instruction dependencies.  Accesses to `mip` would just be another such case.  Second, it's highly unlikey there is any important software whose speed would be impacted by this solution, as there is precious little need for software to read or write `mip`.

It is true that the proposed document change can be interpreted as adding new constraints on cores.  (Although the "new" requirements were actually always intended, just not made as clear before.)  However, I believe the worst impact on cores is to force some additional serialization (pipeline flushes), something cores invariably already do for other situations.  So the cost is less in physical hardware and more in the time lost to those extra pipeline flushes, which I believe shouldn't be very significant in practice.

One might try to provide evidence that I'm wrong about the latter; that in fact there are important cases where the added pipeline flushes will noticeably impact execution speed.

> What about if the (initially-speculative) fetch of the next instruction faults; do we really need to mandate that interrupts have higher priority?

I don't see why this causes any difficulty.  Instruction fetch faults already must be carried through the entire pipeline to the commit stage, which, as you noted, is also the obvious place to decide interrupt traps.  Confronted with both, hardware simply has to prioritize the interrupt trap; no big deal.

jrtc27:
> Jessica Clarke,
> 
> I do not believe that reads of CSR `mip` are the obstacle you suggest. There are actually a couple of simple ways for a complex processor core to handle this case, the obvious one being simply for reads of `mip` to serialize execution, i.e., flush the execution pipeline. This is not at all an extreme solution, for a combination of two reasons: First, it is quite common for complex cores, including out-of-order ones, to flush the pipeline on certain CSR accesses, as a way to obviate what would otherwise be a great deal of complexity over instruction dependencies. Accesses to `mip` would just be another such case. Second, it's highly unlikey there is any important software whose speed would be impacted by this solution, as there is precious little need for software to read or write `mip`.
> 
> It is true that the proposed document change can be interpreted as adding new constraints on cores. (Although the "new" requirements were actually always intended, just not made as clear before.) However, I believe the worst impact on cores is to force some additional serialization (pipeline flushes), something cores invariably already do for other situations. So the cost is less in physical hardware and more in the time lost to those extra pipeline flushes, which I believe shouldn't be very significant in practice.

I am not a fan of defining the architecture in such a way that pipeline flushes are effectively mandated for trivial things like `mip`. Yes, you can do it, and it likely doesn't hurt performance all that much, but why? Software really should not care whether `mip` means the next instruction or some time very soon. Give implementations flexibility where it makes sense. Especially forcing _reads_ to be a pipeline flush is a really sad state to be in; writes at least can have lots of side-effects that are hard to deal with in an out-of-order core, but reads? We have a version of RiscyOO that treats CSR reads like normal ALUish instructions but baking into the architecture that that's not allowed for something as simple as this seems completely unnecessary.

> One might try to provide evidence that I'm wrong about the latter; that in fact there are important cases where the added pipeline flushes will noticeably impact execution speed.
> 
> > What about if the (initially-speculative) fetch of the next instruction faults; do we really need to mandate that interrupts have higher priority?
> 
> I don't see why this causes any difficulty. Instruction fetch faults already must be carried through the entire pipeline to the commit stage, which, as you noted, is also the obvious place to decide interrupt traps. Confronted with both, hardware simply has to prioritize the interrupt trap; no big deal.

MIT's RiscyOO injects interrupts in the rename stage as otherwise you needlessly do a whole bunch of speculative work only to throw it away in commit because an interrupt comes in (and may hurt performance if you have any outstanding TLB or cache misses from instructions you now want to throw away). Commit is not _the_ obvious place to inject interrupts, merely _an_ option that comes with certain tradeoffs.

jhauser-us:
> Yes, you can do it, and it likely doesn't hurt performance all that much, but why?

Because it simplifies the execution model for software, and makes it a little more consistent for all implementations.  Rather than claiming that the clarified requirements are difficult to satisfy, your real complaint seems to be that you have an existing implementation that doesn't satisfy them, which makes the spec clarification a big imposition on you and anyone else in your situation.  Right?  I think you might have more success just arguing that straight up.

However, see below about a possible mitigation.

> MIT's RiscyOO injects interrupts in the rename stage as otherwise you needlessly do a whole bunch of speculative work only to throw it away in commit because an interrupt comes in (and may hurt performance if you have any outstanding TLB or cache misses from instructions you now want to throw away). Commit is not `the` obvious place to inject interrupts, merely `an` option that comes with certain tradeoffs.

Okay, fair enough, but that's hardly an argument that RiscyOO can't prioritize interrupts over instruction fetch faults, is it?  I thought that was the real point.

Anyway, there is another way to handle reads of `mip`.  (I originally said there are a couple of ways, though I only gave the example of pipeline flushing.)  Let's call the value you currently have for `mip` the "internal `mip`".  When a CSR instruction reads `mip`, return this value:

    mstatus.MIE ? (internal mip) & (mideleg | ~mie) : (internal mip)

I haven't had time to fully consider all details of your situation, but I suspect if you do this, or some variant, you can restrict pipeline flushes just to CSR writes.  (The expression above can be further optimized a little, but I leave that as an exercise for an implementor such as yourself.)

sorear:
seL4 has code which polls for interrupts during loops operating on critical data structures.  I think the current risc-v implementation does MMIO accesses to the PLIC to check for pending interrupts, but I had a work item to change that to read sip on the theory that a sip read would be much faster than an uncached memory access.

jrtc27:
> > Yes, you can do it, and it likely doesn't hurt performance all that much, but why?
> 
> Because it simplifies the execution model for software, and makes it a little more consistent for all implementations. Rather than claiming that the clarified requirements are difficult to satisfy, your real complaint seems to be that you have an existing implementation that doesn't satisfy them, which makes the spec clarification a big imposition on you and anyone else in your situation. Right? I think you might have more success just arguing that straight up.
> 
> However, see below about a possible mitigation.
> 
> > MIT's RiscyOO injects interrupts in the rename stage as otherwise you needlessly do a whole bunch of speculative work only to throw it away in commit because an interrupt comes in (and may hurt performance if you have any outstanding TLB or cache misses from instructions you now want to throw away). Commit is not `the` obvious place to inject interrupts, merely `an` option that comes with certain tradeoffs.
> 
> Okay, fair enough, but that's hardly an argument that RiscyOO can't prioritize interrupts over instruction fetch faults, is it? I thought that was the real point.

Sure, it _can_ do but it forces the interrupt injection to be in commit not rename. My argument is more general than you say; it's not just about there being existing implementations that this change would break, but also that it's completely reasonable to expect future implementations to want to do that too and you would be constraining there designs in ways that make very little difference for systems software. So yes, that's part of my argument, but I actually think it's _better_ to be arguing about the merits of the change rather than whether it constraints designs as sometimes the change should still be done.

> Anyway, there is another way to handle reads of `mip`. (I originally said there are a couple of ways, though I only gave the example of pipeline flushing.) Let's call the value you currently have for `mip` the "internal `mip`". When a CSR instruction reads `mip`, return this value:
> 
> ```
> mstatus.MIE ? (internal mip) & (mideleg | ~mie) : (internal mip)
> ```
> 
> I haven't had time to fully consider all details of your situation, but I suspect if you do this, or some variant, you can restrict pipeline flushes just to CSR writes. (The expression above can be further optimized a little, but I leave that as an exercise for an implementor such as yourself.)

I don't understand what you're trying to do here; is your intent to just pretend that interrupts are never pending unless masked out for a given mode for whatever reason? That's not exactly in the spirit of things, and if that's the workaround you're encouraging then I would say that's a strong indication that this change really is a bad idea, because you'll just end up forcing performant implementations to lie rather than provide potentially-useful information.

gfavor:
On Mon, Nov 23, 2020 at 1:11 PM sorear <notifications@github.com> wrote:

> seL4 has code which polls for interrupts during loops operating on
> critical data structures. I think the current risc-v implementation does
> MMIO accesses to the PLIC to check for pending interrupts, but I had a work
> item to change that to read sip on the theory that a sip read would be much
> faster than an uncached memory access.
>
Speaking on behalf of a high-performance superscalar out-of-order RISC-V
design, reading and writing CSRs can be tricky.  Less so for some CSRs and
more so for others.  And writing CSRs generally is the trickier thing to do
properly so as to remain architecturally compliant (e.g. resorting to
various degrees of pipeline "synchronization").

Taking our design as a data point, a mip or sip CSR read may sometimes have
more than a one-cycle "cost", but it is far, far cheaper than an uncached
read of a memory-mapped register.  I would like to think that this is
qualitatively representative of many designs (and hence I believe your
optimization would be a good and successful one).

Greg

jscheid-ventana:
This should accommodate External Debug dcsr.stepie in some manner, since that feature may cause a preciseness property to fail.

jhauser-us:
Joshua Scheid wrote:
> This should accommodate External Debug dcsr.stepie in some manner, since that feature may cause a preciseness property to fail.

Please clarify, Joshua, what you mean by "this", and what exactly leads to the "fail" you speak of.

jhauser-us:
I wrote:
> Okay, fair enough, but that's hardly an argument that RiscyOO can't prioritize interrupts over instruction fetch faults, is it? I thought that was the real point.

Jessica Clarke:
> Sure, it can do but it forces the interrupt injection to be in commit not rename.  [...]

That's not at all what I meant.  You may have misunderstood; I was assuming the problem you have with the `mip` read could be handled separately, such as the trick I suggested for making the visible `mip` be different than the internal `mip`.

> I don't understand what you're trying to do here; is your intent to just pretend that interrupts are never pending unless masked out for a given mode for whatever reason? That's not exactly in the spirit of things, and if that's the workaround you're encouraging then I would say that's a strong indication that this change really is a bad idea, because you'll just end up forcing performant implementations to lie rather than provide potentially-useful information.

I disagree with that characterization.  I would say it demonstrates there are multiple, not particularly costly ways to meet the requirement, not nearly as onerous as you have suggested.

There's really no point in informing software (through `mip`) that an interrupt is about to occur "within a bounded number of cycles", as you've proposed it.  What use would software make of this information?  It could wait idly for the interrupt to occur (would've happened anyway), try to stop the interrupt (bizarre, and frequently futile), or ignore it (not really a use).  What else?

joshuascheid:
> 
> 
> Joshua Scheid wrote:
> 
> > This should accommodate External Debug dcsr.stepie in some manner, since that feature may cause a preciseness property to fail.
> 
> Please clarify, Joshua, what you mean by "this", and what exactly leads to the "fail" you speak of.

dcsr.stepie here:

https://github.com/riscv/riscv-debug-spec/blob/b860d530a32e10128a82b80b22b05f25d7d8f8b3/xml/core_registers.xml#L58

If the debugger has set stepie=0 and is single stepping, the behavior from the target mode's perspective can be that the trap does not occur immediately.  So a sequence that would otherwise never occur with the proposed change, could occur: an interrupt could be observed by SW as pending-and-enabled, but not yet taken (single-stepping over a CSR read of `mip`).

jrtc27:
> > Joshua Scheid wrote:
> > > This should accommodate External Debug dcsr.stepie in some manner, since that feature may cause a preciseness property to fail.
> > 
> > 
> > Please clarify, Joshua, what you mean by "this", and what exactly leads to the "fail" you speak of.
> 
> dcsr.stepie here:
> 
> https://github.com/riscv/riscv-debug-spec/blob/b860d530a32e10128a82b80b22b05f25d7d8f8b3/xml/core_registers.xml#L58
> 
> If the debugger has set stepie=0 and is single stepping, the behavior from the target mode's perspective can be that the trap does not occur immediately. So a sequence that would otherwise never occur with the proposed change, could occur: an interrupt could be observed by SW as pending-and-enabled, but not yet taken (single-stepping over a CSR read of `mip`).

Just to be clear, do you intend your message as saying there's another reason why observing an interrupt being pending but not being taken on the next instruction is behaviour that the ISA should permit and in fact already does, i.e. this pull request is overly-restrictive and conflicts with the existing debug spec?

jhauser-us:
Joshua Scheid wrote:
> If the debugger has set stepie=0 and is single stepping, the behavior from the target mode's perspective can be that the trap does not occur immediately.

The debugger is actually a higher authority than the ISA manuals.  Via the debug interface, the execution of a program can be arbitrarily scrambled, by changing register values and causing control jumps without explanation in the software.  Preventing interrupts from being taken for indefinitely long times is already potentially detectable by software, to the point of even causing crashes, regardless of whether pending interrupts are visible in `mip`.

Just as it does concerning some other ISA details (consider the effect of `dcsr` on EBREAK, for example), the RISC-V debug interface may legitimately override the constraints this pull request would add.  We only need to document what behavior the debug interface requires of implementations.  The debugger or its user is then responsible for whatever disruption may be caused while single-stepping a program with `stepie` = 0.

I would describe `stepie` as an additional global interrupt enable, much like MIE and SIE in `mstatus`, but hidden to software running on a core.  With that interpretation, deferred interrupts would accumulate and be visible to software in `mip` when single-stepping with `stepie` = 0, even when MIE = SIE = 1.

jscheid-ventana:
> Just to be clear, do you intend your message as saying there's another reason why observing an interrupt being pending but not being taken on the next instruction is behaviour that the ISA should permit and in fact already does, i.e. this pull request is overly-restrictive and conflicts with the existing debug spec?

It depends on how much the debug architecture is integrated.  I'm saying that correct M-mode code that depends on the precision (there's an `HCF` instruction next instead of a spin loop) will not operate as expected if single-stepped through in this manner.

jscheid-ventana:
John wrote:
> I would describe `stepie` as an additional global interrupt enable, much like MIE and SIE in `mstatus`, but hidden to software running on a core. With that interpretation, deferred interrupts would accumulate and be visible to software in `mip` when single-stepping with `stepie` = 0, even when MIE = SIE = 1.

Yes, this is the type of explicit consideration I was looking for.  I'm unsure how you'd mention this explicitly, other than perhaps that the MEE might have additional mechanisms for disabling/deferring interrupts.

jscheid-ventana:
In the interest in getting this clarification merged, what else needs to be addressed?  Regarding `stepie`, I agree with John's definition that that is just part of the "enabled" logic cone.

jrtc27:
> In the interest in getting this clarification merged, what else needs to be addressed? Regarding `stepie`, I agree with John's definition that that is just part of the "enabled" logic cone.

Well there's still disagreement over a significant amount of the core content. Consensus needs to be reached on what should be specified before you can think about finalising the exact wording and merging it...

jrtc27:
Note that, for example, Armv8 does _not_ define something as strict as this PR, and takes the approach I originally suggested (namely that the interrupt is taken in a bounded amount of time):

> Any interrupt that is pending before a Context synchronization event in the following list, is taken before the first instruction after the context synchronizing event, provided that the pending interrupt is not masked:
> ..
> In the absence of a specific requirement to take an interrupt, the architecture only requires that unmasked pending
interrupts are taken in finite time.
> If an unmasked interrupt was pending but is changed to not pending before it is taken, then the architecture permits the interrupt to be taken, but does not require this to happen. If the interrupt is taken then it must be taken before the first Context synchronization event after the interrupt was changed to not pending.
-- D1.13.4, DDI0487F_b_armv8_arm.pdf

Effectively, "context synchronisation event" forces a pipeline flush so you might as well take the interrupt immediately and it's probably useful to have that barrier, but in the general case there is no good reason why you should force a flush, because software really should not care if a few more instructions execute before the interrupt fires (quite frankly there's rarely a good reason to even look at x[ST]IP). Yes, you can write software that notices the difference, but that's not the point, the point is that there is no legitimate reason I'm aware of where you'd ever write code that cares about the difference.

jhauser-us:
Andrew Waterman has told me he wants to retain the strictness of the proposed text in certain circumstances but not always.  (Possibly, he has in mind something more like the Arm architecture, but I don't know for sure.)  We are now waiting for Andrew to provide his own version, but unfortunately he has also warned it might be a long time before he gets to it.

In the meantime, it can be said, adhering to the totally strict version of this pull request should guarantee conformance with whatever may ultimately be adopted, but the extent to which that might be overkill is uncertain (at least to me).

aswaterman:
It's important that pending interrupts are taken as soon as soon as an xRET is executed.  This is something that software actually expects (e.g., embedded systems that use pending software interrupts for task-scheduling purposes).

The case for requiring the same thing for reads of xIP is admittedly weaker, AFAICS.

jrtc27:
> It's important that pending interrupts are taken as soon as soon as an xRET is executed. This is something that software actually expects (e.g., embedded systems that use pending software interrupts for task-scheduling purposes).
> 
> The case for requiring the same thing for reads of xIP is admittedly weaker, AFAICS.

xRET is fine, exception return is one of the context synchronisation events that Arm explicitly state counts and is likely to require a full pipeline flush anyway (or a _lot_ of complex logic already to be able to speculate across exception levels, which is probably not a good idea in the age of speculative execution attacks anyway). The same is also true for exception entry (both the requirement and that it's not overly restrictive).

gfavor:
On Mon, Mar 1, 2021 at 2:31 PM Andrew Waterman <notifications@github.com>
wrote:

> It's important that pending interrupts are taken as soon as soon as an
> xRET is executed. This is something that software actually expects (e.g.,
> embedded systems that use pending software interrupts for task-scheduling
> purposes).
>
FWIW, x86 has a very strict requirement.  If two consecutive CSR
instructions (in terms of RV ISA) enable and disable interrupts, then
software expects any pending interrupt to be recognized during that tiny
window (and, of course, not outside of it).  This was a classic idiom where
software wanted to occasionally check for and allow recognition of pending
interrupts, but only at controlled points in the code.

ARM is looser and only requires interrupt recognition on the following
context synchronization events:  ISB execution, exception entry and exit,
and debug exit.  Then v8.5 (I think) introduced bits to control whether
exception entry and exit are context synchronization events.

This PR takes a close-to-Intel approach.  Given the strong default
synchronization that RISC-V applies around CSR accesses, this PR for
synchronization around interrupt enable/disable would be
philosophically consistent.

Personally I had interpreted the Priv spec (way back) to intend the
stricter synchronization and hence viewed this PR as just clarifying the
intention of the ratified spec.  In switching to a potentially far looser
requirement, is there a danger that existing software depends on there
being a limited delay - versus being alright if interrupt recognition
doesn't happen until the next xRET is executed?

Maybe we can point at ARM and say that it should be alright for the
architecture to loosen what may have been viewed as a tighter requirement.
But I doubt that there are implementations that exhibit anywhere close to
that allowed behavior (and that there would be some unhappy people if an
implementation took advantage of what the architecture allowed).  It seems
poor for RISC-V architecture to only require what may, in practice, allow
unacceptable behavior.

Greg


> The case for requiring the same thing for reads of xIP is admittedly
> weaker, AFAICS.
>
Is the alternative that a pending enabled interrupt can be visible in xIP
yet remain pending for an indeterminate amount of time (until an eventual
xRET)?

aswaterman:
> Is the alternative that a pending enabled interrupt can be visible in xIP yet remain pending for an indeterminate amount of time (until an eventual xRET)?

Making writes to xip synchronize the taking of interrupts is a less-extreme alternative.  That would handle the x86 case you mentioned, while also keeping CSR reads simple in simple implementations.

gfavor:
This is essentially the interpretation that I have had of what the existing
ratified spec intended.  But irrespective, I would be for this level of
synchronization as striking a reasonable balance between providing
synchronization guarantees to software and being consistent with the
general RISC-V architectural philosophy around CSR writes
and their effects, and with burdening implementations unduly.

Greg

aswaterman:
Unbelievably lengthy discussion superseded by #701

PR #595, opened 2020-10-02
Disabling H extension does not ensure illegal instruction traps

jhauser-us:
The consequences of disabling an extension by clearing a bit in misa is specified with the misa CSR.


aswaterman:
👍

PR #599, opened 2020-10-10
Update PAUSE description.

gfavor:
Updated PAUSE description.


aswaterman:
Thanks, I've incorporated these changes into the other PR.

PR #601, opened 2020-10-13
Both HWBPs and EBREAKs populate mtval

### LINKED ISSUE #600: EBREAK and mtval
The spec remarks that mtval is written with the virtual address on hardware breakpoints.  It is mum about the EBREAK instruction, which currently means mtval should be set to 0.

There are two questions:
- Is this the right definition?  Should EBREAK also be allowed to set mtval to the PC?  (Note that making this change is backwards compatible, since zeroing mtval remains a legal implementation, at least at the ISA level.)
- Do we need to clarify somewhere that EBREAK is not a hardware breakpoint?

aswaterman:
This is a normative change, but it is backwards compatible, since writing 0 to mtval remains a legal option.

Resolves #600

### DIFF
--- a/src/machine.tex
+++ b/src/machine.tex
@@ -2226,7 +2226,7 @@ implementation, though it may be explicitly written by software.  The hardware
 platform will specify which exceptions must set {\tt mtval} informatively and
 which may unconditionally set it to zero.
 
-When a hardware breakpoint is triggered, or an
+When a breakpoint,
 address-misaligned, access-fault, or page-fault exception occurs
 on an instruction fetch, load, or store, {\tt
   mtval} is written with the faulting virtual address.  On an illegal
--- a/src/supervisor.tex
+++ b/src/supervisor.tex
@@ -727,8 +727,9 @@ which exceptions must set {\tt stval} informatively and which may
 unconditionally set it to zero.
 
 
-When a hardware breakpoint is triggered, or an instruction, load, or
-store address-misaligned, access-fault, or page-fault exception occurs, {\tt stval}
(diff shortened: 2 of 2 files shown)


knightsifive:
Is the technical term "breakpoint exception" defined earlier?

gfavor:
Effectively yes - see Table 3.7 in the Priv spec.

>

knightsifive:
To clarify, the "breakpoint exceptions" are precisely the three `mcause == 3` exceptions:

- Instruction address breakpoint
- Environment break
- Load/Store/AMO address breakpoint 

in Table 3.7?

gfavor:
Yes, all the (three) line items with Exception Code = 3.

gfavor:
Which raises the potential need for a clarification?  It's clear that "Environment break" results from an EBREAK instruction, whereas the address breakpoints (resulting from Trigger Module address triggers firing) need to be understood to be "hardware" breakpoints.

knightsifive:
Okay. This matches my understanding from a (long ago) reading of Chapter 10 of the RISC-V Reader (pp.101-102):

> Breakpoint exceptions arise from executing an `ebreak` instruction, or when an address or datum matches a debug trigger

I guess I never read this part of the Priv. spec. that carefully to notice the difference. Good spot, Andrew.

aswaterman:
@allenjbaum gets credit for spotting the inconsistency; I merely proposed the wording of the fix!

aswaterman:
I'm going to merge this since it doesn't seem controversial, but please reopen #600 if further discussion is warranted.

PR #602, opened 2020-10-13
Update README.md

shreepads:
Fix broken link to specifications


aswaterman:
We should fix the broken link on the riscv.org web site.  I'll leave this PR open until that is resolved.

cetola:
This redirect has been fixed. It now works with or without the slash:

https://riscv.org/technical/specifications/

https://riscv.org/technical/specifications

aswaterman:
Thanks for the fix @cetola and for the report @shreepads.  Closing the issue now that it's resolved.

PR #606, opened 2020-10-15
Fix specification link

kvnsmth:
Looks like a website update changed the link.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ The Manual is split up into the following volumes:
 - Volume II: Privileged Architecture
 
 **Official versions** of the specifications are available at
-https://riscv.org/specifications/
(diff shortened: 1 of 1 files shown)


aswaterman:
The riscv.org website should be fixed rather than changing the link here.

@cetola Seems like this is happening again.  This time, the redirect for `/specifications` (without trailing slash) works, but `/specifications/` redirects to `/specifications/technical/specifications/`, which is 404.

kvnsmth:
@aswaterman happy to have y’all fix as you see fit. I’ll go ahead and close this PR.

aswaterman:
Thanks for raising the issue.

kvnsmth:
You’re welcome!

PR #610, opened 2020-11-06
PMP uses physical addresses (not effective addresses)

pdonahue-ventana:


aswaterman:
Thanks for this important correction.

PR #611, opened 2020-11-11
Virtual memory task group updates

daniellustig:
Working draft for the work being performed by the RISC-V Virtual Memory task group.  Not yet ready for freeze or ratification, but open for comments.


aswaterman (changes requested):
I'm just writing this comment so there's a big red X on the merge button for the time being; take no offense :)

src/supervisor.tex
```diff
@@ -1465,16 +1486,18 @@ \subsection{Virtual Address Translation Process}
...
-  \item If this access violates a PMA or PMP check, raise an access-fault exception
-    corresponding to the original access type.
-  \item This update and the loading of $pte$ in step 2 must be atomic; in
-    particular, no intervening store to the PTE may be perceived to have
-    occurred in-between.
+  \item If a store to $pte$ would violate a PMA or PMP check, raise an
+    access-fault exception corresponding to the original access type.
+  \item Perform the following steps atomically:
+    \begin{itemize}
+      \item Compare $pte$ to the value of the PTE at address $a+va.vpn[i]\times \textrm{PTESIZE}$.
```
sorear, line 1496: I'm wondering if it would make sense to, if $pte$ is a NAPOT PTE, allow these three steps to be performed against the memory location of the first equivalent PTE.  The motivation would be a reduction in the number of stores which have to be performed.

src/supervisor.tex
```diff
@@ -1619,8 +1707,18 @@ \subsection{Addressing and Memory Protection}
...
-for future standard use and must be zeroed by software for forward compatibility.
+have the same meaning as for Sv32.
+
+Bit 63 indicates that the PTE uses a custom implementation-specific encoding.
+If bit 63 is set, the algorithm for virtual-to-physical address translation
+is implementation-defined.  If bit 63 is not set, the algorithm for
+virtual-to-physical address translation is described in
+Section~\ref{napotpages}.
+
+Bit 62 indicates the use of naturally aligned power-of-two (NAPOT)
```
sorear, line 1718: IMO these bits should have names so we don't have to say "bit 62" repeatedly.

src/supervisor.tex
```diff
@@ -1643,6 +1741,99 @@ \subsection{Addressing and Memory Protection}
...
+  the pre-existing Sv39 and Sv48 address translation algorithms so as not to
+  disrupt existing implementations or designs that choose not to implement the
+  scheme.
+\end{commentary}
+
+\begin{table*}[h!]
+\begin{center}
+\begin{tabular}{|c|c||c|l|}
+\hline
+Bit 62 & $pte.ppn[i]$      & $pte.napot\_bits$ & Meaning if $i=0$         \\
```
sorear, line 1766: I don't see where napot_bits is defined.  If the intention is for the size to be encoded implicitly by analogy with NAPOT PMP ranges, as suggested by the 1 bit in the encoding, that requires all z to be set to 0.

src/supervisor.tex
```diff
@@ -1741,7 +1932,7 @@ \subsection{Addressing and Memory Protection}
 \instbit{1} &
 \instbit{0} \\
 \hline
-\multicolumn{1}{|c|}{\it Reserved} &
+\multicolumn{1}{|c|}{\it attributes} &
```
sorear, line 1935: Possibly orphaned, there is no other mention of attributes in this draft?

src/supervisor.tex
```diff
@@ -1005,8 +1006,11 @@ \subsection{Supervisor Memory-Management Fence Instruction}
 these data structures; however, these implicit references are ordinarily not
 ordered with respect to explicit loads and stores.  Executing
 an SFENCE.VMA instruction guarantees that any previous stores already visible
-to the current RISC-V hart are ordered before all subsequent implicit references
-from that hart to the memory-management data structures.
+to the current RISC-V hart are ordered before certain implicit references by
+subsequent instructions in that hart to the memory-management data structures,
+and vice versa.  SFENCE.VMA is also used to invalidate entries in the
```
sorear, line 1011: This is a little bit confusing descriptively, since in previous versions of the document (and possibly elsewhere in the current one, I haven't reread other chapters in the new context) the ordering was intended to subsume incoherent caching.  If the ordering is now separate from the incoherent caching that changes the interpretation of various things.  A more formal model would help.

src/supervisor.tex
```diff
@@ -1765,7 +1956,7 @@ \subsection{Addressing and Memory Protection}
 \label{sv48pte}
 \end{figure*}
 
-The PTE format for Sv48 is shown in Figure~\ref{sv48pte}.  Bits 9--0
+The PTE format for Sv48 is shown in Figure~\ref{sv48pte}.  Bits 63--54 and 9--0
```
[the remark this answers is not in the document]
daniellustig: Fixed, thanks

src/supervisor.tex
daniellustig: The virtual memory TG is considering adding other attributes, but since we haven't finalized anything, I've changed this back to Reserved

src/supervisor.tex
daniellustig: Good feedback.  I added a new paragraph of commentary explaining why we did this: https://github.com/riscv/riscv-isa-manual/pull/611/files#diff-1673d75098c5b339d06598ff6bb126f7a429d6bed61593e9b41092ee0587c9b9R1530.  Hopefully that helps.

And yes, integration with the formal model is on the task group's to-do list, and of course if we find any discrepancies along the way, we'll make sure to update this draft before ratification.

src/supervisor.tex
daniellustig: If I understand your suggestion correctly, that would require a change to the address-translation algorithm (and the page table walker in HW) to perform a bunch of extra loads in order to find whichever NAPOT PTE is first?  We wanted to stick with the address-translation algorithm as-is, in order to make it easier for implementations to adopt this idea.

src/supervisor.tex
daniellustig: This table is the definition of napot_bits.

src/supervisor.tex
sorear: No extra loads, just mask off bits corresponding to the decoded napot_bits.

(In open-source implementations and discussions) I've seen ~3 ways to handle the A/D bits:

* Punt and always raise a page fault.  No change is needed.
* Do a page table walk using normal loads, then do a CAS (or LR/SC pair) to update the final PTE was fetched.  In this case we would only need to mask off bits corresponding to napot_bits.
* Do a page table walk using loads with LR semantics, then do a SC to update the PTE.  This case does require (exactly one) additional load, but it may be undesirable to begin with since it requires taking ownership of all PTEs, even if A=D=1 already.

src/supervisor.tex
sorear: OK, that makes sense now.  So we could have larger NAPOT sizes in future extensions by setting the marked bit to 0, but we couldn't have smaller NAPOT sizes without bit permutation.

Years ago there was a proposal that "small" large pages could be implemented microarchitecturally, by observing all PTEs in a cache line when a fetch from memory is done.  That works up to 32 kB for a 4 kB architectural page and 8 pointers per cache line.  So maybe we don't need smaller NAPOT pages.

src/supervisor.tex
sorear: That does help, thank you.

src/supervisor.tex
daniellustig: The intention is that the bit position could in principle shift anywhere to indicate what size NAPOT region is needed, e.g.,

| N | PPN           | NAPOT region size |
|---|---------------|-------------------|
| 0 | `y yyyy yyyy` | 4KiB              |
| 1 | `y yyyy yy10` | 16KiB             |
| 1 | `y yyyy 1000` | 64KiB             |
| 1 | `y yy10 0000` | 256KiB            |
| 1 | `y 1000 0000` | 1MiB              |

We're starting with just one option, and reserving those z bits rather than forcing them to be zero now for NAPOT encoding, in case we decide later we'd rather use those z bits for something else more valuable.

sorear: Makes sense.

src/supervisor.tex
```diff
@@ -1726,16 +1739,31 @@ \subsection{Addressing and Memory Protection}
...
-Section~\ref{napotpages}.
+\begin{table}[h]
+\begin{center}
+\begin{tabular}{|c|c||l|}
+\hline
+C & V & Description \\
+\hline
+0 & 0 & The PTE is invalid; all bits except C are don't cares and can be used freely by software \\
+0 & 1 & The PTE is valid and uses standard address-translation rules \\
+1 & 0 & {\em Reserved for standard use} \\
```
sorear, line 1750: This is a backward incompatible change.  The ratified spec has:

> The V bit indicates whether the PTE is valid; if it is 0, all other bits in the PTE are don’t-cares and may be used freely by software.

On Linux you would need to create a 64 EiB swap file (may be slightly off) to create a PTE with the high bit set, but there may be other supervisors that can create such PTEs under realistic circumstances.

daniellustig: When V=1, bit 63 is currently reserved and must be zeroed by software for forward compatibility.  The PPN field doesn't automatically expand to fill all the upper bits.  So I don't think this one is backwards-incompatible.

When V=0, yes, we would be changing the meaning of bit 63 from don't care to reserved, which is technically a backwards-incompatible change.  We don't currently have a specific plan for the V=0, C=1 combination now marked reserved, so I don't see a problem if we wanted to change that C bit back to don't care when V=0 and maintain compatibility.

sorear: I am referring only to the V=0 C=1 combination, which is non-reserved in the ratified spec.

daniellustig: This table was a mistake on my part; apologies.  Hopefully the new table (with C and N) makes more sense.

jrtc27:
CHERI-RISC-V deliberately allocated its additional PTE bits from the top down so as to try to avoid conflicting with any additional future PTE bits. Is there a reason why you chose to do the same, rather than allocating from the low bits upwards as one would normally expect?

daniellustig:
> CHERI-RISC-V deliberately allocated its additional PTE bits from the top down so as to try to avoid conflicting with any additional future PTE bits. Is there a reason why you chose to do the same, rather than allocating from the low bits upwards as one would normally expect?

The main reason was to allow more easily for future growth of the PPN field, if necessary.  It hadn't yet come up in our group discussions that CHERI-RISC-V already used those upper bits.

jrtc27:
> > CHERI-RISC-V deliberately allocated its additional PTE bits from the top down so as to try to avoid conflicting with any additional future PTE bits. Is there a reason why you chose to do the same, rather than allocating from the low bits upwards as one would normally expect?
> 
> The main reason was to allow more easily for future growth of the PPN field, if necessary. It hadn't yet come up in our group discussions that CHERI-RISC-V already used those upper bits.

Yeah, unfortunately we needed to do something and at the time there was nothing in the spec to accommodate non-standard permission bits (short of defining a new SATP mode entirely which is a bit overkill and there are only 2 of those free but we'd need a variant for each standard mode). If you want to see exactly what bits we're using for what it's in Section 5.3.10 Virtual Memory and Page Tables (page 165 onwards) of [CHERI ISAv8](https://www.cl.cam.ac.uk/techreports/UCAM-CL-TR-951.pdf) (page 167 Figure 5.3 has a bitfield diagram to summarise).

src/supervisor.tex
sorear: New table version looks good to me.

sorear:
Personally I think it makes the most sense for research prototypes and custom embedded systems to use the highest-order PPN bits, which are known to not be needed in the closed environment.  This avoids conflict with future standard extensions, which necessarily use bits that will not conflict with even the largest systems.  I think in many cases this is implicitly allowed by the current spec insofar as the semantics of physical addresses are completely implementation defined, but something like it could be codified.

Tangentially, I/we've been wondering how much longer risc-v is going to make it without defining a 2*XLEN-bit page table entry format (c.f. PAE, LPAE).

jrtc27:
> Personally I think it makes the most sense for research prototypes and custom embedded systems to use the highest-order PPN bits, which are known to not be needed in the closed environment. This avoids conflict with future standard extensions, which necessarily use bits that will not conflict with even the largest systems. I think in many cases this is implicitly allowed by the current spec insofar as the semantics of physical addresses are completely implementation defined, but something like it could be codified.

Using the high PPN bits is not a good idea; it requires changing everything everywhere to mask them out in a non-standard way, and such patches must remain downstream. By using other bits we only need to mess with them in the places we really need to set/get them. For example, simple things like OpenOCD/GDB translation now needs to be patched to throw away bits that the specification says *must* be taken into account, whereas currently we can leave that code entirely untouched.

src/supervisor.tex
```diff
@@ -1487,6 +1519,98 @@ \subsection{Virtual Address Translation Process}
...
+address-translation cache.  Implementations must only perform implicit reads of
+the translation data structures pointed to by the current contents of the {\tt
+satp} register or a subsequent valid (V=1) translation data structure entry.
+
+Speculative executions of the address-translation algorithm behave as
+non-speculative executions of the algorithm do, except that they must not set the
+dirty bit for a PTE, they must not trigger an exception, and they must not create
+address-translation cache entries if those entries would have been invalidated
+by any SFENCE.VMA instruction executed by the hart since the speculative
+execution of the algorithm began.
```
dkruckemyer-ventana, line 1578: I'm not sure this description covers the case of speculation immediately after reset, unless we're saying that the ATC entries cannot be populated until an SFENCE.VMA of the appropriate type is executed. One could certainly interpret the statement "at any time since the most recent SFENCE.VMA that subsumes that address" in line 1524 above as implying ATC entries cannot be established until the first SFENCE.VMA is executed. But I doubt that's what was intended?

src/supervisor.tex
```diff
@@ -1100,6 +1114,11 @@ \subsection{Supervisor Memory-Management Fence Instruction}
 In this case, a similar remark applies: it is unpredictable whether the old
 non-leaf PTE or the new leaf PTE is used, but the behavior is otherwise well
 defined.
+
+Another consequence of this specification is that it is generally unsafe to
+update a PTE using a set of stores of a width less than the width of the PTE,
+as it is legal for the implementation to read the PTE at any time, including
+when only some of the partial stores have taken effect.
```
dkruckemyer-ventana, line 1121: Do we state somewhere that the atomicity of an implicit PTE read (or write in the case of the A/D bits) is the same as the size of the PTE? Or are we not requiring that?

allenjbaum:
There is no requirement that satp.MODE has any particular value at reset,
but regardless of the mode, it is treated as "bare" if you're in M-mode
(and MPRV=0, which reset requires).
So, I'm wondering if the text above should be re-worded to be

as long as *the effective virtual address mode ( indicated by* {\tt
satp}.MODE *and the hart's privilege mode)* indicates that virtual
addressing is enabled.

src/supervisor.tex
gfavor: Daniel and I have a related question outstanding with Andrew whihc potentially would cover this.  Namely, can hardware start using a new satp value right away if it is written in M-mode (and start caching the results), or must hardware wait until execution has switched to the privilege mode relevant to satp (i.e S/U modes)?  (For vsatp and hgatp the same question would be wrt M/HS modes being "ok" versus VS/VU modes being required before hardware can start doing tablewalks using these.)

The latter choice clearly implies that speculative use of satp can't start immediately after reset (and not until M-mode is exited).  The former choice would also limit speculative use of satp until it at least is written by software.

At the same time, more generally, satp is undefined after reset, so it would be bad in any case to allow use of that initial satp value.

gfavor:
To be more precise, while satp is undefined after reset, the "it" that is treated as "bare" is not satp.MODE but that the translation mode while in M-mode is always "Bare" (independent of any *atp registers).

The suggested change by Allen corresponds to one of the two options I mentioned in my last post above, i.e. hardware can't start using a new *atp value until execution enters a relevant privilege mode for that *atp register.

src/supervisor.tex
daniellustig: Resolved in latest commit: translation cannot start until the "effective privilege mode" is U or S (or hypervisor equivalent)

daniellustig:
> Maybe delete the word "current" since it seems at odds with the preceding sentence referring to effective privilege mode (not that saying "current" is now wrong). It seems like it isn't necessary to have any qualifier to "privilege mode".

I think you're referring to the following new text?

```
when checking the U~bit, the current privilege mode is always taken to be U-mode;
```

I used "current" there to match the (pre-existing) wording within the address-translation algorithm itself:

```
  Determine if the requested memory access is
  allowed by the $pte.r$, $pte.w$, $pte.x$, and $pte.u$ bits, given the
  current privilege mode and the value of the SUM and MXR fields of
  the {\tt mstatus} register."
```

kasanovic:
@jrtc27   We have been clear that we will not consider projects using non-standard reserved encodings as a constraint on future standards development.  That said, we should be trying to ensure there is some custom encoding available in each space to allow experimentation alongside standard encodings - though some spaces are valuable and custom is a one-way street as we promise not to reclaim anything once it's made custom.   In this case, where you want to build an orthogonal universe around the standard model and there is not enough custom space available in the space you would like (satp), I'd suggest adding additional custom CSRs holding mode bits and perhaps new instructions to allow those to be set atomically along with custom satp in your prototype.

daniellustig:
For anyone following along: we did a bit of reorganizing.  Svnapot has been factored out into its own branch, #628.  Sv57 (#626) has been manually merged into this branch.  The plan is for the changes in this branch to be reviewed and merged as part of the Priv 1.12 review process.  Meanwhile, Svnapot will go through its own review process as a standalone extension, and that timeline won't have to block anything else in this branch.  Sorry for any logistical confusion.

src/supervisor.tex
```diff
@@ -1092,6 +1092,14 @@ \subsection{Supervisor Memory-Management Fence Instruction}
       rs2}, except for entries containing global mappings.
 \end{itemize}
 
+If the value held in {\em rs1} is not a valid virtual address, then the
+SFENCE.VMA instruction has no effect.  No exception is raised in this case.
+
```
gfavor, line 1097: The following non-normative comment seems inconsistent with the preceding normative statement.  The preceding requires that an SFENCE.VMA with an invalid VA has no effect (as in must have no effect).   Whereas the following comment suggests that such an SFENCE.VMA may in fact have an effect.  So which is the intention?  

The former seemingly would be preferable so that such an SFENCE.VMA could reliably be turned into an SFENCE.VMA that only has ordering effects.  But this assumes that some address (e.g. 0x7FFF_FFFF_FFFF_FFFF) will be invalid across all implementations, for now and forever.  Which seems pretty brittle (starting with wondering how J-ext pointer masking and memory tagging will interact).

So it seems like the above normative statement should be softened to indicate that "no effect" is an allowed behavior (since requiring "no effect" has no real benefit).

src/supervisor.tex
```diff
@@ -1092,6 +1092,14 @@ \subsection{Supervisor Memory-Management Fence Instruction}
...
+\begin{commentary}
+This ensures it remains legal for implementations to simply ignore the
+value in {\em rs1}.
```
gfavor, line 1100: But wouldn't an implementation that ignores the value in rs1 need to over-invalidate, not under-invalidate?  Performing no invalidation would certainly surprise software that's expecting at least the specified addresse to be invalidated.

src/supervisor.tex
```diff
@@ -1440,11 +1516,14 @@ \subsection{Virtual Address Translation Process}
...
 \item Let $a$ be ${\tt satp}.ppn \times \textrm{PAGESIZE}$, and let $i=\textrm{LEVELS} - 1$. (For Sv32, PAGESIZE=$2^{12}$ and LEVELS=2.)
+  The {\tt satp} register must be {\em active}, i.e., the effective privilege
+  mode must be S-mode or U-mode.
 
 \item Let $pte$ be the value of the PTE at address
   $a+va.vpn[i]\times \textrm{PTESIZE}$. (For Sv32, PTESIZE=4.)
-  If accessing $pte$ violates a PMA or PMP check, raise an
-  access-fault exception corresponding to the original access type.
+  If accessing $pte$ violates a PMA or PMP check, or if any reserved
+  bits in $pte$ are set, raise an access-fault exception corresponding to the
```
dkruckemyer-ventana, line 1525: Why does setting reserved bits in a PTE cause an access fault instead of a page fault? What will M-mode software do with this?

Seems a bit inconsistent to me since all other PTE issues end up as page faults, which can easily be delegated to S-mode.

src/supervisor.tex
daniellustig: I don't see a conflict here.  The specified behavior of SFENCE.VMA with invalid address is that it has no effect.

It is also true that SFENCE.VMA with invalid address may over-invalidate.  That's also true of any SFENCE.VMA.  It's also true in general: the implementation can invalidate address-translation cache entries any time it wants to, even when no SFENCE.VMA is executed.  That's an implementation detail at that point.  We shouldn't try to architecturally specify all the possible forms of over-invalidation.

src/supervisor.tex
daniellustig: To paraphrase a prior discussion with @aswaterman, invalid use of defined VM features results in a page fault, while invalid physical addresses cause access faults.  Making invalid use of reserved bits trigger access faults also ensures that the behavior would be consistent if any reserved bits are later repurposed as added PPN bits.

Invalid use of NAPOT PTEs (now on #628) is a page fault because it's invalid use of a defined feature, and it is also consistent with the way misaligned superpages are handled.

In some cases different decisions could have been made, but that's the reasoning for the current choices.

dkruckemyer-ventana: This is the first instance (I think) where the translation mechanism generates an access fault when the PTE is not well-formed (i.e. the reserved bits are non-zero).

The model seems simpler and more modular if the translation mechanism checks the PTEs for proper addresses and permissions and then hands off valid, well-formed PAs to be checked by the PMAs and PMPs. Conversely, if an access raises a page fault, software knows something was wrong with the translation process and data structures, but if an access raises an access fault, software knows that the fault lies somewhere besides the translation process and data structures.

jrtc27: It's then inconsistent if you repurpose the bits later to be flags (e.g. as is done for NAPOT). You can't win either way, but making it be a page fault is more natural than pretending they're address bits but actually aren't, and keeps the semantics from the existing spec version, which I think should always be done unless there's a very good reason why it _can't_ be.

jrtc27:
I hope the Zsn/Svnapot commits and reverts will be removed from the history before this lands?

daniellustig:
> I hope the Zsn/Svnapot commits and reverts will be removed from the history before this lands?

I don't mind cleaning up the history, or rebasing, or whatever the editors would prefer, once we're ready to freeze this or actually merge it.  @aswaterman any preference?

src/supervisor.tex
daniellustig: I'll forward this question to the task group for more discussion.

IIRC in earlier discussions we felt that the existing spec wasn't actually unambiguously clear about this case.  There is wording saying "Bits 63–54 are reserved for future standard use and must be zeroed by software for forward compatibility", but there isn't actually any explicit wording saying what happens if that is violated.  It doesn't show up explicitly in the translation algorithm for example.  It seemed like we needed to actually make a decision one way or the other in order to clarify a hole in the spec.  It wasn't the intention to change existing semantics.

@jrtc27 can you point to the pre-existing spec wording I'm missing that makes this clearly a page fault?  If so, then I agree we should stick with that.  If not, then we should just decide which is better and formalize it.

FWIW, spike also doesn't seem to check for this either way, while the Sail model considers this a page fault.

jrtc27: Hm, I could have sworn it was there, but you're right, I can't find it. All the implementations I know of either ignore the bits or raise a page fault, though, and I do think a page fault would be more natural.

daniellustig: In our discussions it was pointed out that Rocket treats this as an access fault rather than a page fault: https://github.com/chipsalliance/rocket-chip/blob/8f9149faf01fec9725d08c31c9c0bb5341f5e999/src/main/scala/rocket/PTW.scala#L154

So unfortunately we seem to have both behaviors in the wild already.

aswaterman:
> > I hope the Zsn/Svnapot commits and reverts will be removed from the history before this lands?
> 
> I don't mind cleaning up the history, or rebasing, or whatever the editors would prefer, once we're ready to freeze this or actually merge it. @aswaterman any preference?

I'd just squash & merge the PR.  (This obviously destroys more history than requested, but anyone interested in the history can come to this page to see it.)

src/supervisor.tex
daniellustig: After discussing, we have indeed changed this to a page fault.  The expectation is that all misconfigurations of a page table entry (use of reserved bits, misaligned superpages, invalid use of r/w/x, invalid use of Svnapot) will result in page faults, including if/when we start using more reserved bits for other purposes.  Access faults will remain for things like unsupported misaligned accesses or invalid physical addresses, which aren't specific to the translation process.

Note that this rule will apply once priv 1.12 is ratified.  It is not a retroactive change, as implementations raising access faults exist in the wild.

Thanks all for the feedback!

ingallsj: tracking Rocket PTW change request: https://github.com/chipsalliance/rocket-chip/issues/2795

mickflemm: If an implementation today raises a fault when reserved PTE bits are used, then it'll be incompatible with future software making use of those bits. Hw should just ignore reserved bits, or else we 'll need to fill the sw with checks on the critical path so that we choose PTE formats based on priv. spec revisions, or chose PTE formats at compile-time (meaning we can't have one kernel image for all systems).

It also seems a bit inconsistent to me, for CSRs for example WPRI fields ignore writes to reserved bits (hardwired to 0), WARL fields also ignore them ("Implementations will not raise an exception") and WLRL fields "are permitted but not required to raise an exception". If we write an invalid/reserved SATP value, it's ignored, if we use an invalid/reserved bit on sfence.vma 's rs2 it's ignored, and TBH I couldn't find a case on the current Priv. spec where writing a reserved bit/field requires the hw to raise an exception. Another inconsistency is that there are not just reserved PTE bits but also reserved combinations of standard PTEs, e.g. RW=01, or reserved combinations of D,A,U bits on non-leaf PTEs. If we are about to return a fault for invalid PTEs, we should include those cases as well.

daniellustig: This was already debated in the discussion above and in the virtual memory TG.  As of Priv 1.12, once this change is ratified, any invalid encoding (use of upper reserved bits, RW=01, reserved combinations of D,A,U) will trigger a page fault.  Plus, as discussed above, most implementations already raise a page fault or an access fault, and we'd like to make this consistent (as a page fault) moving forward.  Is there new information here that should cause us to open this up again?

What new checks would have to be performed on the critical path?  The OS obviously shouldn't attempt to use the new features if they aren't supported by a given implementation, and in both extensions we've proposed so far the zero(s) encoding is backwards compatible with the existing behavior.

mickflemm: I get it that invalid use of _defined_ VM features should result a fault, but we are talking about features to be defined at a later spec revision, using reserved bits currently unused. Why should we raise an exception on those where we can just ignore them ? And what is sw supposed to do in this case ? We won't even be able to call the trap handler to print the crash info, we'll just hang. We are turning this case to a fatal error for no apparent reason (the hw can work fine by ignoring those bits). As for the critical path and the changes needed, on Linux the core page-table related functions are defined on header files (in our case arch/riscv/include/asm/pgtable.h) so that they can be inlined, and the idea is to keep them as simple as possible (we do mappings all the time, this needs to be as efficient as possible), if we start adding checks there for every new PTE bit we decide to add in the future e.g. on set_pte_at we'll end up having something like powerpc's set_pte_filter (https://elixir.bootlin.com/linux/latest/source/arch/powerpc/mm/pgtable.c#L103) only worse (and that's the only arch I could find that follows this approach, others just use a different header file at compile time, something we want to avoid since we want to have one image that runs everywhere). I'm not saying it's not possible or wrong, I'm saying it's a bit ugly and sub-optimal sw-wise.

As for the consistency, the text still says "For non-leaf PTEs, the D, A, and U bits are reserved for future standard use and must be cleared by software for forward compatibility.", also "Writable pages must also be marked readable; the contrary combinations are reserved for future use", shouldn't we also say that a page fault must occur in those cases like we do for reserved PTE bits on Sv39 ? It's only mentioned on the translation process algorithm.

allenjbaum:
Why shouldn't using features on HW that doesn't implement them be
considered a buggy OS, and get an exception.
Why shouldn't that be considered a corrupt PTE?
A custom extension could supersede that, I suppose.
But - why couldn't the trap handler panic, print an error message, and shut
down the process?
Why will the error message (to console output, I assume) hang? (if it's a
user page table, and not an OS page table - but even kernel panics can
print an error message)
I'm also missing why you expect a single header file, unconditioned by the
architecture type.
I'm not saying you are wrong here - but your arguments need a lot more
backup.

src/priv-preface.tex
```diff
@@ -44,8 +44,9 @@ \chapter{Preface}
 \item PMP changes require an SFENCE.VMA on any hart that implements
   page-based virtual memory, even if VM is not currently enabled.
 \item Allowed for speculative updates of page table entry A bits.
-\item Clarify that PTEs with reserved bits set should trigger page-fault
-  exceptions.
+\item Clarify that PTEs with reserved bits set and non-leaf PTEs with D, A,
+  or U set should trigger page-fault exceptions when accessed by the
```
pdonahue-ventana, line 48: should or must?

daniellustig:
Content merged via #759.  Closing.

PR #614, opened 2020-11-19
Dedicated section for machine-level memory-mapped registers (specifically mtime/mtimecmp)

gameboo:
Even after having read through various discussions, I keep getting confused when reading the privileged document's table of content and seeing `mtime` and `mtimecmp` listed under `3.1 Machine-Level CSRs`: I go and look for their location in the CSR address space and can't find them, and then realise that "ah yes, these are memory-mapped! I should have remembered that".

If I understand correctly, `mtime` and `mtimecmp` are *not* CSRs in that they are not accessible via the various CSR instructions, and only the separate `time`(and `timeh`) CSR provides a window to `mtime` via the CSR space.

This PR suggests conveying this straight from the table of content.

Additionally, I wonder about what a 

> memory-mapped machine-mode read-write register

means when it comes to exceptions. I presume the intended behaviour is that memory accesses to `mtime`/`mtimecmp` from a privilege mode other than `M` should raise an exception (I presume that's what the comment about lower privilege levels not having their own `timecmp` register is hinting at)?
Would it be worth to maybe call this a new kind of PMA? Or possibly described under `3.6.2 Supported Access Type`?

### DIFF
--- a/src/machine.tex
+++ b/src/machine.tex
@@ -1604,120 +1604,6 @@ using the {\tt sie} register.  Otherwise, the corresponding
 bits in {\tt sip} and {\tt sie} appear to be hardwired
 to zero.
 
-\subsection{Machine Timer Registers ({\tt mtime} and {\tt mtimecmp})}
-
-Platforms provide a real-time counter, exposed as a memory-mapped
-machine-mode read-write register, {\tt mtime}.  {\tt mtime} must
-increment at constant frequency, and the platform must provide a
-mechanism for determining the timebase of {\tt mtime}.  The {\tt
-  mtime} register will wrap around if the count overflows.
-
-The {\tt mtime} register has a 64-bit precision on all RV32 and RV64
-systems.  Platforms provide a 64-bit memory-mapped machine-mode
-timer compare register ({\tt mtimecmp}).
-A timer interrupt becomes pending whenever {\tt mtime} contains
-a value greater than or equal to {\tt mtimecmp}, treating the values
-as unsigned integers.
-The interrupt remains posted until {\tt mtimecmp} becomes greater than
-{\tt mtime} (typically as a result of writing {\tt mtimecmp}).
-The interrupt will only be taken if interrupts
-are enabled and the MTIE bit is set in the {\tt mie} register.
-
-\begin{figure}[h!]
-{\footnotesize
-\begin{center}
-\begin{tabular}{@{}J}
-\instbitrange{63}{0} \\
-\hline
-\multicolumn{1}{|c|}{\tt mtime} \\
-\hline
-64 \\
-\end{tabular}
-\end{center}
-}
-\vspace{-0.1in}
-\caption{Machine time register (memory-mapped control register).}
-\end{figure}
-
-\begin{figure}[h!]
-{\footnotesize
-\begin{center}
-\begin{tabular}{@{}J}
-\instbitrange{63}{0} \\
-\hline
-\multicolumn{1}{|c|}{\tt mtimecmp} \\
-\hline
-64 \\
-\end{tabular}
-\end{center}
-}
-\vspace{-0.1in}
-\caption{Machine time compare register (memory-mapped control register).}
-\end{figure}
-
-\begin{commentary}
-The timer facility is defined to use wall-clock time rather than a
-cycle counter to support modern processors that run with a highly
-variable clock frequency to save energy through dynamic voltage and
-frequency scaling.
-
-Accurate real-time clocks (RTCs) are relatively expensive to provide
-(requiring a crystal or MEMS oscillator) and have to run even when the
-rest of system is powered down, and so there is usually only one in a
-system located in a different frequency/voltage domain from the
-processors.  Hence, the RTC must be shared by all the harts in a
-system and accesses to the RTC will potentially incur the penalty of a
-voltage-level-shifter and clock-domain crossing.  It is thus more
-natural to expose {\tt mtime} as a memory-mapped register than as a CSR.
-
-Lower privilege levels do not have their own {\tt timecmp} registers.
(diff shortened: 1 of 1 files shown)


aswaterman:
Seems reasonable to me. @gfavor @kasanovic?

sorear:
As far as I know none of the major open source hardware implementations or simulators do privilege level checks on access to memory-mapped registers (other than what software configures itself via PMPs or page tables).  You're not the first person I've seen interpreting the spec that way, though.

gameboo:
>As far as I know none of the major open source hardware implementations or simulators do privilege level checks on access to memory-mapped registers (other than what software configures itself via PMPs or page tables).

Yes... I don't think I really saw this anywhere, but as it sorta lives in the middles of "definitely protected machine CSRs" in the spec, it is very tempting to read this as "yes, only machine mode can touch it". Not mandating that also sounds fine to me.

All in all, I think this is why I suggest using a PMA-based specification there, allowing for both...

aswaterman:
I’m not aware of any implementations that do so, either, @sorear. (And of course since PMP is a whitelist, it doesn’t take extra effort or additional PMP registers to lock out access to mtime[cmp].)

gameboo:
It would be nice to still have the option to have this protected for systems that chose not to have a PMP maybe?

aswaterman:
I don’t _think_ we need any new spec language to permit hardwired PMAs or PMPs that do lock out less-privileged access to these registers.

gameboo:
Yes, I think you are right.

I was suggesting this simply in the spirit of making it easier on the reader, a bit like the suggested change in this PR is not strictly necessary I suppose.

I was thinking of adding a brief general comment to `3.6.2 Supported Access Type`. Maybe simply change the first sentence to also mention the privilege mode from which the access is issued, and possibly point this out from the mtime/mtimecmp section, and leave it to that.

What do you think?

sorear:
If we don't have new spec language, some people are going to assume it's mandatory and others are going to assume it's forbidden.  (I'm in the "forbidden" camp - the PMA chapter enumerates a large number of possible memory attributes and "only accessible from M-mode regardless of PMP settings" is not one of the allowed options.)

aswaterman:
It would be part of a separate PR, of course, but I think this hole would be closed by making it explicit that PMAs can be a function of privilege mode.

kasanovic:
I'm OK with this.

gfavor:
After catching up with this whole thread, I observe the following few things and reach a conclusion:

The original PR (nor existing spec) did not include a explicit definition of what is a **machine-mode** memory-mapped register.  Subsequent comments suggested that a mode-dependent PMA region could effectuate a meaning for that term.  But it was also rightly pointed out that the current PMA chapter makes no mention of PMA regions possibly having mode-specific behavior.  And if anything the seeming architectural idea currently is that PMA attributes are about the physical nature of the addressed location irrespective of higher-level access control considerations such as taken into account by PMPs and an MMU.

But one can readily choose to enhance PMA arch functionality to also allow their access control to be dependent on the mode of the access being checked (especially since the PMA chapter didn't limit PMAs to only having the documented attributes).

But is one now **required** to implement such "machine-mode-only" PMA region or regions, and **required** that machine-mode memory-mapped registers must be covered by such a region.  Or is this an option or a suggestion?

And why is this job not left for PMPs to handle - since they are specifically about enforcing M-mode-only access control on selected regions of address space?  (I believe there are some people that have inferrred from the Priv spec that PMPs and not PMAs are the place in the architecture to effectuate the access restrictions implied by the term "machine-mode-only" memory-mapped registers.)

This view would mean that there is no need for PMA arch spec change.  Only need for more clearly documenting that mtime/mtimecmp are not CSRs, and providing guidance that PMPs should be used to limit access to "machine-mode" memory-mapped registers to only M-mode).

kasanovic:
Yes, I think we can simply describe these as memory-mapped registers, with commentary on access control via either PMPs or PMAs.

aswaterman:
@gfavor I'm not in favor of mandating an access-control PMA for these registers.  As a separate matter, I do think documenting the concept of access-control PMAs might make sense, though is not strictly required because custom PMAs are already a possibility.  For now, I think the lowest-friction option is to do what @kasanovic suggested and add commentary that either PMP or PMA could provide access control for machine-level memory-mapped registers.

gfavor:
To be clear, I was arguing, if anything, against an access-control PMA (and instead letting a PMP do the job).  But the suggested commentary is fine.

allenjbaum:
PMPs are optional. Mtime is not. Does this imply that if PMPs are not
implemented, then a custom PMA must be?
(our PMA does have mode specific ranges for a variety of reasons)

gameboo:
Thanks for the merge.

> To be clear, I was arguing, if anything, against an access-control PMA (and instead letting a PMP do the job). But the suggested commentary is fine.

For what it is worth, I wanted to re-state what I already suggested which that I like the idea of access control PMA being documented and not mandatory as this provides an escape for implementations who are concerned with access control to memory-mapped registers and do not want to have a PMP.

aswaterman:
@allenjbaum There's still no mandate that mtime be protected; it's a practical concern but not a requirement.

Most systems that don't provide PMPs either don't have multiple privilege modes or implicitly fully trust the next-lowest privilege mode (unlikely for M+U systems, but plausible for M+S+U systems).  In either of those cases, there's no need to protect mtime.

gameboo:
> It would be part of a separate PR, of course, but I think this descriptive hole would best be closed by making it explicit that PMAs can be a function of privilege mode.

Just checking, but, were you suggesting I propose something (most likely short of all the many details that would be interesting) as a new PR, and as a basis for a continued discussion?

aswaterman:
I think we are leaning towards no normative addition to the PMA text for now, but instead add commentary to the memory-mapped register section as Krste suggested.

PR #616, opened 2020-12-29
Allow an optional vendor name filed for non-standard extension names

kito-cheng:
I would like to add an optional field for vendor name, that could be used as kind of namespace to prevent naming collision for non-standard extension.

### DIFF
--- a/src/naming.tex
+++ b/src/naming.tex
@@ -125,10 +125,13 @@ they should be ordered alphabetically.
 
 \section{Non-Standard Extension Names}
 
-Non-standard extensions are named using a single ``X'' followed by an
-alphabetical name and an optional version number.
+Non-standard extensions are named using a single ``X'' followed by an optional vendor
+name field seperated by a single ``-'' an alphabetical name and an optional version
+number.
 For example, ``Xhwacha'' names the Hwacha vector-fetch ISA extension;
 ``Xhwacha2'' and ``Xhwacha2p0'' name version 2.0 of same.
+``Xucb-hwacha'' names the cache extension from UCB,
+``Xucb-hwacha2'' and ``Xucb-hwacha2'' name version 2.0 of same.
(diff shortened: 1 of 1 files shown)


aswaterman:
I don't like the idea of introducing vendors as a semantically significant concept.

Even within vendors, there will be opcode-space conflicts.  So, it will always be better to express extension names fully: e.g., `we shouldn't allow Xacme` as a shortcut to represent `Xacme_anvil` + `Xacme_bomb`, because those two extensions might not be disjoint--nor should they be required to be.

Introducing hyphens as a semantically significant character is also a bit of a wart.

cc @kasanovic

kito-cheng:
My intention is make vendor extension more readable, e.g.  `Xsifive-cache` is more readable than `Xsifivecache` IMO , not a combo set for vendor, e.g `Xsifive` = `Xsifivecache` + `Xsifiveclic`...

However the underline is already used as separator of extension, so the only way is introduce another character for that.

aswaterman:
A vendor-name/extension-name separator seems like too specific of a way to spend the hyphen.

A more general option would be to introduce hyphens as intra-name separators that have no semantic meaning: e.g., `Xacme-anvil` would mean the same thing as `Xacmeanvil`.  This is how underscores worked in a previous version of ISA names.  It also has the advantage of being backwards-compatible with existing custom-extension names.

Anyway, we should wait for KA to chime in.

kersten1:
@kito-cheng Same with this PR. Is it still valid? If so, I can move to adoc file.

kersten1:
@kito-cheng Giving this 2 more weeks and then closing...

aswaterman:
The community has established a de facto standard here of `X{company}{extension}`.  We really can't do any better than a de facto standard, since the namespace is custom, so I consider this problem solved.

PR #617, opened 2021-01-11
Clarify type of timer interrupt

bgasiorzewski:
Clarify that mtime >= mtimecmp results in a *machine* timer interrupt being posted.


bgasiorzewski:
Hi @aswaterman, could you take a look at this PR? It's a very small change, should be easy to review.

PR #619, opened 2021-01-12
Clarify rm field on widening conversions

### LINKED ISSUE #603: FP instructions that depend on rounding mode
This didn't get much traction on the tech-unprivileged mailing list (other than someone who agreed that the spec isn't clear) so I'm posting here:

The F extension says: "The behavior of floating-point instructions that depend on rounding mode when executed with a reserved rounding mode is reserved, including both static reserved rounding modes (101–110) and dynamic reserved rounding modes (101–111)."  How does the specification specify which FP instructions "depend on rounding mode"?

For example: The F extension says: "Note FCVT.D.W[U] always produces an exact result and is unaffected by rounding mode" and "FCVT.D.S will never round."  In a casual sense, that means that those two instructions do not depend on rounding mode.  In a more formal sense, the table in section 14.3 does not list frm as a source register for either of those instructions which also hints that these instructions do not depend on rounding mode (or at least are allowed to not depend on rounding mode).

If an implementation executes one of these instructions either with a reserved value in the rm field or with rm=111 and a reserved value in the frm CSR, is an implementation allowed to raise an illegal instruction exception or must implementations ignore the rm field and the frm CSR on those instructions?

pdonahue-ventana:
Fixes #603

### DIFF
--- a/src/f.tex
+++ b/src/f.tex
@@ -162,8 +162,9 @@ depend on rounding mode when executed with a reserved rounding mode is
 {\em reserved}, including both static reserved rounding modes (101--110) and
 dynamic reserved rounding modes (101--111).  Some instructions, including
 widening conversions, have the {\em rm} field but are nevertheless
-unaffected by the rounding mode; software should set their {\em rm}
-field to RNE (000).
+mathematically unaffected by the rounding mode; software should set their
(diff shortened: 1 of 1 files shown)


aswaterman:
LGTM.

PR #626, opened 2021-02-09
Add Sv57 and Sv57x4

daniellustig:
The virtual memory TG would like to enable Sv57.  Encoding space was already reserved for Sv57, and this commit adds it in the "expected" way.

We expect this to go through the RISC-V fast track review process.  Marking as draft while that process plays out.


aswaterman:
👍

daniellustig:
Manually merged into #611.  Closing.

PR #628, opened 2021-02-22
Svnapot Extension

daniellustig:
This PR contains the Svnapot extension (formerly named Zsn) extension being developed by the RISC-V Virtual Memory Task Group.

This was previously part of #611, but we decided to separate it out into its own PR so that the Svnapot ratification timeline doesn't block the other assorted changes in #611.

This branch is still built off of #611, and therefore I'm creating this as a PR into #611 rather than master in order to highlight the Svnapot extension itself.  I can handle any conflicts with master later when it actually comes time to merge this.

### DIFF
--- a/src/supervisor.tex
+++ b/src/supervisor.tex
@@ -1812,13 +1812,10 @@ root page table is stored in the {\tt satp} register's PPN field.
 The PTE format for Sv39 is shown in Figure~\ref{sv39pte}.  Bits 9--0
 have the same meaning as for Sv32.
 
-The N bit indicates that the page represents a
-naturally-aligned power-of-two range of contiguous translations, as defined in
-the Svnapot extension in Chapter~\ref{svnapot}.
-
-Bits 62--54 are reserved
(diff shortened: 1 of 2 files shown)


src/supervisor.tex
```diff
@@ -2108,3 +2117,144 @@ \subsection{Addressing and Memory Protection}
 The algorithm for virtual-to-physical address translation is the same
 as in Section~\ref{sv32algorithm}, except LEVELS equals 5 and PTESIZE
 equals 8.
+
+\chapter{``Svnapot'' Standard Extension for NAPOT Translation Contiguity, Version 0.1}
+\label{svnapot}
+
+In Sv39, Sv48, and Sv57, when Svnapot is enabled and a PTE has N=1, the PTE represents a
```
jscheid-ventana, line 2124: > when Svnapot is enabled

Does this temporarily need to be stated as "via a TBD mechanism"?  Or is this meant to say "is implemented"?

PR #634, opened 2021-03-24
Clarify relative privilege of VS-mode w.r.t. interrupt global enables

### LINKED ISSUE #633: Hypervisor interrupt prioritization unspecified
See recent discussion on [tech-privileged](https://lists.riscv.org/g/tech-privileged): [Hypervisor interrupt enables](https://lists.riscv.org/g/tech-privileged/topic/hypervisor_interrupt_enables/81068094?p=,,,20,0,0,0::recentpostdate%2Fsticky,,,20,2,0,81068094)

Interrupt enables depend on relative privilege levels. The general rule is that a less-privileged interrupt is always disabled, a same-privilege interrupt depends on an xIE bit (e.g. mstatus.mie or sstatus.sie), and a more-privileged interrupt is always enabled.

But the relative privilege of hypervisor modes (V=1) vs non-hypervisor modes (V=0) is not specified, leading to questions:

1. Are HS-mode interrupts enabled when in VS-mode? Does it depend on sstatus.sie?
2. Are VS-mode interrupts enabled when in U-mode?

This would be resolved if the priorities were specified in the obvious order, though perhaps this order applies only to interrupt enabling: M > HS > U > VS > VU

cc @jhauser-us

scottj97:
Fixes #633.

### DIFF
--- a/src/hypervisor.tex
+++ b/src/hypervisor.tex
@@ -75,6 +75,9 @@ possible operating modes of a RISC-V hart with the hypervisor extension.
 \label{h-operating-modes}
 \end{table*}
 
+For purposes of interrupt global enables, VS-mode is considered less
+privileged than HS-mode and U-mode.
+
 \section{Hypervisor and Virtual Supervisor CSRs}
 
 An OS or hypervisor running in HS-mode uses the supervisor CSRs to interact with the exception,


aswaterman:
Is the second part really true?  It's more that VS-mode is lateral to U-mode.

In any case, given the N extension isn't really a thing, I don't think we need to debate this point.  It would suffice to drop the U-mode part.

gfavor:
This addresses and clarifies one of the questions that were at issue:  Are
VS-mode interrupts globally enabled or disabled while in U-mode (i.e. is
VS-mode more or less privileged than U-mode?)

VS-mode being lower priority than U-mode then implies "globally disabled".
(John confirmed that this is his architectural intention.)

aswaterman:
Ah, yes.  My confusion arises from how this interacts with the N-extension.  As written, it implies that U-mode interrupts are globally enabled when in VS-mode, which is probably _not_ the intention.  But if we don't care about the N-extension, this point is moot.

scottj97:
Yes. I haven't considered interaction with the N-extension. (And that goes for several other things in Hypervisor, iiuc.)

pdonahue-ventana:
I also have concerns about N.  You don't want to trap from VS into U and then have the user handler getting access to regfile state from VS.  And there's probably no way to uret to VS mode.  But I don't fully understand N so I could be off-base.

pdonahue-ventana:
Maybe add another sentence: "If H is implemented then N cannot be implemented."  I think that's basically the intention anyway.

aswaterman:
How about we just rewrite this to remove all doubt:

VS-mode interrupts are globally disabled when executing in U-mode, and U-mode interrupts are globally disabled when executing in VS-mode or VU-mode.

scottj97:
My understanding is that when a hypervisor is in use, U-mode is intended for lesser-privileged parts of the hypervisor software. So perhaps it's not a problem that U-mode can observe VS-mode state.

It seems more likely that the N-extension will be useful to enable VU-mode interrupts.

aswaterman:
Some hypervisors are also OSes, in which case U-mode is used to run normal user programs on the host.  U-mode and VS-mode are equiprivileged in that use case; neither should see the other's state.

scottj97:
> VS-mode and VU-mode interrupts are globally disabled when executing in U-mode, and U-mode interrupts are globally disabled when executing in VS-mode or VU-mode.

We still need to clarify HS-mode interrupts somewhere. So add to that: HS-mode interrupts are globally enabled when executing in U-mode, VS-mode, or VU-mode.

Do you think this still belongs at the end of 5.1? Or somewhere more interrupt-focused?

aswaterman:
Right, I guess I'm now seeing the motivation for your formulation... does the following cover all the bases?

For purposes of interrupt global enables, HS-mode is considered more privileged than VS-mode, and VS-mode is considered more privileged than VU-mode.

VS-mode and VU-mode interrupts are globally disabled when executing in U-mode, and U-mode interrupts are globally disabled when executing in VS-mode or VU-mode.

gfavor:
>
> ... and U-mode interrupts are globally disabled when executing in VS-mode
> or VU-mode.
>
Doesn't this raise the question of whether (future) U-mode interrupts
should be globally enabled or disabled while in VS/VU modes?  It doesn't
seem like the correct answer is trivially obvious and without any
possibility of being wrong.

For example, the OS/hypervisor that executes in HS and U modes is outside
of a VM, so maybe shouldn't both HS and U mode interrupts be enabled while
in VS/VU modes?  I don't claim to know what the "right" answer is, but
unless some subset of you all are certain about the answer for now and into
the future, then why not leave this unspecified until an N extension ever
comes to pass  (which seems very unlikely)?

aswaterman:
OK.  Another take:

For purposes of interrupt global enables, HS-mode is considered more privileged than VS-mode, and VS-mode is considered more privileged than VU-mode.

VS-mode interrupts are globally disabled when executing in U-mode.

\begin{commentary}
This description does not consider the possibility of U-mode or VU-mode interrupts and will be revised if the N extension for user-level interrupts is ultimately adopted.
\end{commentary}

gfavor:
Doesn't the non-normative comment about VU-mode interrupts imply that the normative statement should not include the last clause about VS-mode being more privileged than VU-mode (since that is significant only in the context of there being VU-mode interrupts)?

At which point can't one combine the two normative sentences into one:
   For purposes of interrupt global enables, VS-mode interrupts are globally disabled when executing in HS-mode or U-mode.

Or just:  VS-mode interrupts are globally disabled when executing in HS-mode or U-mode.

aswaterman:
@gfavor Mentioning the VS-VU relationship is necessary even without VU-mode interrupts, because it encodes the fact that VS-mode interrupts are globally enabled when executing in VU-mode.

gfavor:
Fair point.  Although isn't that directly implied by VS-mode being high privilege than VU-mode?  In the same way that M-mode interrupts are globally enabled while executing in HS, U, VS, and VU modes?

aswaterman:
@gfavor It seemed to me that part of the reason for this discussion was that the privilege hierarchy w.r.t. the hypervisor extension isn't clearly stated, even though it seems obvious.  (Isn't also obvious that HS-mode is more privileged than VS-mode, yet we feel the need to explicitly state that here?)  So I thought it would be appropriate to err on the side of being explicit.

gfavor:
Yes, it was considered obvious that priority-wise HS > VS.  More generally the priority chains "HS > VS > VU" and "HS > U" were accepted as being obvious.  But since there wasn't a clear priority between those two chains, that still left the question of whether U > VS or not - which implies whether VS-mode interrupts are globally disabled or not while executing in U-mode.

aswaterman:
... which suggests we're back to only adding the statement "VS-mode interrupts are globally disabled when executing in U-mode."

gfavor:
Yep.

aswaterman:
OK, I'll pull that trigger and close this PR.  If anyone feels that's an insufficient solution, please reopen it or create a new one.

gfavor:
Note that this "final" addition is different than the one Scott started this PR with.

scottj97:
I don't agree. Spike wouldn't have had this bug if it was obvious that VS < HS. And evidently Sandro Pinto had the same confusion in Qemu.

aswaterman:
OK, if that's the case, then I'll continue to recommend my previous suggestion, https://github.com/riscv/riscv-isa-manual/pull/634#issuecomment-806219081

gfavor:
Or just "VS-mode interrupts are globally disabled when executing in HS-mode or U-mode." for the normative text (along with the suggested non-normative text).

scottj97:
> Or just "VS-mode interrupts are globally disabled when executing in HS-mode or U-mode." for the normative text (along with the suggested non-normative text).

I don't think that's as complete. It doesn't explain that HS-mode interrupts are globally enabled (without regard to `sstatus.sie`) when in VS-mode, which was the original confusion.

gfavor:
Good point.

aswaterman:
OK, I merged my suggestion to make progress on this.

scottj97:
Thanks, I'm good with this.

PR #638, opened 2021-04-13
marchid request for Ibex

GregAC:


GregAC:
Updated so requested ID doesn't clash with other pending request https://github.com/riscv/riscv-isa-manual/pull/630

PR #641, opened 2021-04-21
SUM should be hardwired to 0 for cores without paging

aswaterman:
Note that this change does not consider the forthcoming sPMP proposal.  We'll have to update the text to say that SUM should be hardwired to 0 if satp.MODE is hardwired to 0 and sPMP is not provided.

I proposed similar wording in the sPMP proposal: https://docs.google.com/document/d/1x7esOSBFfpcbDHaRPpe5NEWmav1_8der_nB25Hd5hqs/edit#

### DIFF
--- a/src/machine.tex
+++ b/src/machine.tex
@@ -719,7 +719,7 @@ in Figure~\ref{sv32pte}) will fault.  When SUM=1, these accesses are
 permitted.  SUM has no effect when page-based virtual memory is not in effect.
 Note that, while SUM is ordinarily ignored when not executing in S-mode, it
 {\em is} in effect when MPRV=1 and MPP=S.  SUM is hardwired to 0 if S-mode is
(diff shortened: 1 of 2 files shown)


gfavor (approved):
Looks good.

pdonahue-ventana:
@aswaterman: This was merged but I didn't see a new [release](https://github.com/riscv/riscv-isa-manual/releases) of the PDF.

aswaterman:
Our CI flow is broken again. Someone [else] will have to diagnose it and
fix it. Until then, it’s back to building your own PDFs manually.

pdonahue-ventana:
> Our CI flow is broken again. Someone [else] will have to diagnose it and fix it. Until then, it’s back to building your own PDFs manually.

OK.  Just making sure that "someone" knows that things are broken.

PR #642, opened 2021-04-23
Minor mstatus and sstatus layout edits.

steven-bellock:
Remove xlen from mstatus and sstatus bit layouts, since those are already known. Swap the RV32 / RV64 mstatus ordering since 32 < 64 and to be consistent with sstatus. The text editor also removed some trailing whitespace.


aswaterman:
Looks fine, but going forward, please avoid making whitespace changes to other parts of the document that you aren't actively editing.  It can cause extra merge conflicts.

PR #647, opened 2021-05-15
Add Zfinx spec

aswaterman:
This PR contributes a new chapter to the unprivileged spec that describes the Zfinx, Zhinx, and Zdinx extensions, which add floating-point instructions that operate on the x-registers.

### DIFF
--- a/src/zfinx.tex
+++ b/src/zfinx.tex
@@ -16,6 +16,7 @@ minimal implementation cost.
 By eliminating the {\tt f} registers, the Zfinx extension substantially
 reduces the cost of simple RISC-V implementations with floating-point
 instruction-set support.
+Zfinx also reduces context-switch effort.
 
 Unlike most RISC-V extensions, the addition of Zfinx is not backwards
 compatible: software that uses floating-point instructions but assumes the
--- a/src/zfinx.tex
+++ b/src/zfinx.tex
@@ -100,6 +100,7 @@ Register moves need only a single FSGNJ.D instruction, however.
 \section{Zqinx}
 
 The Zqinx extension is defined analogously to the Zdinx extension.
+The Zqinx extension requires the Zdinx extension.
 For RV64, quad-precision operands require two-register alignment.
 For RV32, quad-precision operands require four-register alignment.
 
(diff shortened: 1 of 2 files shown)


src/zfinx.tex
```diff
@@ -0,0 +1,134 @@
...
+
+\begin{commentary}
+Load-pair and store-pair instructions are not provided, so transferring
+double-precision operands in RV32Zdinx from or to memory requires
+two loads or stores.
+Register moves need only a single FSGNJ.D instruction, however.
+\end{commentary}
+
+\begin{commentary}
+In future, an RV64Zqinx quad-precision extension could be defined analogously
```
pdonahue-ventana, line 115:
```suggestion
In the future, an RV64Zqinx quad-precision extension could be defined analogously
```

src/zfinx.tex
```diff
@@ -0,0 +1,134 @@
...
+{\tt mstatus} field FS is hardwired to 0 if the Zfinx extension is
+implemented, and FS no longer affects the trapping behavior of
+floating-point instructions or {\tt fcsr} accesses.
+
+The {\tt misa} bits F, D, and Q are hardwired to 0 when the Zfinx
+extension is implemented.
+
+\begin{commentary}
+A future discoverability mechanism might be used to probe the existence
+of the Zfinx, Zhinx, Zdinx, and Zqinx extensions.
```
pdonahue-ventana, line 133: I don't think that Zqinx should be included here.  Yes, this is talking about the future.  But I thought that it was only talking about the future with respect to the discoverability mechanism and not with respect to an extension that does not currently exist.

src/zfinx.tex
aswaterman: I think the former is a correct Britishism, but yeah, don't see it in American English much.

src/zfinx.tex
aswaterman: This is an editing error.  An earlier draft actually included a Zqinx spec.  Thanks for pointing it out.

aswaterman:
This was manually merged a few weeks ago.

PR #653, opened 2021-06-03
Change to use more common terms.

jscheid-ventana:
Locally define "target instruction" term as a shorthand for the
subsequent instruction to which the HINT applies.


aswaterman:
Much better.  Thanks!

PR #654, opened 2021-06-04
State PMAs that are required for these mappings.

jscheid-ventana:
Explicitly state the PMAs required for this set of mappings to allow C
to run on RISC-V.

### DIFF
--- a/src/memory.tex
+++ b/src/memory.tex
@@ -1230,6 +1230,16 @@ For example, the LR must also be made to respect any data dependencies that the 
 Likewise, the effect a FENCE~R,R elsewhere in the same hart must also be made to apply to the SC, which would not otherwise respect that fence.
 The emulator may achieve this effect by simply mapping AMOs onto {\tt lr.aq;~<op>;~sc.aqrl}, matching the mapping used elsewhere for fully ordered atomics.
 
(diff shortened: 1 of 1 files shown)


aswaterman:
Of course, we mean RVWMO at minimum, not at maximum: we don't want to accidentally give the impression that these C mappings are insufficient for an RVTSO system.

I think we can actually drop that line entirely, because the priv spec says "coherent main memory regions always have either the RVWMO or RVTSO memory model", and we've listed the main memory and coherent PMAs already.

Other than that, LGTM... what say you, @daniellustig?

daniellustig:
RVWMO isn't actually a physical memory attribute in the way the others in that bullet list are, so if we do keep some mention of RVWMO/RVTSO, it should probably move somewhere outside that list.  Otherwise LGTM.

aswaterman:
I just removed RVWMO from the list.  I don't think we need any additional mention of RVWMO/RVTSO, but if Josh disagrees, my recommendation is to add some non-normative text just below, noting that main memory + coherence implies either RVWMO or RVTSO, and hence implies RVWMO compliance.

jscheid-ventana:
> I just removed RVWMO from the list. I don't think we need any additional mention of RVWMO/RVTSO, but if Josh disagrees, my recommendation is to add some non-normative text just below, noting that main memory + coherence implies either RVWMO or RVTSO, and hence implies RVWMO compliance.

No problem.  It wasn't essential.

PR #655, opened 2021-06-04
Remove L and T placeholder chapters from unpriv spec

aswaterman:
See discussion here: https://lists.riscv.org/g/tech-unprivileged/topic/proposed_removal_of_l_and_t/83320037?p=,,,20,0,0,0::recentpostdate%2Fsticky,,,20,2,0,83320037


aswaterman:
Merging after >2 days of no comment on mailing list.

PR #656, opened 2021-06-05
Fix Table 8.1

yang-le:
The quotient of division by zero should have all bits set.


yang-le:
The quotient of division by zero should have all bits set, not 2^L -1. I also checked https://github.com/riscv/riscv-arch-test/blob/master/riscv-test-suite/rv32i_m/M/src/divu-01.S. In the case of "division by zero", the test expect -1 as its result. If the test case was correct, then this table here in the manual may be wrong.

aswaterman:
Because the result is sign-extended, 2^L-1 is the same as -1.

yang-le:
> Because the result is sign-extended, 2^L-1 is the same as -1.

Oh, I see, thank you.

PR #657, opened 2021-06-05
Clarifying FENCE operation behavior for external devices.

kasanovic:


src/rv32.tex
```diff
@@ -1213,7 +1213,24 @@ \section{Memory Ordering Instructions}
   predecessor} set preceding the FENCE.
 Chapter~\ref{ch:memorymodel} provides a precise description of the
 RISC-V memory consistency model.
-  
+
+The FENCE instruction also orders memory reads and writes made by the
+hart as observed by memory reads and writes made by an external
+device.  However, FENCE does not order observations of events made by
+an external device using any other signaling mechanism.
+
```
David-Horner, line 1221: I prefer this inclusive language. Contrasting memory vs other channels is excellent.

aswaterman:
LGTM, thanks.

daniellustig:
Is it worth saying something here about how FENCE orders inter-process interrupts where another hart is the observer, e.g., for the sake of TLB shootdowns?  IPI ordering wrt fences always been somewhat vaguely specified, but one could interpret this new text as more explicitly saying that IPIs are not ordered wrt fences, in which case one could argue that TLB shootdowns are missing that ordering requirement.

kasanovic:
I'm not clear on what is the missing ordering specification for IPIs? A write to a remote msip bit will eventually update the value and will trigger an interrupt sometime in future.  A subsequent FENCE will only order following memory operations so that they are only observable by others after the update to mip is observable by others using reads and writes, but this says nothing about when interrupt is taken.  For TLB case,  I'm assuming the interrupt handler would handshake with the shootdown requester in software to signify shootdown was completed?

daniellustig:
If there's a software handshake already, and the IPI just triggers the receiving thread to start its part of the handshake, then yes, I agree things are fine.  And yes I think the handler would handshake with the requester in software to signal completion.  But what if the IPI itself is intended to be the ordering mechanism in some cases?

e.g., if a producer hart does:
store X
fence
store to a msip bit (MMIO)

...and a consumer hart does:
read mip (CSR)
fence
load X

Is that enough to enforce memory ordering on the accesses to X?  I would think the answer is expected to be yes if the communication is between harts.  If so, I wouldn't want the new text "the FENCE instruction can provide no guarantee on when a change in the interrupt signal is visible to the interrupt controller" to possibly imply that it doesn't.  In other words, the "external device" part of that new text is key.  Or am I misunderstanding the intention?

RVWMO doesn't directly specify what happens when I/O or CSRs are used to communicate between harts instead of memory (although IIUC this case is at least in part using memory-mapped registers instead of directly writing the CSR).  That's all I meant when I said it's currently a little vague from the memory model POV.

aswaterman:
@daniellustig I just got off the horn with @kasanovic on this one, and here's our thinking:
- The sequence you described does need to work.
- The normative text in the CSR chapter in the unpriv spec indicates that this will work, because CSRs are ordered as I/O for the purposes of FENCE.
- The non-normative text in the CSR chapter is misleading in suggesting that only the time/[m]cycle counters are I/O-like: reads of mip/sip are, as well.
- Your example involves an explicit read of mip, but what about the act of taking an interrupt--which, of course, is an implicit read of mip?  I think the current normative text in the CSR chapter is sufficient in this case, too, because the FENCE text doesn't distinguish between implicit and explicit CSR reads.

If all of the above is true, then I think we need no normative changes, but we need to fix the non-normative text in the unpriv spec's CSR chapter.

daniellustig:
Thanks for the summary, SGTM

jscheid-ventana:
It's not clear that all readers of the current text will take away that interrupt handlers need to `FENCE i,?` to establish order with respect to subsequent memory accesses.  In fact, I could see some argument that the ordering should be implicit in the architecture for post-interrupt-trap instructions.

aswaterman:
Part of fixing the non-normative text should involve making this clear.

While I agree reasonable architects might argue that the ordering should be implicit, I don't think it's substantiated by existing normative text, and so I think that would require more serious discussion.

David-Horner:
Should an issue be raised for this.
I will happily do so if y'all agree.

 In that there is substantial discussion on this "merged" pull request, and @aswaterman has suggested "more serious discussion" is required isn't the appropriate place for this an issue that can remain open until all the relevant issues are closed off, either by discussion or a further pull? To work in the workflow??

Indeed, in hindsight, wouldn't this issue have been better served in GitHub by raising as an issue and then after discussion push/pull an appropriate edit to the source? From an opportunity for all to see the process/reasoning, etc. and ...
Workflow speaking as well?

aswaterman:
@David-Horner Thanks for being considerate of the spec editors' view that our unilateral control of the issue tracker is an essential part of our workflow.

My "more serious discussion" comment was an oblique way of saying that, if someone wants to press the issue, they should propose an ISA change to the mailing list.

jscheid-ventana:
I ask here, as I often see PRs fly by without time to comment.  But yes, a closed PR is a poor place to continue discussion.

David-Horner:
Andrew Waterman called Thursday and we discussed many issues including 
challenges with Issues in Github.

We determined that both were unaware of some relevant aspects [neither 
of us intentionally blind];  and neither were exaggerating the our 
concerns [neither of us Chicken Littles].

Andrew volunteered me to look into Github:

    Ffunctionality and features that might diminish some of the
    contention/problems related to "workflow" and "issue expression".

    I consider myself duly deputized to:

       Recommend modifications to a de facto proposal by Krste on how to
    use Github, and

       To make any further suggestions that I consider valuable.


But before I get into that specific concern, there are related concerns

a) timeliness and completion of TG/SIG/other minutes

b) github issues vs group/list discussions

c) Philosophies and World Views.


It is the latter that I want to expound to help us understand the origin 
of some of the conflicts and to plot directions for 
resolution/enhancement in ISA development.


On the call, I admitted to being Process and Enablement oriented,  
Andrew to being Task and Results oriented.


Those familiar with Holism vs Reductionism, know/understand that

1) both views are necessary to achieve significant results

2) at any point [level] in analysis there is value in considering

        i) the holistic nature of the concern/situation/object; its
    behaviour as a whole and response in the context of its environment
    [a sloppy definition, I know] and

        ii) the examination of its components; what it is made of, and
    why the what is needed [if needed] for what it does and doesn't do.
    [an even worse definition].

3) [most] persons have a bias towards one or other mode of analysis, but 
no one [who analyzes] is wholly one or the other.


These points are applicable to Process|Enablement  vs Task|Results.

1) both views are necessary to achieve significant results

2) at any point [level] in accomplishment there is value in considering


    i) the Process|Enablement context; what enables the
    activity/accomplishment to occur.
    How can it be enhanced/leveraged to assist in providing the desired
    outcome.
    Put into place [Enable] the resources to accomplish the goal and let
    the process deliver the results.

    ii) identify the Task/activity that will [hopefully] yield the
    result and work that activity to completion
    Verify the results are as hope for and check the task off the list
    [ or check the task off the list and move to the next Task of
    verifying the result].


3) [most] persons have a bias towards one or other mode of operation.


So, back to Github and Issue Resolution.

A. These different Word Views may conflict at any given point [level] of 
endeavor.


    It occurred in this instance in the conflict of when an issue should
    be closed.

    For the Task|Result oriented, Issue Resolution is closing the issue.
          Task done, move on to next.


    For the Process|Enablement, Issue Resolution is completing all the
    concerns expressed by ensuring a process is in place to address them.

          Closing prematurely aborts that process.


B. Github does not provide  robust Worrkflow Lifecycle in its base 
product for Issues.

    For pull requests there is a support structure with: validations, 
    review requests/responses, and "sign-off" provisions.

    Issue Life-cycle support is bninary/open close status [no
    UnderInvestigation,RequestingConsutation,TentativeResolution,
    Awaiting SignOff].

    It does provide links to/from other issues, and more significantly
    has a CloseRelatedIssue Button when a pull is being applied to the
    repository.

    Andrew mentioned the strong temptation to use this button when
    "finalizing" a pull request.

    External email scripting could provide some functionality that would
    enhance the workflow,

       but the use agreement appears to discourage if not explicitly
    forbid such "non-standard" interfaces.

    Github may have a pay-for-feature process [I remember such before
    its recent acquisition], however,

       I believe we can make do with moderate behavioural changes and
    still be effective in serving all of the RISCV community.


C. Oh. Contrary to Andrew's interference from 
https://github.com/riscv/riscv-isa-manual/pull/657#issuecomment-858481023 ,

           I do not believe unilateral control is desirable nor 
essential to Issue management "workflow".


I will have specific proposals soon.

PR #658, opened 2021-06-07
PMP RWX are collectively WARL, with R=0 W=1 being illegal

aswaterman:
This is technically a semantic change: we are strengthening the R=0 W=1 combination from being "reserved" to being invalid in the WARL sense.

### DIFF
--- a/src/machine.tex
+++ b/src/machine.tex
@@ -3254,9 +3254,10 @@ space, so the RV64 PMP address registers impose the same limit.
 Figure~\ref{pmpcfg} shows the layout of a PMP configuration register.  The R,
 W, and X bits, when set, indicate that the PMP entry permits read, write, and
 instruction execution, respectively.  When one of these bits is clear, the
-corresponding access type is denied.  The combination R=0 and W=1 is reserved
-for future use.  The remaining two fields, A and L, are
-described in the following sections.
+corresponding access type is denied.
+The R, W, and X fields form a collective \warl\ field for which the
+combinations with R=0 and W=1 are reserved.
+The remaining two fields, A and L, are described in the following sections.
 
 \begin{figure}[h!]
 {\footnotesize


kasanovic:
I agree that whole field is WARL, but I don't think we want to rule out using R=0/W=1  for something else in future, so I think it should stay reserved. "Reserved" already implies that implementations that use that encoding before it is later standardized are non-conforming.  They might have non-conforming use of R=0/W=1 which should remain after write.  Conforming implementations will treat reserved encodings as illegal for purposes of WARL.

kasanovic:
Actually it is probably better to describe this throughout section as the permissions field rather than RWX bits (i.e., enum not independent bits).

aswaterman:
So you think the right wording is “permission field is WARL, with the R=0 W=1” combinations reserved”? That would seem to capture both points (current implementations must recode to some other legal value; future standard might repurpose that encoding).

kasanovic:
yes

pdonahue-ventana:
The assumption here seems to be that WARL is defined on a per-field basis and that no WARL field can depend on another field.  I don't think that matches the WARL definition.  The legal value depends on the "architectural state of the hart" and not on only that field.

For instance, mtvec says:
> The value in the BASE field must always be aligned on a 4-byte boundary, and the MODE setting may impose additional alignment constraints on the value in the BASE field.

If low bits of BASE rely on MODE, does the mtvec CSR need to be a single WARL field instead of MODE and BASE?  That doesn't seem reasonable to me.

If mtvec can have MODE affect the legality of certain BASE values, can't the R field in pmpcfg affect the legality of certain W values (or vice versa) without being a single WARL field?  Or is the goal simply to define this as an enum (as Krste later mentioned) and it is not motivated by a per-field WARL assumption?

aswaterman:
@pdonahue-ventana I agree the definition of WARL allows for this sort of coupling already, and it would be possible to obtain this behavior without making RWX a single field, but I also think it's easier to describe this way around.

aswaterman:
Pushed something consistent with Krste's suggestion.  I think refactoring the text to introduce the "permissions" field and updating the rest of the text to refer to it should be left to a separate PR (I'm not going to do that today).

pdonahue-ventana:
OK.  Thanks.

allenjbaum:
In the earlier discussions, I had totally missed Andrew's observation (and so did Andrew, to be fair).
Still, having all 3 bits be simultaneously dependent on the other 2 is less than obvious.
Krste's suggestion seems to be easier to understand, and something more amenable to explanation.

allenjbaum:
Please note that the ePMP extension does define RWX=01x cases explicitly,
so it is reserved for PMP, but not (necessarily) for ePMP.
And, as I recall, that in itself is dependent on another new CSR field for
enablement, which is RW1S, so once set, cannot be cleared.
And the meaning of those configurations (and another) is also dependent on
the L-bit.
My crib sheet is attached for those who are interested.

gfavor:
To be clear, ePMP is backward compatible in this regard.  Only when the mseccfg.MML bit is set, are the meanings of the LRWX bits modified - including giving defined meanings to the previously reserved encodings.  So one should say that the reserved RW encodings still exist in ePMP, but only in certain conditions.

gfavor (approved):
I agree with Andrew's and Krste's perspective on this issue and the change suitably captures that perspective.

aswaterman:
Point about ePMP is noted.  This clarification will clear up some confusion in the short term, and will of course remain releavnt for systems without ePMP.

PR #663, opened 2021-06-13
Svpbmt: Page-Based Memory Types

daniellustig:
Opening a draft PR to enable discussion on the Svpbmt extension being discussed by the RISC-V virtual memory TG.

This extension is created as a branch off of #628 for convenience, and to avoid conflating Svpbmt discussion with other virtual memory TG discussions, but Svnapot and Svpbmt will be reviewed for ratification purposes as independent entities.

PDF versions of the specs are available on the virtual memory TG GitHub page:
https://github.com/riscv/virtual-memory/tree/main/specs

### DIFF
--- a/src/supervisor.tex
+++ b/src/supervisor.tex
@@ -1764,9 +1764,10 @@ quickly distinguish user and supervisor address regions.
 \begin{figure*}[h!]
 {\footnotesize
 \begin{center}
-\begin{tabular}{cY@{}Y@{}Y@{}Y@{}Fcccccccc}
+\begin{tabular}{cF@{}Y@{}Y@{}Y@{}Y@{}Fcccccccc}
 \instbit{63} &
-\instbitrange{62}{54} &
+\instbitrange{62}{61} &
+\instbitrange{60}{54} &
 \instbitrange{53}{28} &
 \instbitrange{27}{19} &
 \instbitrange{18}{10} &
@@ -1781,6 +1782,7 @@ quickly distinguish user and supervisor address regions.
 \instbit{0} \\
 \hline
 \multicolumn{1}{|c|}{N} &
+\multicolumn{1}{c|}{PBMT} &
 \multicolumn{1}{c|}{\it Reserved} &
 \multicolumn{1}{c|}{PPN[2]} &
 \multicolumn{1}{c|}{PPN[1]} &
@@ -1795,7 +1797,7 @@ quickly distinguish user and supervisor address regions.
 \multicolumn{1}{c|}{R} &
 \multicolumn{1}{c|}{V} \\
 \hline
-1 & 9 & 26 & 9 & 9 & 2 & 1 & 1 & 1 & 1 & 1 & 1 & 1 & 1\\
+1 & 2 & 7 & 26 & 9 & 9 & 2 & 1 & 1 & 1 & 1 & 1 & 1 & 1 & 1\\
 \end{tabular}
 \end{center}
 }
@@ -1815,8 +1817,14 @@ have the same meaning as for Sv32.
 The N bit indicates that the page represents a
 naturally-aligned power-of-two range of contiguous translations, as defined in
 the Svnapot extension in Chapter~\ref{svnapot}.
+If the Svnapot extension is not implemented, the N bit must be zeroed by software,
+or else a page-fault exception is raised.
 
-Bits 62--54 are reserved
+Bits 62--61 indicate the page-based memory type, as defined in the Svpbmt extension in Chapter~\ref{svpbmt}.
+If the Svpbmt extension is not implemented, the PBMT bits must be zeroed by software,
+or else a page-fault exception is raised.
+
+Bits 60--54 are reserved
 for future standard use and must be zeroed by software for forward compatibility.
 If any of these bits are set, a page-fault exception is raised.
 
@@ -1923,9 +1931,10 @@ is untranslated.
 \begin{figure*}[h!]
 {\footnotesize
 \begin{center}
-\begin{tabular}{cF@{}F@{}F@{}F@{}F@{}Fcccccccc}
+\begin{tabular}{cF@{}F@{}F@{}F@{}F@{}F@{}Fcccccccc}
 \instbit{63} &
-\instbitrange{62}{54} &
+\instbitrange{62}{61} &
+\instbitrange{60}{54} &
 \instbitrange{53}{37} &
 \instbitrange{36}{28} &
 \instbitrange{27}{19} &
@@ -1941,6 +1950,7 @@ is untranslated.
 \instbit{0} \\
 \hline
 \multicolumn{1}{|c|}{N} &
+\multicolumn{1}{c|}{PBMT} &
 \multicolumn{1}{c|}{\it Reserved} &
 \multicolumn{1}{c|}{PPN[3]} &
 \multicolumn{1}{c|}{PPN[2]} &
@@ -1956,7 +1966,7 @@ is untranslated.
 \multicolumn{1}{c|}{R} &
 \multicolumn{1}{c|}{V} \\
 \hline
-1 & 9 & 17 & 9 & 9 & 9 & 2 & 1 & 1 & 1 & 1 & 1 & 1 & 1 & 1\\
+1 & 2 & 7 & 17 & 9 & 9 & 9 & 2 & 1 & 1 & 1 & 1 & 1 & 1 & 1 & 1\\
 \end{tabular}
 \end{center}
 }
@@ -2064,14 +2074,11 @@ is untranslated.
 \begin{figure*}[h!]
 {\footnotesize
 \begin{center}
-\begin{tabular}{c@{}Y@{}F@{}F@{}F@{}F@{}F@{}Wcccccccc}
+\begin{tabular}{c@{}F@{}Y@{}T@{}Wcccccccc}
 \instbit{63} &
-\instbitrange{62}{54} &
-\instbitrange{53}{46} &
-\instbitrange{45}{37} &
-\instbitrange{36}{28} &
-\instbitrange{27}{19} &
-\instbitrange{18}{10} &
+\instbitrange{62}{61} &
+\instbitrange{60}{54} &
+\instbitrange{53}{10} &
 \instbitrange{9}{8} &
 \instbit{7} &
 \instbit{6} &
@@ -2083,12 +2090,9 @@ is untranslated.
 \instbit{0} \\
 \hline
 \multicolumn{1}{|c|}{N} &
+\multicolumn{1}{c|}{PBMT} &
 \multicolumn{1}{c|}{\it Reserved} &
-\multicolumn{1}{c|}{PPN[4]} &
-\multicolumn{1}{c|}{PPN[3]} &
-\multicolumn{1}{c|}{PPN[2]} &
-\multicolumn{1}{c|}{PPN[1]} &
-\multicolumn{1}{c|}{PPN[0]} &
+\multicolumn{1}{c|}{PPN} &
 \multicolumn{1}{c|}{RSW} &
 \multicolumn{1}{c|}{D} &
 \multicolumn{1}{c|}{A} &
@@ -2099,7 +2103,23 @@ is untranslated.
 \multicolumn{1}{c|}{R} &
 \multicolumn{1}{c|}{V} \\
 \hline
-1 & 9 & 8 & 9 & 9 & 9 & 9 & 2 & 1 & 1 & 1 & 1 & 1 & 1 & 1 & 1\\
+1 & 2 & 7 & 44 & 2 & 1 & 1 & 1 & 1 & 1 & 1 & 1 & 1\\
+\end{tabular}
+
+\begin{tabular}{@{}F@{}F@{}F@{}F@{}F}
+\instbitrange{53}{46} &
+\instbitrange{45}{37} &
+\instbitrange{36}{28} &
+\instbitrange{27}{19} &
+\instbitrange{18}{10} \\
+\hline
+\multicolumn{1}{|c|}{PPN[4]} &
+\multicolumn{1}{c|}{PPN[3]} &
+\multicolumn{1}{c|}{PPN[2]} &
+\multicolumn{1}{c|}{PPN[1]} &
+\multicolumn{1}{c|}{PPN[0]} \\
+\hline
+8 & 9 & 9 & 9 & 9 \\
 \end{tabular}
 \end{center}
 }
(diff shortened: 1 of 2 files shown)


src/supervisor.tex
```diff
@@ -2261,3 +2281,108 @@ \chapter{``Svnapot'' Standard Extension for NAPOT Translation Contiguity, Versio
...
+be respected.  This is not expected to be a common situation.
+
+Note that Svpbmt cannot be used to completely prevent speculative reads from
+being performed to a region of memory for which the PMAs indicate idempotence,
+as speculation can still be performed via M-mode or via Bare mappings, which do
+not use the PBMTs.
+\end{commentary}
+
+Memory accesses are ordered according to the effective memory type for the
+address in question.  A region of memory which is considered main memory by the
```
gfavor, line 2347: I can see how this paragraph (after the first sentence) may be a little confusing for some readers to parse correctly.  Maybe this and the following paragraph can all be condensed to:

The ordering of memory accesses by fence instructions is based on the effective memory type and its 'main memory' versus 'I/O' distinction.  This includes for the purposes of FENCE, .aq, and .rl.

A non-normative comment can provide any desired clarification(s).  Although it seems like the commentary in the two paragraphs all follows directly from the current arch specs and the above statement.


Memory accesses are ordered according to the effective memory type for the address in question.  If the address is

src/supervisor.tex
```diff
@@ -2261,3 +2281,108 @@ \chapter{``Svnapot'' Standard Extension for NAPOT Translation Contiguity, Versio
...
+PMAs but I/O by a PTE will obey channel 0 strong ordering, where for memory
+ordering purposes the ``I/O region'' is considered to be main memory.  In other
+words, each access to such a virtual address will appear in the global memory
+order after all prior operations from the same hart to main memory or to the
+same ``I/O region'', and it will appear earlier in the global memory order than
+all subsequent accesses from the same hart to main memory or to the same ``I/O
+region''.  Such operations are not guaranteed to remain ordered with respect to
```
gfavor, line 2354: For example, this is already true in the current PMA and memory ordering architecture.

src/supervisor.tex
gfavor: P.S. These paragraphs (and some later text) tend to mix together two distinct forms of terminology:  Architectural "main memory" and "I/O" (as PMA attributes), and "main memory" and "I/O" as colloquial terms.  The former refer to how an access is handled or performed; the latter refer to the physical nature of what is at a physical address in a system.  For example, one can "map" an address with a "I/O" memory type to what physically is DRAM "main memory", or an address can have a "main memory" memory type but maps to a RAM buffer within an I/O device.

src/supervisor.tex
```diff
@@ -2261,3 +2281,108 @@ \chapter{``Svnapot'' Standard Extension for NAPOT Translation Contiguity, Versio
...
+
+Likewise, a region of memory which is considered I/O by the PMAs but main
+memory by a PTE will obey RVWMO (or RVTSO if enabled) rather than I/O
+strong ordering rules, and accesses to such pages are considered main
+memory rather than I/O for the purposes of FENCE, {\em.aq}, and {\em.rl}.
+In such cases it is the responsibility of software to insert FENCEs
+or other ordering mechanisms if necessary.
+
+\begin{commentary}
+A device driver written to rely on I/O strong ordering rules will not
```
gfavor, line 2365: This paragraph seems semi-misleading.  While appropriate wrt hyprvisors, an OS is not going to go off on its own and override PMAs, i.e. an OS will apply a pbmt override in response to a request from a device driver.

But I also now realize that this issue can't happen.  If a device driver is relying on I/O strong ordering, then presumably either the PMAs specify "I/O" or the device requested the OS to apply an "I/O" pbmt.  (Otherwise the device driver won't work properly even in a non-virtualized environment.)  In which case, even if the hypervisor maps the address to colloquial "main memory", the address will still be treated as part of an architectural "I/O" region.

src/supervisor.tex
```diff
@@ -2261,3 +2281,108 @@ \chapter{``Svnapot'' Standard Extension for NAPOT Translation Contiguity, Versio
...
+operate correctly if the address range is mapped as main memory by the
+page-based memory types.  Operating systems and hypervisors must take
+care to apply such a combination only when strong ordering is not
+actually needed.
+
+In spite of the caveat above, it will often still be useful to map
+device memory regions in I/O as main memory so that write combining
```
gfavor, line 2372: To some readers this can be confusing or ambiguous since it isn't clear whether "I/O" and "main memory" and "device memory regions" are referring to architectural I/O and main memory regions, or to colloquial physical main memory or memory-mapped I/O.

This is also misleading since it doesn't matter that an address is mapped to colloquial physical DRAM main memory.  the effective memory type will determine how the access is handled/performed.

src/supervisor.tex
```diff
@@ -2261,3 +2281,108 @@ \chapter{``Svnapot'' Standard Extension for NAPOT Translation Contiguity, Versio
...
+likely improve performance when applied with adequate care.
+\end{commentary}
+
+The coherence PMA is not affected by Svpbmt.  Non-coherent I/O address ranges
+re-mapped by PBMTs into main memory must still rely on platform-specific
+mechanisms to enforce coherence with respect to external devices.  However,
+RVWMO and RVTSO still require eventual visibility of writes from one harts to
+other harts in the system.
+
+When two-level paging is enabled within the H extension, the page-based memory
```
gfavor, line 2383: This text doesn't spell out how one combines the pbmt from VS-stage and G-stage PTEs when both are specifying overrides, i.e. which one wins out.

If one had names for the pbmt encodings, one could say something like:

_The combined PTE memory type is the more conservative of the VS-stage and G-stage memory types, i.e. IO wins out over NC which wins out over PMA.  The result is always one of IO, NC, or PMA._

gfavor: Conceptually I would view the PMA and G-stage PTE attributes being combined first.  This intermediate set of attributes then represents what is seen by a VM as PMA attributes - which the guest OS can then choose to condiitonally override.

This also provides an easy solution to specifying how VS-stage and G-stage pbmt's are combined.  One applies the earlier description - for how a PTE modifies PMA attributes - first with the G-stage pbmt, and then take that and apply the VS-stage pbmt.

src/supervisor.tex
```diff
@@ -2261,3 +2281,108 @@ \chapter{``Svnapot'' Standard Extension for NAPOT Translation Contiguity, Versio
...
+
+In Sv39, Sv48, and Sv57, bits 62--61 of the page table entry indicate the use
+of page-based memory types that override the PMA(s) for the associated memory
+pages.  The encoding for the PBMT bits is captured in Table~\ref{pbmt}.
+
+\begin{table*}[h!]
+\begin{center}
+\begin{tabular}{|r|l|}
+\hline
+Value  & Page-Based Memory Attributes \\
```
gfavor, line 2299: To avoid any potential misunderstanding by a reader, it might be better to say "Page-Based Memory Attribute Combination".  (Which is what a "page-based memory type" is.)

src/supervisor.tex
gfavor: I somewhat misspoke.  A hypervisor could turn PMA "I/O" into effectively "noncacheable main memory".  But the motivating concern is when a hypervisor maps a guest memory-mapped device into physical main memory for purposes of emulating that device.  This would simply involve G-stage mapping the I/O device's VA pages to physical DRAM main memory.  The G-stage PTE would not override the attributes.  Or if it did, and specifically to "noncacheable main memory",  that would be problematic and unnecessary (i.e. to quote Allen: "if it hurts, then don't do it").

src/supervisor.tex
```diff
@@ -1815,8 +1817,14 @@ \subsection{Addressing and Memory Protection}
 The N bit indicates that the page represents a
 naturally-aligned power-of-two range of contiguous translations, as defined in
 the Svnapot extension in Chapter~\ref{svnapot}.
+If the Svnapot extension is not implemented, the N bit must be zeroed by software,
+or else a page-fault exception is raised.
 
-Bits 62--54 are reserved
+Bits 62--61 indicate the page-based memory type, as defined in the Svpbmt extension in Chapter~\ref{svpbmt}.
+If the Svpbmt extension is not implemented, the PBMT bits must be zeroed by software,
+or else a page-fault exception is raised.
```
mickflemm, line 1825: Raising a fault breaks compatibility, software will now have to choose a different PTE format based on hw revision, instead of just implementing the latest Priv. Spec with guaranteed backwards compatibility. I raised the same issue here -> https://github.com/riscv/riscv-isa-manual/pull/611#discussion_r650576954

daniellustig: Let's resolve this on #611 rather than splitting the discussion in two places

src/supervisor.tex
daniellustig: I don't understand...isn't the note about first/second exactly what the text below says?

src/supervisor.tex
daniellustig: I'm not sure this is a "combination" either...it's really the "effective" memory attribute.  To your other comments, maybe we should use that term "effective memory attribute" throughout, so we can contrast it to "physical memory attribute" in a clearer way.

src/supervisor.tex
daniellustig: There should be no imprecise colloquial usage, we should use proper definitions for these things.  I suggested "effective memory attributes" vs "physical memory attributes" above...if we go with that, this would become something like "If a memory access has an effective memory attribute of main memory, it obeys RVWMO (or RVTSO, if enabled).  If it has an effective memory attribute of I/O, it obeys I/O channel 0 ordering rules."

src/supervisor.tex
daniellustig: Channel 0 specifies point-to-point strong ordering with respect to a particular "I/O region".  In this case that PBMTs turn I/O into main memory, we have to define what that "I/O region" means, because there is no underlying physical I/O region.  I decided to declare that "I/O region" to include both main memory and other accesses in the same category, as a seeming natural choice.  It's up for discussion, but that's what the text was getting at.

src/supervisor.tex
daniellustig: Isn't "...then don't do it" exactly what this non-normative text is warning about?  Ultimately either we forbid the combination, or we define what its behavior is.  For now it's not forbidden, so this comment was meant to make the warning explicit rather than leaving it quiet.

src/supervisor.tex
daniellustig: How about "...it will still be useful to map physical I/O regions containing physical device memory into effective main memory regions..."

scottj97:
For non-leaf PTEs, what effect (if any) does the PBMT field have? Does it affect (only?) the next level of tablewalk?

gfavor:
All the Svpbmt spec proposals including the final one (before Dan turned that into a Priv spec PR), have said:

**What this proposal is not trying to address:**
- Support for per-page memory types for the page tables themselves, i.e. having each non-leaf PTE specify memory types for the next level table that it points to.

If the current PR is felt to not be clear that Svpbmt only applies to leaf PTEs (and for all page sizes), then that should certainly be made clear.

src/supervisor.tex
```diff
@@ -2258,3 +2273,97 @@ \chapter{``Svnapot'' Standard Extension for NAPOT Translation Contiguity, Versio
...
+
+\begin{table*}[h!]
+\begin{center}
+\begin{tabular}{|r|l|}
+\hline
+Value  & Page-Based Memory Attributes \\
+\hline
+0      & None \\
+1      & Non-cacheable, idempotent, weakly-ordered (RVWMO or RVTSO), main memory \\
+2      & Non-cacheable, non-idempotent, strongly-ordered (channel 0), I/O \\
```
scottj97, line 2295: Where is "channel 0" explained? The Unprivileged ISA also has a couple of mentions of channels, but no real explanation of what they mean.

src/supervisor.tex
```diff
@@ -2258,3 +2273,97 @@ \chapter{``Svnapot'' Standard Extension for NAPOT Translation Contiguity, Versio
...
+page-based memory types.  As such, this configuration is discouraged.
+
+In spite of this caveat, it will often still be useful to map physical I/O
+regions using PBMT=1 so that write combining and speculative accesses can be
+performed.  Such optimizations will likely improve performance when applied
+with adequate care.
+\end{commentary}
+
+When two-level paging is enabled within the H extension, the page-based memory
+types are applied in two stages.  First, the G-stage PTE PBMT bits (if enabled)
```
scottj97, line 2365: What does "if enabled" mean here? Is there some global enable for Svpbmt? Or does this mean "if the PTE's PBMT bits are nonzero"?

src/supervisor.tex
scottj97: Nevermind...found this in the Privileged ISA section 3.6.4 Memory-Ordering PMAs. (I was expecting it in the RVWMO material in the Unprivileged ISA.)

scottj97:
> If the current PR is felt to not be clear that Svpbmt only applies to leaf PTEs (and for all page sizes), then that should certainly be made clear.

I can't see in this PR where it says anything about that.

daniellustig:
@scottj97 I will push a clarification about non-leaf PTEs, thank you for pointing that out.

src/supervisor.tex
daniellustig: I meant "if hgatp.MODE != 0" and "vsatp.MODE != 0", because otherwise there are no PTE bits to consider at all, but I will reword to clarify.

src/supervisor.tex
```diff
@@ -2258,3 +2273,97 @@ \chapter{``Svnapot'' Standard Extension for NAPOT Translation Contiguity, Versio
...
+operate correctly if the address range is mapped as main memory by the
+page-based memory types.  As such, this configuration is discouraged.
+
+In spite of this caveat, it will often still be useful to map physical I/O
+regions using PBMT=1 so that write combining and speculative accesses can be
+performed.  Such optimizations will likely improve performance when applied
+with adequate care.
+\end{commentary}
+
+When two-level paging is enabled within the H extension, the page-based memory
```
scottj97, line 2364: The H spec calls it "two-stage", not "two-level".

scottj97:
There are four possible combinations of `PBMT={1,2}` and `PMA={main memory, I/O}`. Two of these four combinations are explained in the text. Why only those two? Are the other two combinations too obvious to bother explaining?

daniellustig:
> There are four possible combinations of `PBMT={1,2}` and `PMA={main memory, I/O}`. Two of these four combinations are explained in the text. Why only those two? Are the other two combinations too obvious to bother explaining?

With PMA=main memory and PBMT=1, accesses to the page continue to obey RVWMO (or RVTSO), just as they otherwise would.  With PMA=I/O and PBMT=2, accesses to the page continue to obey I/O ordering rules.  Those are the straightforward combinations, because we're not switching the memory type attribute.

The other two cases are trickier because in those cases we are switching the memory type attribute.  That means we have to define what it means to apply RVWMO to PMA=I/O regions, or to apply I/O ordering rules to PMA=main memory regions.  That's what the text you referred to describes.

src/supervisor.tex
```diff
@@ -2258,3 +2273,103 @@ \chapter{``Svnapot'' Standard Extension for NAPOT Translation Contiguity, Versio
...
+accesses to pages with PBMT=2.  Furthermore, accesses to such pages are
+considered I/O rather than main memory for the purposes of FENCE, {\em.aq}, and
+{\em.rl}.
+
+With Svpbmt enabled, it is possible for multiple virtual aliases of the same
+physical page to exist simultaneously with different memory attributes.  It is
+also possible for a U-mode or S-mode mapping through a PTE with Svpbmt enabled
+to observe different memory attributes for a given region of physical memory
+than a concurrent access to the same page performed by M-mode or when {\tt
+satp}.MODE=Bare.  If accesses are performed simultaneously to the same region
```
jscheid-ventana, line 2330: Is this list incomplete only mentioning `satp`?

src/supervisor.tex
```diff
@@ -2258,3 +2273,103 @@ \chapter{``Svnapot'' Standard Extension for NAPOT Translation Contiguity, Versio
...
+of memory using different attributes, there may be a loss of coherence and/or
+of normal RVWMO, RVTSO, or I/O ordering semantics.  In such cases,
+platform-specific mechanisms must be used to restore corrent coherence and
```
jscheid-ventana, line 2333: `s/corrent/correct/`

src/supervisor.tex
```diff
@@ -2258,3 +2273,103 @@ \chapter{``Svnapot'' Standard Extension for NAPOT Translation Contiguity, Versio
...
+considered I/O rather than main memory for the purposes of FENCE, {\em.aq}, and
+{\em.rl}.
+
+With Svpbmt enabled, it is possible for multiple virtual aliases of the same
+physical page to exist simultaneously with different memory attributes.  It is
+also possible for a U-mode or S-mode mapping through a PTE with Svpbmt enabled
+to observe different memory attributes for a given region of physical memory
+than a concurrent access to the same page performed by M-mode or when {\tt
+satp}.MODE=Bare.  If accesses are performed simultaneously to the same region
+of memory using different attributes, there may be a loss of coherence and/or
```
jscheid-ventana, line 2331: Since "accesses" here can be speculative, this should be changed to describe the architectural scenario, which is something about having conflicting valid mappings exist across all harts logically concurrently.  If SW robustly changes attributes with no overlap (e.g., some mutually globally invalid period), then is no platform-specific mechanism required?

I don't think that this is attempting to say that only explicit architectural accesses can cause this.

daniellustig:
Merged via #763, closing

PR #665, opened 2021-06-16
Clarify the example of an I/O region.

jscheid-ventana:
The previous example of a "device scratchpad RAM" was awkward, possibly
suggesting that such RAMs could not be classified as Memory.

The platform HSC is accommodating Memory across PCIe and so that instigated this cleanup.


kersten1:
@jscheid-ventana Is this change still needed? If so, I can put into the asciidoc version for this chapter.

kersten1:
rewrote to be adoc here - https://github.com/riscv/riscv-isa-manual/pull/1301

PR #668, opened 2021-06-30
First draft of Svinval

daniellustig:
Opening a draft PR to enable discussion on the Svinval extension being discussed by the RISC-V virtual memory TG.

This extension is created as a branch off of #663 for convenience, and to avoid conflating Svinval discussion with other virtual memory TG discussions, but Svnapot, Svpbmt, and Svinval will be reviewed for ratification purposes as independent entities.

PDF versions of the specs are available on the virtual memory TG GitHub page:
https://github.com/riscv/virtual-memory/tree/main/specs

### DIFF
--- a/src/supervisor.tex
+++ b/src/supervisor.tex
@@ -2376,9 +2376,6 @@ used as the final set of attributes.
 \chapter{``Svinval'' Standard Extension for Fine-Grained Address-Translation Cache Invalidation, Version 0.1}
 \label{svinval}
 
-{\bf Warning! This draft specification is likely to change before being
-accepted as standard by the RISC-V Foundation.}
-
 The Svinval extension splits SFENCE.VMA, HFENCE.VVMA, and HFENCE.GVMA
 instructions into finer-grained invalidation and ordering operations that can
 be more efficiently batched or pipelined on certain classes of high-performance
@@ -2411,7 +2408,7 @@ The SINVAL.VMA instruction invalidates any address-translation cache entries
 that an SFENCE.VMA instruction with the same values of {\em rs1} and {\em rs2}
 would invalidate.  However, unlike SFENCE.VMA, SINVAL.VMA instructions are only
 ordered with respect to SFENCE.VMA, SFENCE.W.INVAL, and SFENCE.INVAL.IR
-instructions, defined below.
+instructions as defined below.
 
 \vspace{-0.2in}
 \begin{center}
@@ -2477,10 +2474,6 @@ a hypothetical SFENCE.VMA instruction in which:
   \item reads and writes following the SFENCE.INVAL.IR are considered to be
     those subsequent to the SFENCE.VMA.
 \end{itemize}
-
-The effect is the same if the SFENCE.W.INVAL instruction and/or the
-SFENCE.INVAL.IR instruction in the sequence is replaced by an SFENCE.VMA
-covering the same virtual address and (if provided) ASID as the SINVAL.VMA.
 
 \vspace{-0.2in}
 \begin{center}
@@ -2551,13 +2544,11 @@ the above in U-mode raises an illegal instruction exception.
   effects but no visible side effects.  Trapping of the SINVAL.VMA instruction
   is sufficient to enable emulation of the intended overall TLB maintenance
   functionality.
-\end{commentary}
 
-\begin{commentary}
   In typical usage, software will invalidate a range of virtual addresses in
-  the address-translation caches by issuing an SFENCE.W.INVAL instruction,
-  issuing a series of SINVAL.VMA, HINVAL.VVMA, or HINVAL.GVMA instructions to
-  the addresses (and optionally ASIDs or VMIDs) in question, and then issuing
+  the address-translation caches by executing an SFENCE.W.INVAL instruction,
+  executing a series of SINVAL.VMA, HINVAL.VVMA, or HINVAL.GVMA instructions to
+  the addresses (and optionally ASIDs or VMIDs) in question, and then executing
   an SFENCE.INVAL.IR instruction.
 
   High-performance implementations will be able to pipeline the
--- a/src/supervisor.tex
+++ b/src/supervisor.tex
@@ -2386,3 +2386,199 @@ are applied to the attributes in the PMA to produce an intermediate set of
 attributes.  Second, the VS-stage PTE PBMT bits (if enabled) are applied to
 these intermediate attributes to produce the final set of attributes used by
 accesses to the page in question.
+
+\chapter{``Svinval'' Standard Extension for Fine-Grained Address-Translation Cache Invalidation, Version 0.1}
+\label{svinval}
+
+{\bf Warning! This draft specification is likely to change before being
+accepted as standard by the RISC-V Foundation.}
+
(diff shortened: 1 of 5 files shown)


src/supervisor.tex
```diff
@@ -2386,3 +2386,199 @@ \chapter{``Svpbmt'' Standard Extension for Page-Based Memory Attributes, Version
...
+\multicolumn{1}{c|}{rd} &
+\multicolumn{1}{c|}{opcode} \\
+\hline
+7 & 5 & 5 & 3 & 5 & 7 \\
+SFENCE.INVAL.IR & 0 & 0 & PRIV & 0 & SYSTEM \\
+\end{tabular}
+\end{center}
+
+The SFENCE.W.INVAL instruction guarantees that any previous stores already
+visible to the curent RISC-V hart are ordered befure subsequent SINVAL.VMA
```
pdonahue-ventana, line 2477:
```suggestion
visible to the current RISC-V hart are ordered before subsequent SINVAL.VMA
```

src/supervisor.tex
```diff
@@ -2386,3 +2386,199 @@ \chapter{``Svpbmt'' Standard Extension for Page-Based Memory Attributes, Version
...
+
+When issued in order (but not necessarily consecutively) by a single hart, the
+sequence SFENCE.W.INVAL, SINVAL.VMA, and SFENCE.INVAL.IR has the same effect as
+a hypothetical SFENCE.VMA instruction in which:
+\begin{itemize}
+  \item the values of {\em rs1} and {\em rs2} for the SFENCE.VMA are the same
+    as those used in the SINVAL.VMA,
+  \item reads and writes prior to the SFENCE.W.INVAL are considered to be those
+    prior to the SFENCE.VMA, and
+  \item reads and writes following to the SFENCE.INVAL.IR are considered to be
```
pdonahue-ventana, line 2491:
```suggestion
  \item reads and writes following the SFENCE.INVAL.IR are considered to be
```

src/supervisor.tex
daniellustig: Fixed, thanks!

aswaterman:
This isn't entirely obvious, but there is a relevant pattern in the encoding of these SYSTEM instructions with funct3={0,4}: if instruction bit 25 is clear, the instruction does not read register rs2, making rs2 available as an additional opcode field.  Instructions like ECALL, EBREAK, SRET, and HLV.* take advantage of this property already.

With that in mind, my suggestion is:
- Keep SINVAL.VMA's encoding as currently proposed, since it has bit 25=1 and does read rs2.
- Keep SFENCE.W.INVAL's encoding as currently proposed, since it has bit 25=0 and does not read rs2.
- Change SFENCE.INVAL.IR's encoding to have bits 31..25 the same as SFENCE.W.INVAL.  Then, use rs2=1 to distinguish SFENCE.INVAL.IR from SFENCE.W.INVAL.

This makes the encoding more compact, but since it follows an existing pattern, it doesn't complicate decoding in a new way.

daniellustig:
Thanks @aswaterman.  @gfavor came up with the original encodings...Greg what do you think?

gfavor:
Andrew's suggestions make sense.  Let's go with those modified SFENCE encodings.

quxfred:
It occurred to me that both SFENCEs are more close to FENCE, i.e. no rs2 and don't change persistent states. Would it be a better idea to encode both SFENCEs under opcode MISC-MEM? It could even be on the same funct3 as FENCE. As I last checked there are still 14 fm code points unused.
Yes, I'm aware of the privilege stuff, but does the fences really need supervisor privilege? I'm struggling to find a threat model.
Anyway, it's just a thought.

aswaterman:
Interesting idea, @quxfred.

I was initially worried that moving the new fences into unprivileged land would create a classical virtualization hole (because the hypervisor would have no chance to intercept them).  But I think that concern is actually moot, since the hypervisor will intercept the SINVAL.VMA instructions.  If it emulates SINVAL.VMA in the same manner it emulates SFENCE.VMA, then there is no reason to intercept the new fences.

In fact, this scheme would perform better under classical virtualization, because there would be fewer emulation traps.

We should ask the TG to sanity-check this idea.

src/supervisor.tex
```diff
@@ -2373,3 +2373,199 @@ \chapter{``Svpbmt'' Standard Extension for Page-Based Memory Attributes, Version
...
+\multicolumn{1}{c|}{funct3} &
+\multicolumn{1}{c|}{rd} &
+\multicolumn{1}{c|}{opcode} \\
+\hline
+7 & 5 & 5 & 3 & 5 & 7 \\
+SFENCE.INVAL.IR & 1 & 0 & PRIV & 0 & SYSTEM \\
+\end{tabular}
+\end{center}
+
+The SFENCE.W.INVAL instruction guarantees that any previous stores already
```
jscheid-ventana, line 2463: Should "stores" here be "writes" for consistency?

src/supervisor.tex
```diff
@@ -2373,3 +2373,199 @@ \chapter{``Svpbmt'' Standard Extension for Page-Based Memory Attributes, Version
...
+visible to the current RISC-V hart are ordered before subsequent SINVAL.VMA
+instructions issued by the same hart.  The SFENCE.INVAL.IR instruction
+guarantees that any previous SINVAL.VMA instructions issued by the current hart
+are ordered before subsequent implicit references by that hart to the
```
jscheid-ventana, line 2467: Can the "implicit references" terms in the priv. spec be changed to "implicit accesses" for consistency with usage of the latter in the priv. and base specifications?

src/supervisor.tex
```diff
@@ -2373,3 +2373,199 @@ \chapter{``Svpbmt'' Standard Extension for Page-Based Memory Attributes, Version
...
+  effects but no visible side effects.  Trapping of the SINVAL.VMA instruction
+  is sufficient to enable emulation of the intended overall TLB maintenance
+  functionality.
+\end{commentary}
+
+\begin{commentary}
+  In typical usage, software will invalidate a range of virtual addresses in
+  the address-translation caches by issuing an SFENCE.W.INVAL instruction,
+  issuing a series of SINVAL.VMA, HINVAL.VVMA, or HINVAL.GVMA instructions to
+  the addresses (and optionally ASIDs) in question, and then issuing an
```
jscheid-ventana, line 2561: "and optionally ASIDs or VMIDs".

src/supervisor.tex
```diff
@@ -2373,3 +2373,199 @@ \chapter{``Svpbmt'' Standard Extension for Page-Based Memory Attributes, Version
...
+  SFENCE.INVAL.IR instruction.
+
+  High-performance implementations will be able to pipeline the
+  address-translation cache invalidation operations, and will defer any
+  pipeline stalls or other memory ordering enforcement until an SFENCE.W.INVAL,
+  SFENCE.INVAL.IR, or SFENCE.VMA instruction is executed.
```
jscheid-ventana, line 2567: "or HFENCE.*VMA" to not raise questions about why it's left out.

jscheid-ventana (commented):
Terminology points.

daniellustig:
Merged via #763, closing

PR #669, opened 2021-07-03
CSR should have been used instead of field

jeras:
It would be nice to use a fixed width font for fields too.

Are the reference names in the document or inside https://github.com/riscv/riscv-config/blob/master/riscv_config/schemas/schema_isa.yaml ?

There are probably a few fields starting with a capital letter in the documentation while they are lowercase in yaml.


aswaterman:
I think it's correct as-is.  (0x60a can't fit in the Offset field, after all...)

aswaterman:
(It should say mvendorid CSR instead of mvendorid field, to be clear.)

jeras:
I somehow managed to create a second erroneous fix, the third one (current) should be correct.

aswaterman:
Cool, thanks for following up.

PR #674, opened 2021-07-10
Remove or downgrade more references to N extension

jhauser-us:


aswaterman:
Thanks for the additional fixes!

PR #675, opened 2021-07-13
Allocate distinct addresses for standard read-only CSRs with side effects

jhauser-us:

### DIFF
--- a/src/priv-csrs.tex
+++ b/src/priv-csrs.tex
@@ -28,7 +28,8 @@ The standard RISC-V ISA sets aside a 12-bit encoding space (csr[11:0])
 for up to 4,096 CSRs.  By convention, the upper 4 bits of the CSR
 address (csr[11:8]) are used to encode the read and write
 accessibility of the CSRs according to privilege level as shown in
-Table~\ref{csrrwpriv}.  The top two bits (csr[11:10]) indicate whether
+Tables \ref{tab:CSRRWPriv-notM} and~\ref{tab:CSRRWPriv-M}.
+The top two bits (csr[11:10]) indicate whether
 the register is read/write ({\tt 00}, {\tt 01}, or {\tt 10}) or
 read-only ({\tt 11}).  The next two bits (csr[9:8]) encode the lowest
 privilege level that can access the CSR.
@@ -45,7 +46,7 @@ accesses to be intercepted.  This change should be transparent to the
 less-privileged software.
 \end{commentary}
 
-\begin{table*}[h!]
+\begin{table*}[p]
 \begin{center}
 \begin{tabular}{|c|c|c|c|l|}
 \hline
@@ -57,7 +58,9 @@ less-privileged software.
 \tt   00   &\tt   00  &\tt   XXXX   & \tt 0x000-0x0FF & Standard read/write \\
 \tt   01   &\tt   00  &\tt   XXXX   & \tt 0x400-0x4FF & Standard read/write \\
 \tt   10   &\tt   00  &\tt   XXXX   & \tt 0x800-0x8FF & Custom read/write \\
-\tt   11   &\tt   00  &\tt   0XXX   & \tt 0xC00-0xC7F & Standard read-only \\
+\tt   11   &\tt   00  &\tt   00XX   & \tt 0xC00-0xC3F & Standard read-only \\
+\tt   11   &\tt   00  &\tt   010X   & \tt 0xC40-0xC5F & Standard read-only \\
+\tt   11   &\tt   00  &\tt   011X   & \tt 0xC60-0xC7F & Standard read-only, with side effects \\
 \tt   11   &\tt   00  &\tt   10XX   & \tt 0xC80-0xCBF & Standard read-only \\
 \tt   11   &\tt   00  &\tt   11XX   & \tt 0xCC0-0xCFF & Custom read-only \\
 \hline
@@ -70,7 +73,9 @@ less-privileged software.
 \tt   10   &\tt   01  &\tt   0XXX   & \tt 0x900-0x97F & Standard read/write \\
 \tt   10   &\tt   01  &\tt   10XX   & \tt 0x980-0x9BF & Standard read/write \\
 \tt   10   &\tt   01  &\tt   11XX   & \tt 0x9C0-0x9FF & Custom read/write \\
-\tt   11   &\tt   01  &\tt   0XXX   & \tt 0xD00-0xD7F & Standard read-only \\
+\tt   11   &\tt   01  &\tt   00XX   & \tt 0xD00-0xD3F & Standard read-only \\
+\tt   11   &\tt   01  &\tt   010X   & \tt 0xD40-0xD5F & Standard read-only \\
+\tt   11   &\tt   01  &\tt   011X   & \tt 0xD60-0xD7F & Standard read-only, with side effects \\
 \tt   11   &\tt   01  &\tt   10XX   & \tt 0xD80-0xDBF & Standard read-only \\
 \tt   11   &\tt   01  &\tt   11XX   & \tt 0xDC0-0xDFF & Custom read-only \\
 \hline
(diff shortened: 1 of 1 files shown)


allenjbaum:
Do you actually have a use case for this? I (personally) think side effects on reads should be relegated to legacy IO devices that are impossible to debug.

aswaterman:
I'm OK with the scheme, but I'll let this sit here for a while (including for John to answer Allen's valid question).

David-Horner:
Thank you Allen for expressing more eloquently what I intended.

I considered the original situation as a principled statement that no
standard readonly csr should have side effects, rather than: we don't
happen to have any yet and so neglected to reserve space for them.

jhauser-us:
The Advanced Interrupt Architecture has two read-only CSRs with side effects, `mclaimei` and `sclaimei` (Machine/Supervisor Claim External Interrupt), and it's looking likely that the cryptography Zkr extension will add another, `sentropy`, that returns random bits from an entropy source.

gfavor:
In short, avoiding or minimizing CSRs with read side-effects is a very good goal.  But, like with most things in the real world, there are often some exceptions that are justified - such as the ones that John mentioned.  In addition, the CLIC also has a CSR with read side-effects that is rather analogous to AIA's *claimei CSRs.

David-Horner:
On 2021-07-15 6:51 p.m., gfavor wrote:
>
> In short, avoiding or minimizing CSRs with read side-effects is a very 
> good goal.
>
agreed.
>
> But, like with most things in the real world, there are often some 
> exceptions that are justified - such as the ones that John mentioned.
>
Memory mapped I/O would justify "read side-effects".

But CSRs don't need to be designated Read-Only-with-Side-Effects.

All instructions to [explicitly] read CSR values are also a write 
instructions.

RISCV can use a READ-WRITE CSR to simulate a read-with side-effects, and 
in the process provide a re-readable option.

A READ Only CSR instruction, rs1=0, would read the value without 
advancing or enacting any further side-effects.

A read/write [immediate or otherwise] would enable the side-effect 
[probably with a specific value provided].

> In addition, the CLIC also has a CSR with read side-effects
>
Ouch. I missed that. I will follow up on it..

gfavor:
On Thu, Jul 15, 2021 at 8:37 PM David-Horner ***@***.***>
wrote:

> But CSRs don't need to be designated Read-Only-with-Side-Effects.
>

This is a very broad presumption that no RV extension will ever have a
justifiable need for a CSR read with side-effects.  And that the current
three examples are inappropriate and unjustified.

This is not an inflexibility that RV should impose a blanket
from-now-to-eternity disallowance on.  Conversely, the RV Arch
Review committee will certainly serve to ensure that such CSRs exist only
when properly justified (and, to begin with, TGs should be exercising
proper judgement in this regard).


>
> All instructions to [explicitly] read CSR values are also a write
> instructions.
>

Unless I'm misunderstanding this statement, there ARE certain forms of CSR
instructions that are architecturally designed to only read CSR state.  And
they must be used when accessing a R/O CSR, otherwise a fault will result.


> RISCV can use a READ-WRITE CSR to simulate a read-with side-effects, and
> in the process provide a re-readable option.
>

This isn't a commonly usable "solution".  CLIC's and AIA's read
side-effects affect other arch state.  And this isn't applicable
to read-only registers.


>
> A READ Only CSR instruction, rs1=0, would read the value without
> advancing or enacting any further side-effects.
>
> A read/write [immediate or otherwise] would enable the side-effect
> [probably with a specific value provided].
>

Is a write with side-effects fundamentally any better than a read with
side-effects (although this is moot since some of the existing examples are
read-only registers).

And semantically having a CSR write instruction not write the rs1/imm data
into the CSR in the standard manner generally defined for CSR instructions
would be inconsistent and would arguably represent an arch change.

Greg

David-Horner:
On 2021-07-16 12:15 a.m., gfavor wrote:
> On Thu, Jul 15, 2021 at 8:37 PM David-Horner ***@***.***>
> wrote:
>
> > But CSRs don't need to be designated Read-Only-with-Side-Effects.
> >
>
> This is a very broad presumption that no RV extension will ever have a
> justifiable need for a CSR read with side-effects.

It is not a presumption. I am postulating an alternate formulation of 
read and advance that does not use Read-Only.

If this alternative is accepted, then there will be no need [ever] that 
justifies a CSR read with side-effects.

In other words I am saying RISCV does not need this designation because 
it does not need this mechanism. Controlled advance [of side-effects] 
supplants the need for read [only] triggered advance.  Some 
architectures may need it, mem mapped I/O does, but RISCV does not.

> And that the current
> three examples are inappropriate and unjustified.

The three examples are significant and the functionality is justified. 
But the encoding can be more permissive and reliable.

I am definitely not saying the advance with a read is inappropriate or 
unjustified. Just that it can be better emulated/managed with the 
current R/W CSR mechanism.

>
> This is not an inflexibility that RV should impose a blanket
> from-now-to-eternity disallowance on. 
We can and do allow Custom to do what they want. It is Standardization 
we are needing to define.
> Conversely, the RV Arch
> Review committee will certainly serve to ensure that such CSRs exist only
> when properly justified (and, to begin with, TGs should be exercising
> proper judgement in this regard).

I wish I had that confidence. If I did I would expect TGs to have 
already exercised that proper judgement and disallowed ROwithSideEffects.

Indeed, I understood that to be the case by the architects BECAUSE it 
was not provided in the cannon.

>
>
> >
> > All instructions to [explicitly] read CSR values are also a write
> > instructions.
> >
>
> Unless I'm misunderstanding this statement, there ARE certain forms of CSR
> instructions that are architecturally designed to only read CSR state. And
> they must be used when accessing a R/O CSR, otherwise a fault will result.

True. No argument from me here. But I'm sure you understood what I said 
and what I meant.

All CSR Read also have the write variant. Architecturally there is no 
penalty to include the write. The code size is the same. Microarchs may 
cause write/clear/set variants to be more expensive, take more cycles, 
but once you add side-effects to a R/O CSR uArchs can [find they need 
to] make these more expensive, too.

>
> > RISCV can use a READ-WRITE CSR to simulate a read-with side-effects, and
> > in the process provide a re-readable option.
> >
>
> This isn't a commonly usable "solution". CLIC's and AIA's read
> side-effects affect other arch state. And this isn't applicable
> to read-only registers.

Code from other architectures will have to be ported to RISCV anyway, 
and specifically changed to use the CSR[R][W/S/C][I] instructions.

It is a universally usable "solution". The default could be CSRRSI. And 
the CSRR be used for debugging.

I quote Allen from above about such R/OwithSidEffects:

"legacy IO devices ... are impossible to debug".

>
>
> >
> > A READ Only CSR instruction, rs1=0, would read the value without
> > advancing or enacting any further side-effects.
> >
> > A read/write [immediate or otherwise] would enable the side-effect
> > [probably with a specific value provided].
> >
>
> Is a write with side-effects fundamentally any better than a read with
> side-effects (although this is moot since some of the existing 
> examples are
> read-only registers).

It keeps RO pristine. To me that is significantly better.

We already have RW that have "side-effect" behaviour. And we have a name 
for some of them, especially:  WARL

Other R/W have exotic behaviours.

We may have to re-allocate more RO to RW [with side-effects] if we 
cannot find enough RO candidates and are running out of RW.

>
> And semantically having a CSR write instruction not write the rs1/imm data
> into the CSR in the standard manner generally defined for CSR instructions
> would be inconsistent and would arguably represent an arch change.

It happens already for R/W CSR, what you write is not what you get [WARL].

What would be an arch change is to introduce RO with SideEffects.

allenjbaum:
well, to be fair: entropy would return random bits whether you read it or
not.
The claim registers: IS there ever a case that you write the claim
register?
That is, could you make the side effect occur as the result of a CSRRW
instead of CSRRD?

allenjbaum:
Ah, just read further and saw that David proposed the same thing.
I'm not so much objecting to the idea of a read with side effects as much
as the need for carving out a special space for them in the CSR allocations.
I really don't see a benefit to that from an implementation standpoint.
It's happening behind the scenes, just a UTIME changes values when you read
it (depending on the interval between reads, etc - but still - behind the
scenes state change).
The fact that it triggered by the read vs some other external event (clock
pulse for UTIME, interrupt arrival for xIP, execution of an FP instruction
for MSTATUS.XS) seems totally irrelevant.

aswaterman:
@allenjbaum Assuming we do offer read-only CSRs with read side effects, implementations will benefit from being able to quickly decode that fact, so that they can decide whether the CSR access can be performed speculatively.  Not knowing that fact early will imply pipeline flushes for implementations that can't afford to wait for the implementations.  Hence, we'd rather demarcate the space up-front so that the decode function remains simple.

allenjbaum:
The CSR access can be performed speculatively - it's the update that can't,
but I do see the point.
That also sounds like an argument to force that kind of access to use
CSRRW, so the usual mechanism that holds off writes can be used.

aswaterman:
There is indeed a valid case to be made for rejecting read side effects altogether, instead forcing the use of CSR writes.  John would validly retort that doing so places artificial pressure on the use of the CSR address space.  Reasonable minds can disagree on the relative merits of these points.  For my part, being able to quickly decode this property suffices: the CSR address is effectively an extension of the CSR opcode.

(Personally, I think the fact that the CLIC has gone down the read-side-effects path is borderline irrelevant, since I don't think the CLIC will be a common feature of high-performance cores.)

allenjbaum:
I don't understand the concern of artificial pressure on the use of CSR
addresses.
I view this as just the opposite: instead of confining use of
read-with-side-effect to some newly carved out region of CSR space, it
opens it up to any RW CSR.
And, by the way, while I view read-with-side-effects as an abomination - I
will concede there are legitimate uses cases, and CLAIM is one of them,
as are FIFO read kinds of applications (of which I would put CLAIM). I
don't like it, but that use case is kind of unavoidable.

So, the path that was taken here is:
   - CLAIM is never explicitly written, so put it in readonly region
(despite side effects)
   - for high performance implementations, we can't speculatively update,
so we need to inhibit that
   - we don't need to inhibit speculation for readonly CSRs - but for these
specific CSRs  (with side effects) - you do
   - so let's carve out a non-speculative readonly CSR region.
A side effect of this is we've limited this type of CSR to an even more
limited region, and we need to specially inhibit speculation for this
carved-out region.

The alternative approach is
   - CLAIM has side effects on read, so put it in (any) RW region
   -  have the implementation ignore the write. This cost is zero. Less
than zero, perhaps.
   - You're done, because now you don't have to do anything special for
speculation that you're not already doing.
   - (not sure it's useful, but now you might be able to read without side
effect also by specify X0 as the destination register as well)

jhauser-us:
So that standard RISC-V CSRs can always be safely read for debugging purposes, an informal conclusion has been reached that no standard CSRs will be defined to have side effects on reads.  This pull request is withdrawn by the author.

PR #677, opened 2021-07-18
Rename hypervisor extension "operating mode" to "privilege mode"

jhauser-us:
For the hypervisor extension, define "privilege mode" as the combination of virtualization mode and the new concept of "nominal privilege mode".

### DIFF
--- a/src/hypervisor.tex
+++ b/src/hypervisor.tex
@@ -53,30 +53,33 @@ When V=1, the hart is either in virtual S-mode (VS-mode), or in virtual U-mode
 When V=0, the hart is either in M-mode, in HS-mode, or in U-mode atop an OS
 running in HS-mode.
 The virtualization mode also indicates whether two-stage address translation
-is active (V=1) or inactive (V=0).  Table~\ref{h-operating-modes} lists the
-possible operating modes of a RISC-V hart with the hypervisor extension.
+is active (V=1) or inactive (V=0).  Table~\ref{tab:HPrivModes} lists the
+possible privilege modes of a RISC-V hart with the hypervisor extension.
 
 \begin{table*}[h!]
 \begin{center}
 \begin{tabular}{|c|c||l|l|l|}
   \hline
-   Virtualization & Privilege & \multirow{2}{*}{Abbreviation} & \multirow{2}{*}{Name} & Two-Stage \\
-   Mode (V)       & Encoding  &                               &                       & Translation \\ \hline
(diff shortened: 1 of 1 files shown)


ethindp:
I'm not up to speed on RISC-v developments but I feel like the name is ambiguous. To me, "privilege mode" (or the past tense of the phrase) means "any mode that is capable of executing privileged system instructions". I feel like individualizing the modes makes it clearer as to what mode you are referring to, instead of calling any privileged execution mode or combination of modes a "privilege mode". Could you provide a rationale for this change? (I'm not sure what the "nominal privilege mode" does or does not allow compared to the "virtualization mode", so I might be missing something.)

jhauser-us:
The RISC-V ISA manual Volume II makes clear that U (user) is considered one of the privilege levels, and U-mode is one of the privilege modes for a hart.  This has been the case from the beginning.  This pull request makes no change in that regard.

ethindp:
I don't think I ever implied that it did. I am simply trying to understand why this PR is necessary.

gfavor:
I'll bite.  Most people I interact with in RISC-V already say "privilege mode" when referring to what is currently called "operating mode.  That is conceptually natural and consistent with the idea or intuition that the H extension simply adds two more privilege modes to the existing three.  So why introduce a new term, i.e. operating modes.

The architecture already distinguishes (encoding-wise in various CSRs) between the Virtualization Mode and the Privilege Level. 
 (Btw, the current Priv spec notes that priv modes and levels are often tied together or equated with each other, but also explains the fine distinction between the two.)  So maintaining "privilege mode" as being the overall privilege mode (one of five) maintains greater consistency with the current Priv arch, and then using the term Nominal Privilege instead of Privilege Encoding is (in my mind) a good clarification or better way to refer to that.  Especially since this mirrors the idea - with this virtualization architecture (unlike ARM's) - that there are three "base" privilege levels/modes and then H introduces virtualized variants of two of those (i.e. VS and VU versus HS and U modes).

In short, at least from my own perspective, this change is a clearer expression of what is going on with privilege levels and  modes, and does so without introducing a new third term (operating modes).

ethindp:
@gfavor I understand now, and now that I'm able to look at the diff (I wasn't able to earlier -- my screen reader was making it a bit difficult :P) I do see the clarification. Thanks for the explanation.

kasanovic:
I think people are confusing "privilege mode" (what your current mode is) with "privileged mode" (one that has privileges).

PR #678, opened 2021-07-18
mstatus/hstatus GVA bit is set for additional traps

jhauser-us:


src/hypervisor.tex
```diff
@@ -322,15 +322,20 @@ \subsection{Hypervisor Status Register ({\tt hstatus})}
...
 \begin{commentary}
-For memory faults, GVA is redundant with field SPV (the two bits are set
+For memory access traps, GVA is redundant with field SPV (the two bits are set
 the same) except when the explicit memory access of an HLV, HLVX, or HSV
 instruction causes a fault.
 In that case, SPV=0 but GVA=1.
+
+If an instruction address misaligned or breakpoint trap writes zero to
+{\tt stval} instead of the faulting virtual address, then GVA=0 even if
+SPV=1.
```
scottj97, line 338: Why does this comment refer only to 2 of the 5 types of traps listed just above? (And does this refer only to instruction breakpoint traps and not data breakpoint traps?)

Does this mean that if an access fault, page fault, or guest-page fault that writes 0 to `stval` then GVA can still be 1?

jhauser-us: For the other types of traps (load/store address misaligned, access fault, page fault, or guest-page fault), the hypervisor extension will require that the appropriate `*tval` CSR must be written with the faulting virtual address.  (This used to be required of all RISC-V cores, but at some point was weakened.)  Because there will be no option to write zero to `*tval` for those traps, GVA will never be zero if SPV = 1.

scottj97: Gotcha...what about data breakpoint vs instruction breakpoint? If this refers to both, how about reversing the two:

> If a breakpoint or instruction address misaligned trap writes zero to {\tt stval} instead of the faulting virtual address, then GVA=0 even if SPV=1.

jhauser-us: Okay, done.

aswaterman:
LGTM, thanks.

PR #683, opened 2021-07-21
mstatush is not optional in priv-1.12 for RV32

aswaterman:
But of course it can be hardwired to 0 in most implementations.

cc @jhauser-us


jhauser-us:
LGTM.

PR #684, opened 2021-07-22
Improve labels of CSR categories

jhauser-us:
In particular, relabel "user CSRs" to be "unprivileged and user CSRs".


aswaterman:
LGTM.

PR #685, opened 2021-07-22
Resolve contradiction in mtval definition

aswaterman:
The definition gives the implementation the flexibility of hardwiring mtval to 0, but also says that mtval is always written on certain exceptions.  This is, at a minimum, confusing since if mtval is hardwired to 0, those writes are fruitless.

Resolve by conditioning the sentence on whether mtval is hardwired to 0.

@jhauser-us -- this is not the resolution we discussed last night, but I think it does the trick, right?

### DIFF
--- a/src/machine.tex
+++ b/src/machine.tex
@@ -2123,8 +2123,10 @@ software in handling the trap.  Otherwise, {\tt mtval} is never written by the
 implementation, though it may be explicitly written by software.  The hardware
 platform will specify which exceptions must set {\tt mtval} informatively and
 which may unconditionally set it to zero.
+If the hardware platform specifies that no exceptions set {\tt mtval} to a
+nonzero value, then {\tt mtval} is hardwired to zero.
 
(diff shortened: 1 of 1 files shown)


jhauser-us:
That edit seems okay, but later in the same section there is also:
> For misaligned loads and stores that cause access-fault or page-fault exceptions, `mtval` will contain the virtual address ....  For instruction access-fault or page-fault exceptions on systems with variable-length instructions, `mtval` will contain the virtual address ....

Plus, we have:
> If the hardware platform specifies that no exceptions set `mtval` to a nonzero value, then it may be hardwired to zero.

The way you're going, you better say that it *is* hardwired to zero, not just that it may be.

aswaterman:
Thanks for the review, @jhauser-us, and yes, I agree on the latter point.  Can you re-review?

jhauser-us:
The new version looks good.

aswaterman:
Thank you!

PR #687, opened 2021-07-22
Standard CSRs don't have read side effects

aswaterman:
No normative change to existing standards in this PR.

### DIFF
--- a/src/csr.tex
+++ b/src/csr.tex
@@ -165,9 +165,11 @@ be side effects of that write.
   trap, the trap is not considered a side effect of the write but merely
   an indirect effect.
 
-  The CSRs defined so far in this volume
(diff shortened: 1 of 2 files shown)


gfavor (approved):
Looks good overall.

David-Horner:
I am certainly in agreement with this change.

It matches what I initially suggested on the list.

However, I am surprised we didn't reach a consensus on the list before 
this "affirmation of the status quo".

The remaining concern I have is the weakness of the no sideeffects on reads.

It says:

Standard CSRs *do not* have any side effects on reads.

rather than "must not".

Combined with the "thus far" in the next sentence, one might infer the 
"do not" as also just informative of the state "thus far'.

PR #688, opened 2021-07-23
Extension names

pdonahue-ventana:
- S-level extensions are Sv\* and Ss\*
- H extensions (which are not "H-level" since there is no such level in priv spec table 1.1) are H\*
- M-level extensions are Sm\*
- Extended the underscore requirement to these extensions so we don't end up with rv64gsvinvalsvnapot which likely cannot be parsed by machine and definitely cannot be parsed by humans
- Fixed a couple of typos

### DIFF
--- a/src/naming.tex
+++ b/src/naming.tex
@@ -95,33 +95,40 @@ from other multi-letter extensions by an underscore, e.g.,
 
 \section{Supervisor-level Instruction-Set Extensions}
 
-Standard supervisor-level instruction-set extensions are defined in Volume II,
-but are named using ``S'' as a prefix, followed by an alphabetical name and an
-optional version number.  Supervisor-level extensions must be separated from
-other multi-letter extensions by an underscore.
+Standard extensions that extend the supervisor-level virtual-memory
+architecture are prefixed with the letters ``Sv'', followed by an alphabetical
+name and an optional version number.  Other standard extensions that extend
+the supervisor-level architecture are prefixed with the letters ``Ss'',
+followed by an alphabetical name and an optional version number.  Such
+extensions are defined in Volume II.
 
 Standard supervisor-level extensions should be listed after standard
 unprivileged extensions.  If multiple supervisor-level extensions are listed,
-they should be ordered alphabetically.
+they should be ordered alphabetically.  Supervisor-level extensions must be
+separated from other multi-letter extensions by an underscore.
 
-\section{Hypervisor-level Instruction-Set Extensions}
+\section{Hypervisor Instruction-Set Extensions}
 
-Standard hypervisor-level instruction-set extensions are named like
-supervisor-level extensions, but beginning with the letter ``H'' instead of
-the letter ``S''.
+Standard hypervisor instruction-set extensions are prefixed with the letter
+``H'', followed by an alphabetical name and an optional version number.
+Such extensions are defined in Volume II.
 
-Standard hypervisor-level extensions should be listed after standard
-lesser-privileged extensions.  If multiple hypervisor-level extensions are
-listed, they should be ordered alphabetically.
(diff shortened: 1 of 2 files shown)


aswaterman:
Thanks, Paul.  Since I know @kasanovic has long planned to revisit this topic, ideally he'll be able to review this.

pdonahue-ventana:
Now that all the Sv* and Sm* and Ss* extensions are in public review, I think that this should be merged to make those extension names legit.

gfavor:
One of the upcoming tasks to be done by Krste is documenting the new/updated naming conventions in the Unpriv spec (which subsumes addressing this GH issue).

pdonahue-ventana:
This isn't an issue.  This is a pull request which, if merged, would make all of the Sv*, Sm*, and Ss* extension names legal.  It just is waiting for Krste to read and comment, not to write the spec.

src/naming.tex
```diff
@@ -95,33 +95,40 @@ \section{Additional Standard Extension Names}
...
+Standard hypervisor extensions should be listed after standard unprivileged
+and supervisor-level extensions.  If multiple hypervisor extensions are
+listed, they should be ordered alphabetically.  Hypervisor extensions
+must be separated from other multi-letter extensions by an underscore.
 
 \section{Machine-level Instruction-Set Extensions}
 
-Standard machine-level instruction-set extensions are prefixed with the three
-letters ``Zxm''.
+Standard extensions that only extend the machine-level architecture are
```
lbmeng, line 123: Can you please elaborate what scenario is classified as "only extend the machine-level architecture"?

For example, if an extension adds both machine-level CSRs and supervisor-level CSRs, does it fall into the category?
If not, so [https://github.com/riscv/riscv-state-enable/](Smstateen) should really be named as **Ssstateen**?

lbmeng:
What about debug extension?

Currently there are 2 debug related extensions documented [here](https://github.com/riscv/riscv-debug-spec/blob/master/riscv-debug-stable.pdf): 

- Chapter 4 Sdext ISA Extension
- Chapter 5 Sdtrig ISA Extension

gfavor:
The convention is that S* represents extensions to the Privileged architecture (versus the Unpriv arch). The second letter represents categories of extensions.  'v' for virt-mem; 'd' for debug; 'm' for Machine-level; 's' for Supervisor-level.

The Smstateen extension, when rewritten and integrated into the Priv spec, will be split into the official Smstateen and Ssstateen extensions.  The former will represent the new Machine-level functionality and the latter will represent the new Supervisor-level functionality.

In general, arch extensions that introduce both M and S level functionality will be represented as two privilege level specific extensions (also analogous to how the informally named "Priv 1.12" spec became the official Sm1p12 and Ss1p12 extensions).

henry-hsieh:
Will this change be transferred to adoc version?

wmat:
Note that the .tex version of the spec has been archived.  This change will need to be redone against the naming.adoc if it is still relevant.

pdonahue-ventana:
There definitely needs to be a change.  For years we've been naming extensions with an undocumented naming scheme that conflicts with the documented naming scheme.

After 2+ years, this PR is still not approved.  I'm not going to spend time creating a new adoc PR only to have that sit around for 2 more years.  I encourage others to do so, as long as they are prepared to more forcefully advocate for it.

aswaterman:
@pdonahue-ventana Your consternation isn’t unwarranted. I’ll try to incorporate an updated variant of this PR today. Really, the entire chapter needs to be rewritten, but I can’t do that unilaterally.

aswaterman:
@pdonahue-ventana I have pushed to the main branch a variant of this proposal.  Please file a follow-up issue or PR if you have any suggestions or corrections.

PR #691, opened 2021-07-25
Create separate PDF for the PLIC specification

axel-h:
The PLIC section has been removed from the privileged ISA, build a separate PDF for it to make the content accessible while the still exist, so the content are not lost and accessible to interested readers. I'm not sure if and when https://github.com/riscv/riscv-plic-spec will take over here.

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -8,19 +8,19 @@
 # build directory, copy the makefile there, and change the srcdir
 # variable accordingly.
 #
-# Note that the makefile assumes that the default dvips/ps2pdfwr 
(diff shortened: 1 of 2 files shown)


aswaterman:
There's an effort to more fully specify a PLIC here: https://github.com/riscv/riscv-plic-spec -- we shouldn't have two PLIC specs.

axel-h:
In this case I suggest to delete `plic.tex` from this repo completely, so nobody will look at this when searching for the latest draft. Or even spend time on modifying it.

PR #693, opened 2021-07-25
Remove PLIC documentation and figures

axel-h:
Follow up from https://github.com/riscv/riscv-isa-manual/pull/691

The PLIC chapter has been removed from the privileged specification and will be maintained at https://github.com/riscv/riscv-plic-spec now.


axel-h:
I wonder what is the relation between https://github.com/riscv/riscv-aia and  https://github.com/riscv/riscv-plic-spec now?

gfavor:
No relation between PLIC and AIA - other than they are two different interrupt controller architectures.  The Advanced PLIC in AIA is just one component of AIA and is a different animal in its details from PLIC, despite the similarity in name.

axel-h:
https://github.com/riscv/riscv-plic-spec/pull/25#pullrequestreview-715854745 adds the reference to the last version where the PLAC chapter existed.

kersten1:
None of these files seem to exist any more so I'm closing the PR.

PR #694, opened 2021-07-30
Improved rules for virtual instruction exceptions

jhauser-us:


aswaterman:
LGTM, thanks for the clear explanation of the general case.

PR #696, opened 2021-07-30
Add CSRs henvcfg/henvcfgh to hypervisor extension

jhauser-us:


scottj97:
Where is `senvcfg` defined?

aswaterman:
@scottj97 I've been tasked with writing that spec.  When I get to it, I'll make a single PR that incorporates both that and this one.

aswaterman:
Incorporating this (and shortly the menvcfg/senvcfg specs) into #727, partially because they are conceptually related and partially to avoid merge conflicts.

PR #697, opened 2021-07-31
Add mconfigptr CSR

aswaterman:


src/machine.tex
```diff
@@ -2224,6 +2224,50 @@ \subsection{Machine Trap Value Register ({\tt mtval})}
...
 smaller of XLEN and ILEN.
 
+\subsection{``Smdisc'' Extension for Machine Configuration Pointer ({\tt mconfigptr}) CSR}
+
+This section defines the Smdisc extension, which adds the {\tt mconfigptr} CSR.
+
+{\tt mconfigptr} is an MXLEN-bit read-only CSR, formatted as shown in
+Figure~\ref{mconfigptrreg}, that holds the physical address of a configuration
+data structure.
+Software can use traverse this data structure to discover information about
```
gfavor, line 2234: "... use traverse ..." should just be "... traverse ...".

src/machine.tex
```diff
@@ -2224,6 +2224,50 @@ \subsection{Machine Trap Value Register ({\tt mtval})}
...
+\caption{Machine Configuration Pointer register.}
+\label{mconfigptrreg}
+\end{figure}
+
+The pointer alignment in bits must be no smaller than the greatest supported
+MXLEN: i.e., if the greatest supported MXLEN is $8\times n$, then
+{\tt mconfigptr}[$\log_2n$-1:0] must be hardwired to 0.
+
+\begin{commentary}
+The format of the configuration data structure has yet to be standardized.
```
gfavor, line 2259: Maybe say "The format and schema of ...".

gfavor (commented):
Added two comments about small text tweaks.

aswaterman:
👍

gfavor (approved):
Looks good.

jhauser-us:
I'm unclear why this needs to be a named extension.  We've already added to Priv 1.12 new CSRs `mstatush`, `mcontext`, and `scontext`, and we're planning soon to add `menvcfg`, `menvcfgh`, and `senvcfg`.  The `mconfigptr` CSR seems sufficiently fundamental to me that we should make it mandatory in 1.12, but optionally zero, just like `mvendorid`, etc.

aswaterman:
That approach would seem more consistent.

gfavor:
While this is arguable both ways, I would tend to agree with John.  Initially some had suggested that this should be a fast-track extension, but that was without consideration of what John summarized.

Given that there is no other functionality, that this is a read-only register (nominally hardwired in some or many implementations), and that this can even be hardwired to zero in a design that doesn't want to utilize the associated discovery method, the cost of this as a mandatory feature is quite minimal.  I'm all for simplifying the matter and just making this part of Priv 1.12.

aswaterman:
Yeah, let's go that route.  I've updated the PR.

JamesKenneyImperas:
I'm interested to know more about the rationale and intent of this `Smdisc` extension (for example, why it is better to have a memory-mapped block for description of features rather than using read-only CSRs, and how this fits with the existing `misa` CSR). Is there anywhere I can look to find out more?

Thanks.

gfavor:
>
> I'm interested to know more about the rationale and intent of this Smdisc
> extension (for example, why it is better to have a memory-mapped block for
> description of features rather than using read-only CSRs, and how this fits
> with the existing misa CSR). Is there anywhere I can look to find out
> more?
>
A full answer will be a long answer since there are many aspects to
discover at various levels in the software stack.  So I'll only briefly
mention some things:

- To be clear, the mechanism currently being developed is for "low-level"
discovery.  It does not replace, for example, Device Tree or ACPI discovery
structures - which are more "high-level" discovery methods understood by
"rich" OS's and hypervisors.

- A CSR-based mechanism (e.g. analogous to CPUID in x86) was considered.
It of course has its pros, but it also has its cons.

- Overall there will be many extensions, and a growing list of
extensions over time, that will have discoverable info.  Some may
only require one bit of info, others may require tens of bytes or more.

- Hardware cost and need for compression, especially for cost-sensitive
designs, are important factors.

- Flexibility is needed to accommodate a wide range of needs as far as
amount and types of discoverable info.  (The amount of info for low-end
cost-sensitive embedded systems will likely be cut down from the amount of
info that would be available for "rich OS" platforms.)

- Virtualizing the current approach being developed is easy.  Virtualizing
a register-based approach is not.  (x86 for example unconditionally traps
and software emulates all use of the CPUID mechanism within guest VMs.)
Somewhat similarly, there are issues about controlling what CSR-based
discovery info should be allowed or disallowed to be seen by, for example,
User-level code.

- There are even more considerations, but I'll stop here except to note
that the use of the misa CSR for discovery is quite limited.  There are
more extensions (and will be yet more each year) than there are letters in
the alphabet.  So general discovery of even simply what individual
extensions are supported can't use misa effectively.  For example, there is
not one BitManip extension, but a high single-digit number of bitmanip
extensions heading towards ratification this year, and maybe more to come
next or following years.  So what should the misa.B bit mean?  The latest
position is that it doesn't offer much value and won't be allocated for
bitmanip.  The new low-level discovery method being developed will readily
support this granularity of discovery.  (Ditto for some other misa bits
that haven't been officially allocated yet.)

Greg

lbmeng:
Is the configuration structure format standardized in https://github.com/riscv/configuration-structure?

Is it good to mandate that mconfigptr approach, when implemented, should be interrogated  by software instead of misa register when probing standard extensions?

aswaterman:
The configuration structure is merely a proposal and has not been standardized.

We are _not_ discouraging use of `misa`, but over time the configuration structure will provide more information than `misa`.

PR #699, opened 2021-08-03
Update status of various extensions

jhauser-us:


src/preface.tex
```diff
@@ -39,7 +39,14 @@ \chapter{Preface}
     \em V          & \em 0.7 & \em Draft \\
     \bf Zicsr      & \bf 2.0 & \bf Ratified \\
     \bf Zifencei   & \bf 2.0 & \bf Ratified \\
+    \em Zihintpause & \em 1.0 & \em Draft \\
```
gfavor, line 42: Zihintpause is ratified (as of the beginning of the year).

jhauser-us: Okay, fixed.

I thought previous convention was to change the version number of ratified components to 2.0.

gfavor: That would be a question for Andrew to properly answer.  I wasn't aware one way or the other.

aswaterman: In some sense, we really don't even need version numbers for ratified extensions, since by construction they don't change... but I agree we should stick to the convention at least for now.  I'll take this PR from here and make that change.

aswaterman:
Thanks!

PR #701, opened 2021-08-03
Improve description of interrupt traps

aswaterman:
Supersedes #590

This version is much more relaxed but nevertheless covers the important bases of SRET and explicit toggling of xip/xie.

### DIFF
--- a/src/machine.tex
+++ b/src/machine.tex
@@ -1364,17 +1364,22 @@ MXLEN \\
 \label{miereg}
 \end{figure}
 
-An interrupt \textit{i} will be taken if bit \textit{i} is set in both
-{\tt mip} and {\tt mie}, and if interrupts are globally enabled.  By
-default, M-mode interrupts are globally enabled if the hart's current
-privilege mode is less than M, or if the current privilege mode is M
-and the MIE bit in the {\tt mstatus} register is set.  If bit \textit{i}
-in {\tt mideleg} is set, however, interrupts are considered to be
-globally enabled if the hart's current privilege mode equals the
-delegated privilege mode and that mode's interrupt enable
-bit (\textit{x}\/IE in {\tt mstatus} for mode~\textit{x}) is set,
-or if the current
-privilege mode is less than the delegated privilege mode.
+An interrupt~\textit{i} will trap to M-mode (causing the privilege mode
+to change to M-mode) if all of the following are true:
+(a)~either the current privilege mode is M and the MIE bit in the
+{\tt mstatus} register is set, or the current privilege mode has less
+privilege than M-mode;
+(b)~bit~\textit{i} is set in both {\tt mip} and {\tt mie}; and
+(c)~if register {\tt mideleg} exists, bit~\textit{i} is not set in
+{\tt mideleg}.
+
+These conditions for an interrupt trap to occur must be evaluated in a bounded
+amount of time from when an interrupt becomes pending in {\tt mip}, and must
+also be evaluated immediately following the execution of an {\em x}\/RET
+instruction or an explicit write to {\tt mip} or {\tt mie}.
+
+Interrupts to M-mode take priority over any interrupts to lower privilege
+modes.
 
 Each individual bit in register {\tt mip} may be writable or may be
 read-only.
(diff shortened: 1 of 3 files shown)


src/supervisor.tex
```diff
@@ -368,12 +368,20 @@ \subsection{Supervisor Interrupt Registers ({\tt sip} and {\tt sie})}
...
+following are true:
+(a)~either the current privilege mode is S and the SIE bit in the
+{\tt sstatus} register is set, or the current privilege mode has less
+privilege than S-mode; and
+(b)~bit~\textit{i} is set in both {\tt sip} and {\tt sie}.
+
+These conditions for an interrupt trap to occur must be evaluated in a bounded
+amount of time from when an interrupt becomes pending in {\tt sip}, and must
+also be evaluated immediately following the execution of an SRET instruction
+or an explicit write to {\tt sip} or {\tt sie}.
```
gfavor, line 381: Some questions and observations that all reflect a certain unease over the loosy-goosy-ness of interrupt recognition.

1)  What about before or after an explicit read of sip, i.e. should one be able to directly observe a pending enabled interrupt that remains to be taken?  Further, currently it seems like one could repeatedly read sip and see an enabled interrupt continuing to be pending for a "bounded amount of time".   Although this appears to be the direct intended implication since how else can one know that an interrupt is pending in sip.

2a)  An enabled interrupt can be long pending and then disabled by a sip write without any guarantee of ever being taken.  But then, when the interrupt is re-enabled, there is now an immediate guarantee of being taken.

2b)  It seems a little weird that an enabled interrupt can remain pending for an unknown (but bounded) amount of time, but then by simply writing sie (even if the enable status remains unchanged), the interrupt is now guaranteed to immediately be taken.  This smacks of a bit of logical arbitrariness in having a change to unrelated pending or enable bits force a behavior on taking of a given interrupt.

3a)  Setting an enabled sip bit via an explicit write must immediately be evaluated and taken right after the write instruction - and hence before it could ever be observed via a sip read.  But when an enabled sip bit is set externally, then it isn't required to be taken before or even after it can be observed via a sip read.

3b)  It seems somewhat inconsistent that if an enabled interrupt becomes pending an instant before an explicit sip write, then it must be taken immediately after the write.  Whereas if an interrupt becomes pending an instant after an explicit write, then it can remain pending for a "bounded" amount of time.

Obviously John's proposed definition was clean and simple, but much tighter.  What about just adding the idea that an enabled interrupt cannot explicitly be observed as pending, i.e. via a sip read?  Sort of like quantum mechanics - there are no status guarantees until you make an observation.  There may or may not be a long pending interrupt.  But once you try to observe a pending interrupt, you must only be able to observe a disabled pending interrupt (and an enabled pending interrupt must be taken by or before the point of observation).

This could amount to simply adding to the end of the proposed sentence "..., and before execution of an explicit read of sip."

Maybe also add that an enabled pending interrupt must be taken before a sip/sie write (especially since you're already requiring a tight sync between the instruction's execution and post-recognition of pending enabled interrupts)?  Which leads to "... or immediately before and after execution of an explicit read or write of sip or sie."

In any case, your call.  Sorry for the long rambling unease.

src/machine.tex
```diff
@@ -1364,17 +1364,22 @@ \subsection{Machine Interrupt Registers ({\tt mip} and {\tt mie})}
...
-privilege mode is less than M, or if the current privilege mode is M
-and the MIE bit in the {\tt mstatus} register is set.  If bit \textit{i}
-in {\tt mideleg} is set, however, interrupts are considered to be
-globally enabled if the hart's current privilege mode equals the
-delegated privilege mode and that mode's interrupt enable
-bit (\textit{x}\/IE in {\tt mstatus} for mode~\textit{x}) is set,
-or if the current
-privilege mode is less than the delegated privilege mode.
+An interrupt~\textit{i} will trap to M-mode (causing the privilege mode
+to change to M-mode) if all of the following are true:
```
gfavor, line 1368: The preceding parenthetical comment is inconsistent in that there isn't a similar parenthetical comment for an interrupt trapping to HS-mode.  And, like above for HS-mode, this doesn't seem necessary to say (or otherwise needs to be said in both places).

aswaterman: IIRC, @jhauser-us excluded the corresponding remark from the hypervisor chapter because it's redundant with the one in the supervisor chapter.

src/supervisor.tex
aswaterman: I _want_ to allow the possibility that software reads `sip` and detects the presence of an interrupt that might've been taken but hasn't yet been--exactly the opposite of what you're getting at.  As was discussed on #590, many of us don't perceive this behavior to be a problem.  Allowing this oddity means that reads of `sip` don't need to be special-cased to avoid an inconsistency that doesn't really matter.

You mentioned on #590 an x86 behavior that you understandably want to mimic, and this proposal does satisfy that constraint.

Another prominent architecture seems to have taken the same tack as I'm proposing.  In their parlance, writes of sip/sie are context-synchronizing w.r.t. interrupts.  In @jhauser-us's proposal, _all_ instructions are context-synchronizing w.r.t. interrupts.

aswaterman:
Greg and I discussed the matter this evening.  Our collective unease about the vagueness of this sort of specification is offset by our inability to find compelling contrary use cases, and also by the fact that some other prominent architectures do even more relaxed things than we are proposing.

I'm merging this for the sake of forward progress.  Further argument to follow, no doubt.

PR #703, opened 2021-08-07
Improve rules for virtual instruction exceptions, again

jhauser-us:
One, clarify the rules for virtual instruction exceptions by adding the concept of *HS-qualified*; and two, make a special case for accesses to 32-bit high-half CSRs like `cycleh` and `htimedeltah`.  Also, when WFI is attempted in VU mode, the exception depends on `mstatus`.TW.


aswaterman:
LGTM.

PR #705, opened 2021-08-11
Generalize interrupt trap condition evaluation conditions

aswaterman:
This approach is more extensible, and now implicitly includes writes to sstatus.SIE / mstatus.MIE, as it should.

### DIFF
--- a/src/machine.tex
+++ b/src/machine.tex
@@ -1376,7 +1376,9 @@ privilege than M-mode;
 These conditions for an interrupt trap to occur must be evaluated in a bounded
 amount of time from when an interrupt becomes pending in {\tt mip}, and must
 also be evaluated immediately following the execution of an {\em x}\/RET
-instruction or an explicit write to {\tt mip} or {\tt mie}.
+instruction or an explicit write to a CSR on which these interrupt trap
+conditions expressly depend (including {\tt mip}, {\tt mie}, {\tt mstatus},
+and {\tt mideleg}).
 
 Interrupts to M-mode take priority over any interrupts to lower privilege
 modes.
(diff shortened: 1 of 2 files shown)


jhauser-us:
At the end of the new sentence perhaps add "(including `mstatus`, `mip`, and `mie`)" in the machine chapter, and add "(including `sstatus`, `sip`, and `sie`)" in the supervisor chapter?

aswaterman:
@jhauser-us Sounds good, but to avoid injecting further confusion, I’d like to add mideleg to the first list, since it is also explicitly mentioned in the preceding paragraph. (While it’s of course contrived to use the clearing of an mideleg bit to action the interrupt-taking logic, and while excluding mideleg from this clarifying list has no normative impact, it seems likely that excluding it will invite further questions.)

gfavor:
@aswaterman Not arguing against, but observing that including or excluding mideleg in the list makes a difference to hardware.  If not in the list, hardware is free to not ensure interrupt-taking right after a mideleg write.  If in the list, hardware specifically must ensure interrupt-taking right after a mideleg write.  If mideleg is in the list, one can imagine an eventual ACT testing this case to ensure that interrupt-taking occurs as required.

aswaterman:
Indeed. I’ll briefly note two things: first, John’s list is non-normative, in that it only clarifies which registers are “expressly depended” on. Second, while this does make a difference to HW, most implementations should do it automatically (by flushing the pipe on writes to an almost-never-written register), and this definition keeps the spec cleaner by not creating carve-outs for registers that are “expressly depended” on but don’t require this synchronization.

aswaterman:
Updated as proposed.  I'm going to merge this now to obviate near-term confusion, while acknowledging this might not be the final word on this topic.

PR #710, opened 2021-08-16
Corrections to mstatus in hypervisor chapter

jhauser-us:
1. For RV32, mstatush is now always required to exist.
2. mstatus.MPP is not just one bit.


aswaterman:
Indeed!  Thanks.

PR #712, opened 2021-08-16
VS mode should not see exception code 10

jhauser-us:


scottj97:
Why should VS-mode not be able to receive its own ecalls?

In Spike today, `medeleg` is allowed to delegate S-mode ecalls to S-mode. ([Discussion](https://github.com/riscv/riscv-isa-sim/pull/537))

jhauser-us:
When `medeleg` delegates S-mode ECALLs to S mode, they present as exception code 9.  If somebody wants VS mode to see its own ECALLs (and really they shouldn't), the exception code that VS mode sees should again be 9, not 10.  There's no reason to condone having VS mode see exception code 10 from its own ECALLs.  That would just make clear to the VS-mode software that it's running in a virtual machine.

aswaterman:
John has this one right.

Separately, there is a case to be made that Spike should not support delegating ECALL-from-S to S-mode.

PR #714, opened 2021-08-17
stval already cannot be zero on breakpoints, misaligned addresses

jhauser-us:


YenHaoChen:
Why does stval cannot be zero on breakpoints and misaligned addresses? The spec explicitly says it can be.
image

jhauser-us:
@YenHaoChen:  This is an old commit, and my assertion was (probably?) correct for the version of the document on Aug 16, 2021, but is no longer correct.  I recall there being a debate about this topic at the time.  The current Priv ISA 1.12, which was ratified four months later in December, 2021, changed the comment to say this:
> For breakpoint and memory access traps that write a nonzero value to `stval`, GVA is redundant with field SPV (the two bits are set the same) except when the explicit memory access of an HLV, HLVX, or HSV instruction causes a fault.  In that case, SPV=0 but GVA=1.

So the current ratified Priv spec allows for zero in `stval`.  Furthermore, the comment quoted above remains the same in the latest draft for Priv 1.13 (not yet ratified).

YenHaoChen:
@jhauser-us Thanks for your reply and clarification.

I was searching for the answer: If a breakpoint trap due to an EBREAK instruction writes zero to STVAL from VS/VU-mode, what should the GVA value be?

The current spec states: For any trap (breakpoint, ...) that writes a guest virtual address to STVAL, GVA is set to 1; otherwise, GVA is set to 0. The sentence isn't clear enough to me whether a breakpoint trap due to an EBREAK instruction writing zero to STVAL applies to the condition or not.

While searching for the answer, I found this deprecated commentary that clarifies the question. Since STVAL can be zero on a breakpoint trap due to the EBREAK instruction, the old commentary provides a better description of the GVA during traps.

Should we consider reverting this commit?

jhauser-us:
@YenHaoChen, if you think the current text you quoted isn't clear enough about whether a zero written to `stval` on an EBREAK trap counts as a guest virtual address when the EBREAK instruction's actual guest virtual address is not zero, please open a new GitHub issue requesting clarification on this specific question.  From there, you can point back to this commit.

PR #715, opened 2021-08-17
Clarify priorities of synchronous exceptions

jhauser-us:


pdonahue-ventana:
This seems like a pretty succinct way to specify this.  Thanks!

PR #718, opened 2021-08-19
Relax the naming rule of extension name

kito-cheng:
Current spec only allow alphabetical name, which means we can't use number in
the name, but we already defined serveral extension name with number,
like `zve32x`, `zvl128b`.

So this PR try to make the rule up to date and add one more rule to restrict the name must end with alphabetical to prevent ambiguous with version number.

### DIFF
--- a/src/naming.tex
+++ b/src/naming.tex
@@ -77,10 +77,10 @@ a number.  For example, ``rv32i2p2'' means version 2.2 of RV32I, whereas
 \section{Additional Standard Extension Names}
 
 Standard extensions can also be named using a single ``Z'' followed by an
-alphabetical name and an optional version number.  For example,
-``Zifencei'' names the instruction-fetch fence extension described in
-Chapter~\ref{chap:zifencei}; ``Zifencei2'' and ``Zifencei2p0'' name version
-2.0 of same.
+alphanumeric name, the name must end with an alphabetical, and an optional
+version number.  For example, ``Zifencei'' names the instruction-fetch fence
+extension described in Chapter~\ref{chap:zifencei}; ``Zifencei2'' and
+``Zifencei2p0'' name version 2.0 of same.
 
 The first letter following the ``Z'' conventionally indicates the most closely
 related alphabetical extension category, IMAFDQLCBKJTPV.  For the ``Zam''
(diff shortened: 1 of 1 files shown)


kito-cheng:
@kasanovic @aswaterman

jrtc27:
This renders the parsing of Zfoo1p0 ambiguous

nick-knight:
How about if we require the trailing alphabetical character must not be "p"?

kito-cheng:
Changes:
- Add extra rule to prevent ambiguou extension name with version
  - The second letter from the bottom cannot be a numeric if the
     last letter is ``p''
  - For example, Zfoo1p0 is ambiguou without this extra rule:
     Zfoo1p extension with version 0 or Zfoo extension with version 1.0.
- Note, I know there is a special word for `the second letter from the bottom` called `penultimate letter` after I ask google translation, but it's...too difficult word to non-native speaker (at least to me)

kito-cheng:
> How about if we require the trailing alphabetical character must not be "p"?

There is a zbp extension from bit manipulation, although it's not included in the ratification package this time, so I add more complicated rule for that, and let @aswaterman and @kasanovic to make the final decision.

src/naming.tex
```diff
@@ -77,10 +77,11 @@ \section{Underscores}
...
 Standard extensions can also be named using a single ``Z'' followed by an
-alphabetical name and an optional version number.  For example,
-``Zifencei'' names the instruction-fetch fence extension described in
-Chapter~\ref{chap:zifencei}; ``Zifencei2'' and ``Zifencei2p0'' name version
-2.0 of same.
+alphanumeric name, the name must end with an alphabetical, an optional
+version number, and the second letter from the bottom cannot be a numeric if the
+last letter is ``p''.  For example, ``Zifencei'' names the instruction-fetch
+fence extension described in Chapter~\ref{chap:zifencei}; ``Zifencei2'' and
+``Zifencei2p0'' name version 2.0 of same.
```
eopXD, line 84: looks like a grammatical error?

>  ``Zifencei2'' and ``Zifencei2p0'' name version 2.0 **of** same.

of -> are the

src/naming.tex
```diff
@@ -126,7 +129,9 @@ \section{Machine-level Instruction-Set Extensions}
 \section{Non-Standard Extension Names}
 
 Non-standard extensions are named using a single ``X'' followed by an
-alphabetical name and an optional version number.
+alphanumeric name, the name must end with an alphabetical, an optional
+version number,and the second letter from the bottom cannot be a numeric if the
```
eopXD, line 133: > ,and the second 

There is a space missing before `and`

eopXD (commented):
Minor errors :)

kito-cheng:
@eopXD Thanks :)

kito-cheng:
Changes:
- Minor grammatical fix.

pdonahue-ventana:
Note that #688 also changes this same extension naming section.

kersten1:
@kito-cheng Is this PR still valid? If so, I can move to adoc file.

topperc:
I think this is still relevant. I believe there are checks for this in binutils.

aswaterman:
I agree we need this.  If the PR is rebased onto the most recent spec (and with some minor grammatical cleanups), I'll approve and merge it.

kersten1:
PR is here - https://github.com/riscv/riscv-isa-manual/pull/1662

PR #720, opened 2021-08-24
Fix non-normative text about CSR ordering

aswaterman:
The non-normative text erroneously suggested that only accesses to time/cycle/mcycle need be ordered by FENCEs.  In fact, other CSR accesses are also visible to other harts, including interrupt-pending CSRs.

(The preceding normative text has no such deficiency.)

### DIFF
--- a/src/csr.tex
+++ b/src/csr.tex
@@ -234,15 +234,18 @@ result, the order of CSR accesses with respect to all other accesses is
 constrained by the same mechanisms that constrain the order of memory-mapped
 I/O accesses to such a region.
 
-These CSR-ordering constraints are imposed primarily to support ordering main
-memory and memory-mapped I/O accesses with respect to reads of the {\tt time}
-CSR.  With the exception of the {\tt time}, {\tt cycle}, and {\tt mcycle} CSRs,
-the CSRs defined thus far in Volumes I and II of this specification are not
-directly accessible to other harts or devices and cause no side effects visible
-to other harts or devices.  Thus, accesses to CSRs other than the
-aforementioned three can be freely reordered in the global memory order
-with respect to FENCE instructions
-without violating this specification.
+These CSR-ordering constraints are imposed to support ordering main
+memory and memory-mapped I/O accesses with respect to CSR accesses that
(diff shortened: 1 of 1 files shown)


gfavor (approved):
Looks good.

allenjbaum:
I don't quite understand the comment that CSRs are visible to other harts;
how is MIP visible? Clearing MIP or setting doesn't affect MIP in other
harts, and the only MMIO writes will otherwise affect the value read, so
it's those that must be fenced right?
 I'm trying to envision which other CSRs are visible outside the hart, and
I'm coming up blank outside of some debug CSRs and maybe some performance
counters which might aggregate threads in a core.
I'm clearly missing some context behind this, or misunderstanding something
about the xIP CSRs (both, most likely).

pdonahue-ventana:
The time CSR is a shadow copy of mtime which everyone can see.  The mip CSR is similarly a shadow copy of some interrupt controller state which everyone may be able to see, not directly on their own harts but via MMIO accesses to the interrupt controller.

gfavor:
"Visibility" can go in two directions.  A CSR (such as mip and sip) reflects external activity (e.g. another hart sending it an interrupt request).  That hart did some memory writes, fenced itself, and then sent the interrupt to this hart.  This hart reads the CSR, observes the pending interrupt, fences itself, and then does loads from the memory location that the other hart wrote to.  This hart's fence needs to order the loads after the CSR read (as described in the Unpriv spec).

"Visibility" can also go in the other direction, i.e. a write to a CSR has external effects that can be observed by other harts (or I/O devices).  This hart may do some stores to memory, then need to fence itself before doing the CSR write.  (Although note that currently there are no CSRs in the architecture that have external write effects.)

The existing normative text establishes all this fencing behavior wrt CSR accesses.  It is just the non-normative text that is misleading in saying that only reads of the time CSR can reflect external activity and hence must be subject to fencing, where in fact all the various CSRs that can reflect external interrupt pending status (i.e. reflecting an interrupt being sent to this hart that is now pending) must also be subject to fencing.



Adding on, CSR reads that can reflect external (to the hart) activity

PR #723, opened 2021-08-28
Replace "EEI" with "execution environment"

jhauser-us:
The term *execution environment interface* (EEI) appears to be intended to refer to the most applicable "binary interface" standard for the execution environment, such as the Linux RISC-V ABI or the RISC-V SBI.  However, such standards never fully specify every detail that is relevant to determining RISC-V instruction execution.  In particular, while a specific version of the Linux RISC-V ABI may require that certain features are implemented and others are not implemented, there are always some optional features or behaviors, or ISA extensions, about which the ABI is ambivalent.  As a consequence, in almost every instance where the base ISA manual says that something is specified by the EEI, it should really say, "specified by the EEI, or, if not, then determined by the underlying execution environment".

This construction can be simplified to just "determined by the execution environment".  Hence, this pull request drops the term *execution environment interface* and its abbreviation *EEI* and replaces them everywhere by just *execution environment*.  And because the term *EEI* has been used inconsistently throughout the manuals, this change also serves to consistently settle on *execution environment* instead.

In Section 1.2 ("RISC-V Software Execution Environments and Harts"), the existence of standards for RISC-V execution environments is still noted.

### DIFF
--- a/src/f.tex
+++ b/src/f.tex
@@ -379,7 +379,8 @@ is naturally aligned.
 FLW and FSW do not modify the bits being transferred; in particular, the
 payloads of non-canonical NaNs are preserved.
 
-As described in Section~\ref{sec:rv32:ldst}, the EEI defines whether
+As described in Section~\ref{sec:rv32:ldst},
+the execution environment defines whether
 misaligned floating-point loads and stores are handled invisibly or raise
 a contained or fatal trap.
 
--- a/src/intro.tex
+++ b/src/intro.tex
@@ -116,14 +116,16 @@ isolation between subsystems.
 \section{RISC-V Software Execution Environments and Harts}
 
 The behavior of a RISC-V program depends on the execution environment
-in which it runs.  A RISC-V execution environment interface (EEI)
+in which it runs.  The execution environment
 defines the initial state of the program, the number and type of harts
 in the environment including the privilege modes supported by the
 harts, the accessibility and attributes of memory and I/O regions, the
 behavior of all legal instructions executed on each hart (i.e., the
-ISA is one component of the EEI), and the handling of any interrupts
+ISA is one component of the execution environment),
+and the handling of any interrupts
 or exceptions raised during execution including environment calls.
-Examples of EEIs include the Linux application binary interface (ABI),
+An execution environment may conform to a common standard,
+such as the Linux application binary interface (ABI)
 or the RISC-V supervisor binary interface (SBI).  The implementation
 of a RISC-V execution environment can be pure hardware, pure software,
 or a combination of hardware and software.  For example, opcode traps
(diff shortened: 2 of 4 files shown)


aswaterman:
Yeah, this is a valid point.  I'm going to merge as-is, but will encourage @kasanovic to take a look asynchronously.

David-Horner:
I found the EEI abstraction to be of considerable value in 
distinguishing Execution Environments from  Interfaces.

The execution environment can potentially morph through execution, but 
the EEI can be more robust and accommodate the many changes in the 
environment.

Conversely the Interface can remain constant even when the supporting 
hardware and software layers evolve over generations, dynamically mutate 
with hardware/software configurations, or are substituted during execution.

The distinction between EE and EEI allows for such subtle discrimination 
and assignment of characteristics.

Further, as a coined phrase it has become for me an abstraction that has 
allowed me to think outside the current implementation limitations.

And lastly, the variously defined AEE, SEE and HEE do have their own 
EEIs or portions of the EEI.. More of this later.


Consider this excerpt from
Preface to Document Version 20190608-Base-Ratified in Volume I.


> Changed description of misaligned load and store behavior.*The 
> specification now allows visi-**
> **ble misaligned address traps in /execution environment interfaces,/ 
> rather than just mandating**
> **invisible handling of misaligned loads and stores in user mode.* 
> Also, now allows access-fault
> exceptions to be reported for misaligned accesses (including atomics) 
> that should not be
> emulated.

I suggest the *Interface *aspect of this notice clarifies the changes 
and highlights the value of considering an EEI as a distinct entity.


  \begin{commentary}
-  A bare hardware platform can be considered to define an EEI, where
+  A bare hardware platform determines a complete execution environment, where
    the accessible harts, memory, and other devices populate the
    environment, and the initial state is that at power-on reset.

DS Note: These are radically different.
     The original serves to help define what an EEI is.
     The revised text either defines "bare hardware platform" or implies an equivalence of the two.
        It raises the question "What is an incomplete execution environment?".

-  Generally, most software is designed to use a more abstract
-  interface to the hardware, as more abstract EEIs provide greater
-  portability across different hardware platforms.  Often EEIs are
-  layered on top of one another, where one higher-level EEI uses
-  another lower-level EEI.

DS Note:
The definition of EEI is continued above and gives the understanding that, as a concept EEI is a composite of many interface aspects.

+  Generally, most software is designed to conform to a more abstract
+  standard for a RISC-V execution environment, such as the Linux ABI,
+  to provide greater portability across different hardware platforms.
+  Often these execution environment standards are layered on top of
+  one another, with a higher-level standard constraining the generality
+  permitted by one or more lower-level standards.

DS Note:

Here the naming has shifted to "standards" for execution environments in their totality, with no distinction for the separate "Interface" aspects.
Further, the RISCV execution environment considerations are abdicated to to other ABIs,*such as *Linux,*but also, *of necessity, inclusive of all other ABIs with their conflicting requirements/characteristics,
Were we targetting only a limited set of current ABIs. then this might be adequate and manageable. But we are intending on build something that will evolve and outlast many of the current conventions while continuing to support all those remaining meaningful.


On 2021-08-28 6:07 p.m., John Hauser wrote:
>
> The term /execution environment interface/ (EEI) appears to be 
> intended to refer to the most applicable "binary interface" standard 
> for the execution environment, such as the Linux RISC-V ABI or the 
> RISC-V SBI.
>
As I tried to explain above, I believe this is a misrepresentation of 
what EEI is intended to mean. It is certainly not what I understand it 
to mean.
>
> However, such standards never fully specify every detail that is 
> relevant to determining RISC-V instruction execution.
>
Agreed, you have made an assertion that fails. The EEI itself is 
extensible independent of the Linus Foundation or other standard bodies.

To the degree EEI needs to be defined for RISCV, specifically, it can 
augmented other standards and apply to areas outside their perview.,

> In particular, while a specific version of the Linux RISC-V ABI may 
> require that certain features are implemented and others are not 
> implemented, there are always some optional features or behaviors, or 
> ISA extensions, about which the ABI is ambivalent. As a consequence, 
> in almost every instance where the base ISA manual says that something 
> is specified by the EEI, it should really say, "specified by the EEI, 
> or, if not, then determined by the underlying execution environment".
>
I inherently disagree with this assertion. It follows from a flawed 
understanding of what EEI is, can be and is intended to be.

PR #725, opened 2021-08-29
Make mtime[cmp] and msip memory-mapped registers optional

aswaterman:


aswaterman:
We'll revisit this topic soon, but possibly in a different form.

PR #727, opened 2021-08-30
Add mseccfg and *envcfg CSRs

aswaterman:


jhauser-us:
The new text has:
> If U-mode is not supported, FIOM is hardwired to zero.

If U mode isn't implemented, I would say `menvcfg` doesn't exist.  This register's mandate is to control "certain characteristics of the execution environment for modes less privileged than M".

I have no other complaints at this time.

aswaterman:
s/complaints/comments/ right? :P

Anyway, you’re absolutely right.

scottj97:
Does `senvcfg` affect VU-mode also? I can imagine a virtualized guest OS in VS-mode setting it.

jhauser-us:
> Does senvcfg affect VU-mode also?

Yes.  See the hypervisor extension chapter, Section 5.2, "Hypervisor and Virtual Supervisor CSRs":
> Some standard supervisor CSRs (...) have no matching VS CSR. These supervisor CSRs continue to have their usual function and accessibility even when V=1, except with VS-mode and VU-mode substituting for HS-mode and U-mode.

jhauser-us:
If S mode isn't implemented (but U mode is), we should allow bit FIOM to be hardwired to zero.

aswaterman:
Ack.

jhauser-us:
We should probably also allow FIOM to be hardwired to zero if S mode is implemented but only Bare address translation is supported.

aswaterman:
Indeed.

aswaterman:
I made that remark for both M and S modes, which I think is the right thing to do.

jhauser-us:
FWIW, I think this pull request is ready to merge.

aswaterman:
If only to reduce the complexity of juggling multiple outstanding pull requests, I'm going to merge it now.  As usual, that's not an indication that further criticism is unwelcome.

src/priv-csrs.tex
```diff
@@ -197,6 +197,10 @@ \section{CSR Listing}
 \tt 0x105 & SRW  &\tt stvec      & Supervisor trap handler base address. \\
 \tt 0x106 & SRW  &\tt scounteren & Supervisor counter enable. \\
 \hline
+\multicolumn{4}{|c|}{Supervisor Configuration} \\
+\hline
+\tt 0x10A & MRW  &\tt senvcfg    & Supervisor environment configuration register. \\
```
scottj97, line 202: Shouldn't the privilege here be SRW, not MRW?

src/priv-csrs.tex
```diff
@@ -243,6 +247,11 @@ \section{CSR Listing}
 \tt 0x64A & HRW  &\tt htinst     & Hypervisor trap instruction (transformed). \\
 \tt 0xE12 & HRO  &\tt hgeip      & Hypervisor guest external interrupt pending. \\
 \hline
+\multicolumn{4}{|c|}{Hypervisor Configuration} \\
+\hline
+\tt 0x60A & MRW  &\tt henvcfg    & Hypervisor environment configuration register. \\
+\tt 0x61A & MRW  &\tt henvcfgh   & Additional hypervisor env. conf. register, RV32 only. \\
```
scottj97, line 253: Shouldn't these two be HRW, not MRW?

aswaterman: Yes, copy-paste screwup.

PR #728, opened 2021-09-01
Clarify widths of privileged CSRs

jhauser-us:
When different privilege modes have different values for XLEN, the terms _RV32_ and _RV64_ are not sufficient to indicate the width of a privileged CSR that is accessed from a more-privileged mode (e.g. when 32-bit `vsatp` is accessed from 64-bit HS mode).  Where needed, replace "RV32" and "RV64" with more specific references to the correct XLEN.


aswaterman:
Thank you!

PR #764, opened 2021-11-03
remove word 'original' as this is confusing

axel-h:
The value in 'rs2' is not modified by the operation, thus the word "original" is confusing and implies there could be something besides the original value.

### DIFF
--- a/src/a.tex
+++ b/src/a.tex
@@ -428,14 +428,28 @@ AMOMIN[U].W/D & \multicolumn{2}{c}{ordering} & src & addr & width & dest & AMO  
 
 \vspace{-0.1in} The atomic memory operation (AMO) instructions perform
 read-modify-write operations for multiprocessor synchronization and
-are encoded with an R-type instruction format.  These AMO instructions
-atomically load a data value from the address in {\em rs1}, place the
-value into register {\em rd}, apply a binary operator to the loaded
-value and the original value in {\em rs2}, then store the result back
-to the original address in {\em rs1}. AMOs can either operate on 64-bit (RV64
-only) or 32-bit words in memory.  For RV64, 32-bit AMOs always
-sign-extend the value placed in {\em rd}, and ignore the upper 32 bits
-of the original value of {\em rs2}.
+are encoded with an R-type instruction format. A data value is atomically
+loaded from the address in {\em rs1} and placed into {\em rd}, then a
+binary operator is applied to the loaded value and the original value in
+{\em rs2}, finally the result is stored back to the original address.
+
+\begin{commentary}
+If {\em rs1} and {\em rd} refer to the same register, the address value
+in there is overwritten with the data value read from this address.
+However, this does not affect the address used for the final write
+operation, as AMOs always operate on a single memory address only.
+
+If {\em rs2} and {\em rd} refer to the same register, the original
+value in the register is used for the binary operation, before the
+register content is overwritten with the original memory content.
+
+At the and of the AMO, {\em rd} will hold the original memory content,
+unless it refers to {\tt x0}, then the original content is lost.
+\end{commentary}
+
+AMOs can either operate on 64-bit (RV64 only) or 32-bit words in memory.
+For RV64, 32-bit AMOs always sign-extend the value placed in {\em rd},
+and ignore the upper 32 bits of the original value of {\em rs2}.
 
 For AMOs, the A extension requires that the address held in {\em rs1}
 be naturally aligned to the size of the operand (i.e., eight-byte
--- a/src/a.tex
+++ b/src/a.tex
@@ -255,7 +255,7 @@ with both bits clear, but may result in lower performance.
 \begin{figure}[h!]
 \begin{center}
 \begin{verbatim}
-        # a0 holds address of memory location 
+        # a0 holds address of memory location
         # a1 holds expected value
         # a2 holds desired value
         # a0 holds return value, 0 if successful, !0 otherwise


scottj97:
Shouldn't line 438 have the same change applied?

pdonahue-ventana:
Isn't the word "original" necessary in case rs1 and rd are the same register?

axel-h:
Thanks, that is the thing I have been missing. I will reword things to make this more clear.

scottj97:
NACK on af87f42ea3badf1bce8de34d9110938a656fcfba. The rewording is unnecessary, overly verbose, and full of grammar errors/typos. I don't see anything confusing about the original.

aswaterman:
I'm also fine with the original.

axel-h:
Thanks for the feedback. I've added a commentary section now to explain things a bit more in detail and clarify the behavior for the non-trivial case. 
Since I'm not a native speaker, any help on grammar/spelling is appreciated.

src/a.tex
```diff
@@ -428,14 +428,28 @@ \section{Atomic Memory Operations}
...
-value into register {\em rd}, apply a binary operator to the loaded
-value and the original value in {\em rs2}, then store the result back
-to the original address in {\em rs1}. AMOs can either operate on 64-bit (RV64
-only) or 32-bit words in memory.  For RV64, 32-bit AMOs always
-sign-extend the value placed in {\em rd}, and ignore the upper 32 bits
-of the original value of {\em rs2}.
+are encoded with an R-type instruction format. A data value is atomically
+loaded from the address in {\em rs1} and placed into {\em rd}, then a
+binary operator is applied to the loaded value and the original value in
+{\em rs2}, finally the result is stored back to the original address.
```
scottj97, line 434: I can't see how this rewording clarifies anything.

src/a.tex
```diff
@@ -428,14 +428,28 @@ \section{Atomic Memory Operations}
...
+
+\begin{commentary}
+If {\em rs1} and {\em rd} refer to the same register, the address value
+in there is overwritten with the data value read from this address.
+However, this does not affect the address used for the final write
+operation, as AMOs always operate on a single memory address only.
```
scottj97, line 440: This paragraph can be reduced to just a few words: "original address in rs1".

src/a.tex
```diff
@@ -428,14 +428,28 @@ \section{Atomic Memory Operations}
...
+
+If {\em rs2} and {\em rd} refer to the same register, the original
+value in the register is used for the binary operation, before the
+register content is overwritten with the original memory content.
```
scottj97, line 444: This paragraph can be reduced to just a few words: "original value in rs2".

src/a.tex
```diff
@@ -428,14 +428,28 @@ \section{Atomic Memory Operations}
...
+
+At the and of the AMO, {\em rd} will hold the original memory content,
+unless it refers to {\tt x0}, then the original content is lost.
+\end{commentary}
```
scottj97, line 448: This is true of any operation with x0 as a destination. No need to repeat this explanation on every single instruction.

scottj97 (changes requested):
NACK 666ffcc for the same reasons.

src/a.tex
axel-h: Well, this is the whole thing this PR is about. Unless you are are RISC-V expert, this is nothing really trivial. I think the docs should also help non-experts.

scottj97: The spec assumes you understand how CPUs generally work. In my review here, I'm trying to explain how the original wording was already complete and precise. If the spec had to spell out all these possibilities in such painstaking detail, it would be 4000 pages long. Brevity is important.

axel-h: I disagree here, the spec is also something used by average developers and students to just look up things. However, I can see your point. Seems there is a certain need to to versions then:
* a "pure" specs, that is technically unambiguous and ideally formalized. It can be used as input for tools then also.
* the verbose RISC-V Handbook" for humans. 

Note that I've added my comment in a commentary section on purpose, it's not pure specs but useful background information for human seeking help.

src/a.tex
axel-h: Point taken, will drop this change

src/a.tex
nick-knight: This is a long-standing debate over the RISC-V ISA Manual. Some argue for more commentary, background, rationale, etc., and others want to cut all the fat. I acknowledge both sides. All I will say here is my observation that the community strongly resists to making changes in this document, especially ratified portions, without a clear argument that they address a defect. In this case, I don't see a convincing argument that there exists a defect.

I'd also like to respond to your respond to your "two versions" comment:

- In case you aren't aware, there is an ongoing effort, [RISC-V Sail Model](https://github.com/riscv/sail-riscv), to express the RISC-V in a formal language. (Some might also argue that the [Spike RISC-V ISA simulator](https://github.com/riscv-software-src/riscv-isa-sim) technically qualifies as a formal-language specification.)

- Patterson-Waterman also wrote a small book, [The RISC-V Reader](http://riscvbook.com/), which is intended to serve as a friendlier introduction. I believe they are in the process of producing a second edition.

It's tempting to propose an open, crowd-sourced "RISC-V Handbook", where detailed commentary like in this PR would certainly be welcome. But this repository is the wrong venue to pursue that.

axel-h: Thanks for the feedback. I'm closing this PR based on the comments. In case there is an option to contribute to [The RISC-V Reader](http://riscvbook.com) I'd be willing to contribute there.

allenjbaum:
I think Nick is right on the money.
The spec is different than a book that students use to understand
the corners of the architecture.
The Sail model is supposed to be the golden-model, since English can be so
ambiguous even when you're trying to be exact.
 Andrew & Dave's book came to mind and of course the newer versions of
Hennessy and Patterson (Patterson and Hennessy? I've lost track) uses
RISC-V as examples.
That is a bit different than focusing on RISC-V though; that's focusing on
concept and RISC-V is incidental.
I know a college professor who is working on a book on RISC-V processor
design for undergraduates, and that might be what you're looking for.

PR #766, opened 2021-11-09
Add draft of RNMI spec

aswaterman:
This isn't yet complete, and in particular the inability to directly manipulate the `rnmie` bit may be a deficiency.  (As it stands, it's impossible to perform the standard pattern of save state; enable interrupts; do something; disable interrupts; restore state.  The open question is whether or not this is important to provide for RNMIs, vs. always requiring RNMI handlers to execute with RNMIs masked.)

### DIFF
--- a/src/machine.tex
+++ b/src/machine.tex
@@ -2884,6 +2884,8 @@ Writable PMP registers' A and L fields are set to 0, unless the platform
 mandates a different reset value for some PMP registers' A and L fields.
 If the hypervisor extension is implemented, the {\tt hgatp}.MODE and
 {\tt vsatp}.MODE fields are reset to 0.
+If the Smrnmi extension is implemented, the {\tt mnstatus}.NMIE field
+is reset to 0.
 No \warl\ field contains an illegal value.
 All other hart state is \unspecified.
 
--- a/src/priv-csrs.tex
+++ b/src/priv-csrs.tex
@@ -358,6 +358,13 @@ Number    & Privilege & Name & Description \\
 \hline
 Number    & Privilege & Name & Description \\
 \hline
+\multicolumn{4}{|c|}{Machine Non-Maskable Interrupt Handling} \\
+\hline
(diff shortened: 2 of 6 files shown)


allenjbaum:
Does the RNMI handlre have to be able to distinguish between U and VU, S
andd VS modes (which would require 3bits of mode), or can the V bit just be
retained from the source of the interrupt?
This leaves unsaid wheter all the other bit in mstatus woud just be used by
nmstatus.

Also, you say "Traps in the RNMI handler can only be resumed if they occur
while the handler was servicing an interrupt that occurred outside of
machine-mode."
which could ve interpreted as
   "Traps in the RNMI handler can only be resumed if they occur while the
handler was servicing *(*an interrupt that occurred outside of machine-mode
*)*."
or
  "Traps in the RNMI handler can only be resumed if *(*they occur while the
handler was servicing an interrupt that occurred) outside of machine-mode)."

The former sound more obvious,( and grammatica)l, but  you have to think a
bit to determine whether the exception source was outside of machine mode,
or that the handler is outside of machine mode. It's kind of a long
sentence, and breaking it up may make it more clear.

aswaterman:
@allenjbaum This spec is indeed missing an `mnstatus.MPV` field (it was written not considering the H extension).

aswaterman:
It is needed; using hstatus.SPV doesn't work, either, because the NMI might occur during HS-mode trap entry, before the hypervisor has had a chance to save that field.

pdonahue-ventana:
Is mnstatus.NMIE a global enable or a local enable for purposes of deciding whether a WFI is allowed to sleep?

aswaterman:
It's meant to be analogous to mstatus.MIE (i.e., global).

aswaterman:
Manually merged via https://github.com/riscv/riscv-isa-manual/commit/869dcc608e11f9680e950bcb20a9b8294d2b82bd

PR #768, opened 2021-11-11
Rewrite most instances of "hardwire" as "read-only"

jhauser-us:


aswaterman:
For others' context, the rationale is that "hardwired" leads to the impression that the value is immutable for all time.  In actuality, the execution environment might be partially responsible for setting the constant value.  So, while the expectation should be that the value is constant for an entire process's (or OS's) lifetime, "hardwired" still lends a stronger impression than is appropriate.

PR #770, opened 2021-11-12
Rename hstatus.HU

jhauser-us:


aswaterman:
For others' context, there was a concern that "Hypervisor User mode" sounded too much like a new privilege mode.  This tweak makes it pretty clear that's not the case.

PR #771, opened 2021-11-12
CSR mideleg masks hideleg, hip, and hie

jhauser-us:


scottj97:
@jhauser-us please provide some context here, for those of us watching this repo for spec updates. If this has been discussed elsewhere already please provide a link here in the PR.

In this case, is this a semantic change, or just a clarification of already-specified behavior?

jhauser-us:
This change clarifies behavior that had always been intended.

Zero bits in `mideleg` cause the corresponding bits in `sip` and `sie` to be masked as read-only zeros, and they once also masked bits in `sideleg`, before `sideleg` was deleted from the document along with the never-completed N extension for user-level traps.  Likewise, zero bits in `hideleg` cause the corresponding bits of `vsip` and `vsie` to be masked.

In the same way, it was always intended that zero bits in `mideleg` would mask the corresponding bits in `hideleg`, `hip`, and `hie` (and the draft Advanced Interrupt Architecture for RISC-V assumes this is the case); however, public review of the hypervisor extension made clear that we forgot to say so in the document.

scottj97:
What about `hvip`? Among bits 15:0, only bits 10, 6, and 2 are writable, and those bits are hardwired to 1 in mideleg, so this doesn't apply. But what about the nonstandard bits above bit 15?

scottj97:
Should `hedeleg` be masked by `medeleg`?

jhauser-us:
Scott Johnson asked:
> What about `hvip`?

Good question, but `hvip` is intentionally not masked by `mideleg`.

The draft Advanced Interrupt Architecture (AIA) for RISC-V uses `hvip` as part of a mechanism to inject a subset of "local" interrupts into a virtual machine (VS mode) when the interrupts are not delegated from M mode, i.e., when the corresponding bits in `mideleg` are 0.  Hence, to maximize forward compatibility of software with the AIA, it's best that we not require that `mideleg` mask `hvip`.  We could, but the AIA would then need to override that requirement.

As I read the document, if any of the custom/platform-designated bits of `mvip` are writable, the effects of setting those bits must be implementation/platform-defined, which causes no problems.

jhauser-us:
Scott Johnson asked:
> Should `hedeleg` be masked by `medeleg`?

Another good question, but no, absolutely not.  Note the table of required writable/read-only bits for `hedeleg` in the hypervisor chapter, in the subsection titied "Hypervisor Trap Delegation Registers (hedeleg and hideleg)".

The way that zero bits in `mideleg` etc. cause the interrupt not to exist at lesser privilege levels is a precedent that was set early on, and we're sticking with it for consistency in the base privileged ISA.  The Advanced Interrupt Architecture extends the ISA to provide a bypass of this precedent for custom and "local" interrupts.

A corresponding precedent was never established for `medeleg` etc., and I won't willingly start one now, as it would just add extra steps for handling some traps.

aswaterman:
This looks ready to merge, but feel free to continue the discussion here if need be.

scottj97:
Is my understanding correct that this change is intended to affect only the nonstandard bits above bit 15?

Because the only bits that are defined for `hip`, `hie`, or `hideleg` in the standard bits (15:0) are 12, 10, 6, and 2, and those 4 are hardwired to 1 in `mideleg`.

But something's fishy...`hideleg` bit 12 is read-only 0 (before and after #772); therefore `hip` and `hie` bit 12 (defined as `sgeip`/`sgeie`) must always be read-only 0?

jhauser-us:
Scott Johnson wrote:
> Is my understanding correct that this change is intended to affect only the nonstandard bits above bit 15?

Yes, plus presumably any additional bits that extensions might define among the standard ones, 15:0.  (Although no additions within 15:0 are expected, with or without the future Advanced Interrupt Architecture.)

scottj97:
So what should happen if a bit in `mideleg` is set, then the corresponding bit in `hideleg` is set, then `mideleg` is cleared, then set again?

E.g.:
```
csrw mideleg, 1<<13
csrw hideleg, 1<<13
csrw mideleg, 0
csrr x4,hideleg // must read 0
csrw mideleg,1<<13
csrr x4,hideleg // okay to read 1<<13?
```

Is it okay if the previously-set bit reappears in `hideleg`? Or must it be 0 at the end of this sequence?

I'm trying to determine if the implementation must write `hideleg` any time `mideleg` is written with 0, or if masking `hideleg` on read is sufficient.

And similar question if I try to write to `hideleg` while `mideleg` is 0, e.g.:
```
csrw hideleg,0
csrw mideleg,0
csrw hideleg,1<<13
csrr x4,hideleg // must be 0
csrw mideleg,1<<13
csrr x4,hideleg // okay to read 1<<13?
```

This will tell me if the implementation must mask `hideleg` on every write too.

jhauser-us:
Scott Johnson asked:
> So what should happen if a bit in `mideleg` is set, then the corresponding bit in `hideleg` is set, then `mideleg` is cleared, then set again?

It's been a longstanding issue with the document that it doesn't answer questions like this.  As it happens, we are in the middle of patching that problem right now.  The spec update should be coming within a week, tops.

jameshippisley:
Given that rules for these CSR legal value inter-dependencies are proving difficult to agree on, it occurs to me that one way to mitigate the problem is not to introduce the inter-dependencies in the architecture in the first place, except where they are necessary. So I want to double check, is there a strong requirement to make hideleg writable bits be masked by mideleg?

I'm imagining a hypothetical scenario, not sure whether this is realistic, but it seems possible: Machine mode temporarily disables delegation of a custom interrupt to the hypervisor. Hypervisor tests what interrupts are delegable to guest by writing a 1 to the custom interrupt hideleg bit in question. It reads back 0, and concludes that this interrupt is not delegable. Later, machine mode changes the interrupt back to delegable, and as a side effect, this actually delegates the interrupt in hideleg because the trial write of the value 1 is now visible. This seems like bad behavior, but it would be the correct behavior under one proposed csr r/w inter-dependency scheme.

If there was no inter-dependency of hideleg writable bits on mideleg, then when hypervisor writes the bit in hideleg it would learn that the interrupt is delegable, and can do what it wants with this information. It does not learn that the interrupt is in fact delegated from machine mode, it will have to figure this out by some other mechanism. But the hypervisor knows it has to do that, because that is how the architecture was written. I wonder if this would be a preferable situation, compared to trying to come up with reasonable and universal rules for how changes in mideleg and hideleg are supposed to propagate with respect to each other?

JamesKenneyImperas:
If any bit is masked then the rule elsewhere is that writes are _ignored_ and reads return zero, so your scenario shouldn't be possible, I think.

scottj97:
> If any bit is masked then the rule elsewhere is that writes are _ignored_ and reads return zero, so your scenario shouldn't be possible, I think.

@jameshippisley's scenario describes one of the three proposals [being discussed](https://lists.riscv.org/g/tech-privileged/topic/87022963) on the tech-privileged mailing list.

JamesKenneyImperas:
Thanks for the pointer, @scottj97 - I wasn't aware of that discussion.

allenjbaum:
That may be a bit nuanced. The _appearance_ is that the writes are ignored.
But if the modifier field is restored and causes the field to be unmasked,
 - do writes made while the field is masked show the last write attempt, or
 - does it show the last unmasked write value, or
 - can it be any legal value (including zero, the masked value, obviously)?
-----------------
In general, for fields dependent upon modifiers in other (CSR) fields,
there are two primary scenarios, and one more complicated:
 - one is the straight masking case, where the field is masked as if it
doesn't exist; as described, it appears as if it is read-only zero.
 - the other is where the field is still writable, but the values that are
legal changes.

The first scenario is the case that the proposed change was addressing; it
is a bit simpler and more common than the second.
I still believe the simplest implementation in this case  - and what John
proposed, is output masking only.
That is, writes are performed to the underlying state as if it were
unmasked (no logic changes), but will have no effect on the output. This is
output masking
Inhibiting the writes while masked isn't much more work, but still requires
output masking (sticky bits are an example of this, but with a masked value
of the bit itself)
The third case (forcing the field state to the masked value) is the input
masking case (

Asserting "freedom of implementation" as reason to not define a behavior is
a completely bogus argument (IMHO, of course)
There are a handful of gates difference between these behaviors (and not
even handfulls plural).
There is no cost to mandating one or the other; but there is a cost for
mandating none of them.

The second scenario (the field remains writable and changes are visible) is
the one I brought up.
It is a little more complex, but only because the masking logic is more
complicated (the masking function isn't a constant)
Whether writes are inhabited and whether it is input or output mapped
doesn't seem like it make any substantial difference in implementation
There may even be cases that require both input and put masking, and that
should (must?) be avoided

The elephant in the room is backwards compatibility, of course, which would
lead to leaving the behavior implementation defined.
That may be necessary and unfortunate.

I believe that future extensions that have fields like this should mandate
some specific implementation and it really doesn't matter which; the cost
is zero to minimal, and there SW benefits.
From a spec point of view, mandate one specific default behavior, and if
the extension really cares, specify the specific behavior desried in the
extension spec.

scottj97:
> It's been a longstanding issue with the document that it doesn't answer questions like this. As it happens, we are in the middle of patching that problem right now. The spec update should be coming within a week, tops.

See #782

jhauser-us:
@jameshippisley wrote:
> So I want to double check, is there a strong requirement to make hideleg writable bits be masked by mideleg?

By the time you asked this question (Nov 19), it's my understanding the hypervisor extension had already been approved by the RISC-V Board of Directors (Nov 18).  At this point, dropping `mideleg`'s masking of `hideleg` wouldn't be just a clarification but a normative change.

Might the spec have been slightly better without the masking?  I'm not sure.  It certainly wouldn't have made a significant difference to hardware, considering how the unmasking question got resolved (an unmasked value becomes unspecified).

> Later, machine mode changes the interrupt back to delegable, and as a side effect, this actually delegates the interrupt in hideleg because the trial write of the value 1 is now visible.

I can't think of a reason for machine level to be regularly changing the delegation of major interrupts, but assuming it does, in the scenario you describe, either it needs to let the OS that's running in HS mode know, or it needs to zero that delegation bit in `hideleg` itself after setting it in `mideleg`.  (Or both.)

PR #776, opened 2021-11-15
Memory access traps may write zero to stval

jhauser-us:
Account for the fact that, once again, memory access traps may write zero to `stval`.


aswaterman:
Good catch.

PR #777, opened 2021-11-15
Separate transformation for HLV instructions

jhauser-us:


aswaterman:
Just confirming that the intent here just to distinguish the fact that HLV has a funct5 field in place of HSV's rs2.

pdonahue-ventana:
Doesn't the definition of HLV/HLVX in section 8.3.1 still contain the rs2 field instead of funct5?

jhauser-us:
Paul Donahue wrote:
> Doesn't the definition of HLV/HLVX in section 8.3.1 still contain the rs2 field instead of funct5?

Good point.

Andrew, given that the encodings of a few other instructions (CSR and FCVT) are also shown as using the rs2 field to encode something other than a register operand, should we just back out of this commit entirely?  It's in response to a public comment that bits 24:20 were labeled "rs2" when they aren't actually a register for HLV/HLVX.  Yet there's precedent that field "rs2" isn't always a register.

aswaterman:
Agreed.  Reverted via 714c50aa294e0c56a2d5adc3b7559a6f200b3044.  If we ever decide to revisit this topic throughout the ISA, we should revisit this, too.

PR #780, opened 2021-11-19
Extension is "implemented", not "enabled" (in Svinval)

jhauser-us:


scottj97:
Guys, this Git repo -- or at least this GitHub repo -- needs to record some reason for why every change is made, if it's not plainly obvious. PRs with no explanation leave subscribers like me wondering why this change is happening. Was this discussed on a mailing list? Discussed in an offline conversation? Was there a problem that is being solved? Were alternatives discussed and rejected? The repo needs to have some kind of record of this. Is that an unreasonable request?

In this case, it sounds like a semantic change. Does this mean if I write misa.H=0 these two instructions should still work? If so, why are they different than `hfence.vvma` and `hfence.gvma` in that regard?

aswaterman:
No, it's not too much to ask.

The intent is to standardize on the concept of extensions either being implemented or not.  That is, there aren't three distinct states (unimplemented, implemented but disabled, implemented and enabled); the first two states are the same.  See https://github.com/riscv/riscv-isa-manual/blob/efac2a5c8f2c3992a584941011125e46b9f4959d/src/machine.tex#L101 -- this isn't supposed to be a semantic change.

aswaterman:
@jhauser-us despite the time crunch, can you make an effort to provide some context/motivation for these PRs (even something extremely terse)?  I'll try to do better to fill in the details if missing and/or act as the enforcer before merging.

scottj97:
I understand, and agree that's a useful goal, but I read the quoted line differently. It uses the term "disabled", which makes the extension operate "as if" it were not implemented. It doesn't make it unimplemented.

The three states you mention all actually exist. We can say the first two must behave the same but we can't say there aren't three distinct states. (Though perhaps substitute something else for the word "implemented".) Of the three states, only the third can be interpreted as "enabled", while the term "implemented" covers both the second and third states. So IMO the original text here is clearer than the replacement.

aswaterman:
(I'll let John take it from here)

gfavor:
On Fri, Nov 19, 2021 at 3:57 PM Scott Johnson ***@***.***>
wrote:

> I understand, and agree that's a useful goal, but I read the quoted line
> differently. It uses the term "disabled", which makes the extension operate
> "as if" it were not implemented. It doesn't make it unimplemented.
>
I think from an architectural perspective (which is all that the arch specs
are concerned with specifying), a physically unimplemented extension and a
physically implemented but disabled extension are equivalent.  It's only
when one starts talking about microarchitectural ideas (like
enabling/disabling a physically implemented extension) that the
unimplemented versus disabled distinction comes into play.

But again, John is the better one to respond.  It may be that a better word
than "unimplemented" is needed, or that word with an adjective in front of
it is needed, to avoid unintended interpretations of the term.  Or there
just needs to be an explicit "glossary-style" definition of the term that
establishes an unambiguous intended meaning.

Greg

scottj97:
But enabling/disabling is specified in the architectural spec (under `misa`)! Right where Andrew quoted it above. It's not a microarchitectural detail at all.

In this case, "enabled" already has that unambiguous, intended meaning, since it obviously cannot be enabled unless it is physically implemented.

jhauser-us:
I have long argued that "implemented but disabled" is a paradox in the standard.  If a feature is *implemented*, it necessarily must satisfy the rules that the standard defines for it.  If a feature is *disabled* it clearly doesn't do that, so then it can't be *implemented* from the perspective of the standard.

I encourage you to search for how the documents regularly use the word "implemented".  Here are some examples:

> If the C extension is **implemented**, the appropriate compressed floating-point load and store instructions must be provided whenever the relevant standard floating-point extension (F and/or D) is also **implemented**.

> *x*PP fields are **WARL** fields that can hold only privilege mode *x* and any **implemented** privilege mode lower than *x*.  If privilege mode *x* is not **implemented**, then *x*PP must be read-only 0.

> When S-mode is **implemented**, then executing WFI in U-mode causes an illegal instruction exception, unless it completes within an implementation-specific, bounded time limit.

> If the F extension is **implemented**, the FS field shall not be read-only zero.

> If Svnapot is not **implemented**, bit 63 remains reserved and must be zeroed by software for forward compatibility, or else a page-fault exception is raised.

It doesn't say "implemented and enabled" or "enabled".  It says "implemented".

There are some exceptions, and the description of `misa` is one of those places.  However, there is this sentence:

> When a standard extension is disabled by clearing its bit in `misa`, the instructions and CSRs defined or modified by the extension revert to their defined or reserved behaviors as if the extension is not implemented.

That sentence by itself provides all the connection we need between bits of `misa` and the concept of implemented extensions.

I understand that everyone wants to use the word *implemented* to mean that the transistors are on the die, or the lines of code exist in the program, even if that "implementation" is totally dormant because it's disabled.  But again, for the standard, a dormant/disabled implementation doesn't do what an implementation is required to do, so it's not actually an implementation.  It may as well be a rock.

Anyway, if the community's consensus is that the term *implemented* shall also encompass "disabled implementations", then the documents need to be consistently updated to switch from saying "implemented" and say instead "enabled" or "implemented and enabled".

The purpose of this pull request was to make this one sentence more consistent with how the rest of the document is currently written.

scottj97:
Thank you for the explanation. I understand the confusion about the terminology and I agree that the spec is inconsistent. But I don't think this is an improvement.

The line about `misa` does not say that setting `misa.H=0` makes the hypervisor extension unimplemented. It says it makes it behave _as if_ it were unimplemented. If hypervisor is "implemented but disabled" then it's still following all the rules that the standard defines for hypervisor -- including the rule about `misa.H`.

I suspect many of the spec's uses of the word "implemented" were written without due consideration of the possibility of those extensions being disabled.

(And I apologize for asking you to spend so much time on a seemingly trivial change, but precision is important in a spec like this.)

aswaterman:
Thanks, @jhauser-us.

allenjbaum:
Hmm, this does bear on the issue of WARL fields that are masked, and then
unmasked, where the masking bit is a MISA bit.
I won't count the case where "masking" is really removing an entire CSR (so
that accesses to it trap).

But if a field of an existing CSR depends on a misa bit, and that masking
misa bit transitions 1->0->1, the spec implies that the field is in the
reset state:
    an undefined but legal value (unless otherwise defined by the extension
or by a platform spec).
Which is a pity, but it is what is is.
IF the modifier bit is outside misa, then the other discussion is still
relevant.

Does that sound correct?

PR #782, opened 2021-11-25
Clarify effect of changing the set of legal values for a CSR field

jhauser-us:


src/priv-csrs.tex
```diff
@@ -471,6 +471,42 @@ \subsection*{Write Any Values, Reads Legal Values (WARL)}
 deterministically depend on the illegal written value and
 the architectural state of the hart.
 
+\section{CSR Field Modulation}
+
+If a write to one CSR changes the set of legal values allowed for a
+field of a second CSR, then unless specified otherwise, the second
+CSR's field immediately gets an unspecified value from among its new
+legal values.
+This is true even if the field's value before the write reamins legal
```
pdonahue-ventana, line 480:
```suggestion
This is true even if the field's value before the write remains legal
```

src/priv-csrs.tex
```diff
@@ -471,6 +471,42 @@ \subsection*{Write Any Values, Reads Legal Values (WARL)}
 deterministically depend on the illegal written value and
 the architectural state of the hart.
 
+\section{CSR Field Modulation}
+
+If a write to one CSR changes the set of legal values allowed for a
+field of a second CSR, then unless specified otherwise, the second
+CSR's field immediately gets an unspecified value from among its new
```
pdonahue-ventana, line 478: Do you want `\unspecified` here?

src/priv-csrs.tex
```diff
@@ -471,6 +471,42 @@ \subsection*{Write Any Values, Reads Legal Values (WARL)}
...
+the value of the field may be changed in consequence of the write to
+the controlling CSR.
+
+\begin{commentary}
+As a special case of this rule, the value written to one CSR may
+control whether a field of a second CSR is writable (with multiple
+legal values) or is read-only.
+When a write to the controlling CSR causes the second CSR's field
+to change from previously read-only to now writable, that field
+immediately gets an unspecified but legal value, unless specified
```
pdonahue-ventana, line 491: `\unspecified`?

src/priv-csrs.tex
```diff
@@ -471,6 +471,42 @@ \subsection*{Write Any Values, Reads Legal Values (WARL)}
...
+\end{commentary}
+
+\begin{commentary}
+Some CSR fields are, when writable, defined as aliases of other CSR
+fields.
+Let $x$ be such a CSR field, and let $y$ be the CSR field it aliases
+when writable.
+If a write to a controlling CSR causes field $x$ to change from
+previously read-only to now writable, the new value of $x$ is not
+unspecified but instead immediately reflects the existing value of its
```
pdonahue-ventana, line 502: `\unspecified`?

scottj97:
[Relevant discussion](https://lists.riscv.org/g/tech-privileged/topic/87022963)

allenjbaum:
Just an observation: if a controlling field changes the legal value such
there is only one legal value, that is defacto read-only, and it is
defacto specified.

allenjbaum:
Greg has convinced me that for fields that have legal values that are
dependent on some controlling field,
specifying the values after the controlling field (or fields, even) are
changed to be unspecified (but legal) removes the need for tests to know
the illegal->legal mapping behaviot.
We may want to check that it is legal (or not illegal), but not exactly
which legal value it is (with the exception that if there is only one legal
value, it is effectively read-only and we can check it.
One way to check that an implicit write (after a controlling bit(s) change)
is being done correctly is to read the field after changing the controlling
bit(s)
Then re-write that same value and re-read it again. IF the first read
doesn't match re-read, then the spec is being violated.

aswaterman:
> Just an observation: if a controlling field changes the legal value such there is only one legal value, that is defacto read-only, and it is defacto specified.

Indeed, this has always been the case.

PR #786, opened 2021-11-30
Rework the description of exceptions for Svinval

jhauser-us:
Restructure the listing of exceptions raised for SINVAL.VMA, HINVAL.VVMA, and HINVAL.GVMA, as follows:

  - List exceptions that occur absent the hypervisor extension first.

  - Replace obscure "when V=1" with "in VS-mode or VU-mode".

  - Add that attempts to execute SINVAL.VMA in VU-mode raise a virtual instruction exception.


aswaterman:
Yeah, this reads a little better.  Thanks.

PR #797, opened 2021-12-16
Hypervisor: use HFENCE.GVMA after PMP write to synchronize VS-enabled-G-disabled

ingallsj:
A HFENCE.GVMA is still required after changing PMPs affecting VS-stage translation when vsatp.mode=ON and hgatp.mode=BARE.

### DIFF
--- a/src/hypervisor.tex
+++ b/src/hypervisor.tex
@@ -2786,16 +2786,11 @@ the virtual memory system.
 For HS-level address translation, this is accomplished by executing in M-mode
 an SFENCE.VMA instruction with {\em rs1}={\tt x0} and {\em rs2}={\tt x0}, after
 the PMP CSRs are written.
-If G-stage or VS-stage address translation is in use and is not Bare,
-synchronization with its data
-structures is also needed.
-When PMP settings are modified in a manner that affects the physical
-memory that holds guest-physical page tables, the physical memory to which
-guest-physical page tables point, or when G-stage translation MODE is Bare,
-the physical memory that holds VS-stage page tables or the physical memory to
-which VS-stage page tables point, then an HFENCE.GVMA instruction with
-{\em rs1}={\tt x0} and {\em rs2}={\tt x0} must be executed in M-mode after the
-PMP CSRs are written.
+Synchronization with G-stage and VS-stage data structures is also needed.
+Executing an HFENCE.GVMA instruction with {\em rs1}={\tt x0} and
+{\em rs2}={\tt x0} suffices to flush all G-stage or VS-stage
+address-translation cache entries that have cached PMP settings
+corresponding to the final translated supervisor physical address.
 An HFENCE.VVMA instruction is not required.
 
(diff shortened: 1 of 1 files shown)


aswaterman:
Thinking about this some more, I think there is some additional subtlety.  Assuming vsatp.MODE=Sv39 and hgatp.MODE=Bare:
- If you're trying to enforce that the PMP settings corresponding to the final PPN are respected, then HFENCE.GVMA should suffice (because the properties of the GPA space have changed), as should HFENCE.VVMA (because it'll force a re-walk).
- If you're trying to enforce that the implicit accesses for the page-table walks observe the new PMP settings, then HFENCE.VVMA is needed for the VS-stage side, because otherwise the walks might've been performed early and cached.  HFENCE.GVMA doesn't suffice, because it isn't guaranteed to flush pure VS-stage address-translation caches.

scottj97:
> * If you're trying to enforce that the implicit accesses for the page-table walks observe the new PMP settings, then HFENCE.VVMA is needed for the VS-stage side, because otherwise the walks might've been performed early and cached.  HFENCE.GVMA doesn't suffice, because it isn't guaranteed to flush pure VS-stage address-translation caches.

If the cache is caching PMP settings then I would argue that it is no longer a "pure VS-stage address-translation cache". It's caching things that can change when G-stage translations change.

aswaterman:
I agree that if PMP information is being cached, that cache isn’t a pure VS-stage translation cache. I think HFENCE.GVMA is necessary and sufficient for PMP changes wrt. permissions changes to the final PPN.

In response to your edit, my second bullet meant to address that case. I think it’s necessary to execute an HFENCE.VVMA if the perms might’ve changed on the PTEs themselves and if software wants to force VS-stage walks’ PTE accesses to see the new perms. (Which might not be the common case… maybe software doesn’t care about speculative/early walks’ PTE accesses having seen the old permissions.)

aswaterman:
Flushing address-translation caches following a PMP change is an M-mode operation, and M-mode software can't (in general) know what address-translation mode the hypervisor is using, since the hypervisor can change the mode at runtime.  Pragmatically, that means M-mode software should be able to use the same instruction sequence for the G-off case as the G-on case.  This argues in favor of the PR's intent.

I don't think the PR is phrased quite right, though.  The second change seems to me to unintentionally change the semantics for the G-on, VS-on case: it suggests to me that HFENCE.GVMA also synchronizes wrt. the guest's page tables in the G-on case (unintentionally implying an HFENCE.VVMA).  Does the text need to be split explicitly into G-on and G-off cases to clarify what's going on here?

jhauser-us:
I believe I can live with the new version of this PR.

It seems to me what we really want to say is, after modifying some PMP CSRs, you need to execute a general HFENCE.GVMA unless it's certain not to matter.  Unfortunately, I can't think of a better way to nail down "certain not to matter" than the text that's currently proposed, messy though it is.

PR #802, opened 2021-12-28
add dummy GitHub action

cetola:
In order to test the new build of the Asciidoc using a GitHub action off
the "convert2adoc" branch, the master branch must have a GitHub action
with the same name.

Signed-off-by: Stephano Cetola <scetola@linuxfoundation.org>

### DIFF
--- a/.github/workflows/build-pdf.yml
+++ b/.github/workflows/build-pdf.yml
@@ -1,0 +1,9 @@
+name: Build Document PDF
+
+on:
+  workflow_dispatch:
+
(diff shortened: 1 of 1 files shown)


cetola:
See https://github.com/riscv/riscv-isa-manual/pull/801 for the GitHub action PR that actually does something, namely building the Unprivileged Spec PDF from the asciidoc source.

aswaterman:
The workaround is "interesting", but whatever works!

PR #804, opened 2021-12-29
Update the "Dummy GitHub Workflow" with a proper build workflow

cetola:
This action should only be run on a branch that contains Asciidoc and a makefile for building said Asciidoc. Today that is only the convert2adoc branch.

We want this workflow in the main branch so that any other branches created with the intent of converting LaTeX to Asciidoc can take advantage of GitHub actions for building.


cetola:
Will remove pointless merge commit and resubmit a PR. Sorry for the noise.

PR #805, opened 2021-12-29
Update GitHub action to build Asciidoc branch

cetola:
Currently there is a branch which contains the unprivileged
specification code in asciidoc format. This action should only be run on
that branch. Once the conversion of the privileged specification is
complete as well, this action can be updated to build the complete set
of specs.

Signed-off-by: Stephano Cetola <scetola@linuxfoundation.org>


aswaterman:
I can't say I understand why this needs to go on the master branch, but if it works, it's fine for now.

PR #809, opened 2022-01-07
Fix ordering in A.1 litmus test forbidden execution explanation

hasheddan:
Updates the second possible scenario in which a0=1 to state that (c)
appears before (b) would result in a0=1 but is forbidden due to
violation of the Load Value Axiom. The current description states that
a0=1 if (a) appears before (b) in global memory order, which would not
be the case as (b) stores the immediate 2 to the same address 0(s0)
after (a) stores the immediate 1.

Signed-off-by: hasheddan <georgedanielmangum@gmail.com>

cc @daniellustig @aswaterman 

This is based on my own understanding of the Load Value Axiom and the litmus test in question. Please feel free to close if my interpretation is incorrect.

### DIFF
--- a/src/memory.tex
+++ b/src/memory.tex
@@ -108,7 +108,7 @@ This notation is explained in Table~\ref{tab:litmus:key}.
 Of the listed relations, {\sf rf} edges between harts, {\sf co} edges, {\sf fr} edges, and {\sf ppo} edges directly constrain the global memory order (as do {\sf fence}, {\sf addr}, {\sf data}, and some {\sf ctrl} edges, via {\sf ppo}).
 Other edges (such as intra-hart {\sf rf} edges) are informative but do not constrain the global memory order.
 
-For example, in Figure~\ref{fig:litmus:sample}, {\tt a0=1} could occur only if one of the following were true:
+For example, in Figure~\ref{fig:litmus:sample}, {\tt a0=1} could occur only if (c) reads the value written by (a) and one of the following were true:
 \begin{itemize}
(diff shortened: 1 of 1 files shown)


daniellustig:
Thanks for the review @hasheddan.

There are many ways to think about these examples, but the intention of the original text was to show that 1) it's clear from the definition of GMO that either (a) is before (b) or (b) is before (a), and so 2) if both of those are ruled out, the proposed outcome is impossible.

If rewritten as proposed, it would IMHO be less clear that all possibilities have been ruled out.  Also, (b) and (c) would not be described has having a coherence relationship, because (c) is a store.  Does that make sense?

So I would leave this text as is.

hasheddan (commented):
@daniellustig thank you for following up! Great point about the coherence, I should have removed the parenthetical portion of the sentence with the change as proposed. I can see the motivation for exhausting possible outcomes with the two examples. The sentence preceding the two outcomes still is somewhat confusing to me as it sounds like the second option is saying that `(a)` preceding `(b)` in GMO by itself would result in `a0=1`. Would you be in support of a slight revision to the second option that changed it to read: "(a) appears before (b) in global memory order (and in the coherence order), and (c) reads the value written by (a)."?

Thanks again for your time and explanation.

daniellustig:
I would be fine with the new proposal that just adds `and (c) reads the value written by (a)`, except really that applies to both cases.  Perhaps it would be better to add it here: `a0=1 could occur only if (c) reads the value written by (a) and one of the following were true`.

hasheddan:
@daniellustig sounds reasonable to me. Updates made in latest patch. Thanks again.

daniellustig:
LGTM

aswaterman:
Thanks, y'all.

PR #810, opened 2022-01-10
Add Zihintntl spec

aswaterman:
Non-temporal locality hints.  See text.

### DIFF
--- a/src/zihintntl.tex
+++ b/src/zihintntl.tex
@@ -10,17 +10,22 @@ Three variants are provided:
 The NTL.L1 instruction indicates that the target instruction
 does not exhibit temporal locality within the capacity of the innermost level
 of cache in the memory hierarchy (e.g., an L1 cache).
-NTL.L1 is encoded as \mbox{ADD {\em x0, x0, x1}}.
+NTL.L1 is encoded as \mbox{ADD {\em x0, x0, x2}}.
 
 The NTL.L2 instruction indicates that the target instruction
-does not exhibit temporal locality within the capacity of the next-innermost
-level of cache in the memory hierarchy (e.g., an L2 cache).
-NTL.L2 is encoded as \mbox{ADD {\em x0, x0, x2}}.
+does not exhibit temporal locality within the capacity of the two innermost
+levels of cache in the memory hierarchy.
+NTL.L2 is encoded as \mbox{ADD {\em x0, x0, x3}}.
+
+The NTL.L3 instruction indicates that the target instruction
+does not exhibit temporal locality within the capacity of the three innermost
+levels of cache in the memory hierarchy.
+NTL.L3 is encoded as \mbox{ADD {\em x0, x0, x4}}.
 
 The NTL.LLC instruction indicates that the target
 instruction does not exhibit temporal locality within the capacity of the last
 level of cache in the memory hierarchy.
-NTL.LLC is encoded as \mbox{ADD {\em x0, x0, x3}}.
+NTL.LLC is encoded as \mbox{ADD {\em x0, x0, x5}}.
 
 \begin{commentary}
 The NTL instructions can be used to avoid cache pollution when streaming data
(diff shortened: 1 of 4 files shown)


allenjbaum:
Nice. but...corner cases.
Is the intent that if the load/store is a vector load/store, and traps in
the middle, that it affects only (and all of) the first part? (so, multiple
cache lines)
IF it is an unaligned store with HW support, yet still traps, and the
implementation is such that it will store the first part before trapping,
    then does first part of the access treated as non-temporal, but not the
second? (a restricted version of the above)
Both of those seem kind of obvious.

IS there any sidechannel leakage that occurs because of that behavior, i.e.
timing an interrupt causing something to be cached vs. not cached,
with the resulting timing differences being seen later?

I note that you refer to zicbop, which is not ratified yet, and referred to
nowhere in (this) spec version.
 Assuming there will eventually be different flavors of prefetch (read
intent, write intent, instruction)
- do they get treated any differently from each other beyond what is
described here?

src/zihintntl.tex
```diff
@@ -0,0 +1,202 @@
...
+
+If the C extension is provided, compressed variants of these HINTs are also
+provided:
+C.NTL.P1 is encoded as \mbox{C.ADD {\em x0, x2}};
+C.NTL.PALL is encoded as \mbox{C.ADD {\em x0, x3}};
+C.NTL.S1 is encoded as \mbox{C.ADD {\em x0, x4}};
+and C.NTL.ALL is encoded as \mbox{C.ADD {\em x0, x5}}.
+
+The NTL instructions affect all memory-access instructions, unless otherwise
+specified.
```
dkruckemyer-ventana, line 142: They must not affect management CBOs (Zicbom). Zero and prefetch CBOs (as you state below) can be affected.

src/zihintntl.tex
```diff
@@ -0,0 +1,202 @@
...
+
+\begin{commentary}
+As of this writing, there are no exclusions to this rule, and so
+the NTL instructions affect all memory-access instructions
+defined in the base ISAs and the A, F, D, Q, C, and V standard extensions,
+as well as those defined within the hypervisor extension in Volume II.
+
+We also expect that the NTL instructions will affect some cache-maintenance
```
dkruckemyer-ventana, line 150: "management" is the preferred term... :)

src/zihintntl.tex
```diff
@@ -0,0 +1,202 @@
...
+operations.
+For example, NTL.PALL followed by a cache-line-zero instruction might indicate
```
dkruckemyer-ventana, line 152: CBO.ZERO instead of "cache-line-zero" to be precise?

pdonahue-ventana:
> I note that you refer to zicbop, which is not ratified yet

https://wiki.riscv.org/display/TECH/Recently+Ratified+Extensions says that Zicbop was ratified in November.

allenjbaum:
Huh -I based that comment on searching for it in the latest 20191213 unpriv
spec and the 20211203 priv spec  but failed to check the Recently Ratified
Extensions page.
My apologies.

aswaterman:
> Is the intent that if the load/store is a vector load/store, and traps in the middle, that it affects only (and all of) the first part? (so, multiple cache lines)

Because it's a HINT, this is up to the implementation as long as there isn't an architecturally visible effect, and so we're unburdened from having to specify this.  In practice, if an implementation follows our recommendations, the NTL will stop taking effect upon trapping.  Then, when execution resumes at the faulting load or store, the NTL won't be re-executed, and so won't apply to the remainder.

> IS there any sidechannel leakage that occurs because of that behavior, i.e. timing an interrupt causing something to be cached vs. not cached, with the resulting timing differences being seen later?

Yes, but this could've been observed more readily through other means (e.g., reading the timer before and after).

src/zihintntl.tex
aswaterman: Ack.

src/zihintntl.tex
aswaterman: Ack.

src/zihintntl.tex
aswaterman: Ack.

allenjbaum:
OK, I reread it, and notice recommendations and discouragements, but no
real requirements or prohibitions.
An implementation with decoded Icache could even encode the hint there, so
re-execution or re-start could re-use the hint.
I'm not totally convinced with the side-channel argument, but because I
don't have that malware mindset, I can't justify that non-conviction.
Looks good.

PR #811, opened 2022-01-15
hypervisor: Add Sv57 to &x4 address witdth list

dgreid:
The Sv57 example wasn't included in the list of expanded widths for the
*x4 modes.

Signed-off-by: Dylan Reid <dgreid@rivosinc.com>


aswaterman:
Thanks.

PR #812, opened 2022-01-15
Zihintntl: Fix number of hint instruction variants

a4lg:
It seems too minor but this extension defines four hint variants, not three.

1. `NTL.P1` / `C.NTL.P1`
2. `NTL.PALL` / `C.NTL.PALL`
3. `NTL.S1` / `C.NTL.S1`
4. `NTL.ALL` / `C.NTL.ALL`

Possible old list of "three" hint instructions follow:

1. `NTL.L1`
2. `NTL.L2`
3. `NTL.LLC`


aswaterman:
Oops.  This text went through multiple revisions, and I neglected to update that part of the intro.  Thanks!

PR #813, opened 2022-01-20
Refine HINT Instruction references (including tables)

a4lg:
Assuming Zihintntl extension to be frozen, ratified and merged, this pull request changes refines various locations related to HINT instructions.  Note that there are some changes *not* involving Zihintntl extension.

## Commit 1,2: Reflect "PAUSE" instruction and formatting

Commit ea9410a6a5ea by @aswaterman added "PAUSE" instruction and changed semicolon to colon.
Commit 1 and 2 ports this change to RV64I and RVC (RV64I change is a must).

## Commit 3: Change compressed hint example

From section 18.7 "HINT Instructions" (RVC):

> RVC HINTs do not necessarily expand to their RVI HINT counterparts. For example, `C.ADD x0, t0` might not encode the same HINT as `ADD x0, x0, t0`.

However, Zihintntl extension actually defines `C.NTL.ALL = C.ADD x0, t0[x5]`, which encodes the same hint as `NTL.ALL = ADD x0, x0, t0[x5]`.  This commit changes `t0[x5]` to `a0[x10]` to use unused encoding space available (suitable for an example).

## Commit 4,5: Add Zihintntl hints

I used new compressed format in the tables for a good reason.  Because it can cause various problems when a table exceeds the size of one page (and need major changes to the table design), I chose that design (in commit 4) to avoid using new package (`longtable` for example).

I hope new AsciiDoc documentation avoids this problem by design.

Commit 5 removes "no hints are defined" text in RVC, reflecting added Zihintntl hints.  This change is separate from commit 2 so that we can easily decide whether or not to merge Zihintntl-related changes.

# TODO 1: "RV32/64G Instruction Set Listings" (Chapter 26)

I didn't change "RV32/64G Instruction Set Listings" because it's based on output from [riscv-opcodes](https://github.com/riscv/riscv-opcodes).

# TODO 2: Zicbop HINT instructions

Recently ratified Zicbop HINT instructions (`ORI`-based) are not in the table (because it's not in the ISA Manual). If CMO instructions are to be added in the ISA Manual, they must be included later.  For this, [riscv-opcodes](https://github.com/riscv/riscv-opcodes) changes are not required because I already made changes.

However, do not replace `instr-table.tex` from riscv-opcodes yet because it lacks Zfh extension support.

### DIFF
--- a/src/c.tex
+++ b/src/c.tex
@@ -1229,7 +1229,13 @@ HINTs will ever be defined in this subspace.
   C.LI                    & {\em rd}={\tt x0}                           & 64          & \\ \cline{1-3}
   C.LUI                   & {\em rd}={\tt x0}, {\em nzimm}$\neq$0       & 63          & \\ \cline{1-3}
   C.MV                    & {\em rd}={\tt x0}, {\em rs2}$\neq${\tt x0}  & 31          & \\ \cline{1-3}
-  C.ADD                   & {\em rd}={\tt x0}, {\em rs2}$\neq${\tt x0}  & 31          & \\ \hline \hline
+  C.ADD                   & {\em rd}={\tt x0}, {\em rs2}$\neq${\tt x0}, {\em rs2}$\neq${\tt x2}--{\tt x5} & 27   & \\ \hline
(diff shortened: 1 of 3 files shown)


aswaterman:
Thank you.  I had been holding out on incorporating the NTL hints until the spec became frozen, because changing those tables is very painful.

PR #815, opened 2022-01-21
Update Instruction Tables based on riscv-opcodes

a4lg:
Now, the entire instruction tables are generated by [riscv-opcodes](https://github.com/riscv/riscv-opcodes).

Related: riscv/riscv-opcodes#99

But CMO instructions are manually deleted because they are not in the ISA Manual yet.

Relatively large diff came from removing trailing spaces (riscv/riscv-opcodes#98). Others are came from slight instruction order modifications (on Zfh, hypervisor and Svinval instructions).

### DIFF
--- a/src/instr-table.tex
+++ b/src/instr-table.tex
@@ -78,28 +78,28 @@
 &
 \multicolumn{10}{c}{\bf RV32I Base Instruction Set} & \\
 \cline{2-11}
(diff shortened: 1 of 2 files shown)


aswaterman:
Thanks again for helping to reunify riscv-opcodes.

PR #831, opened 2022-03-28
Add "N" to extension category list again

a4lg:
Despite that "N" extension itself is dropped (and no Zn* extensions are neither ratified nor proposed), canonical extension order should keep "N" to avoid future incompatibilities.

It effectively reverts half of commit f8d27f805b65 ("Remove or downgrade more references to N extension (#674)").

### DIFF
--- a/src/naming.tex
+++ b/src/naming.tex
@@ -83,7 +83,7 @@ Chapter~\ref{chap:zifencei}; ``Zifencei2'' and ``Zifencei2p0'' name version
 2.0 of same.
 
 The first letter following the ``Z'' conventionally indicates the most closely
-related alphabetical extension category, IMAFDQLCBKJTPV.  For the ``Zam''
+related alphabetical extension category, IMAFDQLCBKJTPVN.  For the ``Zam''
 extension for misaligned atomics, for example, the letter ``a'' indicates the
 extension is related to the ``A'' standard extension.  If multiple ``Z''
 extensions are named, they should be ordered first by category, then


aswaterman:
I'm on the fence about this.  On the one hand, adding `N` back is more consistent with the fact that we include e.g. `B` and `K` in this list, despite the fact that they do not exist.  On the other hand, I'd rather resolve the inconsistency by deleting all of them from this list.

gfavor:
I would argue for removing all three since there is no current expectation of these bits being assigned to those extensions (and no current expectation of some new version of the N extension being developed).  It's better to make these bit officially Reserved.  Then at whatever future point in time that assignment of one of these to an extension comes up, it can be weighed whether to assign the bit to that extension or to keep it reserved for future use for bitmanip, crypto, or a new N extension.

a4lg:
I'm against removing `B` and `K` (even if this pull request is rejected).
They are already used as prefixes of ratified Zb* and Zk* extensions and removing `B` and `K` from canonical order list requires something new to prevent problems. I think keeping `B` and `K` in the list is the least destructive way to maintain consistency.

`N` is not that important but I didn't want a simple problem to become complicated later (that's why I pull-requested this when I noticed it).

aswaterman:
I've made the executive decision to resolve this issue in favor of only imposing constraints on extensions that are actually defined.  f5f9c27010b69a015958ffebe1ac5a34f8776dff

a4lg:
That's unexpected move but also makes sense.

I asked you on gcc-patches but let me ask here again.

For instance, `Zi*` extensions must be placed before `Zv*` extensions (the same rule as before).

Then, how about `Zk*` and `Zv*`? Can `Zk*` be anywhere (as long as all `Z*` extensions are grouped together) or, like LLVM, `Zk*` must be placed after all `Z[CH]*` extensions (with single-letter extension `[CH]` existing)? I think you assume former but clarification would be better.

aswaterman:
We’ll also discuss this topic in architecture review when we discuss the H ordering matter.

For a brief preview, my recommendation will be to abandon ordering constraints for Z* extensions, because they are somewhat arbitrary, add extra complexity to the tooling, and reduce composability of these strings. The goal should be that we can blindly concatenate Z* extension strings. But this is just my opinion, and the rest of the committee may well disagree.

aswaterman:
As a final note, re: “unexpected”, @kasanovic has long planned to rip up and redo the naming chapter, and it’s virtually guaranteed that some of the rules will change in that process (but only with good reason).

aswaterman:
We discussed this topic in the architecture-review meeting today.  I'll summarize the group's thoughts below.

First, some historical context.  There are "ISA strings" and "canonical ISA strings".  The original intent was that canonical ISA strings are easily human-readable, with one important property being that it's easy to check with a glance whether an extension is present, since it always appears in the same place in the string.  Canonical ISA strings are what you'd use in product marketing material, are what a pretty-printer might emit, etc.

Originally, the goal wasn't that ISA-string parsers  should enforce canonical ordering, because, among other reasons, doing so reduces composability.  A desirable composition property, for example, is to take `$my_isa_string` and append `"_zfh"` to create a new ISA string that contains `Zfh`--even if the original string already contained Zfh, and even if canonical order would be violated.  This property simplifies meta-tooling (i.e. scripting) because the canonicalization rules don't need to be encoded all over the place.

The plan going forward is to normatively define what constitutes a legal ISA string--even though that string might not be canonical.  These strings are what software tools should accept in the long term.  We'll also non-normatively define the canonicalization scheme, which will define how human-readable strings _should_ (but not _must_) be constructed.  These spec changes will be made as part of the rewrite of the naming chapter.

Since existing canonical ISA strings will continue to be legal ISA strings, future tools will correctly continue to accept strings that current tools accept, i.e., there's no backwards incompatibility.  But, over time, ISA-string parsers will need to relax their rules to accept other legal but non-canonical ISA strings.

PR #836, opened 2022-04-19
Remove any references of HBI/HEE

atishp04:
RISC-V Hypervisor design has been evolved not to have a separate hypervisor
specific privilege mode. Thus, there is no concept of HBI or HEE anymore.

Remove all the references to it and update the diagram.

Signed-off-by: Atish Patra <atishp@rivosinc.com>


kasanovic:
Unless every SEE always also has the hypervisor extension, the distinction still exists and is useful.

There are effectively more privilege modes with the hypervisor extension, not fewer.

pdonahue-ventana:
It's clear that when the hypervisor doesn't exist then the interface that M mode has to implement is the SBI.  When the hypervisor exists, does M mode need to implement HBI calls?  In what ways are HBI and SBI different?

kasanovic:
We can have an SBI or HBI without M-mode - a RISC-V M-mode is only one way to implement an SBI or an HBI.  I think part of the confusion is that "the" SBI (as supported by OpenSBI) is only one possible SBI ECALL interface, and other OS stacks might have their own SBI (in the same way different OS can have their own user ABI).

An EEI is not only the set of  ECALLs, e.g., user ABI includes ISA, user-visible CSRs, as well as calling conventions and ECALLs.  A hypervisor EE provides the additional instructions and CSRs defined by the hypervisor spec - this is different than a non-hypervisor supervisor mode. A HBI ECALL interface might provide additional calls needed to support the hypervisor above and beyond what is needed with a supervisor only.

kersten1:
@pdonahue-ventana Does this PR need further review? If it is still valid, I would be happy to move it to the ASCIIDOC version.

pdonahue-ventana:
My understanding is that there conceptually is a HEE (and HBI).  The fact that there's no existing implementation of that doesn't negate the fact that it exists in theory.  The existing document seems correct and this PR can just be closed.

kersten1:
Closing PR as per previous comments.

PR #840, opened 2022-04-23
D extension is implied by V extension

a4lg:
... as commented by Andrew Waterman:
https://github.com/riscv/riscv-isa-manual/pull/838#issuecomment-1107473631

Related: #838


aswaterman:
Thanks!

PR #842, opened 2022-05-03
Clarify what is meant by "implemented"

jhauser-us:
And explain why "implemented and disabled" is considered a contradiction, while "implemented and enabled" is redundant.

### DIFF
--- a/src/machine.tex
+++ b/src/machine.tex
@@ -100,6 +100,36 @@ extensions, and I shall be selected over E if both are available.
 
 When a standard extension is disabled by clearing its bit in {\tt misa}, the instructions and CSRs defined or modified by the extension revert to their defined or reserved behaviors as if the extension is not implemented.
 
+\begin{commentary}
+For a given \mbox{RISC-V} execution environment, an instruction,
+extension, or other feature of the \mbox{RISC-V} ISA
+is ordinarily judged to be \emph{implemented} or not
+by the observable execution behavior in that environment.
+For example, the F extension is said to be implemented for
+an execution environment if and only if the instructions that the
+\mbox{RISC-V} Unprivileged ISA defines for F execute as specified.
+
+With this definition of \emph{implemented}, disabling an
+extension by clearing its bit in {\tt misa} results in the
+extension being considered \emph{not implemented} in M-mode.
+For example, setting {\tt misa}.F=0 results in the F extension
+being not implemented for M-mode, because the F extension's
+instructions will not act as the Unprivileged ISA requires
+but may instead raise an illegal instruction exception.
+
+We acknowledge that defining the term \emph{implemented}
+based strictly on the observable behavior can conflict
+with other common understandings of the same word.
+In particular, although common usage may allow
+for the combination ``implemented but disabled,''
+in this document it is considered a contradiction of terms,
+because \emph{disabled} implies execution will not behave
+as required for the feature to be considered \emph{implemented}.
+In the same vein, ``implemented and enabled''
+would be redundant here;
+either ``implemented'' or ``enabled'' is sufficient.
+\end{commentary}
+
 The design of the RV128I base ISA is not yet complete, and while much of the remainder of this
 specification is expected to apply to RV128, this version of the document
 focuses only on RV32 and RV64.
@@ -164,8 +194,9 @@ interface without burdening hardware designs.
 \end{commentary}
 
 The ``E'' bit is read-only.  Unless {\tt misa} is all read-only zero, the ``E''
-bit always reads as the complement of the ``I'' bit.  An implementation that
-supports both RV32E and RV32I can select RV32E by clearing the ``I'' bit.
+bit always reads as the complement of the ``I'' bit.
+If an execution environment supports both RV32E and RV32I,
+software can select RV32E by clearing the ``I'' bit.
 
 If an ISA feature {\em x} depends on an ISA feature {\em y}, then attempting
 to enable feature {\em x} but disable feature {\em y} results in both features


aswaterman:
Thanks, @jhauser-us.

aswaterman (approved):
Thanks, @jhauser-us.  Hopefully this clarifies our parlance.

allenjbaum (commented):
I'm going to have to respectfully disagree that Implemented but disabled is a contradiction.
There are 4 possiblities, not just Implemented (as you define it) or disabled.

Implemented, to me, simply says the hardware exists
Unimplemented says the hardware doesn't exist, so enable isn't posible.
Implemented but disabled implies (to me) that it is possible to change the configuration to be implemented and enabled - e.g.in the case if F-ext, that misa.F is RW, and not fixed at zero or 1. 
I would describe an extension as:
  -Unimplemented (no qualifier needed - the hardware doesn't exist
  -Implemented (with no qualifier implies it cannot be disabled)
  -Implemented and enabled  (the hardware exists, and misa.F is RW and set
  -Implemented and disabled (the hardware exists, and misa.F is RW and clear

scottj97:
I tend to agree with @allenjbaum that this is an unnecessary and confusing conflation of terms. Why not have the spec use "enabled" instead of its current uses of "implemented"? It is clear that if the hardware doesn't exist, then the extension is not enabled. If the hardware exists, then it may or may not be enabled, depending on e.g. `misa`.

It seems useful to be able to distinguish between "hardware exists but has been turned off by software" versus "hardware doesn't exist", but that's no longer possible with this oddball terminology.

gfavor:
I think part of the point behind this change is addressing confusion between "enabled" versus "implemented" at an architectural level, and that architecturally the distinction does not matter, i.e. the architecture intends to make no distinction between the two.  A design will of course care about the distinction, but that is an implementation level matter (transparent to the architecture).

scottj97:
@gfavor I understand that, but wouldn't that confusion be resolved by substituting "enabled" instead of "implemented" in most places?

The architecture does care about the distinction in a few places, such as [explaining](https://github.com/riscv/riscv-isa-manual/blob/55cd85c891f21302cdd9d1bfa60ee87a7a241f93/src/machine.tex#L101) what happens when an extension is disabled via `misa`:

> When a standard extension is disabled by clearing its bit in `misa`, the instructions and CSRs defined or modified by the extension revert to their defined or reserved behaviors as if the extension is not implemented.

This sentence does not conform to the new definition of "implemented", and I don't see a good way to reword this sentence given that new definition.

allenjbaum:
If the distinction is strictly microarchitectural, we need non-normative text that says that" implemented" and can be a dynamic property that depends on MISA (or other architectural state ...). Guaranteed to confuse people that have a conventional idea of what "implemented" usually means. (unless we use @scottj97 suggestion above. But it, as is pointed out above it isn't. If you can write code to determine whether it can be enabled and disabled (or implemented and unimplemented or whichever terms are used) - that seem to indicate that it is architectural.

jhauser-us:
Scott Johnson wrote:
> Why not have the spec use "enabled" instead of its current uses of "implemented"?

One major reason is because we'd have to do so also in the Unprivileged ISA document, which doesn't have any concept of an extension being enabled/disabled.  And there are intentions to reorganize the Privileged ISA doc so that the descriptions of supervisor-level environments can stand alone without any reference to machine-level environments, and again, supervisor-level environments have no concept that an extension can be disabled.  So, in both cases, how would we explain saying "enabled" instead of the current "implemented"?

> It seems useful to be able to distinguish between "hardware exists but has been turned off by software" versus "hardware doesn't exist"

Actually, the normative ISA text shouldn't even speak about hardware, because an implementation might be entirely in software, as it is with QEMU for example.  (Currently, the documents do occasionally speak of hardware normatively, but in my opinion those are almost all mistakes that should eventually get fixed.)

scottj97:
> So, in both cases, how would we explain saying "enabled" instead of the current "implemented"?

Call me dense but it seems easily explained. "_Enabled_ means that an extension is implemented and has not been disabled by [for? in?] the execution environment."
> Actually, the normative ISA text shouldn't even speak about hardware, because an implementation might be entirely in software, as it is with QEMU for example. (Currently, the documents do occasionally speak of hardware normatively, but in my opinion those are almost all mistakes that should eventually get fixed.)

Fair enough; substitute "feature" for "hardware" and the distinction still seems useful.

You didn't address the `misa` text. How should the spec explain `misa` when "enabled" and "implemented" mean the same thing?

jhauser-us:
Allen Baum wrote:
> we need non-normative text that says that" implemented" and can be a dynamic property that depends on MISA (or other architectural state ...).

At this time, the only architecturally defined mechanism for disabling an extension is `misa`, visible to M mode only.  Currently, the ISA documents attempt to contain the concept of "disabling an extension" to only within the subsection describing the `misa` CSR.  In that subsection, I believe this very commit adds the nonnormative text you suggest, explaining that "implemented" can be a dynamic property that depends on `misa`.

jhauser-us:
Scott Johnson wrote:
> Call me dense but it seems easily explained. "Enabled means that an extension is implemented and has not been disabled by [for? in?] the execution environment."

As I see it, that adds mental complexity everywhere, for the sole benefit of upholding a particular view of the word *implemented* in relation to CSR `misa`, relevant only to the subset of readers interested in M mode.  (And arguably only to hardware implementations of M mode, which the ISA docs should be agnostic to.)  In my opinion, the cost-benefit tradeoff weighs heavily against your suggestion.

The `misa` subsection says:
> When a standard extension is disabled by clearing its bit in misa, the instructions and CSRs defined or modified by the extension revert to their defined or reserved behaviors as if the extension is not implemented.

Scott:
> This sentence does not conform to the new definition of "implemented", [...]

I think that's debatable.  But even if I were to concede that there is a single inconsistent use of the word *implemented* which is then immediately clarified by a comment, that seems to me to be easily justified if it allows us to simplify the concept of *implemented* everywhere else in the ISA documents.

scottj97:
> As I see it, that adds mental complexity everywhere, for the sole benefit of upholding a particular view of the word _implemented_ in relation to CSR `misa`, relevant only to the subset of readers interested in M mode.

It strikes me as peculiar that by conflating the terms "enabled" and "implemented", which naturally mean different things, you've concluded that you're _reducing_ mental complexity.

But I understand that I (a hardware verification engineer) am not the only target audience, or even the primary one. Perhaps for other audiences this reduces mental complexity, so I won't belabor the point any further.

allenjbaum:
ditto. As someone who has to supply tests for this property. (Though, grankly, I wouldn't mind if it was either not possible to dynamically enable/disable.

jhauser-us:
Scott Johnson wrote:
> But I understand that I (a hardware verification engineer) am not the only target audience, or even the primary one. Perhaps for other audiences this reduces mental complexity, so I won't belabor the point any further.

Yes, I'd have to agree that for anyone who must pay close attention to `misa`, there's no actual simplification of concepts; possibly the opposite.  But we're consciously favoring the much larger set of readers who will be oblivious to `misa`, or rarely need to think about it.

PR #843, opened 2022-05-05
M-mode's next implemented privilege mode should be "next-lower"

YenHaoChen:
M-mode's next implemented privilege mode should be "next-lower" instead of "next-lowest" mode.

From the content, the mcounteren controls the availability to the next implemented privilege mode (S-mode if implemented, otherwise U-mode).  The "next-lowest" does not match the concept, and the "next-lower" makes more sense, as shown below.
##############################
Implemented modes |  MSU   |  MU
     next-lowest  | S-mode | M-mode
     next-lower   | S-mode | U-mode
##############################
Screenshot 2022-05-05 133823

### DIFF
--- a/src/machine.tex
+++ b/src/machine.tex
@@ -1773,7 +1773,7 @@ counter, and writes change only bits 63--32.
 
 The counter-enable register {\tt mcounteren} is a 32-bit register that
 controls the availability of the hardware performance-monitoring counters to
-the next-lowest privileged mode.
+the next-lower privileged mode.
 
 \begin{figure*}[h!]
 {\footnotesize


aswaterman:
To my knowledge, the current wording is the standard English for this concept.

YenHaoChen:
For a design without S-mode (MU only), the lowest mode is U-mode, and the next-lowest mode is M-mode, which does not match the concept.
I think "next-lower" or "next-highest" makes more sense.

aswaterman:
The text means “the next-lowest mode [relative to the mode currently being described]”. The mode being described is M, and so the next-lowest mode in an MU system is U.

aswaterman:
> Then, the next-lowest mode relative to M-mode in an MSU system should also be U (since U is lower than S), which does not match the concept.

This seems like a misunderstanding of the English term “next-lowest”. Relative to M, the next-lowest mode in an MSU system is S.

YenHaoChen:
Thank you for your clarification

PR #861, opened 2022-06-22
Remove a defunct commentary on reservation set sizes

palmer-dabbelt:
As of f6fe734 ("More LR/SC updates") these rules are in the privileged
spec, thy're not in the platforms specs.


aswaterman (approved):
Good catch, thanks.

PR #862, opened 2022-06-27
Add OP-V to base opcode map

a4lg:
`V` extension corresponding new base opcode OP-V (`0b1010111`) is now ratified and this table is updated.

Although vector instructions are not listed in Chapter 26, noting that this base opcode is used seems good to me. If `Zp*` extensions are ratified, OP-P (`0b1110111`) can be listed here, too.


aswaterman (approved):

aswaterman:
Yeah, makes sense to me.

Nelson1225:
Seems OP_P also need to be updated here?

PR #870, opened 2022-07-15
Svpbmt: leaf PTE reserved value=3 page fault

ingallsj:
@gfavor pointed out that this is already stated elsewhere:

> fundamentally, the tablewalk pseudo code says "if any bits or encodings that are reserved for future standard use are set within pte, stop and raise a page-fault exception corresponding to the original access type."

https://github.com/riscv/riscv-isa-manual/blob/f518c259c008f926eba4aba67804f62531b6e94b/src/supervisor.tex#L1677-L1679

Nevertheless, I think this clarification would be equally as helpful here as the existing adjacent statement for non-leaf PTEs.

### DIFF
--- a/src/supervisor.tex
+++ b/src/supervisor.tex
@@ -2432,6 +2432,10 @@ For non-leaf PTEs, bits 62--61 are reserved for future standard use.  Until
 their use is defined by a standard extension, they must be cleared by software
 for forward compatibility, or else a page-fault exception is raised.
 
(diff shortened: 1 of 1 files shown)


allenjbaum:
Hmm - I wonder what sail does...

aswaterman (approved):
For stylistic consistency, maybe something like "For leaf PTEs, setting bits 62--61 to the value 3 is reserved [...]"?  I don't think we use the value= notation anywhere.

aswaterman:
Merged with my tweak.

PR #877, opened 2022-07-31
fix some typos

cuishuang:


aswaterman:
"Normalisé" is not a typo; it's French.

`listofitems.tex` is a standard LaTeX package, so we shouldn't edit.

PR #888, opened 2022-08-29
Standardize on {\tt pc}, rather than PC

### LINKED ISSUE #887: Undefined abbreviations
1. https://github.com/riscv/riscv-isa-manual/blob/1237e31fc205fdf5d66f1d874ac958a07aec44b7/src/rv32.tex#L600-L602

   I suppose `BTB` stands for *branch target buffer*. `BTB` may not be clear for an average reader. Probably it is better to write it out to assist the reading flow.

2. https://github.com/riscv/riscv-isa-manual/blob/1237e31fc205fdf5d66f1d874ac958a07aec44b7/src/rv32.tex#L45-L46
    
   `pc` was defined in chapter 2.1 as the register name for the program counter, but in chapter 2.4 the abbreviation `PC` pops up:

   https://github.com/riscv/riscv-isa-manual/blob/1237e31fc205fdf5d66f1d874ac958a07aec44b7/src/rv32.tex#L593-L596

   In my opinion `PC` could be replaced with `pc`, because *the* program counter in the register file is mentioned and not the general concept of program counters. I also noticed that `PC` is used several times in next chapters and also in preface. My hunch is that `pc` was introduced later.

   This should be obvious for the reader, but I do not know how stringent the editors are with commentaries.

What do the authors think?

aswaterman:
We were using a mix of the two, with a bias towards the former.

Resolves #887

### DIFF
--- a/src/c.tex
+++ b/src/c.tex
@@ -697,7 +697,7 @@ with $\textit{rs1}{=}\texttt{x0}$ corresponds to the C.EBREAK instruction.
 
 \begin{commentary}
 Strictly speaking, C.JALR does not expand exactly to a base RVI
-instruction as the value added to the PC to form the link address is 2
+instruction as the value added to the {\tt pc} to form the link address is 2
 rather than 4 as in the base ISA, but supporting both offsets of 2 and
 4 bytes is only a very minor change to the base microarchitecture.
 \end{commentary}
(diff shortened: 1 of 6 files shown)


kasanovic (approved):
OK

PR #893, opened 2022-09-18
Add additional information for upper integer registers in commentary 

li3tuo4:
This pull request is regarding the public review of RV32E/RV64E.

Giving more details regarding *upper registers* would help readers to understand better in this context, where *integer registers*, *general-purpose registers*, *upper registers* are all mentioned.


aswaterman:
I don’t think this improves the clarity of the commentary.

PR #906, opened 2022-10-27
Intel tossed TSX

palmer-dabbelt:

### DIFF
--- a/src/a.tex
+++ b/src/a.tex
@@ -125,9 +125,7 @@ with an architected guarantee of eventual forward progress as
 described below.  Another concern is whether the influence of the
 current x86 architecture, with its DW-CAS, will complicate porting of
 synchronization libraries and other software that assumes DW-CAS is
-the basic machine primitive.  A possible mitigating factor is the
-recent addition of transactional memory instructions to x86, which
-might cause a move away from DW-CAS.
+the basic machine primitive.
 
 More generally, a multi-word atomic primitive is desirable, but there is
 still considerable debate about what form this should take, and


aswaterman (changes requested):
Is this not premature given https://en.wikipedia.org/wiki/Sapphire_Rapids?

If the final nail in the TSX coffin does come, we should probably instead write something like the following, to preserve the historical note and avoid "why not use TM?" questions:

"As x86 has dropped support for transactional memory, a move away from DW-CAS appears unlikely."

solomatnikov:
TSX is not the only implementation of transactional memory.

apinski-cavium:
ARM defines theirs here: https://developer.arm.com/documentation/ddi0617/latest
But I don't think any HW is actually going to implement it ...

gfavor:
Having come from the ARM world, ARM developed their TM extension a fair
number of years ago and I'm not aware of any designs implementing it after
all these years - including ARM designs.

Greg

allenjbaum:
IBM still has transactional memory, as far as I know.
 I've always wondered what magic they have that let them avoid the pitfalls
everyone else seems to have run into.

apinski-cavium:
> IBM still has transactional memory, as far as I know. I've always wondered what magic they have that let them avoid the pitfalls everyone else seems to have run into.

IBM removed it in Power10.

brouhaha:
Despite removing it from POWER10, IBM does still has transactional memory. It's in z/Architecture, and was introduced in the Tenth Edition of IBM z/Architecture Principles of Operation, dated September 2012, and first implemented in zEC12 systems announced on August 28, 2012. The 14th edition of the PoO dated May 2022 still includes it, and AFAICT the current z16 generation "Telum" microprocessor still supports it.
It seems premature to declare hardware transactional memory dead, even if it is pretty tricky to get it right. I still would like to see a hardware transaction memory specification for RISC-V. "We choose to define RISC-V hardware transactional memory in this decade and do the other things, not because they are easy, but because they are hard"

strikerdw:
Smiles... as someone with the scar tissue on this one, I laughed a bit at
"even if it is pretty tricky to get it right"... smiles, that doesn't even
begin to cover the pain involved :-).

Especially if your write footprint has ambitions of being large than 8-16
cachelines.

I will say that if you ever do get seriously drunk and try to do TM, do
NOT, repeat NOT, put a suspend mode in it. EVER.

As to the Kennedy quote and the Apollo program, we have nowhere near the
amassed resources that took and possibly not the massed resources this will
take..... It really is ugly and hard. Be sure you want to do this to
yourselves.

Going back to the original email... it has been years since I've looked at
the Intel specs on this, but is TSX the magic thing where they tried to
elide locks and replace critical sections automatically with TM? If so, is
it just that part of TM that Intel is tossing or is it the whole kit and
kaboodle including the explicitly manually coded transactions as well?

Derek Williams

brouhaha:
Intel TSX originally consisted of hardware lock elision (HLE) and restricted transactional memory (RTM).  HLE has been thrown under the bus, but it appears that Intel is still committed (no pun intended) to RTM.

strikerdw:
Fair enough. Thanks for the clarification.

Derek Williams

guoren83:
> Intel TSX originally consisted of hardware lock elision (HLE) and restricted transactional memory (RTM). HLE has been thrown under the bus, but it appears that Intel is still committed (no pun intended) to RTM.

I also hope riscv could keep Intel TSX RTM style and define "xstart/xend/xabort/xtest" is not a big budget for the ISA. For multi-cache lines atomic synchronization, TSX is the best solution for current. The traditional aligned atomic semantics (including DCAS) are limited to one cacheline.

kersten1:
Found this one in the ASCIIDOC so I'm deleting this PR as completed elsewhere.

PR #908, opened 2022-11-02
satp.mode is not WARL

palmer-dabbelt:
satp.mode was described as WARL in the table, but has specific behavior defined by the text of the specification: it requires writes to the whole CSR to be avoided when the MODE field is invalid (and WARL is defined on fields, not CSRs).

---

I don't really care which behavior is desired, but my guess is that the more verbose language ended up there on purpose.

### DIFF
--- a/src/supervisor.tex
+++ b/src/supervisor.tex
@@ -977,7 +977,7 @@ main memory be representable.
 \instbitrange{59}{44} &
 \instbitrange{43}{0} \\
 \hline
-\multicolumn{1}{|c|}{{\tt MODE} (\warl)} &
+\multicolumn{1}{|c|}{{\tt MODE} (see below)} &
 \multicolumn{1}{|c|}{{\tt ASID} (\warl)} &
 \multicolumn{1}{|c|}{{\tt PPN}  (\warl)} \\
 \hline


allenjbaum:
I am missing something then.  The spec says that if a write attempts to set .mode to an illegal value, the write to that field has no effect. That is (from my point of view) the preferred behavior of WARL fields and a completely legal implementation of WARL I only wish that were required for other WARL fields. 
What is different here is that other fields are also not written. That is not standard behavior --unless SATP as a whole is considered to be a single field, and that the illegal value of Mode=illegal, asid=xx, PPN=xx has the defined behavior. 
Note that this is not new behavior; the RWX fields of PMP were defined as individual bits, but changed to a single field because specific combinations where reserved/illegal. I suspect there are a few other lingering cases like that scattered around.

allenjbaum:
Ah, I had trouble interpreting what you wrote, so we are even.
My interpretation is that MODE is WARL, but ASID and PPN were not, and I
thought you were saying that it was MODE that wasn't WARL.
While the diagram for PMPCFG shows RWX as separate bits, it is qualified:
   The R, W, and X fields form a collective WARL field for which the
combinations with R=0 and W=1 are reserved.
I think there are some other odd cases besides just these. WARL fields are
the bane of my existence. Putting limits on them (which is something that
profiles does in several cases) is a good thing.
The other complication is that there are mode bits in one CSR that can
essentially change the format of the same or another CSR, and make some
fields that were formerly valid become invalid (or vice versa).

But, the case that you pointed out is not covered by any WARL definition

palmer-dabbelt:
[Inserting some newlines here, so I can reply in-line.]

> Ah, I had trouble interpreting what you wrote, so we are even. My interpretation is that MODE is WARL, but ASID and PPN were not, and I thought you were saying that it was MODE that wasn't WARL.

I am trying to say "satp.mode is not WARL".  I see WARL defined as

````
\subsection*{Write Any Values, Reads Legal Values (WARL)}

Some read/write CSR fields are only defined for a subset of bit
encodings, but allow any value to be written while guaranteeing to
return a legal value whenever read.  Assuming that writing the CSR has
no other side effects, the range of supported values can be determined
by attempting to write a desired setting then reading to see if the
value was retained.  These fields are labeled \warl\ in the register
descriptions.

Implementations will not raise an exception on writes of unsupported
values to a \warl\ field.  Implementations can
return any legal value on the read of a \warl\ field when the last
write was of an illegal value, but the legal value returned should
deterministically depend on the illegal written value and
the architectural state of the hart.
````

and the definition I have for `satp.mode` includes 

````
Implementations are not required to support all MODE settings,
and if {\tt satp} is written with an unsupported MODE, the entire write has
no effect; no fields in {\tt satp} are modified.
````

My argument is that those are in conflict, specifically the "Implementations can return any legal value on the read of a \warl\ field when the last write was of an illegal value" bit from the WARL definition.

Writes to satp.ASID and satp.PPN don't have similar constraints, so they seem perfectly reasonable to treat as WARL fields to me.

> While the diagram for PMPCFG shows RWX as separate bits, it is qualified: The R, W, and X fields form a collective WARL field for which the combinations with R=0 and W=1 are reserved.

I guess that's also a bit ambiguous.  I opened https://github.com/riscv/riscv-isa-manual/pull/909 to merge them into a single field.

> I think there are some other odd cases besides just these. WARL fields are the bane of my existence. Putting limits on them (which is something that profiles does in several cases) is a good thing.

I haven't really looked at the profile stuff all that closely lately, but I don't see any mentions of WARL or any specific CSR constraints.

> The other complication is that there are mode bits in one CSR that can essentially change the format of the same or another CSR, and make some fields that were formerly valid become invalid (or vice versa).

That's called out in a different part of the spec, I'm seeing

````
\section{CSR Field Modulation}

If a write to one CSR changes the set of legal values allowed for a
field of a second CSR, then unless specified otherwise, the second
CSR's field immediately gets an \unspecified\ value from among its new
legal values.
This is true even if the field's value before the write remains legal
after the write;
the value of the field may be changed in consequence of the write to
the controlling CSR.
````

which as far as I can tell is compatible with WARL, but just different.  It also doesn't apply to this case, where the two fields are in the same CSR (which seems good, as the `satp.MODE` behavior is different than what's specified here).

allenjbaum:
>The definition for satp.mode includes

Implementations are not required to support all MODE settings,and if
{\tt satp} is written with an unsupported MODE,
the entire write has no effect; no fields in {\tt satp} are modified.

>My argument is that those are in conflict, specifically the
"Implementations can return any legal value on
>the read of a \warl\ field when the last write was of an illegal value"
bit from the WARL definition.
I think there are in conflict - but not what I think you are pointing out
For SATP.mode, the returned value is the state of that hart, which includes
the current SATP value.
That perfectly meets the WARL definition that the returned value must be
deterministic,
and depend on the architectural state of the hart (specifically the current
value of SATP.mode is such state).

In fact that is precisely why WARL is defined that way (and I had that
added, so I'm quite sure).

But the ASID/PNN case is different, and is not covered by

If a write to one CSR changes the set of legal values allowed for
afield of a second CSR, then unless specified otherwise, the second
CSR's field immediately gets an \unspecified\ value from among its
newlegal values

Primarily because
a) the controlling field isn't in a different CSR, but also
b) the value being written is legal both before and after the write attempt
with the illegal mode field
That is, this is a statement about a field at rest whose legality is
changed, not one that is being written.

But even if this were changed such that the controlling field was in the
same CSR, then you need to look at whether
the set of legal values for that field has changed - and I'd argue that
either it hasn't changed, or ALL values are now illegal.
And then what do you do? There is nothing that you can replace it with that
isn't illegal, and the definition falls apart.

kersten1:
Closing in favor of https://github.com/riscv/riscv-isa-manual/pull/1378

PR #909, opened 2022-11-02
Mark the PMP R, W, and X bits as a single field

palmer-dabbelt:
As pointed out in
https://github.com/riscv/riscv-isa-manual/pull/908#issuecomment-1301159028 , these are defined as a single WARL field in the text but per-bit fields in the CSR encoding.

---

I haven't built this, so not sure if it's broken the table.


aswaterman (changes requested):
This change is going to confuse people about the encoding of this new "XWR" field.

Maybe an alternative is to delete the WARL annotation on these fields, but keep them as individual fields.  Then, the text supplies the info that they are collectively WARL, so this deletion doesn't change the semantics.

kersten1:
Closing in favor of https://github.com/riscv/riscv-isa-manual/pull/1378

PR #929, opened 2022-12-06
Clarify xRET clearing mstatus.MPRV when leaving M-mode

lbmeng:
Current spec says the following about xRET clearing mstatus.MPRV:

```
  ... and xPP is set to the least-privileged supported mode (U if
  U-mode is implemented, else M). If xPP̸=M, xRET also sets MPRV=0.
```

It may confuse people to think that the condition to set MPRV=0 is
to compare the updated xPP value against M, as this sentence comes
right after the xPP update sentence.

Change 'xPP' to its original value 'y' which avoids the confusion.


aswaterman:
Right, of course the text was referring to the original value of xPP. This clarification makes it unambiguous.

Issue #956, opened 2023-01-12
Zfa half-precision note correction


JamesKenneyImperas:
Section 25.1 of the Unprivileged ISA manual has this note:

> `Additionally, since 2^−16 is a subnormal in half-precision, entry 1 is numerically greater than entry 2 for FLI.H.
> `

Entry 3 (2^-15) is also subnormal, so the note should include that as well.

PR #957, opened 2023-01-13
Fix for issue #956

JamesKenneyImperas:


scottj97 (changes requested):
Please squash the three commits into one and force-push, instead of adding additional commits making corrections.

aswaterman (approved):

PR #959, opened 2023-01-14
Use current date/time in draft builds

jerryz123:
Not sure if the stale date is intentional.


nick-knight:
This PR is probably no longer relevant since the AsciiDoc migration. Suggest closing.

PR #972, opened 2023-01-30
Clarify WFI trapping behavior

### LINKED ISSUE #961: WFI and zero time limit
Section 3.1.6.5 of the Privileged Architecture manual discusses the implementation-specific bounded time limit for WFI. It says:

> The time limit may always be 0, in which case WFI always causes an illegal instruction exception in less-privileged modes when TW=1.

Does this mean that an illegal instruction exception will occur when the time limit is 0 _even if a locally-enabled interrupt is pending but interrupts are disabled at the time the WFI is executed_? Had the implementation-specific time limit been _non-zero_ in this case, the WFI would not have caused a stall at all, because the wake up condition is already true. It feels like the specified behavior with time limit 0 makes that a special case which is not consistent with the non-zero time limit case. Can you please confirm the required behavior?

Thanks.

aswaterman:
Resolves #961.

See discusson on that issue for motivation.

### DIFF
--- a/src/hypervisor.tex
+++ b/src/hypervisor.tex
@@ -278,6 +278,9 @@ instruction exception.
 When VTW=1 (and assuming {\tt mstatus}.TW=0), an attempt in VS-mode to
 execute WFI raises a virtual instruction exception if the WFI does not
 complete within an implementation-specific, bounded time limit.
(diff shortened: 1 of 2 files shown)


jhauser-us:
Okay by me.

PR #978, opened 2023-02-01
Riscv isa asciidoc

wmat:


scottj97:
I subscribe to this repo to keep up to date with any spec changes and discussions. Lately I get lots of spam from these asciidoc PRs.

Is there a way I can unsubscribe from PRs based on their target branch? Or is there a way this could be done without creating all these PRs? Seems like @wmat ought to be given push access to the branch so he can simply push away without having to repeatedly create all these useless PRs.

aswaterman:
I'm not aware of branch-specific following/filtering settings.  I agree pushing directly to the branch is probably the best way to reduce the email traffic.

wmat:
Sorry about that, I'll begin pushing to the branch directly.

PR #985, opened 2023-02-23
Polish the assembly example to avoid potential confusion

DehengYang:


aswaterman (changes requested):
I'll let @daniellustig decide what to do here.  This still isn't legal RISC-V assembly (since the result of `sc.w` needs to be written somewhere), and it has introduced a new bug: the figure caption assumes all of the `sc` instructions are labeled `(c)`.

DehengYang:
Sorry for my mistake. I noticed that the `(c)`  appears only once in the Figure caption and is not used in other places of the context:

> Figure A.4: In all four (independent) code snippets, the store-conditional (c) is permitted but not
guaranteed to succeed

What if we change it to:

>  Figure A.4: In all four (independent) code snippets, the store-conditional (sc.w) is permitted but not
guaranteed to succeed

Please kindly advise. Thank you!

aswaterman:
Some of the snippets use `sc.d`, not `sc.w`, so that isn't quite right either.  How about:

Figure A.4: In all four (independent) code snippets, the final store-conditional instruction is permitted but not guaranteed to succeed

And then to fix the other issue, change all of the `sc` instructions to write a result to `t3`, e.g. `sc.w t3, t2, 0(s0)` in this case.  If Dan approves of that version, we can merge it.

DehengYang:
That looks great! Thank you very much for the advice, which makes the manual more rigorous, sound, and easily understandable.

daniellustig:
Thanks for the PR @DehengYang.  One more quick comment: shouldn't we still pair `lr.w` with `sc.w`, and `lr.d` with `sc.d`?  The first two snippets now pair `lr.d` with `sc.w`.

DehengYang:
@daniellustig Thank you very much for the comment. I have fixed it for your consideration.

aswaterman (approved):

aswaterman:
Thanks, y'all!

Issue #300, opened 2018-12-08
Question regarding reserved category, HINT instructions and illegal exception 


taoj33:
Hi,
In SPEC it defines Reserved as
 > Reserved encoding are currently not defined but are saved for future standard extensions

while for HINT, it is defined as following
> HINTS do not change any architecture visible state,except for advancing the pc and any applicable performance counters.

Is it correct to interpret that HINT is a subset of Reserved category which can be treated as NOP while  all other encodings in the Reserved category should be treated as illegal instructions?

aswaterman:
HINTs are not in the Reserved opcode category; they live within an already-defined standard encoding space.

You are right that HINTs should be treated as NOPs, whereas Reserved opcodes should raise illegal-instruction exceptions on current conforming implementations.

aswaterman:
-

taoj33:
@aswaterman  Thank you for your reply.

I can understand instructions with invalid func/opcode should trigger traps. 

But should instructions with undefined imm/reg raise illegal traps? for example, 
> C.JR  (RES, rs1=0)
> SLLIW,SRIW, and SRAIW encodings with imm!=0 are reserved

aswaterman:
Ah.  I understand the confusion, as the C chapter is written a bit differently and needs to be cleaned up.  The two examples you quoted are in different classes:

- The shift instructions with imm=0 are marked HINT in Tables 16.6 and 16.7, so they should execute as NOPs (which they do naturally, since 0 is the shift identity).
- C.JR with rs1=0 is marked RES, rather than HINT, so should raise an exception.

taoj33:
@aswaterman 
Sry made a typo 

> SLLIW, SRLIW, and SRAIW encodings with $imm[5] !=0 are reserved.

> Previously, SLLIW, SRLIW, and SRAIW with imm[5]=0 were defined to cause illegal instruction 
> exceptions, whereas now they are marked as reserved.  This is a backwards-compatible change.

So is it safe to conclude all instructions described as reserved should raise exception?

aswaterman:
Yes.  The difference between "is reserved" and "should raise an exception" is that, in the former case, you can put a non-compliant extension in that encoding space.  If you aren't using that space for a non-compliant extension, you should simply raise an exception.

asb:
Where in the spec does it say that instructions in reserved encoding spaces should trap? That's obviously a sensible and desirable thing to do, but I can't see the spec requiring it.

[...]

PR #990, opened 2023-03-14
Disambiguate the use of the term "reserved" for HINT instructions

asb:
As was raised in #300, the overloaded use of the term "reserved" when listing HINT encodings is potentially confusing. As clarified in that thread, these encodings are "reserved" in the intuitive English language meaning of the word, but are _not_ part of the "reserved instruction-set category" as defined in the section 1.3 introduction and further elaborated in section 2.2 "Base Instruction Formats" (which notes how platforms can choose to raise an illegal-instruction exception for reserved instruction encodings).

### DIFF
--- a/src/rv32.tex
+++ b/src/rv32.tex
@@ -1435,7 +1435,10 @@ visible effect.
 Table~\ref{tab:rv32i-hints} lists all RV32I HINT code points.  91\% of the HINT
 space is reserved for standard HINTs.  The
 remainder of the HINT space is designated for custom HINTs: no standard HINTs
-will ever be defined in this subspace.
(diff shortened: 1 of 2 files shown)


asb:
Friendly ping on this as it's been a couple of weeks.

aswaterman:
Maybe we should just avoid use of the word "reserved" in this paragraph and in the tables, and instead use the word "designated".  What do you think @asb @kasanovic?

asb:
That's a good idea, especially as we already say "Designated for custom use" already. I've submitted #1001 as an alternative PR using that approach.

aswaterman:
Superseded by #1001

PR #991, opened 2023-03-14
Fix Known Issues examples

### LINKED ISSUE #795: possible bugs in figures A.20, A.21
I'm copying this from @jrahmeh's issue here: https://github.com/riscv/riscv-v-spec/issues/765

In section A.7 of the unprivileged ISA (Document Version 20191214-draft), both examples (Figures A.20 and A21) seem to be incorrect.

In A.20, for outcome a0=1 to occur, a2 would have to be 1, but a2 is initialized to zero and remains zero after the amoswap at (e). One fix is to replace (h) with: sw t1, 0(s0)

In A.21, the outcome "a0=0, a1=1, a2=0" is not forbidden by operational model, it is obtainable with the global memory order: a b c d e f.
One possible fix is to change the outcome to: "a0=1, a1=1, a2=0" and change (f) to: sw t1, 0(s0)

hboehm:
fixes #795

Fixes are as suggested there by several contributors.


aswaterman (approved):
Thanks again.

Issue #886, opened 2022-08-29
Why setting the rl bit on the LR instruction isn't needed to make the LR/SC sequence sequentially consistent?


patrick-yingxi-pan:
https://github.com/riscv/riscv-isa-manual/blob/1237e31fc205fdf5d66f1d874ac958a07aec44b7/src/a.tex#L236

hboehm:
I think this really depends on the ABI conventions. C++ "sequential consistency" guarantees store->load ordering only if both the load and the store are sequentially consistent. The convention here assumes that stores are followed by a fence or use an RCsc ordering. A.6 assumes the opposite convention. I will suggest improved wording.

aswaterman:
Answered by Hans.

PR #992, opened 2023-03-16
Atomics fixes

hboehm:
This is a fairly minimal fix to the unprivileged spec to prepare us for atomics ABI work. It fixes some oversimplifications that were uncovered during the load-acquire store-release mailing list discussions. Some of the observations behind the changes came from Palmer Dabbelt. Andrea Parri, and others. It also fixes some inconsistencies with A.6, and adds a clarification there.

Note that the use of LR/SC to implement "sequentially consistent" atomics primitives is subtle, and depends on the precise semantics, and how other atomics are implemented. I tried to tread the fine line between inaccuracy and turning this into a memory model tutorial.

#886

### DIFF
--- a/src/a.tex
+++ b/src/a.tex
@@ -229,19 +229,27 @@ correctly on expected common implementations that operate in this manner.
 
 An SC instruction can never be observed by another RISC-V hart
 before the LR instruction that established the reservation.
+
+\begin{commentary}
 The LR/SC
 sequence can be given acquire semantics by setting the {\em aq} bit on
 the LR instruction.  The LR/SC sequence can be given release semantics
-by setting the {\em rl} bit on the SC instruction.  Setting the {\em
-  aq} bit on the LR instruction, and setting both the {\em aq} and the {\em
-  rl} bit on the SC instruction makes the LR/SC sequence sequentially
-consistent, meaning that it cannot be reordered with earlier or
-later memory operations from the same hart.
+by setting the {\em rl} bit on the SC instruction.  Assuming
+suitable mappings for other atomic operations, setting the
+{\em aq} bit on the LR instruction, and setting the
+{\em rl} bit on the SC instruction makes the LR/SC
+sequence sequentially consistent in the C++ {\em memory\_order\_seq\_cst}
+sense. Such a sequence does not act as a fence for ordering ordinary
+load and store instructions before and after the sequence. Specific
+instruction mappings for other C++ atomic operations,
+or stronger notions of ``sequential consistency'', may require both
+bits to be set on either or both of the LR or SC instruction.
 
-If neither bit is set on both LR and SC, the LR/SC sequence can be
+If neither bit is set on either LR or SC, the LR/SC sequence can be
 observed to occur before or after surrounding memory operations from
 the same RISC-V hart.  This can be appropriate when the LR/SC
 sequence is used to implement a parallel reduction operation.
+\end{commentary}
 
 Software should not set the {\em rl} bit on an LR instruction unless the {\em
 aq} bit is also set, nor should software set the {\em aq} bit on an SC
@@ -478,20 +486,6 @@ rl} bit on an AMO makes the sequence sequentially consistent, meaning that
 it cannot be reordered with earlier or later memory operations from the same
 hart.
 
-\begin{commentary}
-The AMOs were designed to implement the C11 and C++11 memory models
-efficiently.  Although the FENCE R, RW instruction suffices to
-implement the {\em acquire} operation and FENCE RW, W suffices to
-implement {\em release}, both imply additional unnecessary ordering as
-compared to AMOs with the corresponding {\em aq} or {\em rl} bit set.
-\end{commentary}
-
-An example code sequence for a critical section guarded by a
-test-and-test-and-set spinlock is shown in Figure~\ref{critical}.  Note the
-first AMO is marked {\em aq} to order the lock acquisition before the
-critical section, and the second AMO is marked {\em rl} to order
-the critical section before the lock relinquishment.
-
 \begin{figure}[h!]
 \begin{center}
 \begin{verbatim}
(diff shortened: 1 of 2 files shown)


src/a.tex
```diff
@@ -528,3 +528,5 @@ \section{Atomic Memory Operations}
 address. A sequentially consistent store can be implemented as an AMOSWAP
 that writes the old value to x0 and has {\em rl} set. However the superfluous
 load may impose ordering constraints that are unnecessary for this use case.
+Specific comilation conventions may require both the {\em aq} and {\em rl}
```
scottj97: Compilation is misspelled

src/a.tex
```diff
@@ -529,5 +529,5 @@ \section{Atomic Memory Operations}
 address. A sequentially consistent store can be implemented as an AMOSWAP
 that writes the old value to x0 and has {\em rl} set. However the superfluous
 load may impose ordering constraints that are unnecessary for this use case.
-Specific comilation conventions may require both the {\em aq} and {\em rl}
+Specific compilation conventions may require both the {\em aq} and {\em rl}
```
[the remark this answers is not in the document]
hboehm: Thanks! Did that, hopefully correctly. I'm not yet very familiar with the workflow here.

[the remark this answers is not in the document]
aswaterman: You got it right :)

aswaterman (changes requested):
My preference is to write no new normative text in the A-extension chapter, but instead make the usage guidance non-normative (i.e. in `commentary` blocks).  This approach avoids conflating the ISA semantics with the ABI mapping.

Beyond that, I have no technical objections.

@daniellustig can you also take a peek?

daniellustig:
LGTM, I'm on board with the general push to add memory model mappings into the ABI so we can over time clean up the lingering discrepancies, and small steps like this are helpful.  Thanks @hboehm.

I also agree with Andrew though.  RVWMO is the normative thing from the ISA POV, so the text being modified here really could all be demoted to commentary, either inline (where I agree we probably don't want to be too verbose) or in Appendix A (where there's plenty of room to elaborate).  Then eventually the ABI will be the thing to give normative guidance for the mappings themselves.

hboehm:
Moved stuff, including a bit of earlier text that didn't appear normative into commentary sections. Also a moved the last figure in the atomics section in the source, so that it doesn't appear in the middle of the commentary. Otherwise it looked weird to me.

src/a.tex
```diff
@@ -511,14 +505,36 @@ \section{Atomic Memory Operations}
 \label{critical}
 \end{figure}
 
+\begin{commentary}
+The AMOs were designed to implement the C11 and C++11 memory models
```
appujee, line 509: Is the C++14 recommendation on prohibiting out-of-thin air values (https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3710.html, https://en.cppreference.com/w/cpp/atomic/memory_order) applicable in this memory model?

hboehm: That's really an orthogonal, purely language-level, issue. And it only affects memory_order_relaxed (and in theory memory_order_consume), where the concern here is with seq_cst and acquire/release ordering.

The architecture avoids out-of-thin-air results due to the dependency rules. This gets messy when you throw in compiler transformations, since "optimizations" can remove the dependencies that save us at the architecture level.

[ More details:]
Currently we're stuck in a situation in which we kind of understand what C++ memory_order_relaxed is supposed to mean, but it's not pretty, and formal definitions that actually seem to say that are still research and rather complicated for a standard. And they could use more scrutiny. (Or we could strengthen memory_order_relaxed a bit. But that costs a smallish amount of performance and hence will not fly in standards committees.)

The history here is unpleasant. C++11 had an out-of-thin-air prohibition attempt that had serious unintended consequences. C++14 replaced this with basically a hand-waving statement that it should be avoided, without really defining what it is. That still stands. And the definition is tricky and controversial. See wg21.link/p1217 .  (Java went through a similar exercise a few years earlier, so C++ started carefully avoided that broken definition, and we started with a different broken definition instead.)

IMO, the uncertainties around memory_order_relaxed make it even more critical that we end up with efficient implementations of the stronger ordering constraints. Which is really the concern here.

src/a.tex
```diff
@@ -229,19 +229,27 @@ \section{Load-Reserved/Store-Conditional Instructions}
...
-later memory operations from the same hart.
-
-If neither bit is set on both LR and SC, the LR/SC sequence can be
+by setting the {\em rl} bit on the SC instruction.  Assuming
+suitable mappings for other atomic operations, setting the
+{\em aq} bit on the LR instruction, and setting the
+{\em rl} bit on the SC instruction makes the LR/SC
+sequence sequentially consistent in the C++ {\em memory\_order\_seq\_cst}
+sense. Such a sequence does not act as a fence for ordering ordinary
+load and store instructions before and after the sequence. Specific
```
aswaterman, line 243: I don't think this is true, since the PPO rules order the acquire with all younger instructions and order the release with all older instructions.

In any case, this is why I'd like to constrain the textual changes to the mapping section in the appendix, rather than modifying the normative text in the A chapter that defines the instructions.  @daniellustig can you spare some time to work with Hans to organize this PR in a way we'd all find agreeable?

hboehm: I may have misunderstood your request. I did enclose this in a commentary block, which presumably is non-normative?

Regarding the technical issue: If I have

s; lr.aq; sc.rl; l

where the lr and sc refer to the same location, but the s and l refer to different locations, I don't believe there is anything to prevent s and the l from being reordered in between the lr and sc, and then past each other. In that sense, the lr;sc does not act as a fence. You get the same phenomenon with programming-language-level locks and relaxed atomic operations (in spite of some amount of code the incorrectly assumes that critical sections are fences)

If you know of a better way to state this, I would be happy to do so. I think this is worth stating in commentary because many people find it surprising, and I think the Linux kernel has tried to strengthen some atomics primitives to avoid it, something I would argue against.

aswaterman: Sorry, @hboehm, I misread the TeX: you did, indeed, enclose this note within the commentary, thus making it non-normative.

On the technical matter, I believe the RVWMO PPO rules contradict your position.  Rule 6 says `s` and `lr.aq` are ordered before `sc.rl`.  Rule 5 says `l` and `sc.rl` are ordered after `lr.aq`.  So `s` and `l` are ordered.

aswaterman: As to the technical matter, I agree on further reflection that I can't find a PPO rule that suggests `s` and `l` are ordered.

aswaterman (approved):
I'm OK with this version.  Thanks, @hboehm.

I'm holding off on merging all PRs until next Thursday's changeover to the asciidoc version of the spec is complete.  The transition will cause some headaches (e.g. this PR will become one giant merge conflict) but I'll lend a hand to manually fix it up.

hboehm:
What's the mechanically best way to move forward here? I have a translation of this to the adoc version.  Abandon this version and generate a new pull request?

aswaterman:
@hboehm Thanks for performing the translation; I was expecting that I'd end up doing it :) yes, I think closing this PR and opening a new one is the way to go.  Don't forget the primary branch is now `main`, not `master`.

hboehm:
This was replaced by #1039, which moves these changes to the adoc version.

PR #995, opened 2023-03-22
Minor improvements on Makefile

rpsene:
* Use the := assignment operator for immediate assignment, to avoid unnecessary recomputations.

* Use $(dir $(abspath $(lastword $(MAKEFILE_LIST)))) to automatically calculate the srcdir relative to the location of the Makefile, without the need to copy the Makefile to a build directory.

* Remove unnecessary whitespace around variable assignments for consistency.

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -1,4 +1,4 @@
-#=======================================================================
+#----------------------------------------------------------------------
 # Makefile for generating LaTeX documents
 #-----------------------------------------------------------------------
 #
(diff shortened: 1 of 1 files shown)


scottj97:
You've described three separate changes; there's no reason that shouldn't be three separate commits. By combining them it makes it unnecessarily difficult to review and understand.

rpsene:
> You've described three separate changes; there's no reason that shouldn't be three separate commits. By combining them it makes it unnecessarily difficult to review and understand.

@scottj97 thanks for the feedback, but the changes are so small and really simple to understand that 3 PRs sounds unnecessary.

scottj97:
> @scottj97 thanks for the feedback, but the changes are so small and really simple to understand that 3 PRs sounds unnecessary.

I did not say three PRs. They are all obviously related so they belong as three separate commits in one PR. [Small commits are beautiful.](https://betterprogramming.pub/why-you-should-write-small-git-commits-c9a042737aa6)

Why would you want to make reviewers' work more difficult for no reason?

PR #1001, opened 2023-03-29
Label HINT encodings intended for future standard use as "designated"

asb:
The term "reserved" could be ambiguous, as it might be understood to refer to the "reserved instruction-set category" as defined in the section 1.3 instruction and further elaborated in section 2.2 "Base Instruction Formats". Avoid this potential confusion by using "designated" instead, which also matches the terminology used for HINT encodings intended for custom use.

This is an alternative to #990, following @aswaterman's suggestion.


aswaterman:
LGTM, thanks.

aswaterman (approved):

PR #1008, opened 2023-04-10
Zihpm and Zicntr are ratifed, give it a explictly version

kito-cheng:
<del>Those two extension has added for a while, but didn't give it version info, I am not sure 1.0 or 2.0 should be used here, 1.0 should be represents that the item has been ratified, but I saw https://github.com/riscv/riscv-isa-manual/commit/7a58119dad5bd43e2171a26b56ef60f9591a1c9c is using `2.0` for `Zihintpause` so I using 2.0 in this PR.</del>

Those two extension has added for a while, so give it a explictly version.

Public review mail:
https://groups.google.com/a/groups.riscv.org/g/isa-dev/c/HZGoqP1eyps/m/7e8iXjJhAQAJ

Status can be found here:
https://wiki.riscv.org/display/HOME/Specification+Status

### DIFF
--- a/src/counters.tex
+++ b/src/counters.tex
@@ -1,4 +1,4 @@
-\chapter{``Zicntr'' and ``Zihpm'' Counters}
+\chapter{``Zicntr'' and ``Zihpm'' Counters, Version 1.0}
 \label{counters}
 
 RISC-V ISAs provide a set of up to thirty-two 64-bit performance counters and
--- a/src/preface.tex
+++ b/src/preface.tex
@@ -38,10 +38,12 @@ The document contains the following versions of the RISC-V ISA modules:
     \em T          & \em 0.0 & \em Draft \\
     \em P          & \em 0.2 & \em Draft \\
     \em V          & \em 1.0 & \em Frozen \\
+    \bf Zicntr     & \bf 1.0 & \bf Ratified \\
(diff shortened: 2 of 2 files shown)


src/counters.tex
```diff
@@ -1,4 +1,4 @@
-\chapter{``Zicntr'' and ``Zihpm'' Counters}
+\chapter{``Zicntr'' and ``Zihpm'' Counters Version 2.0}
```
aswaterman, line 1: `,` after `Counters`

aswaterman:
We might choose to hold off on merging new PRs until the asciidoc changeover occurs.  (I'm not actually sure whether we decided on a policy.)

I think we actually want to use version 1.0 here (assuming we haven't used any version numbers >= 1.0 in the past for these extensions).  Using 2.0 to signify completion is an old convention.

kito-cheng:
@aswaterman thanks, the main goal of this PR is clarify the version of those extensions, it required information for toolchian implementation, it really not matter to be 1.0 or 2.0, but I don't really want to define our own ISA spec as a toolchain folks.

So I think we can treat your reply as confirm they are 1.0?

aswaterman:
Yeah, let’s run with that.

pdonahue-ventana:
I meant preface, not introduction.

PR #1018, opened 2023-04-18
Fix the build issue

rpsene:
This change fixes the build issue, add a release step and update the backend to run the build within a container. It also includes the following:

- Add `create_release` input to `workflow_dispatch` event
- Create and push a git tag when manually triggered with `create_release=true`
- Create a non-draft release with the generated PDF when manually triggered with `create_release=true`

### DIFF
--- a/.github/workflows/build-pdf.yml
+++ b/.github/workflows/build-pdf.yml
@@ -1,75 +1,83 @@
-# This workflow installs dependencies for PDF generation, generates the PDF,
-# and uploads the PDF as an artifact.
-
-name: Build Document PDF
+name: Build RISC-V Privileged Spec PDF
 
 on:
   workflow_dispatch:
-  workflow_call:
-    outputs:
-      name:
-        description: "The base name of the pdf file (without .pdf extensions)"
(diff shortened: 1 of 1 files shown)


aswaterman:
After merging this, exactly which PDFs will be uploaded to the release?

rpsene:
> After merging this, exactly which PDFs will be uploaded to the release?

@aswaterman the PDF generated from the Latex version available in the master branch. Every merged PR will create a draft release and will provide a PDF built (which the commit hash in the name). Then, to create an official release, it will requires a manual trigger. The build process uses this Makefile (https://github.com/riscv/riscv-isa-manual/blob/master/build/Makefile).

aswaterman:
@rpsene I asked because I only saw `riscv-privileged.pdf` in there, but from your description I assume you mean the unpriv spec will also be built and uploaded.

rpsene:
@aswaterman thanks for asking about the PDF, I was missing one. The update now handles both PDFs.

aswaterman (approved):
@rpsene ok, let's give it a shot.

PR #1025, opened 2023-04-27
switch all wavedrom images to output svg

### LINKED ISSUE #1019: Converting riscv-isa-manuals asciidoc results in text-less wavedrom diagrams because of 3rd party bug
I realize the the asciidoc versions of the isa-manual is not complete, and I'm mostly reaching out through this issue to @wmat in hopes there is some secret knowledge about wavedrom-cli that could be shared.

In short, I'm trying to generate the asciidoc html/pdf versions locally after replicating the build environment but I'm getting an annoying artifact in both the pdf and the html versions where the wavedrom diagrams are generated without any text rendered inside them:

image

Trying to render this directly through wavedrom-cli replicated the bug when rendering to PNG only and not with SVG,
and seems to be a [known](https://github.com/wavedrom/cli/issues/22) issue reported in the wavedrom repo and [unfixed in node-svg2img](https://github.com/fuzhenn/node-svg2img/issues/41).

Is there a change that a PR for switching to generating SVGs instead of PNGs would be accepted for the asciidoc version given this (unresolved) bug and SVGs generally being slightly more friendty (IMO) for zooming in / printing / hidpi in general?

damageboy:
fixes #1019 1019

### DIFF
--- a/src/images/wavedrom/atomic-mem.adoc
+++ b/src/images/wavedrom/atomic-mem.adoc
@@ -1,6 +1,6 @@
 //## 9.4 Atomic Memory Operations
 
-[wavedrom, ,]
+[wavedrom, ,svg]
 ....
 {reg: [
   {bits: 7, name: 'opcode',     type: 8, attr: ['7','AMO','AMO','AMO','AMO','AMO','AMO','AMO']},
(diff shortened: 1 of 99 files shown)


wmat:
This commit works around know issue: https://github.com/wavedrom/cli/issues/22

wmat (approved):

PR #1028, opened 2023-04-28
priv-1.13: clarify that MXLEN >= SXLEN; constrain SXLEN >= UXLEN

aswaterman:
@gfavor FYI

@jhauser-us please check my wording.  It's a little obtuse because I was trying to constrain the set of legal values for SXL's WARL purposes.


jhauser-us:
I would probably have:
> If S-mode is implemented, the set of legal values that the UXL field may assume excludes those that would cause UXLEN to be greater than SXLEN.

Because:  (1) Is SXLEN defined without S-mode?  (2) To me it seems more natural to have UXLEN bounded by SXLEN.

aswaterman:
Good points.  PR updated with your proposed wording.

Greg approved via email, but I'll wait to merge until we have the chance to discuss this with the ARC.

PR #1038, opened 2023-05-10
Fix error in RVC Instruction Set Listings: Quadrant 2

### LINKED ISSUE #1037: Error in RVC Instruction Set Listing Q2
I'm following the [latest release](https://github.com/riscv/riscv-isa-manual/releases/tag/riscv-isa-release-5b7252f-2023-04-27) of Unprivilege spec and there seems to be a couple of errors in Quadrant 2 of RVC Instruction Set Listings, described as follows:

1. Encoding of `C.LWSP` under bit-filed [4 : 0] is incorrect. It is `uimm[4|9:6]` which is wrong and this encoding is for `C.LQSP`. Looking at the previous versions of the spec, the correct bitfield of `C.LWSP` should be `uimm[4:2|7:6]`.
2. Definition of `C.LQSP` is missing from Quadrant 2 (Figure 5, Chapter 18).

Abdulwadoodd:
### Relate Issue

This PR fixes #1037 

### Description

1. Fix the definition of `C.LWSP`.
2. Added definition of `C.LQSP`.

### DIFF
--- a/src/images/bytefield/rvc-instr-quad2.adoc
+++ b/src/images/bytefield/rvc-instr-quad2.adoc
@@ -35,6 +35,13 @@
 (draw-box "rd≠0" {:span 5})
 (draw-box "uimm[4|9:6]" {:span 5})
 (draw-box "10" {:span 2})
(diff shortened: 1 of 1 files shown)


wmat:
Good catch.  Thanks for the fix.

PR #1039, opened 2023-05-12
Fixes for Atomics wording (adoc)

hboehm:
This is a fairly minimal fix to the unprivileged spec to prepare us for atomics ABI work. It fixes some oversimplifications that were uncovered during the load-acquire store-release mailing list discussions. Some of the observations behind the changes came from Palmer Dabbelt. Andrea Parri, and others. It also fixes some inconsistencies with A.6, and adds a clarification there.

Note that the use of LR/SC to implement "sequentially consistent" atomics primitives is subtle, and depends on the precise semantics, and how other atomics are implemented. I tried to tread the fine line between inaccuracy and turning this into a memory model tutorial.

This is the adoc version of pull request #992 .

### DIFF
--- a/src/a-st-ext.adoc
+++ b/src/a-st-ext.adoc
@@ -192,19 +192,29 @@ expected common implementations that operate in this manner.
 ====
 
 An SC instruction can never be observed by another RISC-V hart before
-the LR instruction that established the reservation. The LR/SC sequence
+the LR instruction that established the reservation.
+
+[NOTE]
+====
+The LR/SC sequence
 can be given acquire semantics by setting the _aq_ bit on the LR
 instruction. The LR/SC sequence can be given release semantics by
-setting the _rl_ bit on the SC instruction. Setting the _aq_ bit on the
-LR instruction, and setting both the _aq_ and the _rl_ bit on the SC
-instruction makes the LR/SC sequence sequentially consistent, meaning
-that it cannot be reordered with earlier or later memory operations from
-the same hart.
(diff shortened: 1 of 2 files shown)


aswaterman:
Thank you, @hboehm.  Can you review my slightly altered version of this change here? https://github.com/riscv/riscv-isa-manual/commit/d53848e827c974df8e9dd6bc917d9f7aabe87932

I think the sentences about the `aq` and `rl` bits individually need to remain normative.  (The rest follows logically, I think, so keeping it non-normative is desirable.)  @daniellustig can you also be so kind as to weigh in?

hboehm:
I have no strong feelings either way. My reading was that in RVWMO, it's
really memory operations that have acquire and/or release annotations. So
the whole notion of acquire or release semantics of an LR/SC sequence
seemed like informal guidance to me, with the actual semantics following
from the RVWMO section. This is all good, since I'm not sure there is a
priori universal agreement on exactly what it means for LR/SC sequence to
have acquire or release semantics. And the RVWMO definition is much more
precise.

aswaterman:
It's probably true that these statements in the RVWMO spec suffice to define the meaning of the `aq`/`rl` bits: https://github.com/riscv/riscv-isa-manual/commit/0e6df795ced5abd141f4e92b7a935e3b217f0349#diff-2284d757b615fc4f9a380c73afb5735a7a7e56d7a322d3c8a0091844fc26a65dR132-R136

But I think that not explicitly mentioning them in the A chapter is going to lead to confusion.  @daniellustig can you break the tie here?

PR #1052, opened 2023-06-23
A-ext chapter: 64-bit word => doubleword; 32-bit word => word

ved-rivos:
Use _doubleword_ instead of "64-bit word" and _word_ instead of "32-bit word"


aswaterman (approved):

aswaterman:
Thanks!

PR #1059, opened 2023-07-06
Change "Reserved" to "Designated" in RVC HINT instructions table

nananapo:
In https://github.com/riscv/riscv-isa-manual/pull/1001,
I think "Reserved for future standard use" in C extension's table should have been replaced with "Designated for future standard use".


aswaterman (approved):
This is consistent with other uses.

asb:
Good catch, thanks.

PR #1065, opened 2023-07-12
fix missing " in misa description

nananapo:


wmat:
I'm not sure if these should be quoted as the quotes earlier in the paragraph are referring to bits, whereas these instances of I and E are referring to extension names. Also, they were not quoted in the LaTeX version.   I'll defer to @aswaterman to decide on this one.

aswaterman (approved):
I think @wmat is right about the intent here.  With that said, "I" without quotes looks confusingly like the English pronoun, so the proposed version does make it easier on the reader.  So I think we can run with this.

PR #1067, opened 2023-07-13
FLW/FSW should have width=S not width=H

tariqkurd-repo:


wmat:
sp-load-store.adoc is used in the Zfh chapter and the F Standard Extension chapter.  Would "S" apply to both usages?  Note that the LaTeX shows "W" for FLW/FSW. And the LaTeX shows "H" for FLH/FSH.  I'm wondering if we need separate wavedrom files for Zfh vs the F Extension. @aswaterman

aswaterman:
The TeX is right.

wmat:
Closing as per @aswaterman 's comment

PR #1069, opened 2023-07-18
fix formatting

tariqkurd-repo:
`mtval2`'s doesn't render, suggesting a space to fix that but it might not be the best answer


aswaterman:
@wmat

wmat (approved):

PR #1072, opened 2023-07-21
Make "Integer Register-Immediate Operations" Consistent

charlie-rivos:
Properly align the ADDI16SP immediate in wavedrom. Remove the bit brackets from the immediates because the instructions do not uniformly follow the same immediate format.


aswaterman (approved):
@charlie-rivos Thank you for lending a hand fixing these nits.  Keep 'em coming.

PR #1084, opened 2023-08-08
Officially name the hypervisor extension "H"

a4lg:
Note that this is currently the initial change.

It does not replace other references to
*   *the hypervisor extension* to
*   *the "H" extension*.

Still, even without replacing those, it can be a first patch of the series.

### TODO List (for complete change)

- [x] Check whether the name "Standard Extension for Hypervisor" is okay.
    - "Standard Extension for Hypervisor Support" is the official name (as [suggested](https://github.com/riscv/riscv-isa-manual/pull/1084#issuecomment-2026107589) by @aswaterman)
- [ ] Determine where to replace (to *the "H" extension*)
- [ ] Actual replacement (or not)

### DIFF
--- a/src/hypervisor.adoc
+++ b/src/hypervisor.adoc
@@ -1,5 +1,5 @@
 [[hypervisor]]
-== Hypervisor Extension, Version 1.0
+== "H" Extension for Hypervisor Support, Version 1.0
 
 This chapter describes the RISC-V hypervisor extension, which
 virtualizes the supervisor-level architecture to support the efficient


kersten1:
Is this PR moving forward? I can help with the edits if they are approved.

aswaterman:
@jhauser-us should approve or reject.

jhauser-us:
I think something like "Standard Extension for Hypervisor Support" would be a better name.

aswaterman:
The title would be `"H" Standard Extension for Hypervisor Support`, then.  Can you update the PR, @a4lg?

a4lg:
@aswaterman Sorry for late. I'm back and will work on it.

aswaterman:
@a4lg Note that recently we decided to remove the word "Standard" from extension titles, so please make that change here, too.  LGTM otherwise.

a4lg:
@aswaterman Thanks for letting me know (I was off for months and I'm glad to hear such changes to keep up).

aswaterman (approved):

a4lg:
Changed the commit title from "standard" to "base".

The use of the word "base" is likely non-canonical, but this is only in the commit title and reflects the fact that this is the baseline for the hypervisor support (without some extensions expected to be implemented along with the base `H` extension).

aswaterman:
Fine with me.  We do not want "Base" in the chapter title, though, similar to how "Base" does not appear in the M-mode or S-mode chapter titles.

PR #1090, opened 2023-08-10
Maintainance for vector (considering future chapter integration)

a4lg:
*   Remove duplicate files in `src/` (those in `src/images/wavedrom/` are used)
*   Rename too generic `inst-table.adoc` to `v-inst-table.adoc`


wmat:
Makes sense.

PR #1094, opened 2023-08-13
Add OP-V and OP-P to base opcode map

a4lg:
"V" extension corresponding new base opcode OP-V (`0b1010111`) is now ratified and opcode using OP-P (`0b1110111`) is approved for the vector cryptography extensions.

This PR reflects new base opcodes.

Note that OP-V part is an AsciiDoc port of #862.
OP-P part is new in this PR (I didn't notice at the first time but vector cryptography extensions use instructions with OP-P).

### DIFF
--- a/src/rv-32-64g.adoc
+++ b/src/rv-32-64g.adoc
@@ -14,12 +14,12 @@ and RV64G.
 .RISC-V base opcode map, inst[1:0]=11
 [%autowidth.stretch,float="center",align="center",cols=  ">.^4m,  ^.^4m,    ^.^4m,      ^.^4m,    ^.^4m,  ^.^4m,      ^.^4m,           ^.^6m, ^.^4h"]
 |===
-|inst[4:2] .2+|000 .2+|001   .2+|010     .2+|011   .2+|100 .2+|101     .2+|110          .2+|111 (>32b)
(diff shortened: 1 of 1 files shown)


Nelson1225:
Seems like OP_V and OP_P haven't been updated in the spec, they are still reserved.  However, OP_V was already used for vector in the [binutils](https://github.com/bminor/binutils-gdb/blob/master/gas/config/tc-riscv.c#L860).  So it would great if spec can clarify 0x57 is OP_V, and 0x77 is OP_P :-)

a4lg:
Reference: [Vector Crypto Specification (v1.0.0-rc2)](https://github.com/riscv/riscv-crypto/releases/tag/v)

> Appendix A: Crypto Vector Cryptographic Instructions
> OP-P (0x77) Crypto Vector instructions except Zvbb and Zvbc

lists 21 instructions (with 11 `funct6` values) with OP-P.

a4lg:
@aswaterman
Reminder and a question.
Should I separate `OP-P` changes to avoid arguments of P/V allocation?  I thought using the name `OP-P` is already approved because (near ratification) vector cryptography documentation uses `OP-P`... but tell me if not.

aswaterman:
Yeah, I don't want to accept this PR as-is, but I can accept a PR that adds only OP-V.

a4lg:
Understood.  I'll keep this open but make a plain port of #862.

aswaterman:
I'm going to close this with the expectation something similar will be reopened once a P spec is closer to the finish line.

PR #1096, opened 2023-08-15
Mark Zfa as Frozen and Version 1.0

palmer-dabbelt:
While looking at some Zfa patches I couldn't actually find a concrete reference to the spec being frozen, just the commit text of the binutils patches.  I figured this was the easiest way to ask...

### DIFF
--- a/src/colophon.adoc
+++ b/src/colophon.adoc
@@ -40,6 +40,7 @@ h|Extension h|Version h|Status
 |*Zihintpause* |*2.0* |*Ratified*
 |_Zihintntl_ |_0.3_ |_Frozen_
 |_Zam_ |_0.1_ |_Draft_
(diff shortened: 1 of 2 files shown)


aswaterman:
I think should have actually been ratified as of a few days ago — @jjscheel right?

gfavor:
On Tue, Aug 15, 2023 at 11:21 AM Andrew Waterman ***@***.***>
wrote:

> I think should have actually been ratified as of a few days ago —
> @jjscheel <https://github.com/jjscheel> right?
>
The IC/HC committee approval vote to send this to the BoD for artification
is almost complete.  So Frozen but not yet officially Ratified.

aswaterman:
Thanks, @gfavor.  We can merge this for now and update it again in a few days.

aswaterman (approved):

PR #1099, opened 2023-08-21
SLLIW, SRLIW, SRAIW only have 6 fixed bits in imm

charlie-rivos:
SLLIW, SRLIW, SRAIW have 6, not 7, bits in their imm11:0 field.


aswaterman (changes requested):
No, they have 7.  See https://github.com/riscv/riscv-opcodes/blob/d752f193cec46cfd33300abba2a3a0d020c36755/rv64_i#L12-L14

charlie-rivos:
Aww that makes sense. I got thrown off by the differences in these sections.

image

aswaterman:
Gotcha.

PR #1108, opened 2023-09-01
Clarify constraints on when tval might be zeroed

aswaterman:


allenjbaum:
Ick - this is first-class non-deterministic behavior - that's horrible. Why?

aswaterman:
The goal isn't to facilitate non-determinism.  The issue is that there are sometimes multiple exception sources multiplexed onto the same cause code, and only in some of them is useful information available to populate tval.  (Consider e.g. the breakpoint exception, where the EBREAK instruction and the debug trigger module generate the same exception cause code for entirely different reasons.)

@jhauser-us can elaborate.

allenjbaum:
Ok, thanks for the clarification to me, but there should be a clarification
in the spec, otherwise you can't tell when the logic is broken or just
being picky.
E.g. clarify that different sources of the same cause may or may not update
xTVAL, but for any particular cause the action must be consistent.
And: if we know the case where this could happen (as above), I'd like to
see that documented as well.

aswaterman:
Let’s see if @jhauser-us has a suggestion on how to wordsmith it. (I have a feeling he was going to suggest a rewrite, anyway!)

gfavor:
What about:

_The hardware platform will specify which exceptions must set `mtval` informatively, which may unconditionally set it to zero, and which may exhibit either behavior based on different sources for the exception._

ved-rivos:
I think the recommendation that its implementation defined - "specified hardware platform" - makes it not possible for portable SW to rely on the *tval and hints that this can be impdef behavior. The case "The issue is that there are sometimes multiple exception sources multiplexed onto the same cause code, and only in some of them is useful information available to populate tval." on other hand seemed quite deterministic and not impdef. Usually in such cases there would be something else that provides a indicator about the source of the exception since the cause code cannot be used to discriminate between the various contributing sources. Is the reason to state "specified by hardware platform" because such a discriminator of the contributing sources is unspecified in the architecture? 

I agree that documenting known cases would be valuable.

gfavor:
Note that my suggested wording strikes the balance that Ved is arguing for, i.e. that an implementation may exhibit either behavior for a given exception type (aka cause code), but only as a function of differing sources for the exception.  For a given source for an exception, the behavior must be consistent.

allenjbaum:
My desire?requirement? is to have a spec defined way to indicate to Sail
which sources of interrupts have which behavior.
That's easy enough to do using the riscv-config YAML data structure (for
small values of "easy") that Sail already uses - but it requires
 - an architecturally defined list of the TVAL sources for each beyond what
already exists for existing hart[[load/store_amo/inst][illegal
access/pagefault], illegal op] causes
 - adds new requirements to ratification that any extension that adds a new
xTVAL source
   -- add a node to the riscv-config YAML data structure that defines the
new speced xTVAL loading behavior
   -- modify Sail to use that list to select whether or not xTVAL is loaded
I don't think the trap handler cares; it will just save whatever is in
xTVAL in the signature.

aswaterman:
I've adopted a variant of @gfavor's version, which provides the context that the behavior depends on the underlying event that caused the exception.

allenjbaum:
I guess my concern might be the word "indicator", which would distinguish
between the different causes that cause a particular cause value to be
written into the CAUSE CSR.
(I will disavow any suggestion that the above sentence has too many
"cause"s.)
There are three cases (for any specific value in the xCAUSE CSR):
a)    0 is written
b)    either 0 or a value is written. The ACTs require that implementers
indicate, on a cause-by-cause basis , whether a value is stored or zero is
stored
c)   different values may be written for the same cause code, based on some
"indicator"

If that exception source, as Ved has some architecturally defined indicator
(defined in the spec), Sail can be made to handle it.
I am concerned about the word "Usually"  in "Usually in such cases there
would be something else that provides an indicator about the source of the
exception".
I'd like to see that word effectively struck out, because if it is
"Sometimes" we get to the "portable software can't depend on it" case
 - unless it devolves to case b),in which case it is (cause code &
indicator) based  decision, not just cause code based, and Sail is provided
with that decision.
I say implementation decision, not implementation option, because that
condition must be architecturally, unambiguously speced, else it's simply a
nondeterministic choice, as far as Sail is concerned.

So, this is good, but in my opinion, not good enough.

PR #1109, opened 2023-09-01
Define the RV32-only medelegh CSR

aswaterman:


ved-rivos:
should there be a `hedelegh`?

aswaterman:
@ved-rivos Yes.  Added it.

src/priv-csrs.adoc
```diff
@@ -606,15 +612,17 @@ MRW
...
 |Machine status register. +
 ISA and extensions +
 Machine exception delegation register. +
 Machine interrupt delegation register. +
 Machine interrupt-enable register. +
 Machine trap-handler base address. +
 Machine counter enable. +
-Additional machine status register, RV32 only.
+Additional machine status register, RV32 only. +
+Additional machine exception delegation register, RV32 only.
```
jhauser-us, line 625: Change the description of `medelegh` to be:  "Upper 32 bits of `medeleg`, RV32 only".

(CSR `mstatush` gets described differently because it's not a simple alias of bits 63:32 of a 64-bit `mstatus` the way other "high-half" CSRs are.  For RV32, `mstatus` is actually specified as 32 bits in size.)

src/priv-csrs.adoc
```diff
@@ -438,25 +438,29 @@ Supervisor interrupt pending.
...
+`hgeie` +
+`hedelegh`
 |Hypervisor status register. +
 Hypervisor exception delegation register. +
 Hypervisor interrupt delegation register. +
 Hypervisor interrupt-enable register. +
 Hypervisor counter enable. +
-Hypervisor guest external interrupt-enable register.
+Hypervisor guest external interrupt-enable register. +
+Additional hypervisor exception delegation register, RV32 only.
```
jhauser-us, line 463: Change the description of `hedelegh` to be:  "Upper 32 bits of `hedeleg`, RV32 only".

src/priv-csrs.adoc
aswaterman: Yeah, point taken.

PR #1111, opened 2023-09-01
Specify SFENCE/HFENCE requirements when changing PBMTE

aswaterman:


src/hypervisor.adoc
```diff
@@ -1666,6 +1667,22 @@ address-translation cache entries that have cached PMP settings
...
+HFENCE.VVMA with _rs1_=`x0` and _rs2_=`x0` suffices to synchronize with
+respect to the altered interpretation of VS-stage PTEs' PBMT fields for the
+currently active VMID.
+
+NOTE: No mechanism is provided to atomically change `henvcfg`.PBMTE and
+`hgatp` simultaneously.  World switches between guests that have differing
+`henvcfg`.PBMTE settings therefore require execution of HFENCE.VVMA.  Hence,
+for best performance, it is recommended to use the same `henvcfg`.PBMTE
+setting for all guests.
+
```
ved-rivos, line 1685: Though the `henvcfg` and `hgatp` cannot be switched simultaneously, it may not matter as when `V=0` the two stage address translation is not effective. Is the execution of `hfence.vvma` when switching between guests that have differing `henvcfg.pbmte` required?

aswaterman: I do see what you're getting at, but I think it may still matter, for reasons including speculative access of HLV/HSV instructions.

ved-rivos: My understanding of the current definition of HLV/HSV was that the microarchitecture was already required to guard against speculative accesses caused by them. There are other state could affect the HLV/HSV such as `vsatp`, `senvcfg`, `hstatus`, and `vstatus`. While `henvcfg.pbmte` may be amenable to be made mostly static and identical for most virtual machines, it may not be true for other states. For example, would we need to worry about a situation where a PTE gets cached in the address translation cache by speculative HLV from an old VMs page table since `vsatp` has not been switched but gets tagged with the `VMID` of the new VM since `hgatp` has been updated and thereby cause a cross VM leakage?

This recommendation wants to give HLV/HSV uarch the flexibility of observing partial state (but only for `pbmte`?). Should we consider specifying either a) recipe for software to make invocation of HLV/HSV non speculative through the use of suitable fence e.g, write `hstatus` and use a fence to order HLV/HSV to the CSR write OR b) provide an architectural control that enables the use of HLV/HSV - like `SUM` or `HU`.

gfavor:
@aswaterman As suggested by the title of this issue, doesn't there also need to be similar sfence.vma text added for the sake of single-stage translations?

aswaterman:
@gfavor isn't that already covered by the new text in the M-mode chapter?

src/hypervisor.adoc
aswaterman: I'm pretty sure the spec has a comment about how HLV may be speculatively executed and so `hgatp` is always considered active, even when in HS-mode or M-mode.

However, I wonder if we are overthinking this: do we really ever need to change the value of `henvcfg.PBMTE`?  If a guest OS is unaware of PBMT, what's the harm in giving it PBMTE anyway?  It'll just zero those bits in the PTE, which has the desired effect.  So if the hypervisor is OK with any guest having PBMT, then set `henvcfg.PBMTE`=1 unconditionally.

ved-rivos: The specification notes that when privilege is U and `HU=0`, the `hgatp` is not active. This guards against speculating into a HL* as they would be treated as illegal-instructions. In HS, whether an HL* could be executed speculatively with partial state depends on the context switch code construction. If the context switch occurred with interrupts masked and no control transfers between the CSR writes that could mispredict then it may be guarded against this. However, if the context switch code does not keep interrupts masked then it is possible that it might get interrupted, say after the `hgatp` write but before the `vsatp` write and then the ISR execution leads to a speculative HL* instruction execution which may end up observing partial state - for example caching a PTE from old guest with `VMID` of new guest. Since there doesn't exist an `HU` like control, and the architecture does not disallow interrupts during context switch, I thought the expectation was for the microarchitecture to defend against this.

If the hypervisor held `henvcfg.pbmte` then an OS or some framework that relies on causing a page fault due to reserved bits being set in the PTE may be surprised. I am sure Linux does not do this but I am not sure about all other software. At least this could lead to a virtualization hole where the guest may detect if the execution is in `V=1` by accessing a PTE with PMBMT bits non-zero and observing if there is a page fault.

For future extensions that add controls to `henvcfg` or another CSR like `senvcfg` it may not be ideal to expect those controls to be identical across guests and flushing guest address translation caches going between guests where they are not identical may also not be ideal due to performance considerations.

aswaterman: I don’t agree that suffices to guard against speculative execution of HLV. A mispredicted JALR can always direct control flow to any address that’s currently executable. We need to operate under the assumption that HLV is always speculatively executable in HS as currently specified.

I’m not terribly sympathetic to the argument that OSes would rely on those reserved bits trapping.

There is a valid question about whether future additions to henvcfg that reinterpret the PTEs might not tolerate this approach. But I argue that concern should only apply to extensions that change the behavior of existing bits, not those that consume the reserved bits. I’d further argue such extensions should become different satp.MODE values, not henvcfg bits.

ved-rivos: I agree that we should assume that HLV is always speculatively executable in HS. That is why I was stating the two conditions a) no control transfers between CSR writes in the context switch b) no interrupts i.e. unless the HS ensured that the code was straightline and non-interruptible, it has to be prepared for a speculative HLV execution in HS. Could this can then lead to the security issue that if a speculative HLV observes a condition where `vsatp` is that of the old guest and `hgatp` is of the new guest then there could be a address translation in the new guests address space that is based on the old guests page tables. This could be a way for malware to inject a address translation into a victim virtual machine. Since the architecture does not provide a way for HS to defend against this speculative execution based cross VM injection it would have to be defended against and prevented by microarchitecture. Is my observation right?

aswaterman: I agree with the problem statement, but I don't think the conclusion follows.

Can we not fix this by changing the context-switch code to zero `vsatp`, then load the new `hgatp`, then load the new `vsatp`? Similarly, the PBMTE issue could be sidestepped by zeroing `vsatp`, then zeroing `hgatp`, then changing PBMTE, then loading the new `hgatp`, then loading the new` vsatp`.

BTW, I don't think the "no branches, interrupts disabled" rule suffices.  You could still be interrupted into M-mode, whose trap handler mispredicts a JALR, causing it to speculatively execute an HLV instruction.  (We would also be making a stringent assumption about the execution model, namely that only control-flow instructions can speculatively redirect the PC.  Whether or not this is acceptable, it would be a tightening of semantics.)

ved-rivos: Absolutely, the "no branches, interrupts disabled" guideline seems to have shortcomings. Speculation can unpredictably occur either in the M-mode or even as a by-product of higher-level software structures, such as emulated supervisor events that are designed to be delivered when interrupts are masked and cause a transition from M-mode to S-mode.

A well-defined, step-by-step guidance on safe context-switching would be valuable. The sequential zeroing strategy you proposed—first for the `vsatp` and then for the `hgatp` — seems to be a sound approach to counteract potential misbehavior due to speculative execution. Furthermore, adopting this method would also alleviate the need to ensure `PBMTE` uniformity across VMs, as well as the necessity for executing an `hfence.vvma` when there are differences.

aswaterman: OK, I think we have a mutually satisfactory path forward.  I'll amend this PR to provide context-switch guidance, rather than cautioning against using different PBMTE values for different guests.

aswaterman: Actually, I think it is even simpler.  Because `henvcfg.PBMTE` affects only VS-stage, not G-stage, only `vsatp` need be zeroed when changing `henvcfg.PBMTE`.  (By contrast, an M-mode world switch would need to zero both, but that is the less interesting case.)

aswaterman: @ved-rivos can opine on the updated non-normative text?

ved-rivos: I agree with that.

aswaterman: phew!

aswaterman:
Verbally approved by @gfavor

PR #1119, opened 2023-09-07
Allow Misaligned Stores with Page Fault to partially succeed

ingallsj:
1. Extend the >XLEN decomposition relaxation to vector stores by deleting the word “floating-point” in the description of PMP faults.
2. Add similar text copied from the PMP section to the Sv32 section for virtual memory page faults.

> This is just a clarification, since it follows logically from misaligned accesses possibly being decomposed into smaller accesses.
> "Notably, instructions that reference virtual memory are decomposed into multiple accesses."

### DIFF
--- a/src/machine.adoc
+++ b/src/machine.adoc
@@ -3046,9 +3046,8 @@ fetches may also be decomposed into multiple accesses, some of which may
 succeed before an access-fault exception occurs. In particular, a
 portion of a misaligned store that passes the PMP check may become
 visible, even if another portion fails the PMP check. The same behavior
-may manifest for floating-point stores wider than XLEN bits (e.g., the
-FSD instruction in RV32D), even when the store address is naturally
-aligned.
+may manifest for stores wider than XLEN bits (e.g., the FSD instruction
+in RV32D), even when the store address is naturally aligned.
 
 [[pmp-vmem]]
 ==== Physical Memory Protection and Paging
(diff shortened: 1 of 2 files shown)


aswaterman:
cc @gfavor — this is already the case (it follows from the legality of decomposing misaligned accesses, as might be the case in HW and is definitely the case in trap-and-emulate), but this makes it explicit with respect to the virtual-memory system. OK to merge?

allenjbaum:
I am wondering if you want to be (non-normatively) explicit about the
corner cases that can cause a naturally aligned FSD on an RV32 to partially
store.
That can only happen if a PMP top of range entry ends on an odd address
(TOR entry or  4 byte entry), right?.
It can't happen if granularity is >0

aswaterman:
@allenjbaum Also FSD+NA4 (for RV32 with G=0), and also FSQ+TOR (for RV32 or RV64 with G <= 1).  It's a little ungainly to exhaustively specify.

@gfavor Since this has marinated a while, and it's just a clarification (due to the rule that accesses can be decomposed down to the byte level to facilitate emulation), I'm going to merge it as-is.  LMK if you have any concerns after the fact.

aswaterman (approved):

gfavor:
Yes, this is fine.

allenjbaum:
Agreed. I'd love a bit more clarity on the order of those accesses, which is what really makes it complicated, but we've discussed that. We'll deal with deterministic cases, but someday I suspect we will need to deal with it just like the vector spec does.

PR #1122, opened 2023-09-11
Add missing words

KubaO:
Some words were missing.


aswaterman (approved):

aswaterman:
Actually, this is wrong.  The first missing phrase was "39 bits", not "32 bits".  I'll fix it manually.

aswaterman:
Thanks for pointing it out, though!

PR #1123, opened 2023-09-12
Clarify REM wording

hdelassus:
I think this was obviously in the writers' intent that only nonzero results have the same sign as the dividend. The non-normative note is also updated to reflect the special case of signed division overflow.


aswaterman:
At least we got this right in the table, so it wasn’t actually ambiguous…

PR #1131, opened 2023-09-16
Standardize on no hyphen in illegal instruction exception

aswaterman:
Based on casual grepping, we omit the hyphen in 46 cases but retain it in 7.  Standardize on the more common one.

### DIFF
--- a/src/hypervisor.adoc
+++ b/src/hypervisor.adoc
@@ -885,7 +885,7 @@ widest supported width not wider than the new VSXLEN.
 
 When V=1, both `vsstatus`.FS and the HS-level `sstatus`.FS are in
 effect. Attempts to execute a floating-point instruction when either
-field is 0 (Off) raise an illegal-instruction exception. Modifying the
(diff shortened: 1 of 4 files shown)


gfavor (approved):
Hard to find fault in replacing a few hyphens with spaces. :)  I haven't checked that all of the offending hyphens were found, but I'll place trust in grep. So, LGTM.

kasanovic:
Technically, it should be hyphenated.  The phrase should be parsed as "(illegal Instruction) exception" not "illegal (instruction exception)".  Probably most native speakers can disambiguate correctly, but wouldn't surprise me if the other reading didn't cause confusion at some point.

aswaterman:
Unless sufficiency frequent usage has made “illegal instruction” a compound noun rather than a noun phrase. Our inconsistent usage was a bug, regardless. We can chat briefly about this one (and its cousin virtual instruction exception) in ARC. I don’t mind re-fixing it in the opposite direction if that’s what we think is best.

PR #1133, opened 2023-09-20
Hyphenate "virtual-/illegal-instruction exception"

aswaterman:
Reverts #1131 

One might argue that "illegal instruction" is a compound noun in the context of computer architecture, and so hyphenating "illegal instruction exception" would not be necessary.  But there are certainly other noun phrases we use as exception types where the hyphen would be mandatory.  This scheme is also friendlier to readers, since it provides explicit phrase grouping.  So, standardize on hyphenation.

Also, fix a couple places where "trap" was used where "exception" was more appropriate.


aswaterman:
(Verbal approval for this PR given by ARC)

Issue #1103, opened 2023-08-29
What is the step-by-step process for two-stage address translation


kdockser:
In the latest pdf build (5/23/23) of the Privileged Architecture, the step-by-step description in **5.3.2 Virtual Address Translation Process** is clear and concise for single stage translation. However, when we move to two-stage translation, the differences specified in  **9.5.1 Guest Physical Address Translation** (e.g., hgatp substitutes for the usual satp) seem to be incomplete.

For example, applying step 7 in **5.3.2**  (as updated by **9.5.1**) "if the original memory access is a store, also set pte.d to 1" is a bit confusing when trying to apply it to G-stage translations. The intent is _probably_ for all traversed G-stage PTEs to have A=1 and, if the operation that kicked off the original translation was a STORE, the final G-stage PTE (i.e. pointing to the SPA of the page of the STORE) needs D=1. We should have a clarification of some sort for this.

Furthermore, the updated steps in **5.3.2** do not provide any guidance on how to update the G-stage PTE that points to the VS-stage leaf cell. Presumably, if the VS-stage leaf cell is updated, this G-stage that points to it must have D=1.

It appears to me that the various models  (SPIKE, RISC-V, and QEMU) are missing this last step (actually, SPIKE is currently being corrected based this issue being raised in  [Are G-stage D bits set correctly as a result of setting a VS-stage D bit? #1447](https://github.com/riscv-software-src/riscv-isa-sim/issues/1447) with the changes in
[A/D updates in G-stage PTE #1448](https://github.com/riscv-software-src/riscv-isa-sim/pull/1448/commits/ae9888e50806837a3b6a39293c1014d9c7c37c56))

I suggest that we add clarifications to the Privileged Architecture in the form of a _two_-stage table in **9.5.1 Guest Physical Address Translation**. Alternatively, we  could add a normative description that explicitly calls out of these differences in two-stage translations,

ved-rivos:
> Alternatively, we could add a normative description that explicitly calls out of these differences in two-stage translations.

The section 9.5.1 does call out these differences:

- `hgatp` substitutes for the usual `satp`;
- For the translation to begin, the effective privilege mode must be VS-mode or VU-mode;
- When checking the `U` bit, the current privilege mode is always taken to be U-mode; and
- Guest-page-fault exceptions are raised instead of regular page-fault exceptions.
[...]

PR #1139, opened 2023-10-05
Clarify A/D setting in G-stage PTE to support VS-stage address translation

ved-rivos:
See issue #1103


ved-rivos:
@gfavor @kdockser

ved-rivos:
Made a bad push. Closing and will resubmit.

PR #1142, opened 2023-10-07
Clarifications to the Addressing and Memory Protection section

ved-rivos:
These clarifications were created as part of the Svadu extension development.

### DIFF
--- a/src/supervisor.adoc
+++ b/src/supervisor.adoc
@@ -1266,11 +1266,10 @@ other.
 The PTE updates due to memory accesses ordered-after a FENCE are not
 themselves ordered by the FENCE.
 
-Simpler implementations that cannot precisely order the PTE update
-before subsequent explicit memory accesses to the associated virtual
-page by the local hart may simply order the PTE update before all
-subsequent explicit memory accesses to any virtual page by the local
-hart.
+Simpler implementations may order the Page Table Entry (PTE) update
+to precede all subsequent explicit memory accesses, as opposed to
+ensuring that the PTE update is precisely sequenced before subsequent
+explicit memory accesses to the associated virtual page.
 
 Prior versions of this specification required PTE A bit updates to be
 exact, but allowing the A bit to be updated as a result of speculation
--- a/src/supervisor.adoc
+++ b/src/supervisor.adoc
@@ -1217,30 +1217,61 @@ Two schemes to manage the A and D bits are permitted:
 
 * When a virtual page is accessed and the A bit is clear, or is written
 and the D bit is clear, a page-fault exception is raised.
-* When a virtual page is accessed and the A bit is clear, or is written
-and the D bit is clear, the implementation sets the corresponding bit(s)
-in the PTE. The PTE update must be atomic with respect to other accesses
-to the PTE, and must atomically check that the PTE is valid and grants
-sufficient permissions. Updates of the A bit may be performed as a
-result of speculation, but updates to the D bit must be exact (i.e., not
-speculative), and observed in program order by the local hart.
-Furthermore, the PTE update must appear in the global memory order no
-later than the explicit memory access, or any subsequent explicit memory
-access to that virtual page by the local hart. The ordering on loads and
-stores provided by FENCE instructions and the acquire/release bits on
-atomic instructions also orders the PTE updates associated with those
-loads and stores as observed by remote harts.
-+
-The PTE update is not required to be atomic with respect to the explicit
-memory access that caused the update, and the sequence is interruptible.
-However, the hart must not perform the explicit memory access before the
-PTE update is globally visible.
+* When a virtual page is accessed and the A bit is clear, the PTE is
+  updated to set the A bit.  When the virtual page is written and the D
+  bit is clear, the PTE is updated to set the D bit. When G-stage address
+  translation is in use and is not Bare, the G-stage virtual pages may be
+  accessed or written by implicit accesses to VS-level memory management
+  data structures, such as page tables.                                 +
+                                                                        +
+  When two-stage address translation is in use, an explicit access may
+  cause both VS-stage and G-stage PTEs to be updated. The following rules
+  apply to all PTE updates caused by an explicit or an implicit memory
+  accesses.                                                             +
+                                                                        +
+  The PTE update must be atomic with respect to other accesses to the
+  PTE, and must atomically perform all tablewalk checks for that leaf
+  PTE as part of, and before, conditionally updating the PTE value.
+  Updates of the A bit may be performed as a result of speculation, even
+  if the associated memory access ultimately is not performed
+  architecturally. However, updates to the D bit, resulting from an
+  explicit store, must be exact (i.e., non-speculative), and observed in
+  program order by the local hart. When two-stage address translation is
+  active, updates of the D bit in G-stage PTEs may be performed as a
+  result of speculative updates of the A bit in VS-stage PTEs.          +
+                                                                        +
+  The PTE update must appear in the global memory order before the
+  memory access that caused the PTE update and before any subsequent
+  explicit memory access to that virtual page by the local hart. The
+  ordering on loads and stores provided by FENCE instructions and the
+  acquire/release bits on atomic instructions also orders the PTE updates
+  associated with those loads and stores as observed by remote harts.   +
+                                                                        +
+  The PTE update is not required to be atomic with respect to the memory
+  access that caused the update and a trap may occur between the PTE
+  update and the memory access that caused the PTE update. If a trap
+  occurs then the A and/or D bit may be updated but the memory access
+  that caused the PTE update might not occur. The hart must not perform
+  the memory access that caused the PTE update before the PTE update is
+  globally visible.                                                     +
+                                                                        +
+  The page tables must be located in memory with hardware page-table
+  write access and _RsrvEventual_ PMA.
 
 All harts in a system must employ the same PTE-update scheme as each
 other.
 
 [NOTE]
 ====
+The PTE updates due to memory accesses ordered-after a FENCE are not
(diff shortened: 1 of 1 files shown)


src/supervisor.adoc
```diff
@@ -1265,6 +1296,9 @@ supervisor software does not rely on accessed and/or dirty bits, e.g. if
 it does not swap memory pages to secondary storage or if the pages are
 being used to map I/O space, it should always set them to 1 in the PTE
 to improve performance.
+
+The PTE updates due to memory accesses ordered-after a FENCE are not
+themselves ordered by the FENCE.
```
aswaterman, line 1301: This sentence appears twice in the same NOTE.  Can you delete one of them?

src/supervisor.adoc
```diff
@@ -1217,30 +1217,61 @@ Two schemes to manage the A and D bits are permitted:
...
 [NOTE]
 ====
+The PTE updates due to memory accesses ordered-after a FENCE are not
+themselves ordered by the FENCE.
+
+Simpler implementations that cannot precisely order the PTE update
+before subsequent explicit memory accesses to the associated virtual
+page by the local hart may simply order the PTE update before all
+subsequent explicit memory accesses to any virtual page by the local
+hart.
```
aswaterman, line 1273: The sentence is a little hard to parse because "implementations that cannot precisely order the PTE update" kinda suggests that ordering the PTE update is not required!

I think reordering the sentence it would fix my concern:
"Simpler implementations may simply order the PTE update before all subsequent
explicit memory accesses by the local hart, rather than precisely ordering the
PTE update before subsequent explicit memory accesses to the associated
virtual page by the local hart."

WDYT?

src/supervisor.adoc
ved-rivos: done

src/supervisor.adoc
ved-rivos: Yes, that is simpler. I pushed and updated with the simpler revised sentence.

src/supervisor.adoc
```diff
@@ -1217,30 +1217,61 @@ Two schemes to manage the A and D bits are permitted:
...
+  The PTE update is not required to be atomic with respect to the memory
+  access that caused the update and a trap may occur between the PTE
+  update and the memory access that caused the PTE update. If a trap
+  occurs then the A and/or D bit may be updated but the memory access
+  that caused the PTE update might not occur. The hart must not perform
+  the memory access that caused the PTE update before the PTE update is
+  globally visible.                                                     +
+                                                                        +
+  The page tables must be located in memory with hardware page-table
+  write access and _RsrvEventual_ PMA.
```
aswaterman, line 1259: Do we need to say what happens when software fails to meet this requirement.  Maybe ", else an access exception may be raised?"  That leaves some wiggle room; I don't think we want to mandate the exception, because uarches might not be using LR/SC as the means to implement this feature.

aswaterman (approved):
OK with me once the last question is addressed.  Thanks, @ved-rivos!

src/supervisor.adoc
ved-rivos: I think that was already addressed by the Address translation process where it states - "If accessing
_pte violates a PMA or PMP check, raise an access-fault exception". Please let me know if I should push an additional statement here.

aswaterman: I'm OK with that interpretation.  We can always add a clarifying statement later if others are confused.

src/supervisor.adoc
```diff
@@ -1217,30 +1217,60 @@ Two schemes to manage the A and D bits are permitted:
...
+  data structures, such as page tables.                                 +
+                                                                        +
+  When two-stage address translation is in use, an explicit access may
+  cause both VS-stage and G-stage PTEs to be updated. The following rules
+  apply to all PTE updates caused by an explicit or an implicit memory
+  accesses.                                                             +
+                                                                        +
+  The PTE update must be atomic with respect to other accesses to the
+  PTE, and must atomically perform all tablewalk checks for that leaf
+  PTE as part of, and before, conditionally updating the PTE value.
```
sorear, line 1234: This appears to say that an implementation must atomically check all intermediate PTEs _in memory_, so on Sv39 an A/D bit update generates a compare-and-compare-and-compare-and-swap bus transaction.  That seems far too strong, is it accurate?

src/supervisor.adoc
```diff
@@ -1217,30 +1217,60 @@ Two schemes to manage the A and D bits are permitted:
...
+  The PTE update must appear in the global memory order before the
+  memory access that caused the PTE update and before any subsequent
+  explicit memory access to that virtual page by the local hart. The
+  ordering on loads and stores provided by FENCE instructions and the
+  acquire/release bits on atomic instructions also orders the PTE updates
+  associated with those loads and stores as observed by remote harts.   +
+                                                                        +
+  The PTE update is not required to be atomic with respect to the memory
+  access that caused the update and a trap may occur between the PTE
+  update and the memory access that caused the PTE update. If a trap
```
sorear, line 1252: Would it be possible to extend this to a "trap or failing sc", so that if a sc causes a PTE update to set the D bit, and that page table update interferes with the sc such that the reservation is lost, the implementation does not need to paradoxically not have updated the page table?

Consider in particular the case where the physical address of the PTE and the physical address written by the sc are the same or in the same cache line, and the page table walker does not have a connection to the L1 data cache.

src/supervisor.adoc
ved-rivos: The statement is stating "checks for that leaf PTE" and  not all intermediate PTEs. There are no A or   D bits in non-leaf PTEs and the table walker is not required to go revisit any non-leaf PTE that  has been already walked in  the past.

src/supervisor.adoc
ved-rivos: The A/D updates occur as part of the address  translation process which is a precursor for the instruction to execute. This process does  not treat instructions that have conditional execution or conditional trapping behavior such as masked vector stores, store conditional, fault-only-first vector loads, or compare-and-swap differently for the purpose of setting A and/or D bit.  So a special mention for sc may not be required.

src/supervisor.adoc
sorear: Possibly unintended interpretation: "must atomically perform all tablewalk checks (all checks in the table walking process, starting from the root) for that leaf PTE (that would be applicable to any translation which uses that leaf PTE and path)"

Possibly intended interpretation: "must atomically perform all tablewalk checks (all checks performed at the last level only) for that leaf PTE (only the checks which directly concern the PTE itself, not those involved in finding it)"

I see your argument, but I think it requires stretching the meaning of the words a bit.

src/supervisor.adoc
sorear: The A extension spec says "If the SC.W fails, the instruction does not write to memory".  I am interpreting that to mean that it does not cause "the virtual page is written" for the purposes of this section.

ved-rivos: Setting D bit is a property of the type of access the instruction could make i.e. a store access and not how the instruction executes. The setting of the D bit is not conditional on the store instructions conditional behavior -  such as the cases mentioned  -  for instance a masked  vector store does not update the elements in memory.

src/supervisor.adoc
ved-rivos: The sentence states "for **that** leaf PTE" and not  for all PTEs leading up to the leaf PTE.

sorear: I'm not objecting to "that", I'm objecting to "for".  "for" can mean "identified by", and the leaf PTE identifies a set of non-leaf PTEs.

If the clause were changed to "atomically perform all tablewalk checks *acting on* that leaf PTE", it would be perfectly clear.

src/supervisor.adoc
sorear: What in the diff supports the interpretation of setting D bits based on accesses an instruction "could" make?

For masked vector stores, especially in the large stride or indexed case, it's important to not require that an implementation perform page table walks for masked-off elements.  The V spec already requires that masked-off elements not generate exceptions, so it would be extremely odd to specify that setting the D bit occurs regardless of the mask value.  We should probably allow imprecise D bit setting in that case, since many implementations will want to perform page table checks for the unit or small stride case optimistically without depending on the mask value.

ved-rivos: This diff does not as this is detailed in the existing section 5.3.2.  The page table walk algorithm and the steps for when the D bit are set  does not include conditionally setting of A and/or D bit based on conditional behavior of instructions.

sorear: 5.3.2 does not mention instructions at all, except for SFENCE.VMA.

ved-rivos: Sorry if I was not being clear. I referred to 5.3.2 to say that the algorithm is agnostic of any conditional execution behavior of the instruction itself i.e. there is no dependence on any conditional execution behavior of instructions for the purpose of setting A and/or D bits

sorear: Is an indexed vector store required to perform address translations independent of the mask values?

Is an indexed vector store required to perform address translations independent of the VL value?

Is an indexed vector store required to perform address translations independent of the LMUL value?

gfavor: Since A&D functionality is orthogonal to any specific load/store instructions, the answer should come from the vector spec.  Also note that the question of updating A&D bits only arises for accesses that pass all MMU/PMP/PMA permission and access restrictions.

So, in short (and assuming one is accessing idempotent memory), "masked" elements (whether due to vstart, mask, or vl) may be accessed.  What the spec generally constrains is "do not raise exceptions and do not update the destination vector register".  Hence masked elements are not required to be translated, but are allowed be translated and accessed - and A&D bits may then be set as a result .

PR #1153, opened 2023-10-16
Fix wrong bit length of rd field

404allen404:
The bit length of rd register should be 5 instead of 3.


wmat:
Good catch, thanks for the fix.

PR #1155, opened 2023-10-30
Fix AsciiDoc error in D-extension chapter

nick-knight:
The underscore was in the wrong place, causing unintentional italicization of subsequent text.

I also added whitespace around the inequality, for consistency with a subsequent sentence.

### DIFF
--- a/src/d-st-ext.adoc
+++ b/src/d-st-ext.adoc
@@ -27,7 +27,7 @@ floating-point precisions supported, including H, F, D, and Q.
 === NaN Boxing of Narrower Values
 
 When multiple floating-point precisions are supported, then valid values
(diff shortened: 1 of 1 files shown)


nick-knight:
Actually, I see another issue a few paragraphs down. And further down, I see `latexmath:[$n$]` being used instead of `_n_`. Maybe that's a better approach. I'll let the AsciiDoc experts chime in.

aswaterman:
@nick-knight If the second typo you were referring to is FL_n_/FS_n_, that's a case where double underscores are needed for correct parsing.  I can't say I understand the rules; lately, I have been using double underscores exclusively so that I can remain ignorant of them.

Using `latexmath` is not preferable, as the fonts don't match (and, less importantly, the code is ugly).

aswaterman:
Fixed with b46c527

PR #1159, opened 2023-11-05
   delete user-mode interrupt notes in mtvec register section

### LINKED ISSUE #1158: user-mode software interrupt in mtvec section
In Machine-Level ISA v1.12, the section describing Machine Trap Vector Base Address register (`mtvec`), there is a Note about user-mode software interrupt cause. But in the list at table 14 there is no such user-mode software interrupt. Is this right?

yf13:
Drop notes on obsoleted user-mode software interrupts   
Resolves #1158


aswaterman (approved):
Thanks.  If the N extension ever comes back to life, we can always revert this PR.

PR #1165, opened 2023-11-22
Fix MODE field description of mtvec register

HepoH3:
Fix MODE field description in the mtvec register

In Section 1.6, "Exceptions, Traps, and Interrupts" of the unprivileged specification, a distinction is made between exceptions and interrupts.

Table 13 in the privileged specification uses the term "exceptions" as a catch-all expression, encompassing both "exceptions and interrupts" which may not be immediately apparent.

Reading "exceptions" merely as "exceptions" while seeing following word "interrupts" can led to believe that the 0th bit is used to control exception handling behavior, while the 1st bit is used to control interrupt handling behavior.

### DIFF
--- a/src/machine.adoc
+++ b/src/machine.adoc
@@ -1087,7 +1087,7 @@ hand, we wish to allow flexibility for larger systems.
 |Direct +
 Vectored +
 ---
-|All exceptions set `pc` to BASE. +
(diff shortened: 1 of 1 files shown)


aswaterman (approved):
Right, with the direct mode, both exceptions and interrupts set pc to BASE.

PR #1177, opened 2023-12-18
Specify that the supervisor physical address is divided by 4 KiB

charlie-rivos:
It appears that during the conversion from Latex to asciidoc, the 4 KiB divisor for the satp description was dropped. This commit brings it back.


aswaterman (approved):

aswaterman:
Thanks for spotting this transliteration bug.

PR #1178, opened 2023-12-19
Add an information about the potential risk with uninitialized MTVEC

Adam-pi3:
More about the problem can be found here:
https://nvd.nist.gov/vuln/detail/CVE-2021-1104

RISC-V website also posted a small article about it here: https://riscv.org/news/2021/08/video-glitching-risc-v-chips-mtvec-corruption-for-hardening-isa-adam-zabrocki-and-alex-matrosov-def-con-29/

The full DefCon presentation can be found here:
https://www.youtube.com/watch?v=iz_Y1lOtX08

### DIFF
--- a/src/machine.adoc
+++ b/src/machine.adoc
@@ -1115,6 +1115,14 @@ implemented without a hardware adder circuit.
 ***
 
 Reset and NMI vector locations are given in a platform specification.
+
+***
+
+A potential security risk exists (e.g. CVE-2021-1104) if the mtvec CSR
+is not initialized to a suitable defined value before a hart starts
+execution out of reset, i.e. boot firmware initialization of mtvec
+results in a window of execution during which mtvec does not have
+a defined value.
 ====
 
 ==== Machine Trap Delegation Registers (`medeleg` and `mideleg`)


allenjbaum:
Interrupts are disabled at reset. If glitching can cause interrupts to be
enabled, that's a potential problem.
Note that while MTVEC has no architecturally defined reset value, that
doesn't mean that an implementation won't.
Generally they will, else design tools get upset with propagating Xs.

And, people who really care about security will have glitch detectors that
prevent this kind of attack.
We also changed the reset value of PMP entries so they may have defined
reset values, primarily to ensure glitching attacks don't enable bad
behavior.
We don't point out the reason for that spec change, but it means there are
resources available.

So, while I think pointing this out is a good idea, I don't think this is
the right document,
as opposed to one that is more security focused.

jnk0le:
> Interrupts are disabled at reset.

MIE bit doesn't affect exceptions/NMIs which are exploited here.

robotman2412:
That is true, but according to the privileged specification (version 20211203) on page 49, NMIs go to a platform-specific NMI vector (not `mtvec`).
The most likely potential problem of course is a synchronous trap happening before `mtvec` is initialized but that can be completely avoided if the startup code is written correctly.

jnk0le:
"implementation-defined" wchich means that it can use `mtvec` anyway. Ibex for example does actually use `mtvec` for its NMIs.

There is also another hole when application expects vectored mode but exception/NMI arrives before setting up vectored mode in `mtvec`.

allenjbaum:
IBEX uses MTVEC? Then how does it know it was an NMI? IT must use an
otherwise undefined cause code, which could overwrite the cause of the trap
that is currently in process. That sounds architecturally broken, if so (as
you suggest.)
I'm not sure "implementation defined" here means a value that can change on
the fly (i.e. if some code updates MTVEC), as opposed to being a constant
defined by the implementation.
If we actually had a reliable way to test it, I'd label that an errata (but
we can't because the cause of an NMI is also implementation dependent, and
there may be no SW method to signal it - though there should be for
self-test.....).

Given a discussion I had recently, it isn't clear that properly written
startup code will completely avoid it; that's what a glitch attack can do,
and it is demonstrable

jnk0le:
> IBEX uses MTVEC? Then how does it know it was an NMI? IT must use an
otherwise undefined cause code

`mtvec` + 0x7c or custom cause code: https://ibex-core.readthedocs.io/en/latest/03_reference/exception_interrupts.html

> which could overwrite the cause of the trap
that is currently in process. That sounds architecturally broken, if so (as
you suggest.)

Ibex had some hidden CSRs for preserving state, but the "standard" NMIs are indeed unrecoverable. RNMI added explicit state to fix that.

src/machine.adoc
```diff
@@ -1115,6 +1115,14 @@ implemented without a hardware adder circuit.
 ***
 
 Reset and NMI vector locations are given in a platform specification.
+
+***
+
+A potential security risk exists (e.g. CVE-2021-1104) if the mtvec CSR
```
kersten1, line 1121:
```suggestion
Potential security risk (e.g. CVE-2021-1104): If the `mtvec` CSR is not initialized to a defined value before a hart executes out of reset, then there is a window of time during the execution when `mtvec` does not have a defined value. For example, during a boot firmware initialization of `mtvec`.
```

src/machine.adoc
```diff
@@ -1115,6 +1115,14 @@ implemented without a hardware adder circuit.
...
+is not initialized to a suitable defined value before a hart starts
```
kersten1, line 1122:
```suggestion
```

src/machine.adoc
```diff
@@ -1115,6 +1115,14 @@ implemented without a hardware adder circuit.
...
+execution out of reset, i.e. boot firmware initialization of mtvec
```
kersten1, line 1123:
```suggestion
```

src/machine.adoc
```diff
@@ -1115,6 +1115,14 @@ implemented without a hardware adder circuit.
...
+results in a window of execution during which mtvec does not have
```
kersten1, line 1124:
```suggestion
```

src/machine.adoc
```diff
@@ -1115,6 +1115,14 @@ implemented without a hardware adder circuit.
...
+a defined value.
```
kersten1, line 1125:
```suggestion
```

kersten1:
Not getting involved in the whether or not this is needed, but wanted to offer up some edits, just in case it goes.

src/machine.adoc
kersten1, line 1121:
```suggestion
A potential security risk exists (e.g. CVE-2021-1104) if the `mtvec` CSR
```

src/machine.adoc
kersten1, line 1124:
```suggestion
results in a window of execution during which a value is not defined for `mtvec`.
```

src/machine.adoc
kersten1, line 1125:
```suggestion
```

ved-rivos:
A note has been included in the Priv. spec through PR #1208 to guide trap handler design and warn against consequences of exceptions in trap handlers.

gfavor:
Adding on to Ved's comment:  That PR was the result of various discussions among the ARC and others about how best to handle this security risk issue.  In short, that PR represents the resolution of what should be done in the ISA manuals (separate from any additional actions that might be taken by the Security HC).  So this PR should be dropped.

allenjbaum:
The problem with PR#1208 is that is written assuming there is a trap handler - the wording talks about the code path - and the whole point of PR#1178 is point out there is an issue prior to there being a trap handler.  I am asserting that PR#1208 is completely inadequate unless it describes the reset process, which was the entire point of  PR#1178 and what the CVE was all about.

ved-rivos:
> The problem with PR#1208 is that is written assuming there is a trap handler - the wording talks about the code path - and the whole point of PR#1178 is point out there is an issue prior to there being a trap handler. I am asserting that PR#1208 is completely inadequate unless it describes the reset process, which was the entire point of PR#1178 and what the CVE was all about.

"There is no trap handler" is a SW concept. If an interrupt or exception is triggered then the hardware traps to the trap vector. If the trap vector is not appropriately setup to the desired value then should not enable interrupts or cause exceptions. This also follows from the guidance "Do not cause exceptions in your trap handler else they may lead to an infinite loop or traps".

allenjbaum:
Again - there is no trap handler at that point, so that bit of guidance
offers no help

ved-rivos:
I dont understand what it means by there being no trap handler. If the address in the trap vector cannot be executed then it will cause an exception - that is warned against - do not cause exceptions in the trap handler else it will lead to an infinite loop".

pdonahue-ventana:
Allen, is your point that the reset vector points to an address that gets an exception, then we go to mtvec which is uninitialized (i.e. there is no trap handler)?  Ved, is your point that the reset vector is itself a handler that should be written so that it does not get an exception until it has initialized things?

ved-rivos:
> and a situation where the hart going into an infinite loop even though no code is being executed.

An exception caused by first instruction of the trap handler - code pointed to by the trap vector - is also an exception in the trap handler. The guidance states that raising raising exceptions in a trap handler - including its first instruction - could lead to an infinite loop of traps.

allenjbaum:
I think Paul has a distinction I'm looking for.
The reset vector is an address, not an instruction. xTVEC contains an
address, not instructions.
Trap handlers contain.... instructions.
No instructions == no handler.
The code at the reset vector (presumably) contains instructions, and those
instructions may or may not get executed under the scenarios that started
this entire discussion.
And, even if they did get executed, there is no guarantee that that code
can do anything at all to prevent this attack vector; it's outside the
scope of instruction execution. I've heard at least one vendors speak/hint
about the measures they need to take to prevent these specific attacks from
working.
they've been bitten.

This would make a great panel session at the summit, where panellist aren't
all agreeing with each other (those are boring).

PR #1190, opened 2024-01-11
Add marchid for Coreblocks

piotro888:
`marchid` 42 allocation request for Coreblocks. It is an experimental, modular out-of-order RISC-V core generator.
Repository link: https://github.com/kuznia-rdzeni/coreblocks


aswaterman (approved):

aswaterman (approved):

aswaterman (changes requested):
@piotro888 There was a race condition between two PRs, and QEMU won the race (because I accidentally merged it before merging yours).  But that ship has sailed, so can you update the PR to pick the next number?

aswaterman (approved):
Sorry again for the mixup.

PR #1204, opened 2024-01-27
Redefine MXL to be read-only

### LINKED ISSUE #1203: Resolve MXL-writability discussion
See https://lists.riscv.org/g/tech-privileged/message/1761 for context.

aswaterman:
Resolves #1203 (that is, if we choose to resolve it in this manner).


aswaterman:
@gfavor I'm going to tentatively merge this, but we'll revert it if #1203 is resolved in the opposite direction.

PR #1206, opened 2024-01-29
Remove Zam; define misaligned atomicity granule PMA

### LINKED ISSUE #1114: Define misaligned atomicity granule PMA
This is intended to supersede the proposed Zam extension.

aswaterman:
Resolves #1114


ved-rivos:
> reasons of address alignment, and the instruction will give rise to only one memory operation for the purposes of RVWMO--i.e., it will execute atomically.

Should also be same for the purposes of RVTSO.

aswaterman:
@ved-rivos Since RVTSO is specified as "RVWMO plus additional ordering rules", that follows.  In other similar cases, we've avoided mentioning RVTSO to avoid cluttering the text, and I'd like to stick to that discipline.

PR #1207, opened 2024-01-29
Attempts to access non-existent CSRs are reserved

### LINKED ISSUE #1116: Accessing an unimplemented CSR is reserved, not necessarily illegal
The old clause was incorrect, as sometimes a virtual instruction exception must be raised instead.

(It's also inconsistent with the rest of RISC-V, which recommends but does not mandate exceptions for reserved encodings.  And, furthermore, the existence of non-conforming extensions made the old mandate toothless: an implementation could purport to have a non-conforming extension at every otherwise-unused CSR address.)

aswaterman:
Resolves #1116

### DIFF
--- a/src/hypervisor.adoc
+++ b/src/hypervisor.adoc
@@ -1700,6 +1700,7 @@ value, obviating the need to execute an HFENCE.VVMA instruction.
 
 === Traps
 
+[[sec:hcauses]]
 ==== Trap Cause Codes
 
 The hypervisor extension augments the trap cause encoding.
--- a/src/priv-csrs.adoc
+++ b/src/priv-csrs.adoc
@@ -42,10 +42,12 @@ accesses to be intercepted. This change should be transparent to the
 less-privileged software.
 ====
 
-Attempts to access a non-existent CSR raise an illegal-instruction
-exception. Attempts to access a CSR without appropriate privilege level
-or to write a read-only register also raise illegal-instruction
-exceptions. A read/write register might also contain some bits that are
+Instructions that access a non-existent CSR are reserved.
+Attempts to access a CSR without appropriate privilege level
+raise illegal-instruction exceptions or, as described in
+<<sec:hcauses>>, virtual-instruction exceptions.
+Attempts to write a read-only register raise illegal-instruction exceptions.
+A read/write register might also contain some bits that are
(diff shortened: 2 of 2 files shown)


allenjbaum:
If there is an attempt to access a non-existent CSR (or write a
non-existent CSR in the Read-only region) - is there no guarantee even then
that it will get an illegal instruction exception?
Or, it is guaranteed only for implemented CSRs at an insufficient priv
level? (i.e. which has higher priority, insufficient priv or unimplemented)?

I would vote for insufficient priv level here.....

aswaterman:
Unimplemented CSRs don’t have a defined privilege level, so they can’t trap on the basis of privilege (The use of bits 9:8 for privilege is just a convention, as is the use of bits 11:10 for writability.)

allenjbaum:
Hmm; given the description of the regions in the spec, I thought of it as
far more than a convention, but more like a ... spec.

aswaterman:
The title of the section that describes the encoding uses the word "conventions"...

pdonahue-ventana:
Notably, the ratified debug 0.13 spec did not follow that convention.  The mscontext CSR in debug 1.0 continues to not follow that convention but it's now deprecated and replaced by something that follows the convention.

gfavor (approved):
LGTM.

jrtc27:
Both BBL and OpenSBI probe CSR existence by whether they trap, e.g. for PMP registers (but more extensive in OpenSBI). Does this loosening of the spec not break that such that now _conforming_ implementations don't have to trap on accessing those CSRs if they don't implement them, whereas previously it was only non-conforming implementations, due to the CSRs being in the standard encoding space?

aswaterman:
In effect, they are requiring something along the lines of Ssstrict (or Smstrict, really, which hasn't been defined).  That may well be the right thing for them to do, but the assumption should be codified.

jrtc27:
If we're going to add things then we might as well just spell out in the hardware description, as we do for something like the Zicbo* cache block sizes, how many registers there are (or whatever else OpenSBI probes for, which may just be giving extension names to things). But at the moment we're in a bit of a dodgy spot where the spec relaxed something that was relied upon, but there's nothing in its place listed as a requirement by those implementations.

jrtc27:
Cc @atishp04 @avpatel as OpenSBI and platform-y people

PR #1208, opened 2024-01-29
Add note to guide trap handler design

ved-rivos:


ved-rivos:
@gfavor @aswaterman

ved-rivos:
> if there is no code path because there is no code

Then it causes an access fault or a page fault. The guidance says - do not cause exceptions in your trap handler.

> there is no code because the trap handler (or the address of the trap handler) has not been set up yet

There is no "valid" for the `stvec` or `mtvec` - it may be pointing to an address that may cause an access fault or page fault - this follows "write your trap handler to not cause an exception in your trap handler" i.e. do not enable interrupts or trigger exceptions.

allenjbaum:
as above: you keep talking about a trap handler, but there isn't a trap
handler in these situations. No code is executed.

PR #1210, opened 2024-01-30
Add clarifying non-normative comments about [m]time

### LINKED ISSUE #1113: Clarify constraints on [m]time resolution and synchronization
Following are some notes from the ARC minutes, courtesy @gfavor:

- The current normative Unpriv text specifies strong constraints on 'time' values, e.g. the 'time' CSR increments by one with each tick of the real-time clock at a known discoverable frequency, and 'time' must be synchronized across harts to within one ulp.  For example, for a 1 GHz 'time' tick frequency this would require a very tight 1ns synchronization between harts.
- Then there is a non-normative "as-if" note expressing that as long as software cannot discern a more than one ulp difference between two hart's time values, then the synchronization requirement can be considered as satisfied.
- The question is then whether an implementation can, for example, have a 1 GHz 'time' counter (i.e. with a 1ns one ulp resolution) that is only updated at, say 100 MHz, with increments by 10 ulp's.  Literally this violates the normative text.  But is it ok IF software cannot successfully observe this violation - which ties back to the "as-if" text in the spec.
- The net conclusion by the ARC, to strike a pragmatic balance, is that this would be ok within the context of the current spec.  With emphasis that software also must never observe 'time' across harts as ever going backwards.
- A platform spec can then, for example, specify the apparent 'time/mtime' tick frequency (e.g. 1 GHz) and also a minimum frequency (e.g. 100 MHz) at which software is guaranteed to observe updated time values.  Software may read time more frequently, but it should only observe monotonic values and it should observe a new value at least once every 10 ns (corresponding to the 100 MHz in this example).
  - Note that this is different from specifying a minimum update frequency in that this focuses on a minimum rate at which new values can be observed by software.

aswaterman:
Text contributed by @gfavor.

@kasanovic This corresponds to the conclusion of an ARC meeting several months ago, but I thought you'd like the chance to weigh in before we merge this one.

Resolves #1113

### DIFF
--- a/src/counters.adoc
+++ b/src/counters.adoc
@@ -147,6 +147,22 @@ As with other architectural mandates, it suffices to appear "as if"
 harts are synchronized to within one tick of the real-time clock, i.e.,
 software is unable to observe that there is a greater delta between the
 real-time clock values observed on two harts.
+
+If, for example, the real-time clock increments at a frequency of 1 GHz, then
+all harts must appear to be synchronized to within 1 nsec.
+But it is also acceptable for this example implementation to only update the
+real-time clock at, say, a frequency of 100 MHz with increments of 10 ticks.
+As long as software cannot observe this seeming violation of the above
+synchronization requirement, and software always observes time across harts to
+be monotonically nondecreasing, then this implementation is compliant.
+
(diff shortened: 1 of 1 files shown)


kasanovic (approved):
LGTM

PR #1213, opened 2024-01-31
Add a QEMU architecture ID

palmer-dabbelt:
For the cases where QEMU isn't pretending to be something else (for example, when targeting a virtual platform) it'd be nice to set the arch IDs to something other than 0.  That way users can detect they're running on QEMU and act differently, in case they want to work around a QEMU issue of some sort.

Link: https://android.googlesource.com/platform/bionic/+/refs/heads/main/libc/arch-riscv64/dynamic_function_dispatch.cpp#47
Link: https://lore.kernel.org/r/20240131182430.20174-1-palmer@rivosinc.com/


aswaterman (approved):
Agreed.

PR #1217, opened 2024-02-06
Zawrs - Wait-on-Reservation-Set chapter integration.

wmat:
This PR integrates the Zawrs specification into the Unprivileged spec.

### DIFF
--- a/src/zawrs.adoc
+++ b/src/zawrs.adoc
@@ -1,0 +1,131 @@
+== RISC-V Wait-on-Reservation-Set (Zawrs) extension
+
+// Preamble
+[WARNING]
+.This document is in the link:http://riscv.org/spec-state[Ratified]
+====
+No changes are allowed. Any desired or needed changes can be the subject of a
+follow-on new extension. Ratified extensions are never revised
+====
+
+=== Copyright and license information
+This specification is licensed under the Creative Commons 
+Attribution 4.0 International License (CC-BY 4.0). The full 
+license text is available at
+https://creativecommons.org/licenses/by/4.0/.
+
+Copyright 2022 by RISC-V International.
+
+=== Contributors
+
+This RISC-V specification has been contributed to directly or indirectly by:
+
+Aaron Durbin, Abel Bernabeu, Allen Baum, Christoph Müllner, David Weaver, Greg Favor, Josh Scheid, Ken Dockser, Paul Donahue, Phil McCoy, Philipp Tomsich, Tariq Kurd, Ved Shanbhogue
+
+=== Introduction
+The Zawrs extension defines a pair of instructions to be used in polling loops 
+that allows a core to enter a low-power state and wait on a store to a memory 
+location. Waiting for a memory location to be updated is a common pattern in 
+many use cases such as:
+
+. Contenders for a lock waiting for the lock variable to be updated.
+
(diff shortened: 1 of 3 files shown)


ved-rivos (approved):

aswaterman (approved):
As with other cases, this has the downside that it isn't stylistically consistent with the older specs that are already in this document, but it's better to integrate them now and fix the style later.

src/zawrs.adoc
```diff
@@ -0,0 +1,115 @@
+== RISC-V Wait-on-Reservation-Set (Zawrs) extension
+
+=== Contributors
+
+This RISC-V specification has been contributed to directly or indirectly by:
+
+Aaron Durbin, Abel Bernabeu, Allen Baum, Christoph Müllner, David Weaver, Greg Favor, Josh Scheid, Ken Dockser, Paul Donahue, Phil McCoy, Philipp Tomsich, Tariq Kurd, Ved Shanbhogue
+
+=== Introduction
```
ved-rivos, line 9: @wmat - whats the convention we have been following with contributors. If this list gets folded into the top level contributors section then it can be dropped from this chapter. Once that is dropped, the "Introduction" heading can also be removed.

wmat: I've been leaving the contributors section of each chapter as is.  My thinking is that each chapter may have a distinct set of contributors who would only prefer attribution for that chapter.  However, I'm also open to rolling all contributors into the overall contributors list.  I'm not sure which is best.  Another option would be expanding the overall contributors list but breaking it out by chapter.

ved-rivos: For other extensions integrated into the manual (such as Svpbmt, Svinval, Ztso) the manual has not had contributors per chapter. Having contributors per chapter would be new. Maybe something to discuss about maintaining the overall look and feel of the manual.

wmat: Yes, I believe consistency across chapters is important.  And I'm leaning now toward there only being one main contributrs section.

ved-rivos: Agree.

wmat: I have moved contributors to the top level contributors section and deleted the section and introduction title.

ved-rivos: Somehow that commit has not made it in. I still see the contributors section and the introduction title..

wmat: Perhaps you looked right before I pushed?

ved-rivos: Looks good to me. Thanks!

src/zawrs.adoc
```diff
@@ -0,0 +1,107 @@
...
+=== Wait-on-Reservation-Set Instructions
+
+The `WRS.NTO` and `WRS.STO` instructions cause the hart to temporarily stall
+execution in a low-power state as long as the reservation set is valid and no
+pending interrupts, even if disabled, are observed. For `WRS.STO` the stall 
+duration is bounded by an implementation defined short timeout. These 
+instructions are available in all privilege modes. These instructions are not
+supported in a constrained `LR`/`SC` loop.
+
+*Encoding:*
```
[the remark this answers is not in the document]
wmat: Pushed fix.

wmat:
Merging this PR as per Andrew & Ved's confirmation.  Original repo to be archived.

PR #1218, opened 2024-02-06
Integration of Smstateen extension chapter.

wmat:
Added the applicable adoc content from the Smstateen repository to the smstateen.adoc file and included in the Priv header.  Set chapter anchor, titled to be consistent with other extension chapter titles in priv and set header levels appropriately.

### DIFF
--- a/src/smstateen.adoc
+++ b/src/smstateen.adoc
@@ -1,0 +1,335 @@
+[[smstateen]]
+== "Smststeen" State Enable Extension
+
+=== Motivation
+
+The implementation of optional RISC-V extensions has the potential to open
+covert channels between separate user threads, or between separate guest OSes
+running under a hypervisor. The problem occurs when an extension adds processor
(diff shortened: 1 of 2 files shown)


src/smstateen.adoc
```diff
@@ -0,0 +1,335 @@
...
+
+bit 57 - `scontext`
+
+bits 60:58 - Reserved for the RISC-V Advanced Interrupt Architecture
+
+bit 61 - Reserved for a possible `senvcfg2`
+
+bit 62 - `senvcfg`
+
+bit 63 - `sstateen0`
```
ved-rivos, line 255: Should be converted into a register layout diagram.

wmat: Prior to merge or can this be done post merge?  If pre-merge, someone will have to do it.

wmat: @ved-rivos so what do you think, can the register layout diagram be added post chapter merge?

PR #1219, opened 2024-02-06
Smepmp - Integration of PMP Enhancements chapter.

wmat:
Adding smepmp chapter to the Privileged Specification.


sorear:
Even if the current version of asciidoctor can't use it for whatever reason, I'd like to have the SVG in the repository so that we aren't stuck using bitmapped graphics forever.

Suggest renaming MMWP to MMDD (Machine Mode Default Deny) to make the meaning of 0 and 1 more immediately obvious, and also to comply with the RISC-V Friendly Terminology Policy.

PR #1222, opened 2024-02-09
How should xPIE be set ?

OccupyMars2025:
## In the next paragraph, the manual says:
```
An MRET or SRET instruction is used to return from a trap in M-mode or
S-mode respectively. When executing an __x__RET instruction, supposing
__x__PP holds the value _y_, __x__IE is set to __x__PIE;
```
## This sentence reinforces my belief that my revision may be correct


aswaterman:
I can understand why you’d find this to be a thorny aspect of the spec, but it’s correct as written.

OccupyMars2025:
Thank you for your information. So much to learn.

PR #1239, opened 2024-02-22
Explicitly allow side effects for a failed SC

sorear:
Since PTE updates are specified to occur as a result of "memory accesses" it is possible to interpret the old wording as requiring no PTE update for a failed SC, since there is no memory access.  However, the PTE update is part of the translation process, which on natural implementations will occur before the reservation validity check can occur.

This is consistent with the wording in "Addressing and Memory Protection" which allows traps to prevent the memory access without preventing the PTE update.

(Filing this in a dedicated issue so it can be considered for 1.13.)

### DIFF
--- a/src/a-st-ext.adoc
+++ b/src/a-st-ext.adoc
@@ -62,10 +62,11 @@ if the reservation is still valid and the reservation set contains the
 bytes being written. If the SC.W succeeds, the instruction writes the
 word in _rs2_ to memory, and it writes zero to _rd_. If the SC.W fails,
 the instruction does not write to memory, and it writes a nonzero value
-to _rd_. Regardless of success or failure, executing an SC.W instruction
-invalidates any reservation held by this hart. LR.D and SC.D act
-analogously on doublewords and are only available on RV64. For RV64,
-LR.W and SC.W sign-extend the value placed in _rd_.
+to _rd_.  For the purposes of memory protection, a failed SC.W may be
+treated like a store.  Regardless of success or failure, executing an
+SC.W instruction invalidates any reservation held by this hart. LR.D and
+SC.D act analogously on doublewords and are only available on RV64. For
+RV64, LR.W and SC.W sign-extend the value placed in _rd_.
 
 [NOTE]
 ====


aswaterman:
I'll bring this up in the ARC meeting next week.  I certainly agree with the intent with respect to PTE A/D bits.

My first question, as IIRC @gfavor mentioned on another thread, is whether this is already a consequence of something already written in the spec.  (I can't figure out what that might be, but he might know.)

My second question is whether "side effects associated with a memory access by the privileged architecture" is too vague or even too broad.  Do we want to include MMIO side effects in that list?  (After all, they arguably fit that description.)  The answer might well be "we don't care", but it would need to be discussed.

We might prefer wording like "for the purposes of memory protection, a failed SC may be treated like a store".  To me, this would subsume the PMP issue and the A/D bit issue, while being less vague.

gfavor:
I agree that the current text should be clarified, and in the sort of way that is proposed.  I'm not aware of something elsewhere in the spec that makes this matter clear.  Although even if there is something somewhere, by the very fact that it isn't obvious to us suggests that it would still be good to add this clarification.  Especially since a literal reading of the current text indicates that notraps and side effects should occur.

I also agree with Andrew that the added text should not be so vague (for generality sake) about what traps and side-effects are being referred to.  That's just creating a vague "exercise for the reader".  And even if a reader figures out much of what is being referred to, how do they know that they have identified all the things that are being referred to.  Off-hand I like Andrew's suggestion for less vague text.

gfavor:
And, to be clear, I think the current text change is not quite suitable as a final version. Something like Andrew's suggestion needs to be incorporated.

sorear:
I'm fine with Andrew's wording.

aswaterman:
Thanks for updating the PR.  We should still discuss this one in the ARC meeting, but hopefully I can simply click the merge button afterwards.

aswaterman (approved):
The ARC has chosen to proceed with this approach, which clarifies the implementation space and avoids imposing any backwards-incompatible constraints.

aswaterman:
Thank you, @sorear!

PR #1243, opened 2024-02-27
Fix typo: Quotation mark formatting mismatch in intro.adoc.

Yakkhini:
“yield" -> "yield"


Yakkhini:
image

Above screenshot is part of the current content in Page 4, Volume 1.
Maybe this commit could fix that.

aswaterman (approved):

PR #1246, opened 2024-02-27
Clarify when SFENCE.W.INVAL/SFENCE.INVAL.IR are legal

aswaterman:


ingallsj (approved):

src/supervisor.adoc
```diff
@@ -2021,6 +2021,11 @@ or VU-mode, or to execute SINVAL.VMA in VU-mode, raises a
 virtual-instruction exception. When `hstatus`.VTVM=1, an attempt to execute
 SINVAL.VMA in VS-mode also raises a virtual instruction exception.
 
+Attempting to execute SFENCE.W.INVAL or SFENCE.INVAL.IR in U-mode or VU-mode
+raises an illegal-instruction exception.
```
ved-rivos, line 2025: Should raise a virtual-instruction in VU-mode as these instructions are HS-qualified.

gfavor: Good catch.  Otherwise looks good to me.

aswaterman: This is what code review is for!

PR #1248, opened 2024-02-29
Update Smrnmi to account for Smdbltrp extension

aswaterman:
This is a backwards compatible change to support resumable double traps within M-mode when both extensions are implemented.  Although the resumability feature is not terribly interesting, the reuse of the Smrnmi state for M-mode double traps aids debugging (and fault diagnosis for the hardware-error exception case).

### DIFF
--- a/src/rnmi.adoc
+++ b/src/rnmi.adoc
@@ -71,9 +71,15 @@ of holding.
 .Resumable NMI cause `mncause`.
 include::images/bytefield/mncause.edn[]
 
-The `mncause` CSR holds the reason for the NMI, with bit MXLEN-1 set to
-1, and the NMI cause encoded in the least-significant bits or zero if
-NMI causes are not supported.
+The `mncause` CSR holds the reason for the NMI.
+If the reason is an interrupt, bit MXLEN-1 is set to 1, and the NMI
+cause is encoded in the least-significant bits.
+If the reason is an interrupt and NMI causes are not supported, bit MXLEN-1 is
+set to 1, and zero is written to the least-significant bits.
+If the reason is an exception within M-mode that results in a double trap as
+specified in the Smdbltrp extension, bit MXLEN-1 is set to 0 and the
+least-significant bits are set to the cause code corresponding to the
+exception that precipitated the double trap.
 
 .Resumable NMI status register `mnstatus`.
 include::images/bytefield/mnstatus.edn[]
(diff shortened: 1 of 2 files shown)


src/rnmi.adoc
```diff
@@ -71,9 +71,14 @@ of holding.
...
-The `mncause` CSR holds the reason for the NMI, with bit MXLEN-1 set to
-1, and the NMI cause encoded in the least-significant bits or zero if
-NMI causes are not supported.
+The `mncause` CSR holds the reason for the NMI.
+If the reason is an interrupt, bit MXLEN-1 is set to 1, and the NMI
+cause is encoded in the least-significant bits.
+If the reason is an interrupt and NMI causes are not supported, bit MXLEN-1 is
+set to 1, and zero is written to the least-significant bits.
+If the reason is an exception within M-mode that results in a double trap as
+specified in the Smdbltrp exception, bit MXLEN-1 is set to 0 and the
```
gfavor, line 80: exception => extension

src/rnmi.adoc
```diff
@@ -71,9 +71,14 @@ of holding.
...
+least-significant bits are set to the exception cause code.
```
gfavor, line 81: The cause code for which exception?  It's probably sufficient here to just say that the lsb's are set to the exception cause code specified in the Smdbltrp extension.

aswaterman: Yeah, referencing the Smdbltrp spec is probably best.  Using the "unexpected trap" term of art could confuse readers who haven't yet read the other spec.

aswaterman: After writing it, I decided the answer is to do both: "unexpected trap, as defined in the Smdbltrp specification".

aswaterman: I went back and forth, but I think I found a phrasing that is comprehensible to people who haven't read the other spec.

aswaterman:
Are you looking at the old version somehow?  It looks right to me (although apparently I'm dyslexic with respect to these words, so I'm not a reliable witness).

gfavor:
Right now I'm only on a phone and can only see the text included in the
email from GitHub - which apparently is not showing the latest changes that
you're describing.  Later I'll check properly from my laptop and presumably
will see everything looking hunky dory. :)

Greg

gfavor:
Strange how github emails show changes, but apparently not these latest
changes.  (I haven't noticed this before, but then again usually there
isn't change on top of change on top of an original change.)  Now looking
at the changes directly within github, they look good.

Greg

PR #1256, opened 2024-03-12
Clarify behavior of CSR access side effects

aswaterman:
See https://lists.riscv.org/g/tech-unprivileged/message/717


aswaterman:
Approved out-of-band by @jhauser-us.

gfavor:
FWIW, LGTM as well.

Greg

PR #1270, opened 2024-03-19
Add marchid for rrv32

SolraBizna:
This PR assigns ID 45 to rrv32, because that is the next `marchid` in line. If permissible, I would prefer to use 1495879317 instead. This larger ID encodes a memorial to my father, who passed away 13 months ago. He set and guided me on the 25-year path that led to the creation of rrv32. That ID still has the high bit unset, which means it is technically within the open source `marchid` space. (However, I could easily see such a huge ID being inconvenient and annoying for those who want to make a simple lookup table for `marchid` values.)


SolraBizna:
Is there a procedure I have not followed, a requirement I have not met, or a problem I have created?

aswaterman (approved):
@SolraBizna Sorry, these PRs fall through the cracks sometimes.

PR #1272, opened 2024-03-19
Integrate zacas chapter.

wmat:
Pull in the zacas spec.  Update bibliography and contributors.

### DIFF
--- a/src/zacas.adoc
+++ b/src/zacas.adoc
@@ -1,0 +1,260 @@
+== Atomic Compare-and-Swap (CAS) instructions (Zacas), Version 1.0.0
+
+=== Introduction
+
+Compare-and-Swap (CAS) provides an easy and typically faster way to perform
+thread synchronization operations when supported as a hardware instruction. CAS
+is typically used by lock-free and wait-free algorithms. This extension proposes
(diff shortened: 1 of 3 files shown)


src/zacas.adoc
```diff
@@ -0,0 +1,260 @@
+== Atomic Compare-and-Swap (CAS) instructions (Zacas), Version 1.0.0
+
+=== Introduction
```
ved-rivos, line 3: We could remove the "Introduction" subheading.

src/zacas.adoc
```diff
@@ -0,0 +1,260 @@
...
+  amocas.q.aqrl a6, a4, (a0)
+  ret                    # Enqueue done
+move_tail:               # Tail was not pointing to the last node
+  addi a3, a7, 1         # Try to swing Tail to the next node
+  amocas.q.aqrl a6, a2, (a0)
+  j    enqueue           # Retry
+
+====
+
+=== Additional AMO PMAs
```
ved-rivos, line 244: This section should ideally be part of AMO PMA subsection in the Priv specifcation. 
When moved into the AMO PMAs section, we can remove the first sentence.  
@gfavor - what do you think?

PR #1278, opened 2024-03-20
Reorder priv spec chapters and fix other ToC issues

aswaterman:


src/riscv-privileged.adoc
```diff
@@ -83,12 +83,12 @@ include::priv-csrs.adoc[]
 include::machine.adoc[]
 include::smstateen.adoc[]
 include::smepmp.adoc[]
+include::smcntrpmf.adoc[]
 include::rnmi.adoc[]
 include::supervisor.adoc[]
```
aswaterman, line 88: Note to @gfavor to help with reviewing this PR: `supervisor.adoc` also contains Svpbmt, Svinval, and Svadu, so they follow the supervisor chapter and precede Sscofpmf.

gfavor: Overall looks good.  But shouldn't indirect-csr go near the top, right after priv-csrs?

And just a nit, but maybe put sstc before (instead of after) sscofpmf?  In some sense, sstc is related to more central supervisor functionality (versus sscofpmf is "just" counter-related functionality).

aswaterman: I think indirect-csr could be moved earlier, but it shouldn't come before the machine chapter.  I'll do that.

Agreed re: Sstc.

PR #1279, opened 2024-03-20
Reorder unpriv chapters more logically

aswaterman:
The basic principle is to group all Zy extensions together, following Y.


kasanovic:
I think we need to put the basic base extensions zifence, zicsr, counters right after the base ISAs before the other Zi* extensions.  Otherwise looks mostly OK, though it will all obviously change as we iterate and condense.

aswaterman:
SGTM.  Merging after making that adjustment.

PR #1287, opened 2024-03-25
Integrate B standard extension text

ved-rivos:
Should wait for ratification prior to merge.
Integrates text from: https://github.com/riscv/riscv-b/blob/main/b.adoc

### DIFF
--- a/src/b-st-ext.adoc
+++ b/src/b-st-ext.adoc
@@ -1,6 +1,9 @@
 [[bits]]
 == "B" Standard Extension for Bit Manipulation, Version 1.0.0
 
+The B standard extension comprises instructions provided by the Zba, Zbb, and
+Zbs extensions.
+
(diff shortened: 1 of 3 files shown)


gfavor:
Is there a need to wait all the way until ratification.  We don't do this for other extensoins (e.g. we include not only frozen extensions but also draft extensions).

From a community perspective, it would seem more desirable to integrate B sooner than later (with of course the standard up top marking as to its official state).

ved-rivos:
The process discussed in the chairs was to provide a PR as a step in the Ratification Ready milestone qualification. This extension completed public review and is on its way to IC/HC chairs for Ratification Ready vote. I do not think integrating now in Frozen state and updating the state later to Ratified would be an issue.

aswaterman (approved):

PR #1288, opened 2024-03-25
Add Smcdeleg/Ssccfg

bcstrongx:
The ratification-ready PR for Smcdeleg/Ssccfg


aswaterman (commented):
Thanks for taking the time to properly integrate the spec with the mainline priv spec.

Don't you need to add an `include` to `riscv-privileged.adoc` to slurp in `smcdeleg.adoc`?

bcstrongx:
Yes!  I had one locally that I guess I failed to commit.  Will fix.

aswaterman (approved):

PR #1290, opened 2024-03-26
Fix typo in sscofpmf.adoc

huxuan0307:


aswaterman (approved):
Thanks for spotting this.

PR #1304, opened 2024-03-28
Integrate Zaamo and Zalrsc text

ved-rivos:
Spec: https://github.com/riscv/riscv-zaamo-zalrsc
Status: Frozen

### DIFF
--- a/src/a-st-ext.adoc
+++ b/src/a-st-ext.adoc
@@ -1,7 +1,7 @@
 [[atomics]]
-== "A" Standard Extension for Atomic Instructions, Version 2.1
+== "A" Extension for Atomic Instructions, Version 2.1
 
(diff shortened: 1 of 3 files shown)


src/a-st-ext.adoc
```diff
@@ -18,6 +18,10 @@ appear to have finally settled on release consistency as the standard
 memory consistency model and so the RISC-V atomic support is built
 around this model.
 ====
+
+The A standard extension comprises instructions provided by the Zaamo and Zalrsc
```
wmat, line 22: I believe we can remove the word standard here, as well as from the section headings in order to be consistent with the rest of the spec.

ved-rivos: Updated.

aswaterman (approved):

PR #1307, opened 2024-03-28
Swap scalar and vector crypto chapters

nibrunieAtSi5:
Shouldnt' scalar crypto "Volume I"
> Cryptography Extensions Volume I: Scalar & Entropy Source Instructions, Version 1.0.1
appear before vector crypto "Volume II"
> RISC-V Cryptography Extensions Volume II: Vector Instructions, Version 1.0.0

and maybe we could unify the titles


wmat:
This makes sense, thanks.

PR #1315, opened 2024-04-03
Sstc editorial fixes.

wmat:
The second to last paragraph starts "When STCE in menvcfg is zero, ...".  This should be clarified to say "When this extension is implemented and STCE in menvcfg is zero, ...".  And ditto for the beginning of the last paragraph.

### DIFF
--- a/src/sstc.adoc
+++ b/src/sstc.adoc
@@ -179,12 +179,12 @@ enables stimecmp for S-mode when set to one, and the same bit of henvcfg
 enables vstimecmp for VS-mode. These STCE bits are WARL and are hard-wired to 0
 when this extension is not implemented.
 
-When STCE in menvcfg is zero, an attempt to access stimecmp or vstimecmp in a
+When this extension is implemented and STCE in menvcfg is zero, an attempt to access stimecmp or vstimecmp in a
 mode other than M-mode raises an illegal instruction exception, STCE in henvcfg
 is read-only zero, and STIP in mip and sip reverts to its defined behavior as
-if this extension is not implemented.
+if this extension is not implemented. Further, if the H extension is implemented, then hip.VSTIP also reverts its defined behavior as if this extension is not implemented.
 
-When STCE in menvcfg is one but STCE in henvcfg is zero, an attempt to access
+When this extension is implemented and STCE in menvcfg is one but STCE in henvcfg is zero, an attempt to access
 stimecmp (really vstimecmp) when V = 1 raises a virtual instruction exception,
 and VSTIP in hip reverts to its defined behavior as if this extension is not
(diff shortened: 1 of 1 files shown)


src/sstc.adoc
```diff
@@ -179,12 +179,12 @@ enables stimecmp for S-mode when set to one, and the same bit of henvcfg
...
 
-When STCE in menvcfg is zero, an attempt to access stimecmp or vstimecmp in a
+When this extension is implemented and STCE in menvcfg is zero, an attempt to access stimecmp or vstimecmp in a
 mode other than M-mode raises an illegal instruction exception, STCE in henvcfg
 is read-only zero, and STIP in mip and sip reverts to its defined behavior as
-if this extension is not implemented.
+if this extension is not implemented. Further, if the H extension is implemented, then hip.VSTIP also reverts its defined behavior as if this extension is not implemented.
 
-When STCE in menvcfg is one but STCE in henvcfg is zero, an attempt to access
+When this extension is implemented and STCE in menvcfg is one but STCE in henvcfg is zero, an attempt to access
```
aswaterman, line 187: We can undo this last change because menvcfg.STCE cannot be 1 if this extension is not implemented.

gfavor: Andrew is correct.  My mild concern is that this leaves an asymmetry between this and the preceding statement - that are saying symmetric things.  Hence creating an implied "exercise for the reader" to realize why there is the asymmetry.
For a concise compromise - that creates a sense of the second statement being a continuation of the first statement - how about starting the second statement "But when STCE in menvcfg is one and STCE in henvcfg is zero, ..."?

Andrew, if you don't like that, then we can stick with dropping the change.  (Although I do also like replacing the 'but' with an 'and' in what is a logical 'A and B' conjunction.)

gfavor: Bill, the other two changes look good.

aswaterman: I like your proposed alternative, @gfavor.

gfavor: Bill, you can go ahead and switch the original change to this modified change.  And then you can merge this whole PR of three changes.

gfavor (approved):
Assuming that change update, I'm good overall with this PR.

aswaterman (changes requested):
I'll impose a "request changes" just so I get the chance to review the final outcome.

gfavor:
Certainly.  I just wanted to get one approval out of the way so Bill can proceed as soon as you give approval.

src/sstc.adoc
wmat: I've made this change .

aswaterman (approved):

PR #1316, opened 2024-04-03
Sscofpmf editorial fixes.

wmat:
For the beginning of the descriptions of bits 58 and 59, "If set, ..." should instead be "If set and the H extension is implemented, ...".

In the last paragraph, "In M and S modes, ..." should instead say "In S mode, ...".

### DIFF
--- a/src/sscofpmf.adoc
+++ b/src/sscofpmf.adoc
@@ -62,9 +62,9 @@ bit [61] +++SINH+++ - If set, then counting of events in S/HS-mode is inhibited
 
 bit [60] +++UINH+++ - If set, then counting of events in U-mode is inhibited
 
-bit [59] +++VSINH+++ - If set, then counting of events in VS-mode is inhibited
+bit [59] +++VSINH+++ - If set and the H extension is implemented, then counting of events in VS-mode is inhibited
 
-bit [58] +++VUINH+++ - If set, then counting of events in VU-mode is inhibited
+bit [58] +++VUINH+++ - If set and the H extension is implemented, then counting of events in VU-mode is inhibited
 
 bit [57] 0 - Reserved for possible future modes
 
@@ -183,7 +183,7 @@ M-mode).
 
 Read access to bit _X_ is subject to the same mcounteren (or mcounteren and
 hcounteren) CSRs that mediate access to the hpmcounter CSRs by S-mode (or
-VS-mode). In M and S modes, scountovf bit _X_ is readable when mcounteren bit
+VS-mode). In S modes, scountovf bit _X_ is readable when mcounteren bit
 _X_ is set, and otherwise reads as zero. Similarly, in VS mode, scountovf bit
 _X_ is readable when mcounteren bit _X_ and hcounteren bit _X_ are both set,
 and otherwise reads as zero.


src/sscofpmf.adoc
```diff
@@ -62,9 +62,9 @@ bit [61] +++SINH+++ - If set, then counting of events in S/HS-mode is inhibited
 
 bit [60] +++UINH+++ - If set, then counting of events in U-mode is inhibited
 
-bit [59] +++VSINH+++ - If set, then counting of events in VS-mode is inhibited
+bit [59] +++VSINH+++ - If set and the H extension is implemented, then counting of events in VS-mode is inhibited
 
-bit [58] +++VUINH+++ - If set, then counting of events in VU-mode is inhibited
+bit [58] +++VUINH+++ - If set and the H extension is implemented, then counting of events in VU-mode is inhibited
```
aswaterman, line 67: @gfavor these seem odd.  Shouldn't these bits be read-only zero when the H extension isn't implemented, rendering these two changes unnecessary?  (Perhaps their being read-only zero is what needs to be clarified.)

aswaterman: (I double-checked that Spike does it the way I suggested.)

src/sscofpmf.adoc
```diff
@@ -183,7 +183,7 @@ M-mode).
 
 Read access to bit _X_ is subject to the same mcounteren (or mcounteren and
 hcounteren) CSRs that mediate access to the hpmcounter CSRs by S-mode (or
-VS-mode). In M and S modes, scountovf bit _X_ is readable when mcounteren bit
+VS-mode). In S modes, scountovf bit _X_ is readable when mcounteren bit
```
aswaterman, line 186: @gfavor Don't we need to say what happens in M-mode, e.g. add a sentence before this one like "In M-mode, scountovf bit _X_ is always readable."

src/sscofpmf.adoc
gfavor: You're correct as to the not so appropriate wording.  (This change wording came from an issue raised a long time ago.)

But what does still need to be specified is that bits 58 and 59 are read-only-zero when H isn't implemented.  The existing sentence should remain unchanged, and then a second sentence added that says the preceding.

So how about (using bit 58 as the example):

_If set, then counting of events in VU-mode is inhibited.  But if the H extension is not implemented, then this bit is read-only-zero._

src/sscofpmf.adoc
gfavor: Yes re scountovf.  @wmat , the overall change should be replacing the "In M and S modes, ..." sentence with the following two sentences:

_In M mode, scountovf bit X is always readable.  In S/HS modes, scountovf bit X is readable when mcounteren bit X is set, and otherwise reads as zero._

src/sscofpmf.adoc
aswaterman: SGTM.

gfavor (approved):
Assuming Bill makes the updated change, I'm good overall with this PR.

aswaterman (changes requested):
👍  I'll make a final review before merging.

src/sscofpmf.adoc
wmat: I've made the change as described, however, is S/HS correct?  Should it be S/VS?
```suggestion
VS-mode). In M mode, scountovf bit X is always readable, in S/HS modes, scountovf bit _X_ is readable when mcounteren bit
```

src/sscofpmf.adoc
```diff
@@ -183,7 +183,7 @@ M-mode).
 
 Read access to bit _X_ is subject to the same mcounteren (or mcounteren and
 hcounteren) CSRs that mediate access to the hpmcounter CSRs by S-mode (or
-VS-mode). In M and S modes, scountovf bit _X_ is readable when mcounteren bit
+VS-mode). In M mode, scountovf bit X is always readable, in S/HS modes, scountovf bit _X_ is readable when mcounteren bit
```
aswaterman, line 186: Italicize the X, and start a new sentence after "readable" rather than using a comma.  LGTM otherwise.

aswaterman: Fixed manually, as I need to make some more edits to this chapter and I don't want to create merge conflicts.

aswaterman (approved):

PR #1323, opened 2024-04-05
Integrate Sscofpmf into priv spec more smoothly

aswaterman:
Some of the text was written as a delta against the priv spec.  Incorporate those deltas and delete the old text.  (Some of the deltas had already been incorporated, but the old text hadn't been deleted.)


gfavor (approved):

gfavor (approved):
Good generalization re the *INH bits becoming read-only-zero.

gfavor (approved):

gfavor (approved):
Good cleanup.

gfavor (approved):

gfavor (approved):

Issue #1322, opened 2024-04-04
CSR tables in Priv Ch. 2 are misformatted


aswaterman:
Because the first and third column use a taller font than the second and fourth column, the rows don't line up.  See below for an example (but it affects all of the other tables, too).

Changing the fonts to have the same height would probably work, but it's a little hacky; really, these should be laid out as tables so that the rows are automatically aligned, regardless of font.

![Screenshot 2024-04-04 at 2 53 14 PM](https://github.com/riscv/riscv-isa-manual/assets/1031106/f62b01be-3d73-4aa0-b273-ae7f5a88fc93)

aswaterman:
Thanks for the fix, @wmat!

PR #1326, opened 2024-04-05
Removing font-size setting of 11.5 on codespan.

wmat:
Commented out the font-size: 11.5 in the codespan section.  Fixes issue #1322


aswaterman (approved):
Much better!

PR #1339, opened 2024-04-09
Chapter titles cleanup

wmat:
Cleaned up chapter titles, version numbers, etc.

### DIFF
--- a/src/smepmp.adoc
+++ b/src/smepmp.adoc
@@ -1,5 +1,5 @@
 [[smepmp]]
-== PMP Enhancements for memory access and execution prevention on Machine mode (Smepmp)
+== "Smepmp" Enhancements for memory access and execution prevention on Machine mode
 === Introduction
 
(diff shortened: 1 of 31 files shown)


src/zacas.adoc
```diff
@@ -1,4 +1,4 @@
-== Atomic Compare-and-Swap (CAS) instructions (Zacas), Version 1.0.0
+== "Cas" and "Zacas" Extensions for Atomic Compare-and-Swap Instructions, Version 1.0.0
```
ved-rivos, line 1: Should be -  "Zacas" Extension for Atomic Compare-and-Swap (CAS) Instructions, Version 1.0.0

@wmat

src/smepmp.adoc
```diff
@@ -1,5 +1,5 @@
 [[smepmp]]
-== PMP Enhancements for memory access and execution prevention on Machine mode (Smepmp)
+== "Smepmp" Enhancements for memory access and execution prevention on Machine mode, Version 1.0.0
```
aswaterman, line 2: "Enhancements" should be changed to "Extension" (and it should be singular).

aswaterman (commented):
LGTM aside from my one comment.

src/smepmp.adoc
gfavor: Note that Smepmp makes enhancements on top of the PMP spec.  So maybe the title should be:

"Smepmp" Extension for PMP enhancements for memory access and execution prevention in Machine mode, Version 1.0.0

aswaterman: Works for me.

wmat: Fixed.  Thanks.

Issue #1352, opened 2024-04-18
Ordering of names in titles


kersten1:
Noticed this as I was looking through the TOC. Some register names are like this: 

"Supervisor Status Register (sstatus) 

and then others look like this: 

"Supervisor Trap Value (stval) Register"

I like the second better but not enough that I'll argue if everyone else likes the first.  Let me know which is preferred and I can fix up titles.

kasanovic:
I'll vote for the second version also.

aswaterman:
Likewise.

gfavor:
Ditto.

Greg

kersten1:
Most of these are done. Have to catch the errant ones as we go.

PR #1357, opened 2024-04-18
More register name clean up

kersten1:
More register name fixes - https://github.com/riscv/riscv-isa-manual/issues/1352

### DIFF
--- a/src/machine.adoc
+++ b/src/machine.adoc
@@ -53,7 +53,7 @@ knowing the register width (MXLEN) of the hart. The base width is
 given by __MXLEN=2^MXL+4^__.
 
 The base width can also be found if `misa` is zero, by placing the
-immediate 4 in a register and then shifting the register left by 31 bits at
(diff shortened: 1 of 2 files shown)


src/machine.adoc
```diff
@@ -53,7 +53,7 @@ knowing the register width (MXLEN) of the hart. The base width is
 given by __MXLEN=2^MXL+4^__.
 
 The base width can also be found if `misa` is zero, by placing the
-immediate 4 in a register then shifting the register left by 31 bits at
+immediate 4 in a register and then shifting the register left by 31 bits at
```
aswaterman, line 56: I think this reads better without the “and”.

aswaterman (approved):

src/machine.adoc
kersten1, line 56:
```suggestion
immediate 4 in a register, then shifting the register left by 31 bits at
```

aswaterman (approved):

PR #1380, opened 2024-05-02
Integrate Zicfiss and Zicfilp extensions into Priv. and Unpriv. ISA spec.

ved-rivos:


aswaterman:
@ved-rivos the new wavedrom-based figures look wonderful, and the coding style looks much easier to maintain.  Thank you!

@wmat @kersten1 please take a look.

aswaterman (approved):

PR #1283, opened 2024-03-22
Maintain acronym and update acronym expansion.

wmat:
Keep MMWP as is but update it's expansion to Machine-Mode When-no-PMP-match Policy.  This fixes Issue https://github.com/riscv/riscv-isa-manual/issues/1280

PR #1381, opened 2024-05-03
Updating "whitelist" to "allowlist" in text

kersten1:
Plus a couple other clean up items. In general, "may" is a permission thing. "Can" is better word in this instance. "using" is also a hard word to translate on its own. If you making it "by using" translation is easier.


kersten1:
Also ties in with https://github.com/riscv/riscv-isa-manual/pull/1283

kersten1:
sigh - didn't realize that other PR got the other word. Closing.

Issue #1390, opened 2024-05-09
Clarify the ordering of Ztso in ISA Extension Naming Conventions


cyyself:
Since Ztso belongs to Standard Extension, [the description of the ordering of this extension in the current manual is](https://github.com/riscv/riscv-isa-manual/blob/7b8ddc9f58b858265525b59f45bfca961f805b6b/src/naming.adoc?plain=1#L86):

```
The first letter following the "Z" conventionally indicates the most
closely related alphabetical extension category, IMAFDQCVH. For the
"Zfa" extension for additional floating-point instructions, for example, the letter "f"
indicates the extension is related to the "F" standard extension. If
multiple "Z" extensions are named, they should be ordered first by
category, then alphabetically within a category—for example,
"Zicsr_Zifencei_Zam".
```

In this case, the Zicsr and Zifencei are categorized as Additional Integer Standard Extension since they appear before Zam in Additional Atomic Extension. Thus, the order of "IMAFDQCVH" is correct.

However, the [ISA naming table shows Ztso there](https://github.com/riscv/riscv-isa-manual/blob/7b8ddc9f58b858265525b59f45bfca961f805b6b/src/naming.adoc?plain=1#L157). But 'T' is not in the "IMAFDQCVH" ordering string.

Also, other extensions like B (Bitmanip) and P (Packed-SIMD) are not in the naming string.

I want to fix it, but I need to know where to place T for Ztso.

cyyself:
I proposed a draft PR #1391, which categorized `Ztso` and any other extensions that do not relate to any alphabetical
extension category as the "others" category and placed it in the last place among the ordering of all Z* extensions.

cyyself:
There is another concern about `T` extension standards for transactional memory. Since TSO memory ordering is not related to transactional memory, can we reuse the [canonical order defined in RISC-V Linux User ABI](https://docs.kernel.org/arch/riscv/uabi.html)? If yes, I think it's OK now.

cyyself:
Fixed in #1391

PR #1391, opened 2024-05-09
Clarify the ordering of Additional Standard Unprivileged Extensions

cyyself:
Fixed some problems mentioned in #1390

### DIFF
--- a/src/naming.adoc
+++ b/src/naming.adoc
@@ -74,16 +74,16 @@ if it follows a number. For example, "rv32i2p2" means version 2.2 of
 RV32I, whereas "rv32i2_p2" means version 2.0 of RV32I with version 2.0
 of the P extension.
 
-=== Additional Standard Extension Names
+=== Additional Standard Unprivileged Extension Names
 
-Standard extensions can also be named using a single "Z" followed by
+Standard unprivileged extensions can also be named using a single "Z" followed by
 an alphabetical name and an optional version number. For example,
 "Zifencei" names the instruction-fetch fence extension described in
 <<zifencei>>; "Zifencei2" and
 "Zifencei2p0" name version 2.0 of same.
 
 The first letter following the "Z" conventionally indicates the most
-closely related alphabetical extension category, IMAFDQCVH. For the
+closely related alphabetical extension category, IMAFDQLCBKJTPVH. For the
 "Zfa" extension for additional floating-point instructions, for example, the letter "f"
 indicates the extension is related to the "F" standard extension. If
 multiple "Z" extensions are named, they should be ordered first by
@@ -190,11 +190,9 @@ e.g., RV32IMACV is legal, whereas RV32IMAVC is not.
 
 |Hypervisor Extension |H |
 
-|Control and Status Register Access |Zicsr |
+3+|*Additional Standard Unprivileged Extensions*
 
-|Instruction-Fetch Fence |Zifencei |
-
-|Total Store Ordering |Ztso |
+|Additional Standard unprivileged extensions "abc" |Zabc |
 
 3+|*Standard Supervisor-Level Extensions*
 


cyyself:
I don't know where to place `T` for Ztso, so I converted it to draft now.

src/naming.adoc
```diff
@@ -74,16 +74,16 @@ if it follows a number. For example, "rv32i2p2" means version 2.2 of
...
-Standard extensions can also be named using a single "Z" followed by
+Standard unprivileged extensions can also be named using a single "Z" followed by
 an alphabetical name and an optional version number. For example,
 "Zifencei" names the instruction-fetch fence extension described in
 <<zifencei>>; "Zifencei2" and
 "Zifencei2p0" name version 2.0 of same.
 
 The first letter following the "Z" conventionally indicates the most
-closely related alphabetical extension category, IMAFDQCVH. For the
+closely related alphabetical extension category, IMAFDQLCBKJTPVH. For the
```
topperc, line 86: Is this basically a revert of https://github.com/riscv/riscv-isa-manual/commit/f5f9c27010b69a015958ffebe1ac5a34f8776dff

cyyself: For this, I think the answer is Yes. I reviewed the comments in that commit and understand what happened. If we can’t decide in time, I think the other part of this patch that clarifies the Z* is unprivileged and the order of Zicsr, Zifencei, and Ztso should also be removed to avoid confusion for readers.

aswaterman: Note `H` is _not_ an unprivileged extension so does not belong here.

topperc: unfortunately the order of H needs to be listed because it governs the sorting of Zhinx Zhinxmin

aswaterman: Yeah, those probably should've been named Zfhinx... Seems like the H is not in the ideal place in the list, then?  Shouldn't it be somewhere around Q?

cyyself: > Note `H` is _not_ an unprivileged extension so does not belong here.

Will Zh* exist someday for hypervisor additional extension? If yes, I think "where to place it" should be defined.

aswaterman: No, hypervisor additional extensions use the `Sh` prefix (like `Sm` for machine or `Ss` for supervisor)

cyyself:
@aswaterman I have force-pushed a commit that deletes the `H` in the string. But the ordering of `Ztso` remains unknown. As I said from [#1391](https://github.com/riscv/riscv-isa-manual/issues/1390#issuecomment-2102239010), TSO memory ordering is not related to transactional memory.

aswaterman:
@cyyself I think @topperc's point was that we need to keep H in the list because of Zhinx.  It's unfortunate, but I agree with him

I agree with you that the T/Ztso situation is somewhat analogous to the H/Zhnix situation.  So I guess we need to allocate T somewhere in this list, too.  I guess we can just put it at the end; Ztso isn't especially similar to any other extension.

(The other option is to say that there is no canonical order anymore.  But I still think it makes sense to have a _recommended_ order for tools that print out ISA strings, even if tools that parse ISA strings don't enforce canonicity.  Either way, we'd need an ordering.)

topperc:
For Zhinx and Ztso, gcc and clang have both been printing using the order that was in this.

aswaterman:
@topperc not sure what to read into the "was" in that sentence.  Is the current incarnation of the PR a problem, or should I merge it as-is?

topperc:
> @topperc not sure what to read into the "was" in that sentence. Is the current incarnation of the PR a problem, or should I merge it as-is?

I wrote "was" between the time H was removed and when it was restored. Its good to merge.

aswaterman:
Gotcha, thanks.

Issue #12, opened 2017-02-12
Consider adding mbadinst CSR


aswaterman:
If omitting features like misaligned memory accesses and certain CSRs is going to be the most common implementation strategy, it may be worth provisioning an optional mbadinst register that holds the instruction word that most recently caused an illegal instruction trap.  This saves several instructions and a likely D$ miss in the emulation path.

A similar template in H/S modes could improve virtualization performance for the same reason.

If the instruction is >XLEN bits, the register would only reflect the LSBs.

(Implementation note: if illegal instructions are detected early & in-order in the pipeline, it's not necessary to pipeline the whole instruction word down; one microarchitectural shadow copy suffices.)

aswaterman:
Possible alternative: use mbadaddr for this purpose, since it is redundant with mepc on illegal-instruction traps.

aswaterman:
Need to be careful about what shows up in MSBs; simplest thing to do is to zero them.

David-Horner:
How precious are these CSRs, really?
The extra real-estate for flip-flops in the core for 32 (or 64 bits) is indeed significant.
However, reserving and allocating the register number is not a significant burden, is it?

Re:
>Possible alternative: use mbadaddr for this purpose, since it is redundant with mepc on illegal instruction traps.

I'd suggest making mbadaddr undefined for trapped instructions that do not have an addressing component.
And that mbadinst contain the trapping instruction in the LSB with any MSB undefined.
As an implementation optimization mbadinst is alias with mbadaddr and the instruction be injected into mbadinst if that overall reduces complexity/cost/etc.

I suggest MSB be undefined as it could allow a future extension to, say inject the actual compressed instruction there - or to indicate a fused instruction, etc.. I only foresee this as significant on RV64/RV128.

Further, how is an illegal compressed instruction shown?
Indeed, if an instruction has the low order bit 00,01 or 10 and it is not a valid Compressed Instruction, is it an illegal Compressed instruction, or "just" an illegal instruction.
e.g. is x"00000000" one or two illegal instructions - does it depend upon C extension being implemented?

This does raise the question of how to handle an illegal exception raise by the hart before the remainder of the instruction is read (for example at the end of a cache line).

 I would recommend that at least the 4 LSBytes as available to the decoder be written to mbadinst, whether the format would otherwise be 16bit on a Compressed Extension implementation or not, and whether all the instruction had been loaded for the decode yet.
[...]

PR #1398, opened 2024-05-12
Added support for EPUB3

Ioan-Cristian:
### Description

This pull requests adds support for building EPUB3 versions of the RISC-V ISA manuals. To achieve this, `asciidoctor-epub3` extension has been used.

### Motivation

eBook readers are convenient devices for reading. They are based on e-Ink screens, reducing fatigue while reading. EPUB3 is an open standard eBook format supported by a wide range of hardware readers. It is also well supported by `asciidoctor` through the aforementioned extension, making it an ideal file format for RISC-V ISA manuals.

### Testing

Manually built using Ubuntu 24.04 and viewed files on PocketBook InkPad 3.

### Known issues

Some eBook readers are not capable of rendering embedded images. The solution to this problem is disabling embedded image altogether. The `README.md` file provides more details on how to achieve this.

### Related pull requests

riscv-docs-base-container-image: [#12](https://github.com/riscv/riscv-docs-base-container-image/pull/12)


aswaterman:
@wmat I think we need to partially revert this.  Trying to build this within the usual docker container, the build fails because of a lack of `asciidoctor-epub3`.

aswaterman:
Nevermind.  I see the docker container has been updated accordingly and I just needed to download a newer version.  In case anyone else stumbles across this, don't forget to `docker pull docker.io/riscvintl/riscv-docs-base-container-image:latest`

aswaterman:
Nevermind once again.  The epub build is still broken inside the docker container, but for a different reason:

```
asciidoctor: WARNING: riscv-privileged.adoc: invalid reference to unknown anchor: LP_INST
(sass):384: Invalid US-ASCII character "\\xE2" (Sass::SyntaxError)
```

I'll disable building epubs within the docker container for now.

Ioan-Cristian:
Probably you've run `cd build && make`. You need to run `export LANG=C.utf8; cd build && make` as suggested by the `README.md` and workflow files.

aswaterman:
The command “make docker” is supposed to just work, regardless of environment. So you have not fixed up the Makefile properly. When you have, let us know.

Issue #1402, opened 2024-05-14
Should Vector chapter version be bumped to 2.0?


wmat:
The versioning of the Vector chapter is causing some confusion.  The introduction to the chapter states that "This document is version 1.1-draft of the RISC-V vector extension.

NOTE: This version holds updates gathered after the start of the public review.  The spec will have a final update to version 2.0 at
time of ratification."

How should we proceed with versioning this chapter?

camel-cdr:
I think the original idea was to bump the spec version to 2.0 once ratified, but that idea got dropped. (Somebody correct me if I'm wrong)

Considering that RVV 1.0 is used in a lot of marketing, slides, articles, and conversations retroactively renaming it to RVV 2.0 would be very problematic.
Especially considering the history of XTheadVector ("RVV 0.7.1" based) being marketed as RVV support adding another version to the mix that doesn't change anything would just cause confusion.


I think the best course of action is to just remove that NOTE.

gfavor:
After checking with people involved with RVV development (including the TG chair - Krste), it is recognized that there are these bits of text that are a ready source of confusion - and are now being removed from the spec as part of the current ISA manual conversion/integration/cleanup efforts that are occurring.  The net of the cleanup will clearly show that the current version is v1.0.  (Also note that the text mentioning v1.1 was anticipating a new version of RVV that never happened.)

camel-cdr:
So whats the verdict?
I think this needs to be addressed, and it sounds like the solution is to remove that passage.
Is this change still in some kind of internal review?

gfavor:
The latest (integrated) vector chapter has been cleaned up to properly
reflect that it is v1.0.  Various bits of text that looked forward to
possible or probable changes that would warrant updating to v2.0 or v1.1,
have been removed since those changes never occurred.

Greg

aswaterman:
What @gfavor said. I neglected to close this issue when the spec integration occurred, but now’s as good a time as any.

PR #1406, opened 2024-05-14
Remove outdated RVV 2.0 notice from document #1402

camel-cdr:
For reasons I [describe here](https://github.com/riscv/riscv-isa-manual/issues/1402#issuecomment-2110699739), I think it's best to remove the RVV 2.0 version notice from the specification, there does seem to be no indication that what the notice says is actually planned.

I replaced it with text derived from the Vector Crypto Introduction, both should ideally be reworded before an actual release, but this should remove the confusion for now.

See also past discussion on the v-spec repo: https://github.com/riscv/riscv-v-spec/pull/946

It sounded like the change would be merged, but the repo was archived before that got through. My understanding is that the riscv-v-spec repo wasn't supposed to be a source of truth, hence it was archived.

### DIFF
--- a/src/v-st-ext.adoc
+++ b/src/v-st-ext.adoc
@@ -11,17 +11,14 @@ domains._
 
 === Introduction
 
-This document is version 1.1-draft of the RISC-V vector extension.
+This document describes the Vector extensions to the RISC-V Instruction Set Architecture.
 
-NOTE: This version holds updates gathered after the start of the
-public review.  The spec will have a final update to version 2.0 at
-time of ratification.
 
-This spec includes the complete set of currently frozen vector
-instructions.  Other instructions that have been considered during
-development but are not present in this document are not included in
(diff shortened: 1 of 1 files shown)


aswaterman:
I'm dropping the extensive warning about what Ratified means, as it's a global property of RVIA policy, not of this extension.  Yes, I understand there's a tortured history in this case, but I'd rather not flog a dead horse.

aswaterman:
Looking at the other extensions (with the exception of vector crypto), the consistent thing to do is to remove any draft/freeze warnings and update the preface.  That's what we'll do here.

camel-cdr:
> I'm dropping the extensive warning about what Ratified means, as it's a global property of RVIA policy, not of this extension

Exactly, that was the motivation behind this change. Everything I've heard from RVI in any official capacity other than this NOTE referred to RVV 1.0 being ratified. Maybe I've missed some part of the communication.

PR #1408, opened 2024-05-14
Remove "vacant" class of memory region

aswaterman:
Represent them as never-accessible main memory or I/O regions instead.


aswaterman:
Approved verbally by the ARC.

PR #1411, opened 2024-05-15

muks:
"... where these are the only places where misalignment can occur." was far away from the "branch or jump" making it difficult to follow.


aswaterman:
The change of emphasis detracts from the meaning.

muks:
What are the places where misalignment can occur?

PR #1412, opened 2024-05-15

muks:
"allocated" should be followed by "so" or "in this way" or "in this manner".


aswaterman:
This introduces a grammatical error and does not, in my opinion, improve clarity.

muks:
@aswaterman The sentence does not read right in the current form. If you don't agree, please ask for another opinion from a colleague.

aswaterman:
I stand by my statement that your PR introduces a grammatical error. I would accept a version of a patch that replaces of “so” with “so as”.

muks:
You could have suggested replacing "so" with "so as" instead of closing it.

I'll push an update to the PR with this change.

PR #1413, opened 2024-05-15

muks:
I've updated the wording so it's clearer. Please let me know if you prefer something else.

### DIFF
--- a/src/rv32.adoc
+++ b/src/rv32.adoc
@@ -170,9 +170,9 @@ registers at the same position in all formats to simplify decoding.
 Except for the 5-bit immediates used in CSR instructions
 (<<csrinsts>>), immediates are always
 sign-extended, and are generally packed towards the leftmost available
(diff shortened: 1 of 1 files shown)


aswaterman:
I don’t agree that this improves readability.

muks:
Without the "in this way" qualification, it reads as:

>  "immediates ... have been allocated to reduce hardware complexity."

instead of

>  "immediates ... have been allocated in this way to reduce hardware complexity."

.. the "in this way" referring to "generally packed towards the leftmost available bits."

Please do consider a colleague's opinion. I created this PR because it was not readable to me.

aswaterman:
Packing the immediates to the left does not reduce hardware complexity. The sentence means the right thing as written.

muks:
The correct meaning is "immediates reduce hardware complexity" ? In that case I have misread it.

PR #1414, opened 2024-05-15
Fixed Makefile to allow building EPUB manuals with docker rule

Ioan-Cristian:
This pull request fixes `docker` Makefile rule to allow building EPUB manuals.


aswaterman (approved):
Thanks.  Confirmed this works on my side.

PR #1417, opened 2024-05-16
Slightly clean up *envcfg bit spec

ved-rivos:


ved-rivos:
@aswaterman

aswaterman (approved):
Looks good, thanks.

PR #1419, opened 2024-05-22
Update B extension description to remove mention of sub extensions grouped for review

jordancarlin:
- Remove heading stating "Bit-manipulation a, b, c and s extensions grouped for public review and ratification"
- Slightly reword intro to remove mention of lack of a larger extension now that B is ratified

### DIFF
--- a/src/b-st-ext.adoc
+++ b/src/b-st-ext.adoc
@@ -5,12 +5,11 @@ The B standard extension comprises instructions provided by the Zba, Zbb, and
 Zbs extensions.
 
 [[preface]]
-=== Bit-manipulation a, b, c and s extensions grouped for public review and ratification
+=== Zb* Overview
 
 The bit-manipulation (bitmanip) extension collection is comprised of several component extensions to the base RISC-V architecture that are intended to provide some combination of code size reduction, performance improvement, and energy reduction.
 While the instructions are intended to have general use, some instructions are more useful in some domains than others.
-Hence, several smaller bitmanip extensions are provided, rather than one large extension.
-Each of these smaller extensions is grouped by common function and use case, and each has its own Zb*-extension name.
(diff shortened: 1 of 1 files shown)


src/b-st-ext.adoc
```diff
@@ -5,12 +5,11 @@ The B standard extension comprises instructions provided by the Zba, Zbb, and
 Zbs extensions.
 
 [[preface]]
-=== Bit-manipulation a, b, c and s extensions grouped for public review and ratification
+=== Bit-manipulation small extensions
```
aswaterman, line 8: "small extensions" isn't a term/concept we should proliferate.  I suggest changing the title to " Bit-manipulation Extensions Zba, Zbb, Zbb, and Zbs"

aswaterman (changes requested):

src/b-st-ext.adoc
jordancarlin: I wonder if it makes sense to include a title for this part at all since it’s really an intro to all the Zb* extensions (including Zbk*). Maybe just having it under the main chapter title is fine. 

If we do want a title, what about “Zb* Extensions Overview” or something along those lines?

aswaterman: If you think ditching the title altogether works, that’s fine with me, too.

jordancarlin: Looking at it more closely, the Zc* chapter uses "Zc* Overview" to title a similar section, so I went with that. If you'd rather I removed it let me know and I can make that change.

aswaterman (approved):
Thank you. Works for me.

Overall, there are several rough edges associated with integrating these specs into a single document. The intro to this chapter is probably due for a total rewrite, but improvements like this are a useful step.

Issue #1420, opened 2024-05-22
Does mcountinhibit Inhibit Only Unprivileged Counters, or Both Machine and Unprivileged Counters?


cebarobot:
I'm confusing about the `mcountinhibit` CSR. Does `mcountinhibit` inhibit only Unprivileged Counters, or both Machine and Unprivileged Counters?

The Privileged ISA manual says:
> When the CY, IR, or HPMn bit in the mcountinhibit register is clear, the `cycle`, `instret`, or `hpmcountern` register increments as usual. When the CY, IR, or HPM_n_ bit is set, the corresponding counter does not increment.

It only refers to **Unprivileged Counters** (`cycle`, `instret`, or `hpmcountern`) and not mensions the **Machine Counters** (`mcycle`, `minstret`, or `mhpmcountern`) at all.

As `cycle`, `instret` and `hpmcountern` is a read-only shadow of `mcycle`, `minstret` and `mhpmcountern`, there is an intuitive understanding based on the text: "mcountinhibit` only inhibits Unprivileged Counters":
> When the CY, IR, or HPM_n_ is set, the hart makes a data copy of `mcycle`, `minstret` and `mhpmcountern` at that moment.
> Reading Unprivileged Counters returns that copy, while machine counters keep going.

In that understanding, the purpose of `mcountinhibit` (reduce energy consumption) make no sense, which leads to more confusion.

I also checked some open-source implementation, and they did as below:
* [Spike (riscv/riscv-isa-sim)](https://github.com/riscv/riscv-isa-sim): `mcountinhibit` is [read-only zero](https://github.com/riscv-software-src/riscv-isa-sim/blob/c81d8e73daf1875d79a54f28f67df150fba0e44c/riscv/processor.cc#L292).
* [BOOM](https://github.com/riscv-boom/riscv-boom): `mcountinhibit` is not implemented.
* [Rocket](https://github.com/chipsalliance/rocket-chip): `mountinhibit` will inhibit [Both Machine and Unprivileged Counters](https://github.com/chipsalliance/rocket-chip/blob/dbcb06afe1c76d1129cb6d264949322a34c37185/src/main/scala/rocket/CSR.scala#L584).
[...]

PR #1422, opened 2024-05-22
Clarify the behavior of mcountinhibit CSR

cebarobot:
Use mcycle, minstret, mhpmcountern to replace cycle, instret, hpmcountern in mcountinhibit spec. This will not changing any meaning because of "read-only shadow".

See #1420 for details.


aswaterman:
Merging, as this is a useful improvement to the description that has no normative impact.

PR #1431, opened 2024-05-28
Clarify CBO access size

dkruckemyer-ventana:
Define access size to be size of cache block for PMP/PMA checks


aswaterman (approved):
Nevermind.  After reading the discussion on https://github.com/riscv/riscv-CMOs/issues/60, I agree this can be argued to just be a clarification.

PR #1433, opened 2024-05-29
Clarify CBO for unaligned rs1

### LINKED ISSUE #1263: Specify mtval for unaligned CMOs
If you do `cbo.zero` with an unaligned address, it aligns the address. But if there's a memory exception what value should be written to `mtval`? IMO it should be the aligned address, but it isn't specified.

Timmmm:
Rework the text to be much more explicit about the behaviour when rs1 is not aligned to the cache block size.

I also moved & slightly reworded the note about the assembly syntax since it's not relevant to the instruction semantics.

Fixes #1263

### DIFF
--- a/src/cmo.adoc
+++ b/src/cmo.adoc
@@ -385,8 +385,8 @@ translation, the instruction does _not_ check the accessed and dirty bits and
 neither raises an exception nor sets the bits.
 
 When a page fault, guest-page fault, or access fault exception is taken, the
-relevant *tval CSR is written with the faulting effective address (i.e. the same
-faulting address value as for other causes of these exceptions).
+relevant *tval CSR is written with the faulting effective address (i.e. the
+value of _rs1_).
 
 [NOTE]
 ====
@@ -924,10 +924,14 @@ Encoding::
 Description::
 
 A *cbo.flush* instruction performs a flush operation on the cache block whose
-effective address is the base address specified in _rs1_. The offset operand may
-be omitted; otherwise, any expression that computes the offset shall evaluate to
-zero. The instruction operates on the set of coherent caches accessed by the
+that contains the address specified in _rs1_. It is not required that _rs1_ is
+aligned to the size of a cache block. On faults, the faulting virtual address
+is considered to be the value in rs1, rather than the base address of the cache
+block. The instruction operates on the set of coherent caches accessed by the
 agent executing the instruction.
+
+The assembly _offset_ operand may be omitted. If it isn't then any expression
+that computes the offset shall evaluate to zero.
 
 Operation::
 [source,sail]
@@ -959,12 +963,17 @@ Encoding::
 Description::
 
 A *cbo.inval* instruction performs an invalidate operation on the cache block
-whose effective address is the base address specified in _rs1_. The offset
-operand may be omitted; otherwise, any expression that computes the offset shall
-evaluate to zero. The instruction operates on the set of coherent caches
-accessed by the agent executing the instruction. Depending on CSR programming,
-the instruction may perform a flush operation instead of an invalidate
-operation.
+that contains the address specified in _rs1_. It is not required that _rs1_ is
+aligned to the size of a cache block. On faults, the faulting virtual address
+is considered to be the value in rs1, rather than the base address of the cache
+block. The instruction operates on the set of coherent caches accessed by the
+agent executing the instruction.
+
+Depending on CSR programming, the instruction may perform a flush operation
+instead of an invalidate operation.
+
+The assembly _offset_ operand may be omitted. If it isn't then any expression
+that computes the offset shall evaluate to zero.
 
 [NOTE]
 ====
@@ -1004,10 +1013,14 @@ Encoding::
 Description::
 
 A *cbo.zero* instruction performs stores of zeros to the full set of bytes
-corresponding to the cache block whose effective address is the base address
-specified in _rs1_. The offset operand may be omitted; otherwise, any expression
-that computes the offset shall evaluate to zero. An implementation may or may
-not update the entire set of bytes atomically.
+corresponding to the cache block that contains the address specified in _rs1_.
+It is not required that _rs1_ is aligned to the size of a cache block. On
+faults, the faulting virtual address is considered to be the value in rs1,
+rather than the base address of the cache block. An implementation may or
+may not update the entire set of bytes atomically.
+
+The assembly _offset_ operand may be omitted. If it isn't then any expression
+that computes the offset shall evaluate to zero.
 
 Operation::
 [source,sail]


src/cmo.adoc
```diff
@@ -385,8 +385,8 @@ translation, the instruction does _not_ check the accessed and dirty bits and
 neither raises an exception nor sets the bits.
 
 When a page fault, guest-page fault, or access fault exception is taken, the
-relevant *tval CSR is written with the faulting effective address (i.e. the same
-faulting address value as for other causes of these exceptions).
+relevant *tval CSR is written with the faulting effective address (i.e. the
+value of _rs1_).
```
aswaterman, line 389: This is the part that I endorse.  My initial reaction is the other changes are not necessary, and maybe not even desirable, but I'll leave that decision to @dkruckemyer-ventana.

Timmmm: I think at least the bits about not requiring `rs1` to be aligned are needed because that isn't specified anywhere currently (unless I missed it).

src/cmo.adoc
```diff
@@ -931,10 +931,13 @@ Encoding::
...
 
 A *cbo.flush* instruction performs a flush operation on the cache block whose
-effective address is the base address specified in _rs1_. The offset operand may
-be omitted; otherwise, any expression that computes the offset shall evaluate to
-zero. The instruction operates on the set of coherent caches accessed by the
-agent executing the instruction.
+that contains the address specified in _rs1_. It is not required that _rs1_ is
+aligned to the size of a cache block. On faults the address stored in the *tval
+CSR is _rs1_; not the base address of the cache block. The instruction operates
+on the set of coherent caches accessed by the agent executing the instruction.
```
ved-rivos, line 937: A CMO to flush/invalidate/clean should take effect even if the cache block is cached in a coherent cache that is not directly accessible to the agent executing the corresponding CMO. For instance, the cache block may be cached in the private cache of a hart should be flushed/invalidated/cleaned if a CMO addressing that cache block was executed by another hart.

src/cmo.adoc
ved-rivos: I did not think there was an intent to support a CMO for flush/invalidate/clean to operate on multiple cache blocks. Its not clearly documented but it is more reasonable to think of a CMO.flush/invalidate/clean as having an access width of 1 i.e. it targets the cache block addressed by `rs1` and that cache block may be addressed by specifying any byte of that cache block.

Timmmm: I agree, that's what I meant it to say. Does it sound like it's saying it would operate on multiple cache blocks?

src/cmo.adoc
Timmmm: It does say this elsewhere:

> Additionally, if a coherent agent in the set executes a CBO instruction that specifies the cache block, the resulting operation shall apply to any and all of the copies in the caches that can be accessed by the load and store operations from the coherent agents.

So I think it's reasonably clear, but if you want to clarify it more here I think it can go in a separate PR.

ved-rivos: "that can be accessed by the load and store operations" - it could be interpreted that CMO will not operate on private caches of other harts because they cannot be "accessed" by load/store operations performed on the hart that executes the CMO. The other interpretation would be that "accessed" includes forwarding of data from private caches of remote harts to the private cache of the hart executing the load/store. I would like to hear @dkruckemyer-ventana interpretation.

Timmmm: Yeah I think that's just an oversight here. The other bit I quoted is much more explicit and clearly agrees with what you said.

Timmmm:
@ved-rivos @dkruckemyer-ventana any objection to merging this? I think the points Ved raised above are pre-existing issues (unless I've misunderstood) and therefore can be addressed in another PR.

src/cmo.adoc
```diff
@@ -931,10 +931,13 @@ Encoding::
...
+
+The assembly _offset_ operand may be omitted. If it isn't then any expression
+that computes the offset shall evaluate to zero.
```
aswaterman, line 940: This doesn't belong in the ISA spec; delete it.

aswaterman: Ah, I see this has been factored out of the previous paragraph and isn't a new addition.  I still think it doesn't belong here, but since it's better to keep the topic of the PR restricted, we should leave it alone.

src/cmo.adoc
aswaterman: Please rewrite "On faults the address stored in the *tval CSR is _rs1_; not the base address of the cache block." as "On faults, the faulting virtual address is considered to be the value in _rs1_, rather than the base address of the cache block."

This improves separation of concerns, because faulting virtual addresses might end up in places other than *tval.

src/cmo.adoc
```diff
@@ -966,12 +969,16 @@ Encoding::
...
 A *cbo.inval* instruction performs an invalidate operation on the cache block
-whose effective address is the base address specified in _rs1_. The offset
-operand may be omitted; otherwise, any expression that computes the offset shall
-evaluate to zero. The instruction operates on the set of coherent caches
-accessed by the agent executing the instruction. Depending on CSR programming,
-the instruction may perform a flush operation instead of an invalidate
-operation.
+that contains the address specified in _rs1_. It is not required that _rs1_ is
+aligned to the size of a cache block. On faults the address stored in the *tval
+CSR is _rs1_; not the base address of the cache block. The instruction operates
```
aswaterman, line 974: Same comment as https://github.com/riscv/riscv-isa-manual/pull/1433/files#r1693650461

src/cmo.adoc
```diff
@@ -1011,10 +1018,14 @@ Encoding::
...
 
 A *cbo.zero* instruction performs stores of zeros to the full set of bytes
-corresponding to the cache block whose effective address is the base address
-specified in _rs1_. The offset operand may be omitted; otherwise, any expression
-that computes the offset shall evaluate to zero. An implementation may or may
-not update the entire set of bytes atomically.
+corresponding to the cache block that contains the address specified in _rs1_.
+It is not required that _rs1_ is aligned to the size of a cache block. On
+faults the address stored in the *tval CSR is _rs1_; not the base address of
+the cache block. An implementation may or may not update the entire set of
```
aswaterman, line 1024: Same comment as https://github.com/riscv/riscv-isa-manual/pull/1433/files#r1693650461

aswaterman (commented):
I'm OK with it after the minor suggestions I made, but I still want @dkruckemyer-ventana to sign off.

src/cmo.adoc
Timmmm: Done

src/cmo.adoc
Timmmm: Done

src/cmo.adoc
Timmmm: Done

Timmmm (commented):
Sorry for the delay - finally got back to this! (It came up on the mailing list recently, so I thought I'd better get it done.)

I've applied your suggestion. Hopefully this is good to go now.

aswaterman (approved):
Thanks, @Timmmm.

Issue #1395, opened 2024-05-11
Merge privileged and unprivileged specs


Timmmm:
When I first started learning RISC-V I started by only reading the unprivileged spec, and trying to make a simple processor based on that. That's turned out to be extremely confusing because you have to read both documented. There are loads of features where the details are spread over both documents. In particular:

* exceptions
* CSRs

Information from the privileged spec can also affect behaviours from the unprivileged spec. For example `xenvcfg.FIOM` affects fences, even though the fence instruction description makes no mention of this.

So is there any benefit to splitting the spec like this? The only one I can think of is document length, but I would say that isn't a good trade-off.

aswaterman:
This will be done as part of a large-scale spec reorg.

Aman071106:
@Timmmm  I think there is one reasong why splitting is there, modularity is the thing required if someone wants to change some specific part of only one part (priveleged or unpriveleged)  , even distributing them over actual hardware would be easier task if they are splitted like this. Isn't that? Would love correction if I'm going in wrong direction.

PR #1434, opened 2024-05-29
Update PMP spec to use "memory operation"

Timmmm:
This part of the spec was written before the RVWMO section which introduced the "memory operation" nomenclature.

This updates the section to talk about memory operations instead of memory accesses.

I also reworked the text about CMOs and PMPs because it didn't quite make sense - the access size is irrelevant for PMPs; it's the memory operation size that matters, and that can be anything for CMOs since they are allowed to be decomposed.


src/machine.adoc
```diff
@@ -3233,47 +3233,49 @@ R/W/X permissions apply only to S and U modes.
 
 ===== Priority and Matching Logic
 
+On some implementations, misaligned loads, stores, and instruction
+fetches may be decomposed into multiple memory operations, some of which may
+succeed before an access-fault exception occurs (see <<memorymodel>>).
```
aswaterman, line 3238: <<memorymodel>> is probably a broken link since it's in the unpriv spec.  Maybe replace the parenthetical with ", as described in the RVWMO specification."

src/cmo.adoc
```diff
@@ -337,8 +337,16 @@ _This specification assumes that the above constraints will typically be met for
 main memory regions and may be met for certain I/O regions._
 ====
 
-Additionally, for the purposes of PMP and PMA checks, the access size of a CMO
-instruction equals the size of the cache block accessed by the instruction.
+[NOTE]
+====
+The access size for CMO instructions is equal to the size of the cache
+block, however that access can be decomposed into multiple memory operations.
```
aswaterman, line 343: I'm not sure we want to make this general of a statement.  It's clearly true that `cbo.zero` may give rise to multiple memory operations, but I don't think we want to imply that's the case for the clean/flush/inval.  Need @dkruckemyer-ventana to weigh in here.

aswaterman (commented):
Thanks. OK with me after my comments are addressed and @dkruckemyer-ventana weighs in.

src/cmo.adoc
Timmmm: Ah good point, I'll change it to `cbo.zero`.

src/machine.adoc
Timmmm: Ooops I did notice this but then forgot (wrote this a while ago). I will update it. (*cough* #1395 :-D)

Timmmm: Done

src/cmo.adoc
Timmmm: Done, but I used a vague "in some cases" so it doesn't add any new specifications for clean/flush/inval. Would be good to know what the actual behaviour intended there is. I would guess it's the same but lets wait for @dkruckemyer-ventana .

src/machine.adoc
aswaterman: #1395 is the long-term plan but we have bigger documentation fish to fry in the meantime.  It'll be a more involved effort than just concatenating the documents.

Timmmm:
@dkruckemyer-ventana any opinion?

aswaterman (approved):
I think we should proceed with this one.

PR #1443, opened 2024-05-31
Integrate Ssdbltrp and Smdbltrp into Priv spec

ved-rivos:


aswaterman (commented):
Thank you, @ved-rivos.  Can you add these two extensions to the table at the top of `priv-preface.adoc`, and add a one-line ?  LGTM otherwise.

ved-rivos:
> Can you add these two extensions to the table at the top of priv-preface.adoc, and add a one-line ? LGTM otherwise.

@aswaterman - I have updated the table in the Preface section as Draft and will send another PR to update once they are moved to Ratified state. Could you please say more about the one-line.

aswaterman:
Actually, don't worry about that, I think we can just merge this as-is.

aswaterman (approved):

Issue #1448, opened 2024-06-06
use docs-resources as submodule to pull in shared resources (pdf theme, fonts, etc)


kbroch-rivosinc:
relates to:

* https://github.com/riscv-admin/docs-sig/issues/48
* https://github.com/riscv-non-isa/riscv-iommu/pull/342#issuecomment-2150847986

wmat:
Recolved by https://github.com/riscv/riscv-isa-manual/pull/1449

PR #1449, opened 2024-06-07
add docs-resources submodule to get pdf theme,fonts, and global attri…

kbroch-rivosinc:
…butes

related to #1448


wmat:
The github action is failing to find the docs-resources. Perhaps the include path needs adjustment. I'd do it but I'm on my phone atm.

kbroch-rivosinc:
> The github action is failing to find the docs-resources. Perhaps the include path needs adjustment. I'd do it but I'm on my phone atm.

yes, The docker build is working fine locally but  I'm not appreciating some difference in the GH actions and paths.
I'm going to try to simplfy the make script and align it more with what's in docs-spec-template (without changing any of the build targets/outputs)

wmat (approved):
LGTM.  Thanks for this.

aswaterman:
@wmat @kbroch-rivosinc This should not have been merged as-is.  The build system is now less functional than it used to be.  In particular, dependence tracking no longer works properly.

We shouldn't be forcing consistency with other repos just for consistency's sake.  There really isn't a good reason this repo looks identical to the other ones; it has a different set of concerns and constraints than one-off repos for new extensions.

I will fix the problems that have been re-introduced, but going forward, please involve me when you make significant changes to stuff I've been working on.

PR #1453, opened 2024-06-12
Correct typos

wissygh:
The original expression of this sentence is ambiguous.  According to 'Parter 2.5 Access control by the state-enable CSRs' in The RISC-V Advanced Interrupt Architecture Version 1.0, bit 59(AIA) in stateen controls access to all other state added by the AIA and not controlled by bits 60(CSRIND) and 58(IMSIC).But “is not controlled by either the CSRIND or the IMSIC bits of mstateen0“ seems more like it's restricting "the AIA bit" rather than "state".


aswaterman (approved):
FYI @jhauser-us

Issue #122, opened 2018-01-01
Pseudoinstruction Name Consistency


steven-bellock:
If the words "pseudoinstruction", "pseudo-instruction", and "pseudo-op" all refer to the same thing we should pick one form for consistency's sake. I think "pseudoinstruction" is fine.

kasanovic:
Standardized on pseudoinstruction.

PR #1469, opened 2024-06-21
Make uses of 'pseudo' consistent

ThinkOpenly:
There are a couple of classes of misuse of "pseudo" within the document that conflict with uses elsewhere in the overall RISC-V ISA Specification:

1. All uses of "pseudo-ops" should instead be "pseudoinstructions".
1. "pseudosomething" sometimes has a space or hyphen separating "pseudo" from the rest of the word. We should pick one usage. I think the most common is with no separation. "pseudorandom" is in the dictionary, as an example.

### DIFF
--- a/src/hypervisor.adoc
+++ b/src/hypervisor.adoc
@@ -2300,7 +2300,7 @@ earlier.
 Standard +
 Instruction
 ^.>|Custom Value
-^.>|Pseudo-instruction Value
+^.>|Pseudoinstruction Value
 |Instruction address misaligned |Yes |No |Yes |No
 |Instruction access fault +
 Illegal instruction +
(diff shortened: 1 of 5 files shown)


nibrunieAtSi5 (approved):
`vectorcrypto.adoc` LGTM

@kdockser, @ben-marshall, @mjosaarinen for your review

aswaterman (approved):
A couple years ago we cleaned up most of the pseudo-instruction vs. pseudoinstruction inconsistencies, so this is moving in the right direction.

nick-knight:
See also https://github.com/riscv/riscv-isa-manual/issues/122 , and other issues/PRs/commits that reference it.

kdockser (commented):
LGTM

ThinkOpenly:
May need a GitHub CI action that scans for anything "pseudo " or "pseudo-". Unfortunately, something like that would trigger on a reference in the bibliography to "A Simple Unpredictable Pseudo-Random Number Generator". :-/

aswaterman:
I’m not a huge fan of using CI for this sort of thing, including for the sort of reason you mentioned.

Better for us to do a better job reviewing pull requests.

PR #1485, opened 2024-07-01
Update extending.adoc to delete extension name 'L' and 'T'

dzhang28:
The 'L' and 'T' extensions were placeholders but later removed.
Most changes were done by https://github.com/riscv/riscv-isa-manual/pull/655.
So the sentence here should be adapted as well.
Otherwise it would cause confusion for a new hand.
Because 'L' and 'T' extensions are neither complete nor planned standard extensions now.

### DIFF
--- a/src/extending.adoc
+++ b/src/extending.adoc
@@ -31,7 +31,7 @@ categories: _standard_ versus _non-standard_.
 
 * A standard extension is one that is generally useful and that is
 designed to not conflict with any other standard extension. Currently,
-"MAFDQCBPV", described in other chapters of this manual, are either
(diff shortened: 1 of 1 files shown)


aswaterman (changes requested):
We can remove `L` but we can't remove `T`, because it governs how `Ztso` is ordered.  See the lengthy discussion on https://github.com/riscv/riscv-isa-manual/pull/1391

dzhang28:
Thanks a lot for your clarification :)
I will resubmit.

aswaterman (approved):

Issue #481, opened 2020-01-30
Is `htimedelta` mandatory?


palmer-dabbelt:
From reading the draft H extension it sounds to me like all 64 bits of `htimedelta` must be implemented if the H extension is implemented.  This seems like unnecessary CSR state: on systems that are going to trap/emulate timer accesses anyway (via the counter disable bits) the guest's time offset can be stored in memory instead.

Making it mandatory makes my life easier, so I don't really care that much either way.  I just wanted to make sure it's actually that way.

avpatel:
The HTIMEDELTA CSR is tide with TIME CSR because it controls
TIME CSR value read from VS-mode and VU-mode.

TIME (VS-mode or VU-mode) = TIME (HS-mode or M-mode) + HTIMEDELTA

If HW does not implement TIME CSR then HW should not implement
HTIMEDELTA as well. In this case, OpenSBI will trap-n-emulate both
TIME CSR and HTIMEDELTA CSR. If HW implements TIME CSR then HW
has to implement HTIMEDELTA as well.

To summarize, TIME CSR is optional for HW hence HTIMEDELTA is
optional for HW as well. Also, HW designers should either implement
both CSRs or skip both.

aswaterman:
Implementations that trap into M-mode for access to the time CSR would
presumably trap accesses to htimedelta in the same manner, so there should
be no hardware cost for such designs.

For implementations that don’t trap accesses to the time CSR, this is an
important feature, and the cost seems justified.

Is there another class of implementation you’re thinking of that doesn’t
fit into one of these buckets?

aswaterman:
Consensus via silence--closing :)

PR #1486, opened 2024-07-01
Clarify mandate of htimedelta

cebarobot:
According to #481, htimedelta CSR is tide with time CSR. Implementations should either implement both CSRs or skip both.

### DIFF
--- a/src/hypervisor.adoc
+++ b/src/hypervisor.adoc
@@ -726,6 +726,9 @@ When XLEN=32, `htimedeltah` is a 32-bit read/write register
 that aliases bits 63:32 of `htimedelta`.
 Register `htimedeltah` does not exist when XLEN=64.
 
+If `time` CSR is implemented, `htimedelta` and `htimedeltah` (only for XLEN=32)
+must be implemented.
+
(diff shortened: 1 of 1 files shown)


allenjbaum:
Typo: tide -> tied

aswaterman:
I think the proposed change contradicts current behavior.  If `mcounteren`.TM=0, then HS-mode accesses to `time` will trap, which might ultimately result in a visible trap to HS-mode.  However, `htimedelta` is not conditioned on `mcounteren`.TM.  So it is indeed possible for `htimedelta` to exist from HS-mode's perspective even when `time` doesn't.

aswaterman:
Disagreed. The hypervisor extension chapter is written from the perspective of HS-mode. From that perspective, there is no difference between whether the time CSR is physically implemented or is disabled by mcounteren.

aswaterman:
I would be OK if you deleted the “conversely” clause and kept only the first part.

aswaterman (changes requested):
Putting a hold on this PR until the discussion is resolved.

aswaterman:
Merged manually via 257e5ca096725b1fabc73f674a9a780d76e3e7fc

PR #1489, opened 2024-07-02
vector: Define vtype to be a required assembler field

charlie-rivos:
vtype settings are all partially required assembler fields. LMUL, tail behavior, and mask behavior have been defined to be optional at some point in the history of the spec. However, omitting tail and mask behavior was previously deprecated for vset* instructions. For consistency, a default LMUL should also be deprecated.

Additionally, the LLVM assembler does not support default args for the vset* instructions.

### DIFF
--- a/src/example/sgemm.S
+++ b/src/example/sgemm.S
@@ -73,7 +73,7 @@ c_row_loop: # Loop across rows of C blocks
     mv cnp, cp # Initialize C n-loop pointer
 
 c_col_loop: # Loop across one row of C blocks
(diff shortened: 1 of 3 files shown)


aswaterman:
Thanks.  The NOTE didn't seem necessary to me--the history isn't terribly important, and in any case this document isn't authoritative about the assembly syntax.  So I manually merged the assembly changes via 75e857da51db157814f67e465c7262ae0a74c277.

charlie-rivos:
Thank you!

PR #1494, opened 2024-07-03
Relax behavior of some HINTs when MXLEN > XLEN

### LINKED ISSUE #1487: Proposal to remove RVC HINTs with rd != x0 to remove the paradox in ISA
The encoding of HINTs in RVC contains some instructions with rd != x0. When some CPU implements SXL or UXL change and chooses the simplest implementation to ignore this HINT, as the specification said, this implementation will change the architecture state that the sign-extends some wider GPR in narrow XLEN, causing HINTs to modify the architecture state.

To clarify this paradoxical behavior, the choices are:
1. Remove the following comments
   > This HINT encoding has been chosen so that simple implementations can ignore HINTs altogether, and instead execute a HINT as a regular computational instruction that happens not to mutate the architectural state."
2. Remove RVC HINTs with rd != x0
3. Explicitly allows an implementation to change architectural state when executing HINT instruction optionally

I suggest option 2 because these HINTs haven't been used now, and it breaks the behavior for simple implementation, which may implement modifiable XLEN.

aswaterman:
Resolves #1487

### DIFF
--- a/src/machine.adoc
+++ b/src/machine.adoc
@@ -581,6 +581,21 @@ always be a software bug, but machine operation is well-defined even in
 this case.
 ====
 
+Some HINT instructions are encoded as integer computational instructions that
+overwrite their destination register with its current value, e.g.,
+`c.addi x8, 0`.
+When such a HINT is executed with XLEN < MXLEN and bits MXLEN..XLEN of the
+destination register not all equal to bit XLEN-1, it is implementation-defined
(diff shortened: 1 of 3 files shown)


cyyself:
Thanks for doing this!

aswaterman:
@cyyself no, thank you: this is a special case we failed to properly document.

PR #1504, opened 2024-07-05
Fix C.NOP hint condition

bgasiorzewski:
This patch fixes an error in the instruction listing tables for the C extension. The definition of the C.NOP instruction specifies that using an immediate other than zero encodes a hint. The table was wrongly stating the opposite: that a zero immediate encodes a hint.


aswaterman (approved):
Thanks for the fix.

PR #1507, opened 2024-07-07
Fix funct7 codes in integer register-register inst wavedrom

### LINKED ISSUE #1452: funct7 values of SRA,SUBW,SRAW appear to be incorrect in unpriv section 4.2.2 of version 20240411.
Hi,
I think that the funct7 values in unpriv section 4.2.2 of version 20240411 are incorrect.  
In the V20191213, funct7 values are 7-bits (in section 5.2).  
The values are 0000000 or 0100000.  
image

However, in the version 20240411, funct7 values have only 6-bits; the least significant bits seem to be missing.  
The values are 000000 or 010000.   
image

Is this correct?  

The relevant file is:  
https://github.com/riscv/riscv-isa-manual/blob/353e88f8d5c629be5a42ada8618fcfa9acb5d287/src/images/wavedrom/rv64i_int-reg-reg.adoc?plain=1#L13

ved-rivos:
Closes #1452


aswaterman (approved):

aswaterman:
thanks @ved-rivos!

PR #1510, opened 2024-07-08
Clarify SFENCE.VMA for invalid virtual addresses

Timmmm:
The existing wording sounds like the instruction is *required* to have no effect, but SFENCE.VMA is allowed to over-fence, so it still can have an effect even for invalid addresses.

### DIFF
--- a/src/supervisor.adoc
+++ b/src/supervisor.adoc
@@ -1107,8 +1107,8 @@ virtual address in _rs1_ and that match the address space identified by
 integer register _rs2_, except for entries containing global mappings.
 
 If the value held in _rs1_ is not a valid virtual address, then the
-SFENCE.VMA instruction has no effect. No exception is raised in this
-case.
+SFENCE.VMA is not required to have any effect. No exception is raised
+in this case.
 
 When __rs2__&#8800;``x0``, bits SXLEN-1:ASIDMAX of the value held
 in _rs2_ are reserved for future standard use. Until their use is


Timmmm:
Hmm very good point. Though I still think it's weird to say it *has no effect*, when it actually might do incidentally. Perhaps just moving the NOTE next to this statement would make it better? So it reads like "it has no effect .... But..."

```
If the value held in _rs1_ is not a valid virtual address, then the
SFENCE.VMA instruction has no effect. No exception is raised in this
case.

[NOTE]
====
It is always legal to over-fence, e.g., by fencing only based on a
subset of the bits in _rs1_ and/or _rs2_, and/or by simply treating all
SFENCE.VMA instructions as having _rs1_=`x0` and/or _rs2_=`x0`. For
example, simpler implementations can ignore the virtual address in _rs1_
and the ASID value in _rs2_ and always perform a global fence. The
choice not to raise an exception when an invalid virtual address is held
in _rs1_ facilitates this type of simplification.
====

When __rs2__&#8800;``x0``, bits SXLEN-1:ASIDMAX of the value held
in _rs2_ are reserved for future standard use. Until their use is
defined by a standard extension, they should be zeroed by software and
ignored by current implementations. Furthermore, if
ASIDLEN<ASIDMAX, the implementation shall ignore bits
ASIDMAX-1:ASIDLEN of the value held in _rs2_.
```

aswaterman:
I think you threaded the needle!

Timmmm:
Thanks!

PR #1516, opened 2024-07-10
Add LCOFIE and LCOFIP to mie and mip diagrams.

wmat:
Adds LCOFIE and LCOFIP bits to the mie and mip diagrams in the Hypervisor chapter.


aswaterman (approved):
LGTM.

Issue #1518, opened 2024-07-10
duplicates: xperm.n == xperm4, xperm.b == xperm8, rev.b == brev8


mjosaarinen:
This can be confusing and I don't see why it would be intentional: 
- Sect 30.5.49. xperm.n and Sect 34.4.48. xperm4 seem to define exactly the same instruction with different names. 
- Also 30.5.48. xperm.b === 34.4.47. xperm8.
- Also 30.5.30. rev.b == 34.4.13. brev8.

I recall that (Krste) requested us to change the names of { rev.b, xperm.n, xperm.b } from the bitmanip names in late stages of scalar crypto ratification -- just before the public review. So we changed those. I guess the old names made a comeback from the old bitmanip specification into the merged spec?

Here's the original issue that motivated the change: https://github.com/riscv/riscv-crypto/issues/115

For scalar crypto the change happened for v1.0.0-rc2 in September 2021: https://github.com/riscv/riscv-crypto/releases/tag/v1.0.0-rc2-scalar

aswaterman:
The fundamental problem, IMO, is the duplication of the Zbkb, Zbkc, and Zbkx specs between the bitmanip and scalar crypto chapters.  I think it makes the most sense to consolidate them into the bitmanip chapter, but fix the mnemonics to correspond to the ones in the scalar crypto chapter, as you suggest.  If someone can make a PR that does this, I'd be happy to review it.

wmat:
I will take care of this PR.

aswaterman:
@mjosaarinen My PR fixes the mnemonics: https://github.com/riscv/riscv-isa-manual/pull/1524

Since no good deed goes unpunished, can you sanity-check Zbkb/Zbkc/Zbkx in the bitmanip chapter?  I did some spot checks.  The opcodes look right, but you might want to check them anyway.  The SAIL code is differently formatted, though.  (I didn't check its semantics.)  If you have any recommendations, please follow up with a PR, which I'll be happy to review.

mjosaarinen:
Thanks,

Here are some things from the review:

There's a typo here: https://github.com/riscv/riscv-isa-manual/blob/568e50adbfbaab478153807e0efed6f3bc908671/src/scalar-crypto.adoc?plain=1#L3356

[...]

PR #1522, opened 2024-07-11
Consolidates Zbkb, Zbkc, and Zbkx into he Bitmanip chapter.

wmat:
Deleted duplicate data from Scalar Crypto chapter and added xrefs to the content in the Bitmanip chapter.
Fixed up the IDs and alternate text in the Bitmanip chapter.


aswaterman (approved):
Thanks.  I'll handle the naming issue from #1518 in a follow-up PR.

PR #1524, opened 2024-07-12
Fix mnemonics of Zbkx/Zbkb instructions

aswaterman:
Resolves #1522

### DIFF
--- a/src/b-st-ext.adoc
+++ b/src/b-st-ext.adoc
@@ -912,7 +912,7 @@ common operations in cryptographic workloads.
 
 | &#10003; 
 | &#10003; 
-|  rev.b       
+|  brev8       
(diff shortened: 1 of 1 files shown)


src/b-st-ext.adoc
```diff
@@ -2569,13 +2569,13 @@ Included in::
 
 <<<
 [#insns-revb,reftext="Reverse bits in bytes"]
```
topperc, line 2571: Should the name be fixed in `#insns-revb` too?

aswaterman: Since that conflicts with a similar tag in scalar-crypto, I'm going to leave it alone for now, but someone should definitely come along and fix that up.

Issue #1479, opened 2024-06-27
Many line breaks in code of vector chapter


demin-han:
1. different indents from 0 to 4.  
2. the line is really long along with comment.

for example:
![image](https://github.com/riscv/riscv-isa-manual/assets/135933545/14c9d685-b90f-4f09-ae07-48e810b1cf15)

![image](https://github.com/riscv/riscv-isa-manual/assets/135933545/ad8a322f-790a-4c13-a598-1fdccd5fa4ff)

PR #1536, opened 2024-07-15
fix inconsistent indentation of pseudocode (left justify all if it)

kbroch-rivosinc:
relates to #1479


aswaterman (approved):
LGTM, thanks.

PR #1539, opened 2024-07-16
Add notes about semihosting NOPs; clarify there is only one semihosting exit NOP

aswaterman:
These non-normative clarifications were made in response to architecture review of the semihosting spec.


aswaterman:
Verbally approved in ARC meeting.

PR #1540, opened 2024-07-17
Clarify that only read-only accesses to the seed CSR are forbidden

### LINKED ISSUE #1728: Specify what imprecise traps are
The spec talks in 3 places about exceptions being "precise" or "imprecise":

> [The EEI defines for each trap whether it is handled precisely, though the recommendation is to maintain preciseness where possible. Contained and requested traps can be observed to be imprecise by software inside the execution environment. Invisible traps, by definition, cannot be observed to be precise or imprecise by software running inside the execution environment. Fatal traps can be observed to be imprecise by software running inside the execution environment, if known-errorful instructions do not cause immediate termination.](https://riscv-specs.timhutt.co.uk/spec/20240411/unpriv-isa-asciidoc.html#_exception_handling:~:text=The%20EEI%20defines,cause%20immediate%20termination.)

> [To aid in system debugging, we strongly recommend that, where possible, RISC-V processors precisely trap physical memory accesses that fail PMA checks. Precisely trapped PMA violations manifest as instruction, load, or store access-fault exceptions, distinct from virtual-memory page-fault exceptions. Precise PMA traps might not always be possible, for example, when probing a legacy bus architecture that uses access failures as part of the discovery mechanism. In this case, error responses from peripheral devices will be reported as imprecise bus-error interrupts.](https://riscv-specs.timhutt.co.uk/spec/20240411/priv-isa-asciidoc.html#scause:~:text=To%20aid%20in,bus%2Derror%20interrupts.)

> [PMP violations are always trapped precisely at the processor.](https://riscv-specs.timhutt.co.uk/spec/20240411/priv-isa-asciidoc.html#scause:~:text=PMP%20violations%20are%20always%20trapped%20precisely%20at%20the%20processor.)

However:

1. There's no definition of what "precise" and "imprecise" mean.
2. It doesn't describe what effect imprecise traps should have.

I think there's a lot of detail missing here. I assume the ARM spec (or maybe MIPS?) explains all of this and it's just assumed knowledge, but I'm not familiar with ARM or MIPS & I think the RISC-V spec shouldn't assume I am.

aswaterman:
Write-only accesses are legal (though not useful).

See https://github.com/riscv-software-src/riscv-isa-sim/issues/1728

### DIFF
--- a/src/scalar-crypto.adoc
+++ b/src/scalar-crypto.adoc
@@ -3569,8 +3569,8 @@ The write value (in `rs1` or `uimm`) must be ignored by implementations.
 The purpose of the write is to signal polling and flushing.
 
 The instruction `csrrw rd, seed, x0` can be used for fetching seed status
-and entropy values. It is available on both RV32 and RV64 base architectures
-and will zero-extend the 32-bit word to XLEN bits.
+and entropy values.
+It is available on both RV32 and RV64 base architectures.
 
 Encoding::
 [wavedrom, , svg]
--- a/src/scalar-crypto.adoc
+++ b/src/scalar-crypto.adoc
@@ -3562,14 +3562,15 @@ The 32-bit contents of `seed` are as follows:
 |`15: 0` |`entropy` |16 bits of randomness, only when `OPST=ES16`.
 |=======================================================================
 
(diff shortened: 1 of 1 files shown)


aswaterman:
FYI @mjosaarinen

PR #1541, opened 2024-07-17
Define format of memory-mapped msip register

aswaterman:
This matches ACLINT (and other vendor-specific CLINTs).


src/machine.adoc
```diff
@@ -1407,9 +1407,13 @@ compare register.
 Bits `mip`.MSIP and `mie`.MSIE are the interrupt-pending and
 interrupt-enable bits for machine-level software interrupts. MSIP is
 read-only in `mip`, and is written by accesses to memory-mapped control
-registers, which are used by remote harts to provide machine-level
-interprocessor interrupts. A hart can write its own MSIP bit using the
-same memory-mapped control register. If a system has only one hart, or
+registers, which are used to provide machine-level
+interprocessor interrupts.
+A hart's memory-mapped `msip` register is a 32-bit read/write register,
+where bits 31--1 read as zero and bit 0 is contains the MSIP bit.
```
kasanovic, line 1413: spurious "is" in "...bit 0 is contains..."

aswaterman: thanks, fixed

PR #1552, opened 2024-07-23
Typos fix in naming.adoc and hypervisor.adoc

### LINKED ISSUE #1550: Typos found
While reading the ISA manuals I came across with two typos in distinct parts of the manual.

### Typo 1

In the unprivileged manual, Section 39.8, page 634, path: src/naming.adoc there is a typo in the note "Many augmentations to the hypervisor-level **archtecture**"

### Typo 2

In the privileged manual, Section 19.2.9, page 152, path: src/hypervisor.adoc there is this sentence," `htinst` is a *WARL* register that **need only** be able to hold the values that the implementation may automatically write to it on a trap. " which I believe would be better written as " `htinst` is a *WARL* register that **only needs to** be able to hold the values that the implementation may automatically write to it on a trap. "

AFOliveira:
Fixes #1550

### DIFF
--- a/src/hypervisor.adoc
+++ b/src/hypervisor.adoc
@@ -803,7 +803,7 @@ are documented in <<tinst-vals>>.
 .Hypervisor trap instruction (`htinst`) register.
 include::images/bytefield/htinstreg.edn[]
 
-`htinst` is a *WARL* register that need only be able to hold the values that
+`htinst` is a *WARL* register that only needs to be able to hold the values that
 the implementation may automatically write to it on a trap.
 
 [[hgatp]]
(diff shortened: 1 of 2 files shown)


src/hypervisor.adoc
```diff
@@ -1583,7 +1583,7 @@ are documented in <<tinst-vals>>.
 .Machine trap instruction (`mtinst`) register.
 include::images/bytefield/mtinstreg.edn[]
 
-`mtinst` is a *WARL* register that need only be able to hold the values that
+`mtinst` is a *WARL* register that only needs to be able to hold the values that
```
aswaterman, line 1586: @wmat @AFOliveira this and the above were not typos; it is syntactically correct English.  The new text is fine, so you don't need to revert the PR, but going forward, please don't treat this grammatical pattern as a bug.

aswaterman: Pondering it a bit, I think it reads better the old way around, so I'm going to push a revert of these two.  (The one down below is obviously a correct fix; thanks for that.)

PR #1560, opened 2024-07-24
Add missing "else" in SSAMOSWAP pseudocode

aswaterman:
This doesn't change the meaning of the pseudocode because "raise" is presumably a terminal statement, but it makes the code easier to read.

h/t @tsaiyenting


aswaterman:
@ved-rivos can you sanity-check this one?

ved-rivos (approved):

ved-rivos:
@aswaterman lgtm.
ty @tsaiyenting

aswaterman:
Thanks!

PR #1564, opened 2024-07-25
Add draft of Svukte extension

aswaterman:
Svkt provides a means to make user-mode accesses to supervisor memory raise page faults in constant time, mitigating attacks that attempt to discover the supervisor software's address-space layout.

I plan to submit this as a fast-track soon.


src/supervisor.adoc
```diff
@@ -2270,6 +2282,18 @@ Invalid PTEs using a bounded timer, or making address-translation caches
 coherent with store instructions that modify PTEs.
 ====
 
+[[sec:svkt]]
+== "Svkt" Extension for Address-Independent Latency of User-Mode Translation of Supervisor Addresses, Version 0.2
```
ingallsj, line 2286: The Faults are what is constant time, not the Translation (Table Walks).
```suggestion
== "Svkt" Extension for Address-Independent Latency of User-Mode Faults of Supervisor Addresses, Version 0.2
```

src/supervisor.adoc
```diff
@@ -2270,6 +2282,18 @@ Invalid PTEs using a bounded timer, or making address-translation caches
...
+
+The Svkt extension provides a means to make user-mode accesses to supervisor
```
ingallsj, line 2288: Naming suggestion: it's not the whole translation that occurs with constant time, just the exceptions.
```suggestion
The Svkte extension provides a means to make user-mode accesses to supervisor
```

src/supervisor.adoc
aswaterman: The translation and the faulting are hard to disentangle, since it’s the variable latency of the translation that made the latency of the fault vary. But your version might make the intent clearer; we can try it out.

“faults to” rather than “faults of”, right?

ingallsj: Sure, "faults to".

src/supervisor.adoc
```diff
@@ -824,6 +825,17 @@ The definitions of the CBCFE and CBIE fields will be furnished by the
...
 change prior to the ratification of that extension.
 
+If the Svkt extension is implemented, the SVKT field affects the behavior of
+instruction fetches and explicit memory accesses.
+When SVKT=0, instruction fetches and explicit memory accesses proceed as
+though the Svkt extension were not implemented.
+When SVKT=1, an instruction fetch or explicit memory access with effective
+privilege mode U or VU that attempts to access a virtual address {ge}
+2^SXLEN-1^ raises a page-fault exception corresponding to the original access
+type.
```
AndybnACT, line 835: Do we want more flexibility here? e.g. for 32-bit runtime. AFIAK, 32-bit Linux uses virtual addresses starting from `0xc0000000` as supervisor mapping. Different OSes may define it differently on 32-bit platform though, due to contested address space, e.g. windows fits into this half/half s/u division, IIRC.

deepak0414: Two comments here

1) I am not sure, if it's helpful for 32bit. 2 GB  (or 1GB when supervisor_addr_start=`0xc0000000`) is relatively small address space to have any significant benefit from KASLR.
So may be mandating only for 64 bit, probably makes it cleaner. 
Side note: Not that MSFT Windows should be of concern here but Windows 32bit also have an option to split address space in 3GB/1GB. Anyways 32bit windows is deprecated now.

2) `2^SXLEN-1` doesn't work when pointer masking is enabled because canonical checks are relaxed.

Instead of `2^SXLEN-1`, it should be

```
2^VABITS-1
VABITS = 39 if Sv39
VABITS = 48 if Sv48
VABITS = 57 if Sv57

If Bare mode then this bit has no effect?
```

deepak0414: Never mind about comment # 2.
Shouldn't be a problem, as long as effective address is calculated after masking is performed and generated address is canonicalized again.

aswaterman: @AndybnACT We did discuss that, but the tentative plan is to support Svkt only for RV64 for the reason that @deepak0414 mentioned: KASLR is seemingly much less beneficial on RV32.  But if we did add an RV32 variant, we of course would design it so that it would be usable by the Linux kernel.  This would either mean using a 3G/1G split (most likely) or offering the choice between 3G/1G or 2G/2G (less likely).

@deepak0414 Right, re: # 2, the address transformation is applied before any virtual-address handling, and so there is no problem with pointer masking.

src/supervisor.adoc
aswaterman: Changing the name is a bit of a pain, so I'm going to wait to see if there are further suggestions, then pick the best.

src/supervisor.adoc
aswaterman: Done.

src/supervisor.adoc
aswaterman: I added a note about not having Svkt in RV32. 9a9ba46df4e7bc7f72288c33dc97e020324f59ff

src/supervisor.adoc
```diff
@@ -2270,6 +2279,45 @@ Invalid PTEs using a bounded timer, or making address-translation caches
...
+See <<sec:senvcfg>> and <<sec:henvcfg>> for the definitions of those fields.
+
+The Svkt extension depends on Sv39.
+
+NOTE: Svkt is not defined for Sv32 because the small address space limits the
+available entropy, reducing the effectiveness of address-space layout
+randomization.
+If an Sv32 variant were to be defined, it would need to account for the fact
+that it is more common to reserve only the upper 1 GiB of the virtual-address
+space for for the operating system, leaving the lower 3 GiB for user
```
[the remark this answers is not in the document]
aswaterman: thanks.

src/supervisor.adoc
```diff
@@ -730,6 +730,7 @@ characteristics of the U-mode execution environment.
   {bits:  2, name: 'CBIE'},
   {bits:  1, name: 'CBCFE'},
   {bits:  1, name: 'CBZE'},
+  {bits:  1, name: 'SVKT'},
```
SiFiveHolland, line 733: Since Sekt applies only to RV64 or larger, wouldn't it make sense to put `henvcfg. HUVKT` and `senvcfg.SVKT` at bit offset >= 32, like was done for PMM? This avoids consuming bits in the low word that are irrelevant for RV32.

aswaterman: The ARC was not completely confident that we would never define an RV32 variant of this feature, hence locating the bit down here.

src/supervisor.adoc
```diff
@@ -2270,6 +2279,45 @@ Invalid PTEs using a bounded timer, or making address-translation caches
...
+
+NOTE: Since whether an instruction is Svukte-qualified depends on the _effective_
+privilege mode of the access, even some instructions executed in HS-mode or M-mode
+(e.g. HLV with `hstatus`.SPVP=0, or LW with `mstatus`.MPRV=1 and
+`mstatus`.MPP=U) are Svukte-qualified.
+
+As described in <<sec:henvcfg>>, the `henvcfg`.HUKTE field, rather than the
+`senvcfg`.UKTE field, determines whether HLV, HLVX, and HSV instructions
+executed within U-mode are Svukte-qualified.
+
```
ingallsj, line 2320: Are there `HUKTE` fields in both `henvcfg` and `hstatus`?
```suggestion
If the hypervisor extension is additionally implemented, the `hstatus`.HUKTE
field is also writable.
See <<sec:senvcfg>> and <<sec:hstatus>> for the definitions of those fields.

The Svukte extension depends on Sv39.

NOTE: Svukte is not defined for Sv32 because the small address space limits the
available entropy, reducing the effectiveness of address-space layout
randomization.
If an Sv32 variant were to be defined, it would need to account for the fact
that it is more common to reserve only the upper 1 GiB of the virtual-address
space for the operating system, leaving the lower 3 GiB for user
processes.

When `senvcfg`.UKTE=1, an instruction fetch or explicit memory access whose
effective privilege mode is U or VU is considered to be _Svukte-qualified_.
For any Svukte-qualified memory access, virtual addresses {ge} 2^SXLEN-1^ are
considered to be invalid; hence, an Svukte-qualified access to such an address
raises a page-fault exception corresponding to the original access type.
The timing of an instruction that raises an exception for this reason must be
independent of the faulting virtual address.

NOTE: Since whether an instruction is Svukte-qualified depends on the _effective_
privilege mode of the access, even some instructions executed in HS-mode or M-mode
(e.g. HLV with `hstatus`.SPVP=0, or LW with `mstatus`.MPRV=1 and
`mstatus`.MPP=U) are Svukte-qualified.

As described in <<sec:hstatus>>, the `hstatus`.HUKTE field, rather than the
`senvcfg`.UKTE field, determines whether HLV, HLVX, and HSV instructions
executed within U-mode are Svukte-qualified.

```

aswaterman: Editing error.  As the hypervisor chapter says, the bit lives in `hstatus`.

aswaterman: Fixed.

src/supervisor.adoc
```diff
@@ -2270,6 +2279,45 @@ Invalid PTEs using a bounded timer, or making address-translation caches
...
+randomization.
+If an Sv32 variant were to be defined, it would need to account for the fact
+that it is more common to reserve only the upper 1 GiB of the virtual-address
+space for the operating system, leaving the lower 3 GiB for user
+processes.
+
+When `senvcfg`.UKTE=1, an instruction fetch or explicit memory access whose
+effective privilege mode is U or VU is considered to be _Svukte-qualified_.
+For any Svukte-qualified memory access, virtual addresses {ge} 2^SXLEN-1^ are
+considered to be invalid; hence, an Svukte-qualified access to such an address
```
cwshu, line 2307: > For any Svukte-qualified memory access, virtual addresses {ge} 2^SXLEN-1^ are
> considered to be invalid;

When CPU executes in VU-mode, do we also check 2^SXLEN-1 bit of virtual address (is invalid) in G-stage translation?

aswaterman: I think the spec already says everything it needs to say. Svukte-qualified accesses are affected; the rest aren’t. Page-table walks aren’t Svukte-qualified.

cwshu: Thanks for confirmation!

src/supervisor.adoc
```diff
@@ -2270,6 +2279,45 @@ Invalid PTEs using a bounded timer, or making address-translation caches
 coherent with store instructions that modify PTEs.
 ====
 
+[[sec:svukte]]
+== "Svukte" Extension for Address-Independent Latency of User-Mode Faults to Supervisor Addresses, Version 0.3
+
+The Svukte extension provides a means to make user-mode accesses to supervisor
+memory raise page faults in constant time, mitigating attacks that attempt to
```
ingallsj, line 2286: PR feedback from tech chairs: time invariant based on page table contents, not constant time of the whole exception handling.
```suggestion
memory raise page faults with timing independent of page table data, mitigating attacks that attempt to
```

ved-rivos: I think the point is to not even look at the page table contents. The very act of accessing the page table will make it not constant time and could expose KASLR.

aswaterman: I'm certainly open to other ways of phrasing this, but I agree with Ved that the proposed wording seems to unintentionally weaken the spec by allowing implementations to perform the first level of the page-table walk then discard the result.  (But I think the proposed wording does prohibit accessing subsequent levels of the page table.)

Unlike Ved, I haven't convinced myself that opens up a timing attack, but even so, it wasn't my intent.

Another option would be "with timing independent of the contents of `satp` or the page tables".

Another possibility is "in constant time" => "without accessing the page tables", but that has the downside of shifting the emphasis towards the implementation and away from the high-level intent.

ingallsj: Ooh I like that one: "with timing independent of the contents of satp or the page tables".

aswaterman: Me too, but only if we can find a way to phrase it that includes all active `atp` registers.  "independent of the address-translation configuration and the page tables"?

ved-rivos: Maybe we could leave the normative statement as is and add a short note to guide implementation with some text like "without accessing page tables or address translation caches" so that it provides enough hints that this fault is supposed to be a fault caused by agen, similar to an invalid VA or too-wide GPA case, and not a TLB/PWH caused fault?

ved-rivos: Maybe we could leave the normative statement as is and add a short note to guide implementation with some text like "without accessing page tables or address translation caches" so that it provides enough hints that this fault is supposed to be a fault caused by agen, similar to an invalid VA or too-wide GPA case, and not a TLB/PWH caused fault?

aswaterman: I can live with either my most recent proposal or Ved's suggestion.  Let's see if anyone else chimes in.

src/supervisor.adoc
allenjbaum: This doesn't explicitly say that the PTE.U bit is ignored, but it certainly implicitly does. I would make that explicit, and/or explain in section 10.31.1 that "The U bit indicates whether the page is accessible to user mode." [unless Svutke is enabled] or something

aswaterman: The constraint on the implementation is much stricter than that. The implementation shouldn’t even perform the memory access corresponding to the PTE, and so it shouldn’t even be cognizant as to the value of PTE.U.

Don’t you think the next sentence in the spec (“The timing […] must be independent of the faulting VA”) gets this across? If not, how would you say it? Challenge: you cannot mention the PTE contents, given the machine’s behavior mustn’t depend on the PTE contents.

allenjbaum:
OK, what happens if the extension is enabled, the Ubit in the PTE is set ( so the access would be valid if the extension was disabled). Does that override the U-bit? Implicitly, it does, but I'd like to see that explicit (possibly in this extension spec, and/or in the the description of the U-bit in the spec section on the definition of the U-bit

ved-rivos:
>OK, what happens if the extension is enabled, the Ubit in the PTE is set ( so the access would be valid if the extension was disabled). Does that override the U-bit? Implicitly, it does, but I'd like to see that explicit (possibly in this extension spec, and/or in the the description of the U-bit in the spec section on the definition of the U-bit

The basic premise of Svukte is to take advantage of the convention in all 64-bit OS that user mappings are in positive address space and supervisor mappings are in negative address space. An OS that create user mappings in negative address space - even a single mapping - cannot turn on Svukte or the OS has to provide an emulation. For instance, for x86 based Linux legacy vDSO were located in negative address space and when LASS was introduced the OS provided emulation for such legacy vDSO. Non legacy vDSO on x86 and vDSO on RISC-V are located in positive address space. The point of Svukte is to fault on user access to negative addresses without consulting the address translation caches or doing implicit accesses to the page tables. 

So I think the state of the U bit is a don't care for this extension. For that matter the PTE itself may not be valid at all.

allenjbaum:
I understand that convention of supervisor mappings in negative VA space.
I agree that *implicitly* the state of the U-bit ( or the validity of the
PTE entry) is a don't care.
I am just saying that it should be *explicit*, and not require a careful
reading.
There is nothing in the spec that says that PTE entries *must* not be
fetched - just that, if they are accessed, the amount of time to get to the
leaf PTE should be constant
(it isn't clear how time is measured: realtime, #cycles, etc - but my
assumption is however SW could measure it).
There is nothing in the spec that says an OS that allows U-mode mappings in
negative address space cannot  (or must-not) turn on Svukte.
Bad things may (or may not) happen if it does (as Ved points out), but
certainly there is no HW enforcement.
There is nothing that prevents describing the obvious  implementation
(short circuiting the translation process) into the spec
 ( or even making it a requirement if it is that obvious and simple).
instead of making all implementors puzzle it out / re-invent it from first
principles.

aswaterman:
OK, I'll make it explicit.

src/supervisor.adoc
aswaterman: I went with my version, but we can revisit this later if need be.

allenjbaum:
I would interpret "user-mode accesses to supervisor memory" to mean "where
the U-bit is zero".
"supervisor memory" is not a RISC-V architectural definition, except for
the above.

If there is an obvious implementation of this, why not just come right out
and describe it in a non-normative note?
 I guarantee no one will firebomb your hourse if you do that (but if you
don't - no guarantees)

aswaterman:
I don’t understand what you’re complaining about at this point. The note I added, that the PTE contents don’t matter, is even stronger than the thing you asked for!

lfiolhais:
Hey everyone, just to piggy back on @allenjbaum comment, as I share the same opinion. I also misinterpreted the text on my first reading, and thought the hart would have to check the U bit on the PTE on every U-mode access and deliver an exception in constant-time (even though an implementation like this would be possible). However, after the presentation at the Security HC meeting, I now understand the extension much better. The desired final implementation is to AND three bits: vaddr[2^SXLEN-1], priv == u-mode, and senvcfg.UKTE (or the equivalent for the guest). I think it would be immensely helpful to add this short sentence as an implementation hint to the reader.

aswaterman:
@lfiolhais after discussing offline with Allen, I did add another sentence to the NOTE that directly addresses the matter.  (No TLBs or page-table walks occur => U-bit isn't relevant)

lfiolhais:
@aswaterman Thanks! I think it's better now.

src/hypervisor.adoc
```diff
@@ -652,6 +659,13 @@ The definition of the CBZE field is furnished by the Zicboz extension.
 
 The definitions of the CBCFE and CBIE fields are furnished by the Zicbom extension.
 
+If the Svkt extension is implemented, the HUVKT field determines
+whether the HLV, HLVX, and HSV instructions, when executed in U-mode,
+are Svkt-qualified.
+When one of these instructions is executed in U-mode, it behaves as though
+`senvcfg`.SVKT were set to the value of HUVKT.
+If Svkt is not implemented, HUVKT is read-only zero.
+
```
ingallsj, line 668: Old/duplicate text?

aswaterman: Yes, thanks for catching it.

aswaterman:
ARC has approved this extension, but the PR will stay open until ratification.

stupidfuz:
Hi @aswaterman — we've been studying this PR together with the discussion threads above, the above discussion contains a wealth of information, and we believe it is a valuable extension. We'd like to confirm two points of understanding before we build on them. Apologies if either has already been settled somewhere in the (now collapsed) review threads.

**1. On the note "An Svukte-qualified access to such an address raises an exception even if the underlying PTE would have otherwise allowed the access."**

We read this sentence as describing hart behavior only, and as placing no architectural constraint on how supervisor software constructs its page tables. Concretely:

- Svukte does not prohibit supervisor software from establishing a user-accessible mapping (PTE.U=1) at a virtual address ≥ 2^(SXLEN-1). Such a mapping remains architecturally well-formed, and enabling Svukte does not invalidate it or alter the meaning of the U bit.
- However, once `senvcfg.UKTE`=1 (with `satp.MODE` ≠ Bare), any Svukte-qualified access to such an address unconditionally raises a page-fault exception corresponding to the original access type, with timing independent of the faulting virtual address. The PTE is never consulted, so neither its permission bits nor its validity have any bearing on the outcome or the timing — the same exception is raised whether the address is unmapped, mapped without U, or mapped with U=1, so nothing about the presence or attributes of an upper-half mapping is exposed.
- Consequently, Svukte takes no position on why such a mapping might exist, nor does it provide any means for software to reach it. If an OS genuinely requires user-accessible mappings in the upper half of the address space, its options are to leave Svukte disabled, or to provide a software mechanism such as emulation in the fault handler.

Is that the intended reading?

**2. On the expected implementation of the timing requirement.**

We understand the natural implementation to be one in which the fault is raised purely before address translation is initiated: for a Svukte-qualified access to the upper half, no address-translation cache (TLB or page-walk cache) is consulted and no page-table walk is performed. The fault is therefore of the same character as an existing address-generation-time fault on an invalid (non-canonical) virtual address, rather than one produced by the translation hardware — which is precisely what makes its latency inherently uncorrelated with the faulting address.

We would also appreciate confirmation on the normative status of this: the requirement in the body of the section is stated as timing independence from the faulting virtual address, while "without accessing the page tables or address-translation caches" appears in a non-normative NOTE. Should we read an implementation that does consult the page tables and then equalizes the resulting latency as non-conforming, or merely as strongly discouraged?

Thanks for the work on this extension. We want to be confident that our interpretation lines up with the intent before we commit to it. Any confirmation or correction on the two points above would be much appreciated.

aswaterman:
1) Yes.

2) It was a deliberate choice to make that note non-normative.  Timing independence is the architectural requirement.  The mechanism is up to the implementer (but I expect nearly everyone will follow our suggestion).

nwrousell:
I'm working on the ACTs for the Svukte extension (https://github.com/riscv/riscv-arch-test/pull/2113) and will be ready soon to add coverpoint mappings to normative rules. What's the process for adding normative rule anchors to the .adoc spec? Is this process kicked off after this PR is merged/extension is ratified?

jordancarlin:
> I'm working on the ACTs for the Svukte extension ([riscv/riscv-arch-test#2113](https://github.com/riscv/riscv-arch-test/pull/2113)) and will be ready soon to add coverpoint mappings to normative rules. What's the process for adding normative rule anchors to the .adoc spec? Is this process kicked off after this PR is merged/extension is ratified?

Ideally, they are added by the spec author as part of drafting the specification. They will be the person most familiar with what is normative, so are best suited to tag the spec. If that doesn't happen, anyone else can tag the spec and submit a PR. In this case, if @aswaterman is not planning to tag the spec, you could create a PR targeting this branch with the normative rules.

aswaterman:
I probably won't get around to it any time soon, but if you want to file a PR against this branch, go for it, and tag both Jordan and me for review.

src/priv/hypervisor.adoc
```diff
@@ -191,7 +191,9 @@ tracking and controlling the exception behavior of a VS-mode guest.
   {bits:  1, name: 'VTVM'},
   {bits:  1, name: 'VTW'},
   {bits:  1, name: 'VTSR'},
-  {bits:  9, name: 'WPRI'},
+  {bits:  1, name: 'WPRI'},
+  {bits:  1, name: 'HUKTE'},
```
pmundkur: I'm wondering why `HUKTE` is added to the 32-bit `hstatus` when a Sv39 guest is not supported when `HSXLEN` is 32.

aswaterman: It shouldn't be :)

aswaterman: Fixed, thanks for the bug report.

stupidfuz:
> 1. Yes.
> 2. It was a deliberate choice to make that note non-normative.  Timing independence is the architectural requirement.  The mechanism is up to the implementer (but I expect nearly everyone will follow our suggestion).

Thanks for the clarification, @aswaterman   — that settles both points for us.

We have one further question: does software/hardware prefetching fall within the scope of Svukte's constraints? The Svukte-qualified definition refers to "an instruction fetch or explicit memory access," and we'd just like to confirm whether a prefetch.i/r/w instruction and/or a hardware prefetch operation is covered or not.

aswaterman:
In general, it's hard to circumscribe a complete set of sensible microarchitectural behaviors.  It's pretty obvious that a high-quality implementation of Svukte will not allow the prefetch instructions to leak information about the OS address-space layout to userspace, but how/whether to normatively specify that is less clear.

PR #1570, opened 2024-07-26
Clarify interaction of CBOs and PBMTs

Timmmm:
The behaviour can be technically inferred from this paragraph - as noted in one of the comments on that issue:

> Accessing the same location using different cacheability attributes may cause loss of coherence. Executing the following sequence between such accesses prevents both loss of coherence and loss of memory ordering: `fence iorw, iorw`, followed by `cbo.flush` to an address of that location, followed by a `fence iorw, iorw`.

However this is subtle, which means implementation bugs are likely. Making it explicit removes the risk of confusion and the resulting bugs.

### DIFF
--- a/src/cmo.adoc
+++ b/src/cmo.adoc
@@ -772,6 +772,13 @@ The following instructions comprise the Zicbom extension:
 
 |===
 
+[NOTE]
+====
+_Cache-block management instructions ignore cacheability attributes and operate
+on the cache block irrespective of the PMA cacheable attribute and any Page-Based
+Memory Type (PBMT) downgrade from cacheable to non-cacheable._
+====
+
 [#Zicboz,reftext="Cache-Block Zero Instructions"]
 ==== Cache-Block Zero Instructions
 


Timmmm:
Can I get a review of this? I think it should be uncontroversial.

gfavor:
@ingallsj The existing note about cboz makes a symmetric statement about applying to both cacheable and noncacheable memory, while the newly added note is asymetric and could be taken to suggest that cbom's don't apply to non-downgraded noncacheable memory.

Or, put differently and maybe better, the cboz note talks in terms of the effective memory type, so why wouldn't the cbom note do similar (which avoids any ambiguities and is succinct)?

gfavor:
@ingallsj @Timmmm Changing this new cbom note to be like the cboz note would also address questions raised in recent weeks (elsewhere) about whether cbom applies to noncacheable as well as cacheable memory.

wmat:
I can merge, as I would assume the addition of clarifying non-normative
text to be a good thing?

gfavor:
I confirmed with DavidK that the following succinct and all encompassing statement is correct:  CBOs ignore cacheability.  (Both before and after any potential PBMT override, i.e. the effective cacheable attribute is ignored.)

In light of the sub-case of PMA-cacheable turning into noncacheable after PBMT override being one of multiple sources of ambiguity/confusion, and trying to maintain similar wording to the existing cboz note, I would expand and modify the preceding to say:

Cache-block management instructions ignore cacheability attributes and operate on the cache block irrespective of the PMA cacheable attribute and any Page-Based Memory Type (PBMT) downgrade from cacheable to noncacheable.

If one of you all can update the proposed new note accordingly, that would be great.  Otherwise, once I'm done traveling in a few days and can get back in front of a proper computer, I can make the change myself.

wmat:
Thanks @gfavor.   @Timmmm do you want to add Greg's text?

Timmmm:
Yeah that sounds very clear. I'll update this PR.

Timmmm:
Updated.

PR #1581, opened 2024-08-06
Specify the interaction between CSRRS/C and "reads as" bits

Timmmm:
`pmpaddr` uniquely contains writable bits that "read as" some value, but have an underlying value that can be different. The behaviour when you modify *different* bits in the CSR using `csrr{sc}[i]` was not clear.

This is the behaviour that the Sail model, SPIKE and the open-source Rocket core all implement.

### DIFF
--- a/src/zicsr.adoc
+++ b/src/zicsr.adoc
@@ -47,6 +47,16 @@ treated as a bit mask that specifies bit positions to be cleared in the
 CSR. Any bit that is high in _rs1_ will cause the corresponding bit to
 be cleared in the CSR, if that CSR bit is writable.
 
+[NOTE]
+====
+Since CSRRS and CSRRC perform a read-modify-write, any bits that read as
+a different value to their underlying value may be modified by these
+instructions even if the corresponding bit is not set in _rs1_. For example
+pmpaddr[G-1] may have an underlying value of 1 but read as 0. Executing
+CSRRC or CSRRS to modify a different bit will cause 0 to be read from
+pmpaddr[G-1] and then written back, updating the underlying value to 0.
+====
+
 For both CSRRS and CSRRC, if _rs1_=`x0`, then the instruction will not
 write to the CSR at all, and so shall not cause any of the side effects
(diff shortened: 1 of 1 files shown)


pdonahue-ventana:
IIRC, this also affects mepc[1] if misa.C is writable.

Timmmm:
Ah yes you're right:

> If an implementation allows IALIGN to be either 16 or 32 (by changing CSR misa, for example), then, whenever IALIGN=32, bit mepc[1] is masked on reads so that it appears to be 0. This masking occurs also for the implicit read by the MRET instruction. Though masked, mepc[1] remains writable when IALIGN=32.

For completeness, [here's the discussion on the mailing list](https://lists.riscv.org/g/tech-unprivileged/message/923). The proposal there (which I don't really like) is to make it implementation defined.

src/zicsr.adoc
```diff
@@ -47,6 +47,16 @@ treated as a bit mask that specifies bit positions to be cleared in the
 CSR. Any bit that is high in _rs1_ will cause the corresponding bit to
 be cleared in the CSR, if that CSR bit is writable.
 
+[NOTE]
+====
+Since CSRRS and CSRRC perform a read-modify-write, any bits that read as
```
aswaterman: read-modify-write operation

src/zicsr.adoc
```diff
@@ -47,6 +47,16 @@ treated as a bit mask that specifies bit positions to be cleared in the
...
+a different value to their underlying value may be modified by these
+instructions even if the corresponding bit is not set in _rs1_. For example
```
aswaterman: comma after example

aswaterman (commented):
LGTM after my nits are picked

src/zicsr.adoc
Timmmm: Updated

src/zicsr.adoc
Timmmm: Added

PR #1583, opened 2024-08-06
Use standard "depends upon" verbage rather than "requires"

patrick-rivos:
This fixes the few instances of "requires" I found in the spec.

References:
https://github.com/riscv/riscv-v-spec/issues/723#issuecomment-949542444
https://github.com/riscv/riscv-v-spec/commit/e8594a7b9f4c22d3c02e2cdf81ad5db8d0bb8c01
https://github.com/riscv/riscv-isa-manual/blob/399c9a759eb4540a65c60e2cc236164821ff2346/src/naming.tex#L46

cc @aswaterman


aswaterman (approved):
Thanks.  If you stumble across more of these, please fix them analogously.

PR #1585, opened 2024-08-07
Changing .adoc to .edn for wavedrom files

kersten1:


kersten1:
This PR is the start of renaming all of the wavedrom files from .adoc to .edn. They work with the .adoc file type but .edn is more correct.

mwachs5:
Should this PR title be updated to say `.edn` not `.edu`

aswaterman (approved):

Issue #1567, opened 2024-07-25
Standardize and automate instructions documentation


AFOliveira:
After analyzing the RISC-V manual and consulting many online forums I found out that instructions are one of the parts of this manual that I believe have the most room for improvement. Taking the  [conditional branches](https://github.com/riscv/riscv-isa-manual/blob/fb235792a79a97b18b818bc0315737d08e231b35/src/rv32.adoc?plain=1#L563C1-L570C1) as an example there is only one paragraph and one diagram for all the branch types.

> Branch instructions compare two registers. BEQ and BNE take the branch
> if registers _rs1_ and _rs2_ are equal or unequal respectively. BLT and
> BLTU take the branch if _rs1_ is less than _rs2_, using signed and
> unsigned comparison respectively. BGE and BGEU take the branch if _rs1_
> is greater than or equal to _rs2_, using signed and unsigned comparison
> respectively. Note, BGT, BGTU, BLE, and BLEU can be synthesized by
> reversing the operands to BLT, BLTU, BGE, and BGEU, respectively.

This summarized text approach on part of the documentation makes it difficult on readers (as mentioned in issue #1396 ) and harder to maintain for the community, while also creating a lack of consistency throughout the manual. This problem is even more noticeable when considering all the extensions in the manual and the different maintainers supporting them. It seems that some extensions already follow a different approach like the (["B" Extension for Bit Manipulation](https://github.com/riscv/riscv-isa-manual/blob/main/src/b-st-ext.adoc#orcb)), which is much clearer, and it's also much more in line with the industry standard. My proposal is to migrate this approach to all the base instructions as well. I believe that the "standard" should include:

- Function Name
- Function Synopsis
- Mnemonic
- Instruction Encoding
- Description
- Operation (SAIL format)

This standardized way of describing each instruction, which is common on ISA manuals from other vendors, would also create a great index on the "Table of Contents" for developers trying to quickly find specific instructions description (example below from ARM A64 manual).

![Screenshot 2024-07-23 113300](https://github.com/user-attachments/assets/940f4529-6829-4767-bd2a-d611259ac167)

[...]

Issue #1470, opened 2024-06-22
Need consistent and complete list of pseudoinstructions


ThinkOpenly:
The prior RISC-V ISA Specification (20191213) had a chapter 25, "RISC-V Assembly Programmer’s Handbook" which had a table listing many/most pseudoinstructions. This is no longer present in the most recent RISC-V ISA Specification (20240411).

In many ways, though, I'm actually leaning toward closer association between the pseudoinstructions and their base instructions. This is well done for "zext.w", which is right there in the section for its base instruction "add.uw" (section 28.5.1). Something like this, done more exhaustively, would be ideal.

nick-knight:
I don't understand. We moved the pseudoinstructions to the asm manual
https://github.com/riscv-non-isa/riscv-asm-manual/blob/main/riscv-asm.md#-a-listing-of-standard-risc-v-pseudoinstructions
I thought this was a step in the right direction. Are you wanting them moved back?

ThinkOpenly:
@wmat said (RVI Slack) "There is no assembly programmers manual that we maintain."

Is the link you provided something official? I do now see a reference to it in the bibliography of the RISC-V ISA specification.

One issue I see is that there are plenty of references to pseudoinstructions in the RISC-V ISA specification, but no obvious comprehensive reference for them. Most of the references are thrown in off-handedly. An outlier is for "zext.w", which is well formatted.

It could be that just a simple and obvious pointer to the RISC-V Assembly Programmer's Manual is needed, like "For more information and a listing of pseudoinstructions, see ...".

But, if pseudoinstructions are in the domain of that other document, then I'm not sure why _any_ pseudoinstructions are mentioned in the ISA? There's a fairly lengthy discussion about RDCYCLE, RDTIME, and RDINSTRET pseudoinstructions in the ISA, for example. The Assembly Programmer's Manual even says "Other assemblers might not support the same directives or pseudoinstructions", so arguably there shouldn't be _any_ pseudoinstructions in the ISA spec. Is support for the ones mentioned in the ISA spec mandatory?

[...]

Issue #1396, opened 2024-05-11
Don't write in prose


Timmmm:
One aspect of the spec that makes it quite hard to read is the prose-based style. A good (bad) example of this is CSR fields, e.g. `mstatus`. It's easy enough to find the field diagram:

![image](https://github.com/riscv/riscv-isa-manual/assets/376842/e9f0f197-b9d2-4ece-be18-9b9abbca7346)

But then suppose you want to know what `MPIE` does. You basically have to scan through the entire `mstatus` section and notice it in these big paragraphs:

![image](https://github.com/riscv/riscv-isa-manual/assets/376842/527faf9c-116d-476c-befe-7f0ed4ce90ca)

The style just makes quickly finding information quite tedious. The same is true for instructions:

![image](https://github.com/riscv/riscv-isa-manual/assets/376842/0d181f7e-591d-4db2-9373-628f6ca109ed)

This style just makes it difficult to jump to the information. Very small changes can greatly improve this:

![image](https://github.com/riscv/riscv-isa-manual/assets/376842/27cd843f-7c7a-456a-abe8-7de84ae6fae6)

If written in this style it also makes it easier to hyperlink to instructions, CSRs and CSR fields. For example CSR diagrams could have each field be a link to its definition.

aswaterman:
> Very small changes can greatly improve this

The title of your PR indicates you're on a stylistic bent that we don't necessarily agree with. 
 How about you make a PR to prove your point so we can evaluate it.

Timmmm:
The very small changes I was referring to are demonstrated in the screenshot.

Timmmm:
I'll make a PR for the base instructions as I suggested, and I'll try and do one for `mstatus` too since that is a bigger change and probably more debatable!

aswaterman:
I generally disagree that we want to change this aspect of the manual's style, but others can review the PR and see what they think.

Many of these changes will be superseded by the future addition of a one-instruction-per-page reference appendix.  Both styles serve a purpose; I don't think we want to remove the narrative version.

abrodkin:
I tend to agree with @Timmmm on that it's much harder to digest what particular instructions do.
See for example `SLTI` instruction description (https://github.com/riscv/riscv-isa-manual/blob/main/src/rv32.adoc?plain=1#L302):
[...]

Issue #1369, opened 2024-04-28
Plans for SAIL code integration?


camel-cdr:
A recent article raised the issue that the manual would benefit a lot from having the relevant SAIL code included with the definition of instructions: https://alastairreid.github.io/riscv-spec-issues/

This is already the case for part of the latest draft release, e.g. the bitmanip code and zicond have relevant SAIL definitions.

Notably the R32I/RV64I base instruction sets currently don't have SAIL definitions, I would've expected those to be added first, is this still planned in the future?

Additionally, I'd like to point out a naming discrepancy. Chapter 29 (bitmanip) titles the SAIL code "Operation", while Chapter 12 (zicond) titles it "SAIL code".

I personally much prefer the "SAIL code" title, because it's a lot easier to search for.

Are there plans to keep the SAIL code in the manual synchronised with the SAIL model?

sequencer:
I wonder how to define those undefined behavior in sail for the ISA Manual? SAIL is more like to an implementation to the RISC-V ISA, is there any plan to metaprogramming it to adapt some different uArch?

Timmmm:
We are now doing this for the CHERI spec. Search for `RETIRE_SUCCESS` here:

https://riscv.github.io/riscv-cheri/

It doesn't have the latest `sail-asciidoctor` so no syntax highlighting or hyperlinks to the actual code. The CHERI code is quite complex so some of the definitions are kind of useless without hyperlinks because they just call a function that does all of the work, but some are useful. I think that issue is less common in the base RISC-V code.

It works by `curl`ing the JSON file from GitHub Releases in the `Makefile`, which seems to work pretty well and avoids any extra dependencies.

I will make a PR to add the same system here.

wmat:
@pmundkur just tagging you on this issue so you're aware of it.

PR #1603, opened 2024-08-16
Added Instruction Description

AFOliveira:
Hello!

This PR introduces a continuation of the approach to instruction-per-page documentation for RISC-V that has been followed by newer extensions like "B", Vector crypto and Scalar crypto. We are actively trying to improve the process of documentation generation and believe that aligning the base ISA with these extensions would greatly increase the manual. We recognize that there might be things to improve and that we are fully open to address any feedback.

Key points:

- Enhanced instruction description structure:

  
  - Synopsis
  - Mnemonic
  - Encoding - with attributes
  - Description
  - Arguments
  - Operation (SAIL Code - retrieved from the SAIL repository)
  - Included in - (Extensions)
 - Automated documentation generation:
   As discussed in https://github.com/riscv/riscv-isa-manual/issues/1590 we are also trying to achieve a nice automation process based on existing repos and minimizing our impact as we possible can. Please take note that, in order to achieve the most of this layout, we refined some of the SAIL Code by hand.

 Current status:

- Proof of concept covering a subset of RISC-V instructions (RV32I)
- Example output available in the RISC-V instructions PDF [riscv-unprivileged.pdf](https://github.com/user-attachments/files/16637589/riscv-unprivileged.pdf)

There is a possibility here to impact the 200+ instructions in the manual that aren't still in this format if this approach is welcomed.

We welcome feedback on the approach, output quality, and potential modifications to existing repositories to support this effort.

Related issues: https://github.com/riscv/riscv-isa-manual/issues/1470, https://github.com/riscv/riscv-isa-manual/issues/1369, https://github.com/riscv/riscv-isa-manual/issues/1253, https://github.com/riscv/riscv-isa-manual/issues/1396, https://github.com/riscv/riscv-isa-manual/issues/1567

### DIFF
--- a/src/instrs/instr-encoding-rv_i.adoc
+++ b/src/instrs/instr-encoding-rv_i.adoc
@@ -1,0 +1,2139 @@
+=== Instructions
+
+[[instruction-add]]
+==== ADD
+
+Synopsis:: Add two values
+
+Mnemonic::
+add rd, rs1, rs2
+
+
+Encoding::
+[wavedrom, , svg]
+....
+{reg: [
+{bits: 7, name: 0x33, attr: ["OP"]},
+{bits: 5, name: 'rd'},
+{bits: 3, name: 0x0, attr:["funct3"]},
+{bits: 5, name: 'rs1'},
+{bits: 5, name: 'rs2'},
+{bits: 7, name: 0x0, attr:["funct7"]}
+]}
+....
+
+Description:: Adds rs1 and rs2, writing the result to rd. This basic arithmetic operation is fundamental in almost all computational tasks. It's used in address calculations, loop counters, array indexing, and general arithmetic operations in high-level languages. In more complex mathematical operations, ADD forms the basis for multi-word arithmetic and can be used to implement higher-precision calculations. It's also used in pointer arithmetic and in implementing certain algorithms like checksums.
+
+Arguments::
+[%autowidth]
+[%header,cols="4,2,2"]
+|===
+|Register |Direction |Definition
+|rd |output |Destination register
+|rs1 |input |Source register 1
+|rs2 |input |Source register 2
+|===
+
+Operation:: 
+
+[source,sail]
+--
+function clause execute (RISCV_ADD(rs2, rs1, rd, op)) = {
+  let rs1_val = X(rs1);
+  let rs2_val = X(rs2);
+  let result = rs1_val + rs2_val,
+  X(rd) = result;
+  RETIRE_SUCCESS
+}
+--
+
+Included in:: <<rv32,RV32I>>, <<rv64,RV64I>> 
+
+<<<
+
+[[instruction-addi]]
+==== ADDI
+
+Synopsis:: Add Immediate
+
+Mnemonic::
+addi rd, rs1, imm[11:0]
+
+Encoding::
+[wavedrom, , svg]
+....
+{reg: [
+{bits: 7, name: 0x13, attr: ["OP"]},
+{bits: 5, name: 'rd'},
+{bits: 3, name: 0x0, attr:["funct3"]},
+{bits: 5, name: 'rs1'},
+{bits: 12, name: 'imm[11:0]'}
+]}
+....
+
+Description:: Adds the 12-bit sign-extended immediate to rs1 and writes the result to rd. This is one of the most frequently used instructions in RISC-V, serving multiple purposes. It's used for basic arithmetic with small constants, for adjusting pointers (e.g., accessing structure fields or array elements), and for loading small signed constants into registers. In RV64, it sign-extends the result to 64 bits. ADDI with a zero immediate is used to move values between registers. It's also key in implementing higher-level language constructs like local variable allocation on the stack.
+
+Arguments::
+[%autowidth]
+[%header,cols="4,2,2"]
+|===
+|Register |Direction |Definition
+|rd |output |Destination register
+|rs1 |input |Source register 1
+|imm12 |input |12-bit immediate
+|===
+
+Operation:: 
+
+[source,sail]
+--
+function clause execute (RISCV_ADDI (imm, rs1, rd, op)) = {
+  let rs1_val = X(rs1);
+  let immext : xlenbits = sign_extend(imm);
+  let result = rs1_val + immext,
+  X(rd) = result;
+  RETIRE_SUCCESS
+}
+--
+
+Included in:: <<rv32,RV32I>>, <<rv64,RV64I>> 
+
+<<<
+
+[[instruction-and]]
+==== AND
+
+Synopsis:: AND two values
+
+Mnemonic::
+and rd, rs1, rs2
+
+Encoding::
+[wavedrom, , svg]
+....
+{reg: [
+{bits: 7, name: 0x33, attr: ["OP"]},
+{bits: 5, name: 'rd'},
+{bits: 3, name: 0x7, attr:["funct3"]},
+{bits: 5, name: 'rs1'},
+{bits: 5, name: 'rs2'},
+{bits: 7, name: 0x0, attr:["funct7"]}
+]}
+....
+
+Description:: Performs bitwise AND between rs1 and rs2, writing the result to rd. This instruction is fundamental in bitwise operations and is extensively used in various low-level programming tasks. It's crucial for masking operations, where specific bits need to be isolated or cleared while leaving others unchanged. AND is commonly used in flag manipulation, for extracting specific bitfields from larger words, and in implementing logical operations in boolean algebra. In embedded systems, it's often used for clearing specific bits in control registers. AND is also key in certain optimization techniques where only specific bits of a value are significant, and in implementing bitwise algorithms in cryptography and hash functions.
+
+Arguments::
+[%autowidth]
+[%header,cols="4,2,2"]
+|===
+|Register |Direction |Definition
+|rd |output |Destination register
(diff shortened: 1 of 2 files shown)


wmat:
Where is src/instrs/instr-encoding-rv_i.adoc coming from?  Was it generated somehow or manually created?  What version of the SAIL model is the SAIL code from?

AFOliveira:
src/instrs/instr-encoding-rv_i.adoc is mostly automatically generated. The only part with a manual revision/creation is SAIL code.
 
This is branch with the latest version https://github.com/foss-for-synopsys-dwc-arc-processors/riscv-opcodes/tree/SynopsisDescription (if anynone wants to try please stick to "python3 parse.py rv_i -asciidoc", since the branch is still a WiP).

SAIL model is being directly parsed (both automatically and by hand) from the upstream repository.

wmat:
Or perhaps even a separate document of all instuctions?

psherman42:
This is very nice, keeping each instruction on its own page, with writing brief enough to stay only within that one page.

It makes the ToC horrendously long, as well as breaking up continuity and flow of the rest of the Specification document.

This nice presentation seems much more fitting for another volume, may I suggest "RISC-V Assembly Lamguage Handbook", following the lead of Lance Leventhal's masterpiece, "6502 Assembly Language Handbook"? If you remember it, the layout and presentation is nearly the same as proppsed by this PR.

psherman42:
image

abrodkin:
> Or perhaps even a separate document of all instuctions?

That might very well be a separate document, but then what will be a distinction between what we already have in the ISA manual and that separate "index"? Note, that recently added extensions already have something like that. See, for example:
* 33.15. Vector Mask Instructions
* 34.4. Instructions (crypto instructions)
* 35.2. Extensions Overview (vector crypto)

I.e. will we remove these instruction descriptions from the ISA manual and keep there only general wording about each extension?
Just for the record, our idea was to improve the existing ISA manual with a uniform presentation of all instructions.

AFOliveira:
We're definitely open to suggestions. Of the options presented, I lean towards incorporating this content into an appendix section of the main document. This approach would keep all information easier to acess, rather than yet again creating another external resource. However, if there is room to maintain this as a separate document, I'm open to that solution as well. Our primary goal is to present this information in the best way possible.

abrodkin:
> This is very nice, keeping each instruction on its own page, with writing brief enough to stay only within that one page.
> 
> It makes the ToC horrendously long, as well as breaking up continuity and flow of the rest of the Specification document.
> 
> This nice presentation seems much more fitting for another volume, may I suggest "RISC-V Assembly Lamguage Programming", following the lead of Lance Leventhal's masterpiece, "6502 Assembly Language Programming"? If you remember it, the layout and presentation is nearly the same as proposed by this PR.

Well, there's already "RISC-V Assembly Programmer's Manual", see https://github.com/riscv-non-isa/riscv-asm-manual/blob/main/riscv-asm.md. And frankly speaking I'd rather consider merging it with the ISA manual as otherwise the ISA manual becomes a bit incomplete. In a sense that as a mere mortal software developer I would go check the ISA manual in a hope to get my questions answered fully, while in reality the information of my interest is spread among many documents:
* The RISC-V Instruction Set Manual
  * Privileged &
  * Non-privileged
* RISC-V Assembly Programmer's Manual
* RISC-V ELF psABI Document
* etc etc

I.e. I see value in modularity but we're also getting a problem of fragmentation, ambiguity and incompleteness (as it looks like from the first glance).

psherman42:
Not quite the same. Palmer's manual pointed out above is excellent coverage of best practices. It includes description of all relevant the directives in great detail, and lists all of the pseudoinstructions. Yet it doesn't include any mention of any of the real instruction, no summary and no detail.

That's where this _comprehensive instruction list_ fits in -- following Leventhal. A separate thing to be a nice little pocket reference (for example, see Rick Aster's "Professional SAS Programmer's Pocket Reference").

It could/would even be given a coveted place as a `riscv-isa-manual` as opposed to being contribution simply as `riscv-non-isa`.

AFOliveira:
While there are differences between the ASM manual and this list, I believe there's still value in consolidating the information rather than dispersing it. Given that the instructions are almost entirely automatically generated, maintenance would be minimal in the event of any changes in the ISA.

IMO incorporating the instructions as proposed in the PR would align with the existing standard, as @abrodkin pointed out, thus minimizing changes to the manual. We remain highly receptive to feedback and I'm not opposed to a comprehensive index of all instructions. However, I would suggest organizing that index by extensions as well. This approach would facilitate quick reference to see which instructions each extension adds, although we'd need to address the repeated instructions across extensions.

psherman42:
These are even more good reasons for maintaining that comprehensive instruction list in a separate place. Multiple index lists (one by alphabetical, another by extension) makes even greater length and adds to the "confusion of too many things to look at" when looking at the Unprivigedged Specificatiom to learn its big picture.

Being almost entirely automatically generated, so that maintenance would be minimal, is more of a non-human benefit; it adds little to human readability and overall comprehension. Thus, it's a minor point.

If there are too many lists, and too much detail, it is very hard to see the difference from one thing (instruction, in this case) to the next. That is how bugs and errors are introduced: flipping through to gather and group all desired instructions which might be part of scope of some desired revision, one might not catch them all--the missing ones would then be bugs later needing tombe fixed. A cyclical and painful process.

The goal of a manual or specification document is to say as much as possible, using comparative terms and relational methods and explanations, with the least space used.

anatoly-savchenkov:
In order to answer this question we need to agree on the use model for RISC-V ISA Manual. Is this a book for reading from first page to last page like a novel, or a Reference Manual for a hypothetical CPU which includes all latest versions of ratified RISC-V extensions, or a collection of adoc files grouped by extensions + their versions which IC manufactures can mix and match to create a Programmers' Reference Guide document for their respective chip.

Is the use model for RISC-V ISA Manual defined and documented somewhere?

jalobaba:
I'd like to ensure that this per-instruction documentation is just one view of a database of this information. I'd like to see other views including HTML and then also integration with other RISC-V ISA information like riscv-opcodes or some plan to integrate them together. 

On a technical note, I'd like to see a list of possible exceptions that can be generated by an instruction added. See MIPS example.

image

AFOliveira:
Hello, @jalobaba.
It is just one view of the information of the database for sure. 

> views including HTML

We are using the make method of this repo so it definitely also generates HTML. If you want to check out how that would look, the easiest option would be to open the manual in it's current state and navigate to "BF16" extension since that is the style we are going for.


> integration with other RISC-V ISA information like riscv-opcodes

Our source of information for all of this (except synopsis and description) is the riscv-opcodes repository. That's one of the main advantages of this whole approach since it almost only uses pre-existing information.


> possible exceptions

Since we are parsing from riscv-opcodes, we can't automatically get this information, so it would have to be a manual adittion. We are already seing possibilities for this since the [riscv-unified-db](github.com/riscv-software-src/riscv-unified-db) also has this information and it for sure is a great adittion.

ThinkOpenly:
I haven't heard a very good rationale for the separation of the additional content suggested in this PR (which currently doesn't exist anywhere) into only the 2nd of:
- instruction semantics and encoding (current ISA content)
- instruction assembly syntax reference

Any guide purporting to document how to use the instructions programmatically needs to have the semantics in the immediate vicinity (arguably more so than then encoding). If some separation is required (I don't see why), then there should really be a full reference of instructions with syntax and semantics and a separate encoding guide.

I have seen assertions that the assembly isn't part of the ISA model proper, so it needs to be documented separately, but that also argues for removing all of the assembly from the ISA, and if needed, replaced by an array of encodings. ;-)

AFOliveira:
> I haven't heard a very good rationale for the separation of the additional content suggested in this PR (which currently doesn't exist anywhere) into only the 2nd of:

I could even set this as optional to not cause too much problems on existing document. Basically, set is as a Makefile flag.
We could have normal manual without index of instructions + manual with index. Honestly, I think this would be much better, since we could have both outputs, without creating more scattered repositories, i.e. a step towards fixing the problem, instead of adding even one more.

AFOliveira:
> I think the ISA manual _should_ use a more "instruction per page" style everywhere (like the Vector chapter), but I don't think this PR is currently the right way because:
> 
> 1. It adds new descriptions of all the instructions but doesn't remove the existing ones, so now they are duplicated. That doesn't make any sense; the instructions should be integrated into the chapters not dumped at the end.
> 2. It doesn't group _any_ instructions, even very very similar ones like `blt` and `bge`. The Vector spec does and I think it makes sense to do that in this context.

This PR was an initial approach and totally generated from riscv-opcodes, so there was little information to work with. 

  1. I think we should aim for removal of duplicated information, but I can not remove ratified information, so it seems that the only viable way is to add the appendix and only work in removing duplication afterwards
  2. There is a much newer version that organizes this much better - as you may have seen. However, I think that alphatical order - inside each extension- has always been the preferred method for searching through a large appendix

AFOliveira:
@aswaterman Should we also close this given your comment [here ](https://github.com/riscv/riscv-isa-manual/issues/1567#issuecomment-3632034495)?

aswaterman:
Yeah.  I think we agree on where we want to end up, but I want UDB to do the heavy lifting here, rather than hand-writing a parallel view of the ISA.

AFOliveira:
@aswaterman I agree, this was just an initial experiment and I think it took discussion rolling, so the job of the PR is ready, I'll close this as I am sure a nicer "CI" approach will come and work much better.

PR #1604, opened 2024-08-17
Add missing values to mepc values table

HepoH3:
Apparently, values in the first column were not specified when several rows were added to the table.

Right now, this table looks like this:
<details>

firefox_1BVBnHAuDj

</details>

### DIFF
--- a/src/machine.adoc
+++ b/src/machine.adoc
@@ -1813,6 +1813,9 @@ _Designated for platform use_
 0 +
 0 +
 0 +
+0 +
+0 +
+0 +
 0
 |0 +
 1 +


aswaterman (approved):
Thanks.

@wmat Is there a way to express these tables with all of the cells in a row on the same row of text, rather than flattening out the columns?  We encounter this sort of bug all the time (the extension status table in the Preface is always getting messed up).

ved-rivos:
Keeping all cells in a row on the same row of text makes the source readable. The downside is that asciidoctor-pdf will render a line between rows. I could not find a way to tell asciidoc to not render a line separating the rows while keeping the cells on same row of text.

wmat:
It's possible to keep all cells in a row on the same row by using the [grid=cols] table attribute.  However, this will create a table with no lines between any of the rows.  So table 14 referenced above would display as 1 big column.  Is that what we're after?  I built the tables the way I did to try to match the LaTeX version as closely as possible.

aswaterman:
No, we don't want to change how the table is rendered just to make the code a little nicer.

I think this is yet another case where asciidoc isn't capable enough for our needs, and we might need to write a code generator to get a better tradeoff between expressibility and output quality.

wmat:
Normally, a row's content can be all in the same row of text in the code.  The exception is when you want line breaks in a row cell and have to use the + sign and a carriage return.  I agree that it makes for less than ideal source.

ved-rivos:
The most readable way to code the table in as follows.
```
|===
|HDR1  |HDR2   |HDR3  |HDR4   | HDR5
|R1C1  |R1C2   |R1C3  |R1C4   | R1C5
|R2C1  |R2C2   |R2C3  |R2C4   | R2C5
|==
```
However this will render a horizontal line for each row. To avoid the horizontal line one needs to treat the whole column as a span. However this makes the source unreadable and when there is variable length text makes it hard to find which word/sentence belongs to which cell.
```
|===
|HDR1  |HDR2   |HDR3  |HDR4   | HDR5
|R1C1 + 
  R2C1  
|R1C2 + 
  R2C2
|R1C3 + 
  R2C3
|R1C4 + 
  R2C4
|R1C5 + 
  R2C5
|==
```
If there was a way to code as in the first example where it looks like a table in source as well but have a way to tell asciidoc not to render the horizontal line then that would be great.

wmat:
In asciidoc there isn't.  The + sign requires a carriage return after it in order to format the cell content.

ved-rivos:
> In asciidoc there isn't. The + sign requires a carriage return after it in order to format the cell content.

Even if adoc did nott require a CR it would still become unreadable for a large table. 
It would look as follows:
`|R1C1 + R2C1  |R1C2 +  R2C2  |R1C3 +  R2C3 |R1C4 + R2C4 |R1C5 + R2C5
`

PR #1610, opened 2024-08-22
Clarify determination of yLPE on trap return

ved-rivos:


ved-rivos:
@aswaterman

aswaterman (commented):
This formulation seems clean to me.

I think deleting the redundant Smrnmi text is OK, but it would be better to add something like "See also <<sec:mnret>> for semantics added to the MNRET instruction when this extension is implemented."  Or something similar.  My point is, I want people who have implemented Smrnmi, but are now adding Zicfilp, to realize they have more work to do.

ved-rivos:
@aswaterman updated.

aswaterman (approved):

PR #1615, opened 2024-08-28
Add Smctr/Sscctr

bcstrongx:
Not yet ratified, going to TSC vote soon.


wmat (approved):
The two changes here LGTM, however, don't forget to add the include::smctr.adoc to riscv-privileged.adoc.

bcstrongx:
Good catch!  Added.

aswaterman (changes requested):
Is it really 1.0 given that it hasn't been ratified?  I think the options are to leave the PR unmerged until ratification, or to decrement the version number and add the Frozen warning.

Also, don't forget to update the extension list in the preface.

bcstrongx:
Definitely not 1.0 yet.  I assumed this PR was going to sit around until the extension is ratified, so v1.0 was appropriate.  Do we normally merge frozen but not ratified extensions into the ISA manual?  If so I can modify this.

aswaterman:
I think either way is OK.  Leaving this PR alone until ratification is the minimal-effort thing, so if you're OK with that, let's do that.

bcstrongx:
I believe this PR is ready to merge.  CTR is ratified (v1.0), and I pushed some late clarifications that were added to the CTR repo.  I removed any mention of CTR from the preface, since those entries seem to be made when a new release is released.  I don't know when that will be, so I assume @wmat will create that.  @aswaterman you had a change request related to this, please either approve or let me know if a change is still needed.

aswaterman (approved):
Thanks, @bcstrongx!

PR #1620, opened 2024-08-31
Fix immediate types font size for figure 2.3

HepoH3:
The font size is inconsistent between different types of immediates in the figure below (compare the b and j-types with the others):

<details>

<summary> current figure 2.3 </summary>

firefox_loANPOHf8l

</details>

This might have been done intentionally, but in the next figure, all types have the 'fontsize: 12' setting. Note that even with this size, the brackets of some fields still collide with the bit borders. I've checked at which size the collision doesn't occur, and that size is 10 (in this figure, all types use size 11, except for the j-type, which uses size 10):

<details>

<summary> figure 2.3 size experiments</summary>

firefox_sx1Cbkmblp

</details>

However, I don't think it's worth changing.

---

Additionally, I want to bring to your attention an issue with figure names. Most of them are missing now, even though they were present in the v20191213 version. However, there's a strange figure name in Figure 2.4:

<details>

<summary> current figure 2.4 </summary>

firefox_oUiqUJrxA8

</details>

It appears in an odd location with a strange number. I don't understand how these names are generated, so I can't fix it by myself.

### DIFF
--- a/src/images/wavedrom/immediate-variants.edn
+++ b/src/images/wavedrom/immediate-variants.edn
@@ -11,7 +11,7 @@
   {bits: 5,  name: 'rs1'},
   {bits: 5,  name: 'rs2'},
   {bits: 7,  name: 'funct7'}
-], config: {label: {right: 'R-Type'}}}
+], config: {fontsize: 12, label: {right: 'R-Type'}}}
 ....
 
 [wavedrom, ,svg]
@@ -22,7 +22,7 @@
   {bits: 3,  name: 'funct3'},
   {bits: 5,  name: 'rs1'},
   {bits: 12, name: 'imm[11:0]'},
-], config: {label: {right: 'I-Type'}}}
+], config: {fontsize: 12, label: {right: 'I-Type'}}}
 ....
 
 [wavedrom, ,svg]
@@ -34,7 +34,7 @@
   {bits: 5,  name: 'rs1'},
   {bits: 5,  name: 'rs2'},
   {bits: 7,  name: 'imm[11:5]'}
-], config: {label: {right: 'S-Type'}}}
+], config: {fontsize: 12, label: {right: 'S-Type'}}}
 ....
 
 [wavedrom, ,svg]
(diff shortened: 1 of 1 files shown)


aswaterman (changes requested):
IIRC, the issue is that the `[11]` in the B-type and J-type formats is too wide for the single column it occupies.  If we expand the font size for those, the brackets `[]` extend into the column separator, becoming invisible.  If you can figure out a way to prevent that problem, then I'd support increasing the font size.

aswaterman:
Regarding the final issue (the caption appears in a strange place), I think that's because there's a page break.  See the following screenshot from the PDF rendered in Apple Preview.

I agree it's odd.  Ideally we would not break these figures across pages.  Perhaps @wmat can find a way to do that.

Screenshot 2024-08-31 at 4 05 10 PM

HepoH3:
You're remembering correctly, I understand that. However, this PR is not suggesting an increase in the font size for the B/J rows, but rather a decrease in the font size for the other rows, similar to what has already been done for Figure 2.4:

https://github.com/riscv/riscv-isa-manual/blob/8b9dc50e562af27c9ab22ae0a2d3bf528c6bd41d/src/images/wavedrom/immediate.edn#L7-L23.

HepoH3:
> Regarding the final issue (the caption appears in a strange place), I think that's because there's a page break. See the following screenshot from the PDF rendered in Apple Preview.

I don't think page break cause this. I was referring to the v20240411 specification where I found this problem, but I think you checked some more recent versions of the specification, where this figure just happened to be positioned slightly higher. Last release renders in Firefox just like your screenshot. If we add something that pushes this figure back to page 24, the caption will still appear after the I-immediate:


firefox_86HO04ueYm-1


I can't check it in Apple Preview, but here is the [document itself](https://github.com/user-attachments/files/16828194/riscv-unprivileged.pdf)

Page breaking also doesn't explain why this figure is labeled as `Figure 1` (when it should be `Figure 2.4`). I can't see the captions for the nearest figures, so I think there's something wrong here.

aswaterman:
Yes, the page break is a coincidence.  Anyway, it would be good for @wmat to investigate why the caption shows up in an unexpected place.

aswaterman:
@wmat what do you think about the font-size proposal?  The effect is to make most of the text in the figure slightly more difficult to read, in exchange for consistency between the two figures.  I'll leave this to you.

wmat:
I will look i to this for you tomorrow. I know I played with font sizes at
some points to try to make text fir into boxes.  I’ll check on the missing
figure captions as well.

aswaterman:
@wmat please separate them into separate PRs.  For now, I'll close this PR.

PR #1631, opened 2024-09-06
Separate immediates into separate wavedrom files.

wmat:
This separates the immediates into separate wavedrom files which allows us to add the caption to the last immediate on the page (j-immediate) which puts it after the grouping of wavedroms correctly.


aswaterman (approved):
Looks good, thanks.  There was an unnecessary page break before the figure, so I deleted it as part of this PR.

PR #1643, opened 2024-09-19
Update description of mtval2 in CSR table

arichardson:
While it is always a guest physical address for the hypervisor
traps, when used by a double trap it holds an mcause value.

Change the description to "second trap value" to match the description
used in the section heading. While touching this line also sync the
description of the other CSRs in this part of the table as well as the
*tval registers for the other privilege levels.

### DIFF
--- a/src/priv-csrs.adoc
+++ b/src/priv-csrs.adoc
@@ -273,10 +273,10 @@ SRO
 `stval` +
 `sip` +
 `scountovf`
-|Scratch register for supervisor trap handlers. +
+|Supervisor scratch register. +
 Supervisor exception program counter. +
 Supervisor trap cause. +
-Supervisor bad address or instruction. +
+Supervisor trap value. +
 Supervisor interrupt pending. +
 Supervisor count overflow.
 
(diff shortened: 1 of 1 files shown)


src/priv-csrs.adoc
```diff
@@ -576,7 +576,7 @@ Machine trap cause. +
 Machine bad address or instruction. +
 Machine interrupt pending. +
 Machine trap instruction (transformed). +
-Machine bad guest physical address.
+Machine auxilliary trap information.
```
ved-rivos: "Machine second trap value" is better. 
Maybe could also fix up other labels to match the description in the Machine, Supervisor, and H extension ISA section  (e.g., bad address or instruction -> Machine Trap Value"

arichardson: Will update this to match the names used in the other section headings.

ved-rivos:
I think we can also fix up `stval` and `vstval` - they also have the label from their old names `sbadaddr`/`vsbadaddr`

aswaterman:
Agreed with both of @ved-rivos's comments. Ping me when it's time to merge.

ved-rivos (changes requested):
Could this PR be updated to also fix up `stval` and `vstval` labels.

ved-rivos (approved):

aswaterman (approved):

Issue #1578, opened 2024-08-01
Discrepancy in mhpmevent bits for Sscofpmf


atishp04:
The ratified sscofpmf extension spec[1][2] had bits 56:57 marked as reserved but the integrated priv spec mentioned only 6 bits.
However, the Smcntrpmf extension CSRs (mcyclecfg, minstretcfg) have 8 bits with bits 56:57 marked as reserved citing sscofpmf.

Is this an integration issue or the definition bits actually changed after ratification ?

[1] https://drive.google.com/file/d/1-3rp7Fw69aH26wbRkvwd4lBHangiVRMN/view
[2] https://github.com/riscvarchive/riscv-count-overflow/blob/main/content.adoc

@gfavor @bcstrongx

bcstrongx:
I'll leave this one for @gfavor, Smcntrpmf was simply emulating the Sscofpmf spec.  Either those bits should be reserved in both mhpmeventX and mcyclecfg/minstretcfg, or in neither.

bcstrongx:
See the linked PR that removes that sentence.

atishp04:
@gfavor : I could not find anything about "Bits 57:56 are Reserved in any case" in the latest integrated priv spec. Please let me know if I missed something ?

bcstrongx:
To take this a step further, I'm not sure how we managed to define 63:58 in Sscofpmf, or how we can ever define any bits in future standard extensions, given the current definition:

> The event selector CSRs, mhpmevent3-mhpmevent31, are 64-bit WARL registers that control which event causes the corresponding counter to increment. The meaning of these events is defined by the platform, but event 0 is defined to mean "no event."

I read that as the entire register is reserved for custom use.

ved-rivos:
> I read that as the entire register is reserved for custom use.

My interpretation is that when Sscofpmf is implemented, that is true. When Sscofpmf is implemented the 63:58 are provided by Sscofpmf. So user of `mphmevent*` has to discover availability of Sscofpmf and interpret 63:58 per Sscofpmf only if the extension is present.

atishp04:
Yes. That's what software is implemented with that assumption. But the ratified Sscofpmf marked 63:56 while the latest integrated spec 63:58. If the bits 57:56 are marked reserved somewhere else we should point that out in sscofpmf section. Otherwise, an implementation would use those bits for custom event encoding as well.

gfavor:
[...]

PR #1645, opened 2024-09-23
Adds back bits 57 & 56

wmat:
Adds back bits 57 & 56 to Sscofpmf and Smcntrpmf.  @gfavor and @bcstrongx , can you please make sure I added these bits back correctly?

### DIFF
--- a/src/smcntrpmf.adoc
+++ b/src/smcntrpmf.adoc
@@ -30,8 +30,6 @@ mcyclecfg and minstretcfg are 64-bit registers that configure privilege mode fil
 | UINH | If set, then counting of events in U-mode is inhibited
 | VSINH | If set, then counting of events in VS-mode is inhibited
 | VUINH | If set, then counting of events in VU-mode is inhibited
-| _WPRI_ | Reserved
-| _WPRI_ | Reserved
 |====
 
 When all __x__INH bits are zero, event counting is enabled in all modes.
--- a/src/smcntrpmf.adoc
+++ b/src/smcntrpmf.adoc
@@ -16,10 +16,10 @@ Smcntrpmf remedies these issues by introducing privilege mode filtering for the 
 
 mcyclecfg and minstretcfg are 64-bit registers that configure privilege mode filtering for the cycle and instret counters, respectively.  
 
-[cols="^1,^1,^1,^1,^1,^1,^5",stripes=even,options="header"]
+[cols="^1,^1,^1,^1,^1,^1,^5,^1",stripes=even,options="header"]
 |====
-|63 |62 |61 |60 |59 |58 |57:0
-|0 |MINH |SINH |UINH |VSINH |VUINH |_WPRI_ 
+|63 |62 |61 |60 |59 |58 |57:0 |56
+|0 |MINH |SINH |UINH |VSINH |VUINH |_WPRI_ |_WPRI_
 |====
 
 [cols="15%,85%",options="header"]
@@ -30,6 +30,8 @@ mcyclecfg and minstretcfg are 64-bit registers that configure privilege mode fil
 | UINH | If set, then counting of events in U-mode is inhibited
 | VSINH | If set, then counting of events in VS-mode is inhibited
 | VUINH | If set, then counting of events in VU-mode is inhibited
+| _WPRI_ | Reserved
+| _WPRI_ | Reserved
 |====
 
 When all __x__INH bits are zero, event counting is enabled in all modes.
--- a/src/smcntrpmf.adoc
+++ b/src/smcntrpmf.adoc
@@ -18,7 +18,7 @@ mcyclecfg and minstretcfg are 64-bit registers that configure privilege mode fil
 
 [cols="^1,^1,^1,^1,^1,^1,^5,^1",stripes=even,options="header"]
 |====
-|63 |62 |61 |60 |59 |58 |57:0 |56
+|63 |62 |61 |60 |59 |58 |57 |56
 |0 |MINH |SINH |UINH |VSINH |VUINH |_WPRI_ |_WPRI_
 |====
 
--- a/src/smcntrpmf.adoc
+++ b/src/smcntrpmf.adoc
@@ -16,10 +16,10 @@ Smcntrpmf remedies these issues by introducing privilege mode filtering for the 
 
 mcyclecfg and minstretcfg are 64-bit registers that configure privilege mode filtering for the cycle and instret counters, respectively.  
 
-[cols="^1,^1,^1,^1,^1,^1,^5,^1",stripes=even,options="header"]
+[cols="^1,^1,^1,^1,^1,^1,^5",stripes=even,options="header"]
 |====
-|63 |62 |61 |60 |59 |58 |57 |56
-|0 |MINH |SINH |UINH |VSINH |VUINH |_WPRI_ |_WPRI_
+|63 |62 |61 |60 |59 |58 |57:0
+|0 |MINH |SINH |UINH |VSINH |VUINH |_WPRI_
 |====
 
 [cols="15%,85%",options="header"]
--- a/src/sscofpmf.adoc
+++ b/src/sscofpmf.adoc
@@ -30,10 +30,10 @@ interrupt that is assigned to bit 13 in the mip/mie/sip/sie registers.
 
 The following bits are added to `mhpmevent`:
 
-[cols="^1,^1,^1,^1,^1,^1",stripes=even,options="header"]
+[cols="^1,^1,^1,^1,^1,^1,^1,^1",stripes=even,options="header"]
 |====
-|63 |62 |61 |60 |59 |58
-|OF |MINH |SINH |UINH |VSINH |VUINH
+|63 |62 |61 |60 |59 |58 |57 |56
+|OF |MINH |SINH |UINH |VSINH |VUINH |_WPRI_ |_WPRI_
 |====
 
 [cols="15%,85%",options="header"]
@@ -45,6 +45,8 @@ The following bits are added to `mhpmevent`:
 | UINH | If set, then counting of events in U-mode is inhibited
 | VSINH | If set, then counting of events in VS-mode is inhibited
 | VUINH | If set, then counting of events in VU-mode is inhibited
+| _WPRI_ | Reserved
+| _WPRI_ | Reserved
 |====
 
 For each ``x``INH bit, if the associated privilege mode is not implemented,


wmat:
Fixes https://github.com/riscv/riscv-isa-manual/issues/1578#top

src/smcntrpmf.adoc
```diff
@@ -16,10 +16,10 @@ Smcntrpmf remedies these issues by introducing privilege mode filtering for the
 
 mcyclecfg and minstretcfg are 64-bit registers that configure privilege mode filtering for the cycle and instret counters, respectively.  
 
-[cols="^1,^1,^1,^1,^1,^1,^5",stripes=even,options="header"]
+[cols="^1,^1,^1,^1,^1,^1,^5,^1",stripes=even,options="header"]
 |====
-|63 |62 |61 |60 |59 |58 |57:0
-|0 |MINH |SINH |UINH |VSINH |VUINH |_WPRI_ 
+|63 |62 |61 |60 |59 |58 |57:0 |56
```
gfavor, line 21: I believe "57:0" should now be just "57".

gfavor (approved):
Overall looks fine.  Just one simple "typo" correction to the Smcntrpmf change - for which I created a comment.  Presuming that will be fixed, I approve this set of changes.

src/smcntrpmf.adoc
wmat: Fixed.

ved-rivos: This should be:
```
|63 |62      |61      |60      |59        |58         |57:0 
|0  |MINH    |SINH    |UINH    |VSINH     |VUINH      |_WPRI_ 
```

wmat: So there should be no bit 56?  That contradicts Greg's comment to add them both back into the chapters.

gfavor: Re Bill last post ... Yes, there should be bits 57 and 56 - which are both WPRI and described as Reserved.  The "57:0" should just be corrected to "57" - which still leaves bit 56 as is in the PR.

gfavor: And it's a separate matter (per my post two minutes ago, whether these two specs say something explicitly about bit [55:0] or not.

wmat: The PR has been updated to fix 57:0 and leaves 56, as per @gfavor.

src/smcntrpmf.adoc
```diff
@@ -16,10 +16,10 @@ Smcntrpmf remedies these issues by introducing privilege mode filtering for the
 
 mcyclecfg and minstretcfg are 64-bit registers that configure privilege mode filtering for the cycle and instret counters, respectively.  
 
-[cols="^1,^1,^1,^1,^1,^1,^5",stripes=even,options="header"]
+[cols="^1,^1,^1,^1,^1,^1,^5,^1",stripes=even,options="header"]
 |====
-|63 |62 |61 |60 |59 |58 |57:0
-|0 |MINH |SINH |UINH |VSINH |VUINH |_WPRI_ 
+|63 |62 |61 |60 |59 |58 |57 |56
+|0 |MINH |SINH |UINH |VSINH |VUINH |_WPRI_ |_WPRI_
```
bcstrongx, line 22: If the resolution is just to make 57:56 WPRI, then I'm not sure this file needs any change.  57:0 were already WPRI, hence reserved.  @gfavor do you agree?  This PR leaves 55:0 undefined, which won't work.  Unlike mhpmeventX, those bits are not defined elsewhere.

src/sscofpmf.adoc
```diff
@@ -45,6 +45,8 @@ The following bits are added to `mhpmevent`:
 | UINH | If set, then counting of events in U-mode is inhibited
 | VSINH | If set, then counting of events in VS-mode is inhibited
 | VUINH | If set, then counting of events in VU-mode is inhibited
+| _WPRI_ | Reserved
+| _WPRI_ | Reserved
```
bcstrongx, line 49: In the first column, I would put the bit range/position rather than WPRI.  Could just be a single row:
`| 57:56 | Reserved`

bcstrongx (changes requested):

src/smcntrpmf.adoc
bcstrongx: Ah, I missed this discussion.  These counters count fixed events, so they don't use these cfg registers as event selectors.  57:0 should be WPRI, all those bits are reserved.

src/sscofpmf.adoc
gfavor: But per the last ARC meeting it was specifically resolved that these two bist should be Reserved and declared as WPRI (i.e not just Reserved with no further declaration to software and hardware).

src/smcntrpmf.adoc
gfavor: The Priv spec provides the default definition for all the bits of mhpmevent.  And note that the Priv spec does NOT specify WPRI for the bits of mhpmevent.  It specifies WARL.

The original ratified Sscofpmf extension said nothing about bits [55:0].  Which didn't leave them unspecified, but simply did not override the definition provided by the Priv spec.

src/smcntrpmf.adoc
bcstrongx: Wow, lots of comments rolled in while I was reviewing, so some duplication here.  First, following 57:0 with 56 (as shown above currently) doesn't make sense.  You could have `57 | 56 | 55:0`, but they'd all be WPRI, so I don't see the advantage to that.  

@gfavor, what you're proposing changes the ratified Smcntrpmf ISA.  These *cfg registers are not analogs for mhpmeventX, they only replicate the upper bits.  My understanding for this PR is only that 57:56 should be reserved in both Smcntrpmf and Sscofpmf, so let's stick with that.  Which requires no changes to Smcntrpmf.  If there are other changes to be considered, that should be a separate discussion.  But as stated in my other comments, I don't think any other change to Smcntrpmf makes sense.

gfavor: If Smcntrpmf is going to match Sscofpmf wrt to the top byte of mhpmevent bits (which has been my understanding), then that should include mirroring what is spec'ed for bits [57:56].

src/sscofpmf.adoc
bcstrongx: I'm not disagreeing with that.  The diagram above this lists 57:56 as WPRI, but this table below is giving definitions per bit/field.  So putting 57 and 56, rather than WPRI and WPRI, in the first column makes more sense.  Then it's clear that 57:56 are WPRI and Reserved.

src/smcntrpmf.adoc
gfavor: Per the ARC's final resolution, bits [57:56] for Sscofpmf are to be Reserved and WPRI.  Wasn't Smcntrpmf intending to match whatever Sscofpmf says for the high 8 bits that Sscofpmf spec'ed?

Or is Smcntrpmf specifically not striving to fully match what Sscofpmf says for the high bits?

bcstrongx: And it already does that.  m*cfg[57:0] is listed as WPRI, which matches the change we're making to Sscofpmf to make mhpmeventX[57:56] WPRI.

Note that it doesn't exactly match, since Zicntr counters do not support interrupt-based sampling.  So the OF bit (bit 63) is 0 in m*cfg.

src/sscofpmf.adoc
gfavor: That makes sense.  I'm agnostic to treating these two bist as two bits or as one bit field.  For clarity to Bill (as to what he should do), go ahead and make a comment on the PR to merge these two bits into one Reserved and WPRI bit field.

src/smcntrpmf.adoc
gfavor: I knwo you're still catching up with the posts, but the status quo for now is the the integrated Sscofpmf spec will match the original ratified spec in saying nothing about bits [55:0].  See my earlier post about the difference at issue (and that the Priv spec already provides a definition for all these low bits).

Any change (from the ratified spec) to how the lower bits are spec'ed would be a matter for the ARC to discuss and decide.  In the meantime I assume Sscntrpmf would mirror what Sscofpmf spec's?

ved-rivos: Unlike Sscofpmf which defined the top byte of mhpmevent*, Smcntrpmf defined a new CSRs - mcyclecfg, minstretcfg - and Smcntrpmf made bits 57:0 of that new CSR as WPRI. That should not be changed. So bits 57:0 in mcyclecfg, minstretcfg should remain WPRI.

gfavor: Understood and makes sense.  But just to note, Smcntrpmf is choosing to not mirror what Sscofpmf and the Priv spec says for the low bits of the mhpmevent registers.  But it IS choosing to at least mirror Sscofpmf for bits [63:56].

src/sscofpmf.adoc
gfavor: In which case, I suggest @bcstrongx provide a clear bottom-line net direction - as a comment to the PR - of what the final version of the Smcntrpmf text should look like.

src/smcntrpmf.adoc
ved-rivos: To summarize:
1. The table in Smcntrpmf does not change. The bits 57:0 remain WPRI
2. The statement in Smcntrpmf "Bits 57:56 are reserved for possible future modes." should be replaced by "Bits 57:56 are reserved."

src/smcntrpmf.adoc
bcstrongx: In a side email thread, it was agreed that smcntrpmf.adoc does not need any changes in this PR.  It already meets the requirement of keeping bits 57:56 WPRI and hence reserved.  So let's remove it from this changeset.

src/sscofpmf.adoc
bcstrongx: In our side email thread, we agreed that WPRI implies that the associated bits are reserved.  So I would probably remove rows 48 & 49 above, but don't feel strongly.  If it is preferred to more explicitly call out that these bits are reserved, I would replace lines 48 & 49 with `| 57:56 | Reserved`

wmat:
So everyone is in agreement and this is good to merge now?

bcstrongx:
I don't see any changes based on our comments?  There should only be 1 file changed (sscofpmf.adoc).

ved-rivos:
@bcstrongx for smcntrpmf, there is one change:
"Bits 57:56 are reserved for possible future modes." should be replaced by "Bits 57:56 are reserved."

src/smcntrpmf.adoc
atishp04: @ved-rivos : The original Smcntrpmf spec had 56:57 reserved following the ratified Sscofpmf spec. The decision to make 56:57 as WPRI happened due to the discussion initiated last month.
https://github.com/riscv/riscv-isa-manual/pull/1582

Since we are reverting that decision for Sscofpmf, shouldn't Smcntrpmf follow as well ?

atishp04:
> @bcstrongx for smcntrpmf, there is one change: "Bits 57:56 are reserved for possible future modes." should be replaced by "Bits 57:56 are reserved."

Ahh I see this now. I was referring to this in my earlier comment. Please ignore my previous comment if this is what going to be planned change for smcntrpmf.

bcstrongx:
@ved-rivos I removed that text a while back:
https://github.com/riscv/riscv-isa-manual/commit/340d4b1c168bcf9fdff45593093263a1273b3e0e

57:0 are all reserved in m{instret|cycle}cfg.

src/smcntrpmf.adoc
wmat: Changes to Smcntrpmf have been removed.

src/sscofpmf.adoc
wmat: Thanks, I have removed the changes from Smcntrpmf, with the exception of an extra space at the end of a line.

src/smcntrpmf.adoc
wmat: As per @bcstrongx comments, no changes are required to Smcntrpmf.

wmat:
So only Sscofpmf has changes now, with the exception of the deletion of an extra space at the end of a line in Smcntrpmf.  With approvals, this can be merged now?

ved-rivos:
> So only Sscofpmf has changes now, with the exception of the deletion of an extra space at the end of a line in Smcntrpmf. With approvals, this can be merged now?

LGTM!

PR #1647, opened 2024-09-25
Run root Docker as the current user

Timmmm:
Reapplication of 3539eff338352aba83dbae9e7b310d6a7d9a7d56 but this time we only add the `--user` argument for root Docker. Rootless Docker behaves the same as Podman and doesn't need the `--user` flag; in fact it breaks builds if you add it.

I also added a comment for the `:z` flag.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,9 +8,9 @@
 # SPDX-License-Identifier: CC-BY-SA-4.0
 #
 # Description:
-# 
-# This Makefile is designed to automate the process of building and packaging 
-# the documentation for RISC-V ISA Manuals. It supports multiple build targets 
+#
+# This Makefile is designed to automate the process of building and packaging
+# the documentation for RISC-V ISA Manuals. It supports multiple build targets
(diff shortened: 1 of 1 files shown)


Timmmm:
IMO a better solution than messing around with Docker/Podman is Apptainer, which is explicitly designed for this use case and takes care of exposing mounts etc. for you. Unfortunately people are unlikely to have it installed already so we're probably in an "everyone uses shitty Bash scripts because Bash is already there" sort of situation. :-/

aswaterman:
FWIW, this version works on my Mac.

wmat:
The Release new ISA When merging a PR still fails.  I recognize why this is happening.  We'll try to figure out how to fix this.

Timmmm:
It won't pass because it's running from my fork.

If you run it manually for `master` it should work though. I think the configuration needs to be changed a bit, I'm not a GitHub actions pro though....

wmat:
I fixed up the workflow and it's working as it should now.

PR #1657, opened 2024-09-27
Zicfiss clarifications: CBO and PM

ved-rivos:
@aswaterman


aswaterman (approved):
Thanks, this seems to match the intent.

PR #1675, opened 2024-10-06
Add Ssqosid.

pz9115:
Add Ssqosid privileged extension,ratified.

https://github.com/riscv/riscv-ssqosid


ved-rivos:
@pz9115 - this is already integrated into supervisor.adoc but we had commented it out to wait for Priv 1.13 to be released. Once Priv 1.13 is released (should happen in next couple of weeks) - we will uncomment the code. Thanks much for the PR but this can be closed.

pz9115:
Okay, thank you for mention it.

PR #1696, opened 2024-10-24
Improve table: Compressed instruction formats

cousteaulecommandant:
Table "Compressed 16-bit RVC instruction formats" header indicating the bits corresponding to each column only aligns properly with the first row (CR format), but doesn't reflect the boundaries of other rows (for example, funct3 or rs1'), making it hard to read.

This PR fixes that by distributing the bit indexes on the table header across the "actual" table columns (unmerged cells), so that it's clear which bits each field spans in every compressed instruction format, not just CR.

This is a really minor stylistic edit that you may want to squash into a future commit (or even just apply the change manually in a future edit).


aswaterman (approved):
Yes, this looks better, thanks.

PR #1705, opened 2024-11-01
Integrated Ssqosid extension

ved-rivos:


aswaterman (approved):
I quickly scanned the spec and didn't see other cases where we included in the mainline text the priv version number that added a feature.  I'll strip that from this text and instead add a changelog entry.

I'll also fix a typo (there is a stray `55`).

aswaterman:
Manually merged via 6114bf9a1312f6b9309b32a138ef04c0754eac09

ved-rivos:
Thanks!

ved-rivos:
@aswaterman - "The P1P13 bit in mstateen0 controls access to the hedelegh introduced by Privileged Specification Version
1.13." Was this the text you intended to add to change log? If so it is still in there.

aswaterman:
I created a new changelog with only one entry for P1P14.  (I didn't change the changelog for the ratified spec, which includes the P1P13 entry.)

PR #1707, opened 2024-11-02
Add Ssqosid to extensions table in preface

ved-rivos:
@aswaterman - the table lists extensions in two forms. Some are listed with " Extension" following the extension string and others with just the extension string. Is one preferred? If so I can update the remaining rows to the preferred form in this PR as welll.


aswaterman (approved):

aswaterman:
@ved-rivos Looks like adding "Extension" is what we were doing historically.  I'll merge this PR then change all the entries in this table (leaving the old tables alone).

PR #1710, opened 2024-11-04
Clarify note that compilers may need to change vtype for vmv<nr>r.v

lukel97:
In a similar vein to https://github.com/riscvarchive/riscv-v-spec/commit/856fe5bd1cb135c39258e6ca941bf234ae63e1b1, this note states that compilers don't need to "know or change vtype" for whole vector register moves.

There's some truth to this in that compilers can largely ignore SEW and LMUL, but ultimately they do depend on vtype and the current wording might be misleading. A compiler may in fact need to change vtype to clear vill, see: https://github.com/llvm/llvm-project/issues/114518


aswaterman (approved):
Works for me.

Issue #1671, opened 2024-10-04
Vector CSRs missing from CSR listing table


Timmmm:
The table [Currently allocated RISC-V unprivileged CSR addresses](https://riscv-specs.timhutt.co.uk/spec/20240411/priv-isa-asciidoc.html#ucsrnames) is missing all of the vector registers:

* `vstart`
* `vxsat`
* `vxrm`
* `vcsr`
* `vl`
* `vtype`
* `vlenb`

aswaterman:
Resolved by #1712 / #1714

PR #1712, opened 2024-11-06
Add the Missing Vector CSR in listing table

MrVyM:
Issue #1671 indicates that the listing table does not include the vector registers, so I have added them.

My source : https://github.com/riscvarchive/riscv-v-spec/blob/master/v-spec.adoc#3-vector-extension-programmers-model


aswaterman (approved):
Thanks.  I'm manually merging this via #1714 so that I can make the formatting consistent.

PR #1716, opened 2024-11-07
Minor style fix for packh description

Peter-Herrmann:
The packh description starts with "And", likely copied over from a previous format where this description was shared with  pack.


wmat:
LGTM, good catch.

PR #1722, opened 2024-11-12
Explain `sscratch` is only swapped by software

workingjubilee:
Someone familiar with the interrupt handlers of other ISAs may incorrectly imagine that the current description means that RISCV cores will automatically perform the swap of `sscratch` with a user register.

This is not the case: it is a recommendation disguised as a description. Flesh it out a bit so it is less ambiguous.


aswaterman:
Yes, this is unfortunate use of the passive voice.  Thanks for the suggestion.

aswaterman (approved):

PR #1729, opened 2024-11-15
[CFI-zicfiss] updating pseudocode in ssamoswap unpriv 

kacouane:
as M mode behavior was unclear on priority vs g/page fault ...

### DIFF
--- a/src/unpriv-cfi.adoc
+++ b/src/unpriv-cfi.adoc
@@ -767,9 +767,6 @@ atomically loads a 32-bit data value from address of a shadow stack location
 in `rs1`, puts the loaded value into register `rd`, and stores the 32-bit value
 held in `rs2` to the original address in `rs1`. `SSAMOSWAP.D` (RV64 only) is
 similar to `SSAMOSWAP.W` but operates on 64-bit data values.
-Please see "Shadow Stack Memory Protection" section in Volume II of this manual
-for the exceptions raised by `SSAMOSWAP.W/D` if not already marked as
-illegal/virtual-exception.
 
 .`SSAMOSWAP.W` for RV32 and `SSAMOSWAP.D` (RV64 only) operation
 [listing]
@@ -794,9 +791,6 @@ For RV64, `SSAMOSWAP.W`, if granted access according to page protections rules,
 atomically loads a 32-bit data value from address of a shadow stack location
 in `rs1`, sign-extends the loaded value and puts it in `rd`, and stores the
 lower 32 bits of the value held in `rs2` to the original address in `rs1`.
-Please see "Shadow Stack Memory Protection" section in Volume II of this manual
-for the exceptions raised by `SSAMOSWAP.W` if not already marked as
-illegal/virtual-exception.
 
 .`SSAMOSWAP.W` for RV64
 [listing]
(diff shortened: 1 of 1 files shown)


ved-rivos:
I don't think this edit is required. The memory accesses are performed at the effective privilege level. When the effective privilege level is M, virtual memory system is not active. When effective privilege level is not M, then virtual memory system is always active. Given these two are mutually exclusive, there is no prioritization required.

aswaterman:
I defer to Ved on this matter.

kacouane:
Yes, I agree no virtual mem in machine mode. I meant there might be confusion since we have a sentence defining the M mode behavior in the privilege spec, and it is not reflected here in the unpriv side pseudocode (which explains other privileges). While the (guest) page fault rules descriptions leave in privilege spec ...

Also, I felt we could remind people that some cases might trigger a fault before the memory swap is performed.

So, I added the memory checks in the pseudocode, but I can only add the M mode case if you think it is better.

ved-rivos:
We have a sentence about the effective privilege mode of the access. This instruction can be used in M mode but it cannot be used with effective privilege mode of M. 

This code may be misleading due to that. If `ssamoswap` was invoked with effective privilege mode of S and at privilege mode M, then this pseudocode would incorrectly imply that if the memory access with effective privilege mode of S did not cause a page/guest-page/access-fault at lines 2 then it would cause an access-fault at line 4 due to privilege being M.

```
1.      if page-fault / guest-page-fault / access-fault
2.           raise accordingly
3.       else if privilege_mode == M
4.           raise store/amo access-fault exception
5.       else
```
How about we not add pseudo-code, but we instead add a cross reference to the Priv spec.
> Please see "Shadow Stack Memory Protection" section in Volume II of this manual for the exceptions raised by `SSAMOSWAP.W/D`

kacouane:
The cross-ref seems like a good idea. I will update the PR accordingly

ved-rivos:
Instead of repeating the cross reference twice, could we place it as a single paragraph i.e. between the paragraph beginning with "Just as for AMOs in the A extension..." and the non-normative note that follows it.

kacouane:
Done

src/unpriv-cfi.adoc
```diff
@@ -762,11 +762,11 @@ back_cfi_not_active:
 ], config:{lanes: 1, hspace:1024}}
 ....
 
-For RV32, `SSAMOSWAP.W` atomically loads a 32-bit data value from address of a
-shadow stack location in `rs1`, puts the loaded value into register `rd`, and
-stores the 32-bit value held in `rs2` to the original address in `rs1`.
-`SSAMOSWAP.D` (RV64 only) is similar to `SSAMOSWAP.W` but operates on 64-bit
-data values.
+For RV32, `SSAMOSWAP.W`, if granted access according to page protections rules,
```
ved-rivos, line 765: I am not sure the addition of "if granted access according to page protection rules" is required. All memory accesses are subject to page protection rules, PMP/PMA rules, Svukte rule, and other conditions such as those that cause hardware error exception (attempts to consume poison). Since this principle that memory access occurs only after everything that prevents a memory access is checked, this addition may not be required.

kacouane: You are right; it might not be the right term. I was trying to highlight that the pseudocode below only provides information for some exceptions. But it is not exhaustive, and these SSAMO operations might fail prior to the X(rd) = mem[X(rs1] ...

Maybe the best solution is to split the "illegal/virtual exception part" and the swap /sign-swap pseudocode what do you think?

ved-rivos: Sorry, I somehow dropped this thread. My suggestion is to just drop this addition since its stating the obvious which i.e. memory accesses need to be granted access to be allowed.

PR #1733, opened 2024-11-21
[RVWMO Explanatory] Fix the figure used to illustrate atomicity axiom rule

diantaowang:
In the latest unprivileged version, there is a bug in Figure 4 of A.3.3. Atomicity axiom.

The wrong picture is as follows:

The right picture should be:

Fix it.


aswaterman (approved):
Thanks for the fix!

PR #1742, opened 2024-11-26
Clarify the shadow stack enforcement against superpage missaligned check

kacouane:
When the zicfiss extension is present and we have both a superpage misaligned and a shadow stack enforcement failure we want to report the misalignment as it is a recoverable error and  misaligned superpages logically happen before permission checks (there is no need to know the access type of the transaction check logically happens in the PTW)

> [!NOTE]
> even if we swap the former point 6 and point 5 this change doesn't affect previous systems as the reported errors were both page-faults

### DIFF
--- a/src/supervisor.adoc
+++ b/src/supervisor.adoc
@@ -36,7 +36,7 @@ supervisor-level CSR descriptions.
 ====
 
 [[sstatus]]
-==== Supervisor Status (`sstatus`) Register 
+==== Supervisor Status (`sstatus`) Register
 
 The `sstatus` register is an SXLEN-bit read/write register formatted as
 shown in <<sstatusreg-rv32>> when SXLEN=32
@@ -297,7 +297,7 @@ Additionally, the implementation of an SSE protocol can be considered as an
 optional measure to aid in the recovery from such critical errors.
 ====
 
-==== Supervisor Trap Vector Base Address (`stvec`)  Register 
+==== Supervisor Trap Vector Base Address (`stvec`)  Register
 
 The `stvec` register is an SXLEN-bit read/write register that holds trap
 vector configuration, consisting of a vector base address (BASE) and a
@@ -319,7 +319,7 @@ field.
 |Value |Name |Description
 |0 +
 1 +
-&#8805;2 
+&#8805;2
 |Direct +
 Vectored
 |All exceptions set `pc` to BASE. +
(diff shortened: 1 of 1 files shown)


src/supervisor.adoc
```diff
@@ -1532,14 +1532,24 @@ A virtual address _va_ is translated into a physical address _pa_ as follows:
...
 __a__=__pte__.__ppn__×PAGESIZE and go to step 2.
-. A leaf PTE has been found. Determine if the requested memory access is
-allowed by the _pte_._r_, _pte_._w_, _pte_._x_, and _pte_._u_ bits, given the current privilege mode and the value of the SUM and MXR fields of the `mstatus` register. If not, stop and raise a page-fault exception corresponding to the original access type.
-. If _i>0_ and _pte_._ppn_[__i__-1:0] ≠ 0, this is a misaligned superpage; stop and raise a page-fault exception corresponding to the original access type.
+
+. A leaf PTE has been reached. If _i>0_ and _pte_._ppn_[__i__-1:0] ≠ 0, this is a misaligned superpage; stop and raise a page-fault exception corresponding to the original access type.
+
+. Determine if the requested memory access is allowed by the _pte_._u_ bit, given the current privilege mode and the value of the SUM and MXR fields of the *mstatus* register. If not, stop and raise a page-fault exception corresponding to the original access type.
+
+. Determine if the requested memory access is allowed by the _pte_._r_, _pte_._w_, and _pte_._x_ bits, given Shadow Stack Memory Protection. If not, stop and raise an access exception.
```
ved-rivos, line 1547:
```suggestion
. Determine if the requested memory access is allowed by the _pte_._r_, _pte_._w_, and _pte_._x_ bits, given the Shadow Stack Memory Protection rules. If not, stop and raise an access exception corresponding to the original access type.
```

ingallsj: @ved-rivos can you please reconcile your suggestion with this from the "Shadow Stack Memory Protection" section?

> If a shadow stack (SS) instruction raises an access-fault, page-fault, or guest-page-fault exception that is supposed to indicate the original instruction type (load or store/AMO), then the reported exception cause is respectively a store/AMO access fault (code 7), a store/AMO page fault (code 15), or a store/AMO guest-page fault (code 23). For shadow stack instructions, the reported instruction type is always as though it were a store or AMO, even for instructions SSPOPCHK and C.SSPOPCHK that only read from memory and do not write to it.

ved-rivos: The specification states:
> The access type is classified as a store/AMO in the event of an access-fault, page-fault, or guest-page
fault exception triggered by shadow stack instructions.

ved-rivos: But if that nuance is confusing then I am good with leaving it as "access fault"

kacouane:
I updated the PR with suggestions but I also changed "access exception" to "access-fault exception"

aswaterman (approved):
LGTM, too.

kacouane:
@nibrunieAtSi5

PR #1770, opened 2024-12-12
Fix wording and formatting of x0 special cases

pfusik:
Restore semicolons removed accidentally in aaad788afe0019ebb5c63d3c3193e9d951c3f3e0

### DIFF
--- a/src/c-st-ext.adoc
+++ b/src/c-st-ext.adoc
@@ -306,19 +306,19 @@ These instructions use the CI format.
 C.LWSP loads a 32-bit value from memory into register _rd_. It computes
 an effective address by adding the _zero_-extended offset, scaled by 4,
 to the stack pointer, `x2`. It expands to `lw rd, offset(x2)`. C.LWSP is
-only valid when _rd_&#x2260;x0; the code points with _rd_=x0 are reserved.
+valid only when _rd_&#x2260;x0; the code points with _rd_=x0 are reserved.
 
 C.LDSP is an RV64C/RV128C-only instruction that loads a 64-bit value
 from memory into register _rd_. It computes its effective address by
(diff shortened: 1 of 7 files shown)


wmat:
Thanks for this, those semicolons somehow went missing.  Can I suggest that you also change the "only valid" to "valid only", as it's grammatically an improvement.

pfusik:
How about normalizing these "not equal" signs? Some are UTF-8, some are HTML entities and some are latexmath. I think UTF-8 would be best.

wmat:
I think UTF-8 would work.  Perhaps add attributes to the header docs for UTF-8 replacements, such as:

:le: &#8804;
:ge: &#8805;
:lt: &#60;
:gt: &#62;

Then replace in the asciidoc files with attribute references such as {ge}

pfusik:
What do you mean by "header docs" ? Why use attribute references if we can (and already do) directly use Unicode characters? Especially for ASCII characters such as `<` and `>`.

wmat:
Asciidoc requires header docs, for the ISA manuals, they are riscv-unprivileged.adoc and riscv-privileged.adoc.  At the beginning of each is a set of attributes that asciidoctor requires to build the full documents.  Adding attributes for unicode would but down on a few keystrokes for extension authors.

pfusik:
I see, there's no `≠` on keyboards. But it was already extensively used and WYSIWYG. It's easier to copy the whole phrase than to re-type it. `latexmath` ended up in HTML as SVG - not good for screen readers.

pfusik:
When merging, please select my Samsung email.

aswaterman (changes requested):
The title of this PR does not reflect what it actually does.  It makes several wording changes.

aswaterman (approved):
Much improved, thank you.

PR #1777, opened 2024-12-16
Check Mergeable by Label

rpsene:
Check if can be merged based on the tag set. In this case, PRs with Pending Ratification cannot be merged.


wmat:
This looks good, thanks @rpsene

PR #1784, opened 2024-12-28
Fix formatting in "RISC-V base unprivileged register state" table

radimkrcmar:
I don't understand how those XLENs even got there.

### DIFF
--- a/src/rv32.adoc
+++ b/src/rv32.adoc
@@ -85,10 +85,7 @@ holds the address of the current instruction.
 3+^| [.small]#x29#
 3+^| [.small]#x30#
 3+^| [.small]#x31#
-3+^| [.small]#XLEN#
-| [.small]#XLEN-1#| >| [.small]#0#
-3+^|  [.small]#pc#
-3+^| [.small]#XLEN#
+3+^| [.small]#pc#
 |===
 [NOTE]
 ====


radimkrcmar:
> The table was formatted to match the table from the 2019 spec, see screenshot:

Having XLEN there makes sense, thanks.  The current table is even worse than I thought as it doesn't put the header outside either.

I applied bytefield to make it look just like other register definitions, but that doesn't break on page boundary and would move the whole table the next page, leaving awkward white space:
screenshot-2024-12-30T16:49:26+01:00

Is this solution acceptable?

wmat:
So from a content perspective, is the original table from the 2019 spec incorrect?

radimkrcmar:
> So from a content perspective, is the original table from the 2019 spec incorrect?

No, that one is good and I will add a separate table for the pc register if that is preferred. (Seemed unnecessary.)
The problem is that the current one looks different from 2019.  Table header is inconsistent with other definitions and what happens around the pc register is just confusing:
screenshot-2024-12-31T08:58:17+01:00

aswaterman:
Right, the bug is that two of the XLENs are now cells within the table, whereas in the original they were text _below_ the table.

@wmat I’m OK with any solution that doesn’t make it look like the XLENs are rows in the table. If that requires deleting these two XLENs (but of course retaining the XLEN-1 in the top-left corner), so be it. I’ll leave this to you.

wmat:
@radimkrcmar if you can do as Andrew suggests, that'd be great.

radimkrcmar:
> @radimkrcmar if you can do as Andrew suggests, that'd be great.

The patch does that: remove two lines before the pc register an one line after it, leaving the inconsistent header on top.
I would be happier with the solution presented in https://github.com/riscv/riscv-isa-manual/pull/1784#issuecomment-2565679368, so I'll respin this pull request with it if you don't mind the white space.
(Floating point registers need the same change, so extra PR will land either way.)

PR #1793, opened 2025-01-09
Clarify misa.M behaviour with Zmmul

### LINKED ISSUE #1711: Specify misa.M behaviour with Zmmul
From [this comment](https://github.com/riscv/riscv-isa-manual/pull/1121#issuecomment-1721892829):

> misa spec requires that if M is set, then M is available. If only Zmmul is present misa.M should be clear. If misa.M is writable and M is cleared, the specs say that the hart behaves as if M not present, where behavior includes the opcodes being "reserved". Whether Zmmul is still usable when misa.M is clear is implementation-defined; misa allows the behaviors to be "reserved" in this case - so Zmmul if still available, it acts as a standard extension. This I believe is the only way to have a consistent view of misa behavior matching the ratified specifications. 

To avoid spec-by-github-issue please can we add that information to the actual spec? It is not obvious.

I think it is correct though since it matches the B extension and its sub-extensions. The C extension/sub-extensions work differently.

Ideally a whole section would be added to `misa` explaining how the sub-extensions are supposed to work for M, B, C, etc.

Timmmm:
This clarifies how the combination of misa.M and Zmmul affect the requirements on implementing multiply and division.

Fixes #1711

### DIFF
--- a/src/machine.adoc
+++ b/src/machine.adoc
@@ -192,9 +192,16 @@ supervisor modes respectively.
 
 The "X" bit will be set if there are any non-standard extensions.
 
-When "B" bit is 1, the implementation supports the instructions provided by the
-Zba, Zbb, and Zbs extensions. When "B" bit is 0, it indicates that the
+When the "B" bit is 1, the implementation supports the instructions provided by the
+Zba, Zbb, and Zbs extensions. When the "B" bit is 0, it indicates that the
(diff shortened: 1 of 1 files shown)


aswaterman (approved):

aswaterman:
Thanks for following up on this, @Timmmm.

PR #1799, opened 2025-01-13
Fix parentheses in mstatus SD description

jordancarlin:
Repeated parentheses were not escaped, so they were not displayed correctly.

### DIFF
--- a/src/machine.adoc
+++ b/src/machine.adoc
@@ -1135,7 +1135,7 @@ additional microarchitectural bits might be maintained in the extension
 to further reduce context save and restore overhead.
 
 The SD bit is read-only and is set when either the FS, VS, or XS bits
-encode a Dirty state (i.e., SD=\((FS==11) OR (XS==11) OR (VS==11))). This
+encode a Dirty state (i.e., `SD=(FS==0b11 OR XS==0b11 OR VS==0b11)`). This
 allows privileged code to quickly determine when no additional context
 save is required beyond the integer register set and `pc`.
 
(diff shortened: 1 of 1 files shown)


src/machine.adoc
```diff
@@ -1135,7 +1135,7 @@ additional microarchitectural bits might be maintained in the extension
 to further reduce context save and restore overhead.
 
 The SD bit is read-only and is set when either the FS, VS, or XS bits
-encode a Dirty state (i.e., SD=((FS==11) OR (XS==11) OR (VS==11))). This
+encode a Dirty state (i.e., SD=\((FS==11) OR (XS==11) OR (VS==11))). This
```
radimkrcmar, line 1138: The closing parenthesis has the same issue (unpatched spec):

> image
 
I find three issues with that line aside from the missing parentheses:
* Pseudocode incorrectly uses `11` instead of `3` or `11b` or `0b11`, spec uses both binary formats and I don't know which one is correct :(,
* parentheses around `==` don't add much to the clarity of the snippet (I'd remove them),
* snippet should be formatted as inline code.

Removing the pseudocode is also an option if we improve the English description to clearly state that SD is set if and only if any of FS/XS/VS is dirty.

jordancarlin: I switched the `11`s to be `0b11` to make it clear that they are binary representations and formatted the section as inline code. I also removed the extra parentheses, which makes the escape character unnecessary.

aswaterman (approved):
Works for me.

aswaterman:
Thanks for the fix!

PR #1801, opened 2025-01-14
Added submodule check

bgrayson-mips:
This patch adds a check that the cloning operation was done properly, with the submodule. If that check fails, it will print a message and then do the proper git submodule command so that the docs-resources/ submodule and any other submodules are properly checked out. If desired, the git command could be removed and it could just print the command to invoke -- I'm open to feedback on that.

The phony target `submodule-check` is only added as a dependency on the build, build-container, and build-no-container targets, which should suffice for most oops-forgot-submodule mistakes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,9 +91,17 @@ REQUIRES := --require=asciidoctor-bibtex \
             --require=asciidoctor-lists \
             --require=asciidoctor-mathematical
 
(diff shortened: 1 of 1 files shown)


aswaterman (approved):
LGTM, thanks @bgrayson-mips.

PR #1806, opened 2025-01-16
Fix typos

trdthg:
simple fix, handle some cumulative typos


wmat:
Thanks for the fixes.

PR #1808, opened 2025-01-16
Add jvt CSR to CSR table

trdthg:
- [riscv-unprivileged.html#Zcmt](https://riscv-specs.timhutt.co.uk/spec/riscv-isa-release-f3f8a93-2024-11-28/riscv-unprivileged.html#Zcmt)
- [jvt CSR(0x0017)](https://riscv-specs.timhutt.co.uk/spec/riscv-isa-release-f3f8a93-2024-11-28/riscv-unprivileged.html#csrs-jvt)

### DIFF
--- a/src/priv-csrs.adoc
+++ b/src/priv-csrs.adoc
@@ -196,6 +196,12 @@ Vector register length in bytes.
 |`seed` +
 |Seed for cryptographic random bit generators. +
 
+4+^|Unprivileged Zcmt Extension CSR
+|`0x017` +
+|URW +
+|`jvt` +
+|Table jump base vector and control register. +
+
 4+^|Unprivileged Counter/Timers
 
 |`0xC00` +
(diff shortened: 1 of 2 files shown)


src/priv-csrs.adoc
```diff
@@ -190,6 +190,12 @@ Vector register length in bytes.
 |`ssp` +
 |Shadow Stack Pointer. +
 
+4+^|Unprivileged Zcmt Extension CSR
+|`0x0017` +
```
aswaterman, line 194: should only be a 3-digit number.

trdthg: {1495C447-D086-4C80-91F2-4A61CE793D44}

Oh, I saw it, so this is a typo too?

aswaterman: Yes, please fix that, too.  It isn't technically an error; it's just inconsistent formatting.  CSR addresses are 12-bit numbers, so we represent them with three hex digits by convention.

trdthg: done

aswaterman (approved):
Thank you!

PR #1821, opened 2025-01-23
Undo special formatting of compressed integer reg-reg expansions

pfusik:
No reason for these six instructions to be formatted differently.


wmat:
Good catch, thanks.

Issue #1814, opened 2025-01-17
What does "inlined" mean?


kersten1:
In the 14.1. Specifying Ordering of Atomic Instructions section of the unpriv, there is a para that says this: 

"LR/SC can be used to construct lock-free data structures. An example using LR/SC to implement a compare-and-swap function is shown in Example 2. If inlined, compare-and-swap functionality need only take four instructions."

Inlined isn't a word that I can find. Bill T pointed out that it is probably in reference to the inline assembler. 

Can someone confirm that? 

If so, Does this rewrite make sense?

"LR/SC can be used to construct lock-free data structures. The previous example shows LR/SC implementing a compare-and-swap function. If this example is embedded inline, the compare-and-swap functionality requires only four instructions."

ved-rivos:
"Inlined" refers to code that has been directly inserted into a sequence of instructions rather than being referenced as a separate function. It’s a well-understood term in programming and compiler design, often associated with optimizing performance by avoiding the overhead of function calls. However, if clarity is a concern, "embedded inline" might be a better choice for broader audiences.

aswaterman:
IMO, “inline” is a sufficiently common verb in the programming languages/compilers context that this sentence is fine as-is.

kersten1:
@aswaterman @ved-rivos 

"inlined" is not a word. I can infer meaning from it based on the common usage of "inline" but it doesn't change the fact that "inlined" is not a word! Additionally, the sentence has an unclear antecedent. You must infer that it is the code from the example that is to appear inline. Plus do all compare-and-swap functionalities that are embedded inline require only four instructions or just this example? 

Rather than make the reader infer meaning, we can make this section more clear with a few edits.

We can, however, just remove the "d" from "inlined". It is a less effective fix, but one I can live with.

aswaterman:
It's common practice for technical documents to use terminology that's nonstandard in English but is standard within the technical domain.  In this community, using "inlined" as a past-tense verb or as an adjective is without a doubt accepted: https://scholar.google.com/scholar?hl=en&as_sdt=0%2C5&q=inlined&btnG=

Of course we could rewrite this one sentence, but I'd like to first understand the broader strategy.  This document is chockablock with computer-science jargon.  Rewriting the document to be entirely free of technical jargon would seem to me to be a questionable expenditure that would ultimately produce something that's less readable.

allenjbaum:
I'm going to have to agree with Andrew here; the description of something
[...]

PR #1822, opened 2025-01-23
Move non-normative parts of "A" extensions (atomics) into NOTE block

radimkrcmar:
Most of the series moves non-normative text into notes, but the last patch also removes page breaks before examples as I found it hard to read otherwise.

It doesn't bring the ISA manual any closer to being ISO worthy, but I hope it'll at least be easier to understand until then.
I stumbled upon this issue thanks to https://github.com/riscv/riscv-isa-manual/issues/1814 and tried hard to avoid any controversial changes.

### DIFF
--- a/src/a-st-ext.adoc
+++ b/src/a-st-ext.adoc
@@ -227,8 +227,6 @@ instruction unless the _rl_ bit is also set. LR._rl_ and SC._aq_
 instructions are not guaranteed to provide any stronger ordering than
 those with both bits clear, but may result in lower performance.
 
-<<<
-
 [NOTE]
 ====
 [[cas]]
@@ -446,8 +444,6 @@ order the lock acquisition before the critical section, and the second
 AMO is marked _rl_ to order the critical section before the lock
 relinquishment.
 
-<<<
-
 [[critical]]
 [source,asm]
 .Sample code for mutual exclusion. `a0` contains the address of the lock.
@@ -462,7 +458,7 @@ relinquishment.
         # ...
         amoswap.w.rl x0, x0, (a0) # Release lock by storing 0.
 
-We recommend the use of the AMO Swap idiom shown above for both lock
+We recommend the use of the AMO Swap idiom shown in <<critical>> for both lock
 acquire and release to simplify the implementation of speculative lock
 elision. cite:[Rajwar:2001:SLE]
 ====
--- a/src/a-st-ext.adoc
+++ b/src/a-st-ext.adoc
@@ -229,6 +229,8 @@ those with both bits clear, but may result in lower performance.
 
 <<<
 
+[NOTE]
+====
 [[cas]]
 [source,asm]
 .Sample code for compare-and-swap function using LR/SC.
@@ -250,6 +252,7 @@ those with both bits clear, but may result in lower performance.
 LR/SC can be used to construct lock-free data structures. An example
 using LR/SC to implement a compare-and-swap function is shown in
 <<cas>>. If inlined, compare-and-swap functionality need only take four instructions.
+====
 
 [[sec:lrscseq]]
 === Eventual Success of Store-Conditional Instructions
@@ -434,6 +437,8 @@ both imply additional unnecessary ordering as compared to AMOs with the
 corresponding _aq_ or _rl_ bit set.
 ====
 
+[NOTE]
+====
 An example code sequence for a critical section guarded by a
 test-and-test-and-set spinlock is shown in
 Example <<critical>>. Note the first AMO is marked _aq_ to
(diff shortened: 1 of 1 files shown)


radimkrcmar:
Looking at this again, patches for examples can be squashed.  I'll do that for the next version if there are no objections.

wmat:
What is the reasoning for moving the blocks of text to be non-normative?  In the original spec this text was normative.

gfavor:
I would agree with @wmat.  Changing normative text in a ratified spec to non-normative text by default is not allowed since that represents a semantic arch spec change.

Now I recognize that some of these cases fall in the grey area of, for example, making suggestions and usage observations.  But even if such changes to non-normative text are ultimately deemed acceptable, this will require review by others and probably official review by the Arch Review Committee (as would be true with changes to other ratified specs that aren't simply typo fixes and clarifications).

So, as a first level of triage (to determine whether there's any reasonable chance of approval by the ARC so as to warrant the effort of requesting and conducting an official review), @aswaterman should comment on these changes before we go any further.

src/a-st-ext.adoc
```diff
@@ -221,14 +221,17 @@ same RISC-V hart. This can be appropriate when the LR/SC sequence is
 used to implement a parallel reduction operation.
 ====
 
+[NOTE]
+====
 Software should not set the _rl_ bit on an LR instruction unless the
 _aq_ bit is also set, nor should software set the _aq_ bit on an SC
 instruction unless the _rl_ bit is also set. LR._rl_ and SC._aq_
 instructions are not guaranteed to provide any stronger ordering than
 those with both bits clear, but may result in lower performance.
```
aswaterman, line 230: The second sentence in this paragraph must be normative.  The first sentence is a logical consequence of the second, and so _could_ be non-normative, but I don't think it improves the readability of the spec to make this change.  I would undo this change.

src/a-st-ext.adoc
```diff
@@ -250,6 +253,7 @@ those with both bits clear, but may result in lower performance.
 LR/SC can be used to construct lock-free data structures. An example
 using LR/SC to implement a compare-and-swap function is shown in
 <<cas>>. If inlined, compare-and-swap functionality need only take four instructions.
+====
```
aswaterman, line 256: Sample code should generally be non-normative.  Please post a screenshot of the effect of this change so we can make sure it looks OK.

src/a-st-ext.adoc
```diff
@@ -424,6 +430,7 @@ the AMO in this RISC-V hart. Setting both the _aq_ and the _rl_ bit on
 an AMO makes the sequence sequentially consistent, meaning that it
 cannot be reordered with earlier or later memory operations from the
 same hart.
+====
```
aswaterman, line 433: This paragraph is normative.  Please undo this change.

src/a-st-ext.adoc
```diff
@@ -434,15 +441,15 @@ both imply additional unnecessary ordering as compared to AMOs with the
 corresponding _aq_ or _rl_ bit set.
 ====
 
+[NOTE]
+====
```
aswaterman, line 445: Since this is also sample code, it should also be non-normative (but again, please post a screenshot of the effect).

aswaterman:
@gfavor Indeed, a subset of this PR is the right thing to do, but not the entirety.  Comments above.

gfavor:
@aswaterman Thanks.  Just as I expected.

src/a-st-ext.adoc
radimkrcmar: Before the change:
image
image


After the change:
image
image

radimkrcmar:
For the general motivation.  I heard that the goal is ISO certification, which implies a _lot_ of rewriting, so I find it wise to first properly mark the parts that don't need as much attention.

In this series I only aim to do changes that do not change any behavior.

src/a-st-ext.adoc
radimkrcmar: As per your recommendation, I have removed the change in version 2 of this PR.  Thanks for the review.

radimkrcmar:
I have updated the pull request.

v1->v2:
* Remove [1/5] and [4/5] (aswaterman)
* Squash [2/5] and [3/5]

This removes the controversial changes and the current result can be seen on screenshots that were provided during v1 review.

aswaterman (approved):
This version LGTM; thanks, @radimkrcmar.

src/a-st-ext.adoc
aswaterman: I believe you're right in the context of RVWMO, but this statement may have normative meaning in other contexts.  (Those contexts might not exist yet, of course...)

src/a-st-ext.adoc
aswaterman: I believe you're right that this paragraph could be deleted, but if that is, in fact, the case, it should actually be deleted, rather than made non-normative.  And it should be done in a separate PR.

PR #1834, opened 2025-01-29
vmadc/vmsbc instruction clarification

jacassidy:
There appears to be a typo that suggests the instruction writes back the result of the operation rather than the carry/borrow out


aswaterman (approved):
The intent was that, since these instructions are described as producing a carry or borrow, "result" refers to the carry or borrow, not the sum or difference.  But your version removes all doubt, so let's go with it.

PR #1847, opened 2025-02-10
Simplify the summary of sext/zext instructions

pfusik:


aswaterman (approved):
Yes, this is clearer.

PR #1869, opened 2025-02-19
Clarify that the NMIs referred to in the RNMI spec are RNMIs

aswaterman:
(As opposed to UNMIs.)

Now, all references to "NMI" in this chapter are fully qualified.

Fixes https://github.com/riscv/riscv-debug-spec/issues/1107#issuecomment-2667488240


aswaterman:
cc @pdonahue-ventana

pdonahue-ventana:
I think that you got them all.  Thanks.

PR #1879, opened 2025-03-03
Re-integrate Smepmp

mickflemm:
* Current chapter is the proposal document as-is, clean it up to better fit with the rest of the document.

* Add a link to the proposal document for more info.

* Add references from/to mseccfg definition, and to the PMP chapter. While there also add a reference to the Smmpm chapter on mseccfg definition.

* Rename MMWP to MMAP (Allowlist instead of Whitelist) and update the register field definition and the visual representation fo Smepmp.

* Clarify the reset state of mseccfg.MML/MMAP/RLB in the reset chapter.

* Add the editors of Smepmp (me and Joe Xie) to the list of contributors.


src/machine.adoc
```diff
@@ -2246,10 +2246,10 @@ include::images/wavedrom/mseccfg.edn[]
 The definitions of the SSEED and USEED fields are furnished by the
 entropy-source extension, Zkr.
 
-The definitions of the RLB, MMWP, and MML fields are furnished by the
-PMP-enhancement extension, Smepmp.
+The definitions of the RLB, MMAP, and MML fields are part of the Smepmp
+extension, as defined in <<smepmp>>.
```
Timmmm, line 2250: You can just hyperlink "the Smepmp extension".

```suggestion
The RLB, MMWP, and MML fields are defined in <<smepmp,the PMP-enhancement extension, Smepmp>>.
```

src/machine.adoc
```diff
@@ -2246,10 +2246,10 @@ include::images/wavedrom/mseccfg.edn[]
...
 
-The definition of the PMM field is furnished by the Smmpm extension.
+The definition of the PMM field is furnished by the Smmpm extension in <<Zpm>>.
```
Timmmm, line 2252:
```suggestion
The PMM field is defined in <<Zpm,the Smmpm extension>>.
```

src/machine.adoc
```diff
@@ -2524,9 +2524,9 @@ as described in <<misa>>.  For
 implementations with the "A" standard extension, there is no valid
 load reservation. The `pc` is set to an implementation-defined reset
 vector. The `mcause` register is set to a value indicating the cause of
-the reset. Writable PMP registers’ A and L fields are set to 0, unless
-the platform mandates a different reset value for some PMP registers’ A
-and L fields. If the hypervisor extension is implemented, the
+the reset. Writable PMP registers’ A and L fields, and `mseccfg`.MML/MMAP/RLB
+fields are set to 0, unless the platform mandates a pre-defined PMP ruleset.
```
Timmmm, line 2528: This probably needs a changelog entry?

src/smepmp.adoc
```diff
@@ -2,73 +2,36 @@
...
+The Smepmp extension enhances the xref:pmp[xrefstyle=basic] mechanism to improve Machine mode's security, by providing stronger memory isolation and privilege separation. It enables preventing the execution of code from memory regions not explicitly marked as executable by Machine mode, and restricts Machine mode's access to memory regions designated for Supervisor and User modes. These capabilities help to mitigate control-flow subversion attacks that exploit execution of unauthorized code with Machine mode privileges, and provide security guarantees consistent with the mechanism described in <<sum>> for memory access and execution prevention between Supervisor and User mode.
```
Timmmm, line 5:
```suggestion
The Smepmp extension enhances the xref:pmp[xrefstyle=basic] mechanism to improve Machine mode's security by providing stronger memory isolation and privilege separation. It allows preventing the execution of code from memory regions not explicitly marked as executable by Machine mode, and restricts Machine mode's access to memory regions designated for Supervisor and User modes. These capabilities help to mitigate control-flow subversion attacks that exploit execution of unauthorized code with Machine mode privileges, and provide security guarantees consistent with the mechanism described in <<sum>> for memory access and execution prevention between Supervisor and User mode.
```

Tbh I think the old text was a lot clearer. I would just change the second paragraph to

> To prevent this attack vector, two mechanisms known as Supervisor Memory Access Prevention (SMAP) and Supervisor Memory Execution Prevention (SMEP) can be used. The first one prevents the OS from accessing the memory of an unprivileged process unless a specific code path is followed, and the second one prevents the OS from executing the memory of an unprivileged process at all times. RISC-V includes support for SMAP, through the ``sstatus.SUM`` bit, and for SMEP by always denying execution of virtual memory pages marked with the U bit, with Supervisor mode (OS) privileges, as mandated on the Privilege Spec. This extension adds support for similar mechanisms for Machine mode.

src/smepmp.adoc
```diff
@@ -2,73 +2,36 @@
...
-
-Without being able to protect less-privileged modes from Machine mode, it is not possible to prevent the mentioned attack vector. This becomes even more important for RISC-V than on other architectures, since implementations are allowed where a hart only has Machine and User modes available, so the whole OS will run on Machine mode instead of the non-existent Supervisor mode. In such implementations the attack surface is greatly increased, and the same kind of attacks performed on Supervisor mode and mitigated through SMAP/SMEP, can be performed on Machine mode without any available mitigations. Even on implementations with Supervisor mode present attacks are still possible against the Firmware and/or the Secure Monitor running on Machine mode.
```
Timmmm, line 27: This is all useful text! I don't think it should be deleted; just reworded.

src/smepmp.adoc
```diff
@@ -2,73 +2,36 @@
...
-
-[[proposal]]
-=== Proposal
-
-. *Machine Security Configuration (mseccfg)* is a new RW Machine mode CSR, used for configuring various security mechanisms present on the hart, and only accessible to Machine mode. It is 64 bits wide, and is at address *0x747 on RV64* and *0x747 (low 32bits), 0x757 (high 32bits) on RV32*. All mseccfg fields defined on this proposal are WARL, and the remaining bits are reserved for future standard use and should always read zero. The reset value of mseccfg is implementation-specific, otherwise if backwards compatibility is a requirement it should reset to zero on hard reset.
```
Timmmm, line 32: > All mseccfg fields defined on this proposal are WARL

This seems to have been lost.

src/smepmp.adoc
```diff
@@ -77,95 +40,88 @@ A _Shared-Region_ rule is *enforced* on all modes, with restrictions depending o
...
+This mechanism not only provides a more strict access policy for Machine mode which otherwise has access everywhere unless restricted, it also allows for more efficient use of the available PMP registers, since there is no need to waste a rule for every region that should be blocked such as regions used by peripherals / processors on the system, or e.g. the BootROM after the handoff to firmware.
 ====
-+
-[NOTE]
+
+=== Rule Locking Bypass (RLB)
+
+The ``mseccfg.RLB`` bit provides a mechanism to temporarily modify Locked PMP rules:
+
+* When mseccfg.RLB=1, Locked PMP rules may be modified and locked PMP entries may be edited.
```
Timmmm, line 95:
```suggestion
* When mseccfg.RLB=1 *locked* PMP rules may be modified and *locked* PMP entries may be edited.
```

src/smepmp.adoc
```diff
@@ -2,73 +2,36 @@
...
 
+The ``mseccfg.MML`` bit fundamentally changes the interpretation of the ``pmpcfg.L`` bit defined in <<pmp-locking>>. It is a sticky bit: once set, it cannot be unset until xref:reset[xrefstyle=basic].
 
-.. Adding a rule with executable privileges that either is *M-mode-only* or a *locked* *Shared-Region* is not possible and such ``pmpcfg`` writes are ignored, leaving ``pmpcfg`` unchanged. This restriction can be temporarily lifted by setting ``mseccfg.RLB`` e.g. during the boot process.
+* When ``mseccfg.MML=0``, PMP operates according to the original specification, where the ``pmpcfg.L`` bit marks a rule as *Locked* and applicable in all privilege modes.
 
-.. Executing code with Machine mode privileges is only possible from memory regions with a matching *M-mode-only* rule or a *locked* *Shared-Region* rule with executable privileges. Executing code from a region without a matching rule or with a matching _S/U-mode-only_ rule is *denied*.
+* When ``mseccfg.MML=1``, the semantics of PMP rules change as follows:
 
-.. If ``mseccfg.MML`` is not set, the combination of ``pmpcfg.RW=01`` remains reserved for future standard use.
```
Timmmm, line 68: Has this been lost?

src/smepmp.adoc
```diff
@@ -77,95 +40,88 @@ A _Shared-Region_ rule is *enforced* on all modes, with restrictions depending o
...
 
-.. In case ``mseccfg.MMWP`` is not set, M-mode can still access and execute any region not covered by a PMP rule. Since we try to prevent M-mode from executing malicious code and since an attacker may manage to place code on some region not covered by PMP (e.g. a directly-addressable flash memory), we need to ensure that M-mode can only execute the code segments initialized during firmware / OS initialization.
+=== Smepmp software discovery
+
+Since all fields defined on ``mseccfg`` under this proposal are either locked when set (``MMAP``/``MML``) or locked when cleared (``RLB``), software cannot dynamically query them to determine the presence of Smepmp. It is expected that BootROM will set ``mseccfg.MMAP`` and/or ``mseccfg.MML`` during early boot, before transferring control to the firmware. The firmware can then determine the presence of Smepmp by reading ``mseccfg`` and verifying the state of ``mseccfg.MMAP`` and ``mseccfg.MML``. Alternatively a software-defined discovery mechanism may be used.
```
Timmmm, line 111: BootROM -> "the firmware"?

src/smepmp.adoc
```diff
@@ -77,95 +40,88 @@ A _Shared-Region_ rule is *enforced* on all modes, with restrictions depending o
...
 
-=== Smepmp software discovery
+Shared code regions may be used for exposing vendor-specific functionality (e.g. with custom instructions) to Supervisor/User modes, without exposing that code, or risking the posibility of modifications (no write access).
+====
 
-Since all fields defined on ``mseccfg`` as part of this proposal are locked when set (``MMWP``/``MML``) or locked when cleared (``RLB``), software can't poll them for determining the presence of Smepmp. It is expected that BootROM will set ``mseccfg.MMWP`` and/or ``mseccfg.MML`` during early boot, before jumping to the firmware, so that the firmware will be able to determine the presence of Smepmp by reading ``mseccfg`` and checking the state of ``mseccfg.MMWP`` and ``mseccfg.MML``.
+Additionally, when `mseccfg.MML=1`:
 
-[[rationale]]
-=== Rationale
```
Timmmm, line 106: Feels like a shame to lose all this information too.

Timmmm (commented):
I think I was imagining a lighter touch rewording that would be a bit easier to review. Not saying the new version isn't better, it's just a bit hard to tell if everything the old version said is still present.

src/smepmp.adoc
andrewdellow: imo this is all redundant. it reads like a textbook, not an ISA standard

Timmmm: I don't see why an ISA standard shouldn't have some motivation. There are plenty of notes already in the specification that aren't normative.

It could be condensed though for sure. I think it's at least worth mentioning that this is most useful for M/U machines where supervisor protection mechanisms cannot be used.

gfavor:
Tim and Andrew,

You are both right in that the general practice for RISC-V arch specs is to
focus on being just an architecture spec - with limited and concise key
non-normative notes.  Larger explanations of motivation, the why behind
arch choices, usage guidelines, etc. should be elsewhere.

The "original" ePMP spec contains large amounts of non-normative exposition
- which is what Andrew is correctly commenting on.  And part of the
expected exercise in rewriting the spec - and not just integrating it
largely as-is into the ISA manual - is to move it to the aforementioned
general RISC-V practice for arch specs.  The overall work needs to cover
both stages of integration and rewrite, and then a review by security HC
(and old ePMP TG) members should be conducted.

Greg

src/smepmp.adoc
mickflemm: Protecting Machine mode is always relevant, in any implementation scenario. What we said in the original document was that on systems where the OS runs on Machine mode, the attack surface is greatly increased, not because you don't have S-mode protections in-between, but because you run more complex code on M-mode with a higher probability of a vulnerability being present there that would allow for the exploitation of memory accesses from M-mode to less privileged modes ( tricking M-mode to execute code from less privileged modes). So it's not that Smepmp is most useful for M/U machines, we make this clear in the original document, in the last sentence of the part you quoted "Even on implementations with Supervisor mode present attacks are still possible...". With CoVE for example coming up, where we are going to have more complex code running on M-mode, it makes perfect sense to have Smepmp available, OpenSBI (which targets systems with Supervisor mode available) already uses it.

In any case this further explains the "why" not the "what" or "how", IMHO it makes more sense to keep the extended version of the "why" part in the proposal document which is referenced in case anyone wants to dig deeper, than having it as part of the ISA manual. The short version is still there: "to improve Machine mode's security by providing stronger memory isolation and privilege separation". I agree non-normative text still makes sense which is why I added tips e.g. for "how" shared regions may be used, or "why" the last two restrictions were added to MML, I think it's cleaner to have short tips than whole paragraphs.

src/smepmp.adoc
mickflemm: It seemed redundant, we only override this when MML is set, didn't make much sense to say that when MML is not set, those bits remain reserved. It's also clean from the visual representation, but if you think it's worth re-adding it, I'm ok with it.

src/smepmp.adoc
mickflemm: "BootROM, ...during early boot, before transferring control to the firmware" The idea here is that the BootROM is implementation-specific, while the firmware as a software component may be more implementation-agnostic, such as OpenSBI for example or another third-party firmware (it makes perfect sense to have a reference firmware implementation like OpenSBI, that we can maintain/verify etc, and use that in most situations).

src/smepmp.adoc
mickflemm: I think the important stuff is still there, some as part of the implementation considerations, some as tips/info, and the most important part regarding rlb is mentioned twice. It seems inconsistent with the rest of the document to keep this, it's a bit too much to e.g. explain why we didn't say anything about mprv, or what was our thought process etc. The document is still there for people to dig deeper, it doesn't have to be included in the privilege spec. The chapter was 7 pages long, it's now 4 pages long and still contains what's needed for someone to implement this.

src/smepmp.adoc
Timmmm: I can't agree - understanding why it's requires helps enormously in understanding what it does. Having to dig through old proposals, Github PRs and so on to find out the motivation for bits of spec is not pleasant. Buuut... I'm not going to bang on about it. :-)

Timmmm: (Oh I was going to resolve this thread but I don't have permission, sorry.)

src/smepmp.adoc
Timmmm: Right I just mean the term "BootROM" isn't defined anywhere (or used anywhere else). There's only one other instance of "firmware" too, which also doesn't have a definition.

I think the implicit use of firmware (based on the SBI specs) is pretty much "code running in M mode". Maybe something like this makes more sense:

> It is expected that implementation-specific boot code will set ... before transferring control to an SBI implementation.

src/smepmp.adoc
```diff
@@ -2,73 +2,36 @@
...
-+
-* A _Shared-Region_ rule where ``pmpcfg.L`` is set can be used for sharing code between M-mode and S/U-mode, so is not writeable. Both M-mode and S/U-mode have execute access on the region, and M-mode also has read access if ``pmpcfg.X`` is set. The rule remains *locked* so that any further modifications to its associated configuration or address registers are ignored until a *PMP reset*, unless ``mseccfg.RLB`` is set.
-+
-* The encoding ``pmpcfg.LRWX=1111`` can be used for sharing data between M-mode and S/U mode, where both modes only have read-only access to the region. The rule remains *locked* so that any further modifications to its associated configuration or address registers are ignored until a *PMP reset*, unless ``mseccfg.RLB`` is set.
+[[smepmp-mml]]
+=== Machine Mode Lockdown (MML)
 
+The ``mseccfg.MML`` bit fundamentally changes the interpretation of the ``pmpcfg.L`` bit defined in <<pmp-locking>>. It is a sticky bit: once set, it cannot be unset until xref:reset[xrefstyle=basic].
```
lfiolhais, line 24:
```suggestion
The ``mseccfg.MML`` bit changes the interpretation of the ``pmpcfg.L`` bit defined in <<pmp-locking>>. It is a sticky bit: once set, it cannot be unset until xref:reset[xrefstyle=basic].
```

src/smepmp.adoc
```diff
@@ -2,73 +2,36 @@
...
 
-To prevent this attack vector, two mechanisms known as Supervisor Memory Access Prevention (SMAP) and Supervisor Memory Execution Prevention (SMEP) were introduced in recent systems. The first one prevents the OS from accessing the memory of an unprivileged process unless a specific code path is followed, and the second one prevents the OS from executing the memory of an unprivileged process at all times. RISC-V already includes support for SMAP, through the ``sstatus.SUM`` bit, and for SMEP by always denying execution of virtual memory pages marked with the U bit, with Supervisor mode (OS) privileges, as mandated on the Privilege Spec.
+Additionally this extension also provides a set of mechanisms to allow for more flexible and efficient use of the PMP configuration registers, especially during early boot.
```
lfiolhais, line 7:
```suggestion
Additionally this extension provides a set of mechanisms which allow a more flexible and efficient use of the PMP configuration registers, especially during early boot.
```

src/smepmp.adoc
```diff
@@ -2,73 +2,36 @@
...
+* When ``mseccfg.MML=0``, PMP operates according to the original specification, where the ``pmpcfg.L`` bit marks a rule as *Locked* and applicable in all privilege modes.
 
-.. Executing code with Machine mode privileges is only possible from memory regions with a matching *M-mode-only* rule or a *locked* *Shared-Region* rule with executable privileges. Executing code from a region without a matching rule or with a matching _S/U-mode-only_ rule is *denied*.
+* When ``mseccfg.MML=1``, the semantics of PMP rules change as follows:
 
-.. If ``mseccfg.MML`` is not set, the combination of ``pmpcfg.RW=01`` remains reserved for future standard use.
+** ``pmpcfg.L`` denotes an M-mode-only rule when set (applies only to accesses from M-mode), or an S/U-mode-only rule when clear (applies only to accesses from S/U-mode). When ``pmpcfg.L`` is set the rule is also *Locked*, unless ``mseccfg.RLB`` is also set.
```
lfiolhais, line 30: IMO the text already states that the rules are privilege level specific. I don't think the parentheticals are needed.

```suggestion
** ``pmpcfg.L`` denotes an M-mode-only rule when set, or an S/U-mode-only rule when clear. When ``pmpcfg.L`` is set the rule is also *Locked*, unless ``mseccfg.RLB`` is also set.
```

src/smepmp.adoc
```diff
@@ -2,73 +2,36 @@
...
 
+** The formerly reserved encoding `pmpcfg.RW=01` (with various `pmpcfg.L` and `pmpcfg.X` settings) now encodes Shared-Region rules (that apply to accesses from all privilege modes) with varying permissions.
```
lfiolhais, line 32:
```suggestion
** The formerly reserved encoding `pmpcfg.RW=01` (with various `pmpcfg.L` and `pmpcfg.X` settings) now encodes Shared-Region rules with varying permissions, applied to all privilege modes accesses.
```

src/smepmp.adoc
Timmmm: Yeah I think that would make a lot of sense. Probably there should be a "Programmers Guide" which has that info and also tells you how to write code for RISC-V. There are a few snippets of information for that through the spec (e.g. how to do CAS) but it seems like mostly you'd end up reading Linux/GCC source code if you want those details.

Good luck writing that though - it would be an enormous undertaking!

src/smepmp.adoc
```diff
@@ -77,95 +40,88 @@ A _Shared-Region_ rule is *enforced* on all modes, with restrictions depending o
...
-+
-[NOTE]
+* When mseccfg.MMAP=1, Machine mode cannot access memory regions without a matching PMP rule.
+
+[TIP]
 ====
-For adding _Shared-region_ rules with executable privileges to share code segments between M-mode and S/U-mode, ``mseccfg.RLB`` needs to be implemented, or else such rules can only be added together with ``mseccfg.MML`` being set on *PMP Reset*. That's because the reserved encoding ``pmpcfg.RW=01`` being used for _Shared-region_ rules is only defined when ``mseccfg.MML`` is set, and 4b prevents the addition of rules with executable privileges on M-mode after ``mseccfg.MML`` is set unless ``mseccfg.RLB`` is also set.
+This mechanism not only provides a more strict access policy for Machine mode which otherwise has access everywhere unless restricted, it also allows for more efficient use of the available PMP registers, since there is no need to waste a rule for every region that should be blocked such as regions used by peripherals / processors on the system, or e.g. the BootROM after the handoff to firmware.
```
lfiolhais, line 88:
```suggestion
This mechanism not only provides a more strict access policy for Machine-mode which otherwise has access everywhere unless restricted, it also allows for more efficient use of the available PMP registers, since there is no need to waste a rule for every region that should be blocked such as regions used by peripherals / processors on the system, or e.g. the BootROM after the handoff to firmware.
```

src/smepmp.adoc
```diff
@@ -77,95 +40,88 @@ A _Shared-Region_ rule is *enforced* on all modes, with restrictions depending o
...
+=== Machine-Mode Allowlist Policy (MMAP)
 
-. The current dual meaning of the ``pmpcfg.L`` bit that marks a rule as Locked and *enforced* on all modes is neither flexible nor clean. With the introduction of _Machine Mode Lock-down_ the ``pmpcfg.L`` bit distinguishes between rules that are *enforced* *only* in M-mode (_M-mode-only_) or *only* in S/U-modes (_S/U-mode-only_). The rule locking becomes part of the definition of an _M-mode-only_ rule, since when a rule is added in M mode, if not locked, can be modified or removed in a few instructions. On the other hand, S/U modes can’t modify PMP rules anyway so locking them doesn’t make sense.
+The ``msec.MMAP`` bit changes the default PMP policy for Machine mode when accessing memory regions that don't have a matching PMP rule. It is a sticky bit: once set, it cannot be unset until xref:reset[xrefstyle=basic].
```
lfiolhais, line 80:
```suggestion
The ``msec.MMAP`` bit changes the default PMP policy for Machine-mode when accessing memory regions that don't have a matching PMP rule. It is a sticky bit: once set, it cannot be unset until xref:reset[xrefstyle=basic].
```

src/smepmp.adoc
```diff
@@ -77,95 +40,88 @@ A _Shared-Region_ rule is *enforced* on all modes, with restrictions depending o
...
-+
-An attacker that manages to tamper with a memory region used by S/U mode, even after successfully tricking a process running in M-mode to use or execute that region, will fail to perform a successful attack since that region will be _S/U-mode-only_ hence any access when in M-mode will trigger an access exception.
-+
-[NOTE]
-====
-In order to support zero-copy transfers between M-mode and S/U-mode we need to either allow shared memory regions, or introduce a mechanism similar to the ``sstatus.SUM`` bit to temporary allow the high-privileged mode (in this case M-mode) to be able to perform loads and stores on the region of a less-privileged process (in this case S/U-mode). In our case after discussion within the group it seemed a better idea to follow the first approach and have this functionality encoded on a per-rule basis to avoid the risk of leaving a temporary, global bypass active when exiting M-mode, hence rendering memory access prevention useless.
-====
-+
+* When mseccfg.MMAP=0, Machine mode can access memory regions without a matching PMP rule (default behavior).
```
lfiolhais, line 82:
```suggestion
* When mseccfg.MMAP=0, Machine-mode can access memory regions without a matching PMP rule (default behavior).
```

mickflemm:
Any updates on this one ? Is there anything I can do to help integrate this, anything missing ?

aswaterman:
@mickflemm I can help review it if you fix the merge conflicts.

PR #1890, opened 2025-03-06
Alter HLV/HSV/HLVX description to use effective privilege mode terminology

bcstrongx:


src/hypervisor.adoc
```diff
@@ -1225,12 +1225,10 @@ include::images/wavedrom/hypv-virt-load-and-store.edn[]
 
 The hypervisor virtual-machine load and store instructions are valid
 only in M-mode or HS-mode, or in U-mode when `hstatus`.HU=1. Each
-instruction performs an explicit memory access as though V=1; i.e., with
+instruction performs an explicit memory access with an effective privilege mode 
+of VS or VU; i.e., with
```
bcstrongx, line 1229: I debated whether to keep the "; i.e., with ..." section, since effective privilege mode was already defined in the MPRV section.  Let me know if it would be better to avoid duplication and drop this.  I noticed that effective privilege mode is used in the PMP section without defining it.

aswaterman: Yeah, I could go either way, but my sense is that we should eliminate the duplication.  My reasoning is we might use the effective privilege mode as a hook on which to hang additional semantics (kind of like we did with endianness).  Copies of that list of features affected by effective privilege mode will get out of sync, causing confusion.

pdonahue-ventana: I think that removing the i.e. would be better.  If something later gets added (e.g. circa 2019-2020 when bi-endianness was added, endianness was added to the list of things that effective privilege mode affects) then only the effective privilege mode definition changes without having to also remember to change this.

aswaterman: Lol, jinx, @pdonahue-ventana.

aswaterman (approved):

aswaterman:
Thanks again.

PR #1904, opened 2025-03-13
Change VROR_VI immediate field from imm to uimm

nadime15:
The Sail code for `vror.vi` uses `imm` instead of `uimm`,  even though "uimm" is specified in the encoding.  

A correct reference is `vwsll.vi`, which properly uses `uimm`.

Please cross-check the correctness of this proposed change.


aswaterman (approved):
Thanks.

PR #1905, opened 2025-03-13
Remove note that the Zvl*b extension names may change.

topperc:
The part about the trailing "b" being required is true, but I'm not sure it provides any value in this section.


aswaterman (approved):
Yes, good point.

PR #1907, opened 2025-03-14
Add draft of Svrsw60t59b extension

ved-rivos:
Svrsw60t59b extension provides additional reserved-for-software bits in the PTEs to enable advanced memory management features. This extension is being pursued as a fast-track.


src/supervisor.adoc
```diff
@@ -2284,6 +2284,27 @@ Invalid PTEs using a bounded timer, or making address-translation caches
 coherent with store instructions that modify PTEs.
 ====
 
+[[sec:svrsw60t59b]]
+== "Svrsw60t59b" Extension for PTE Reserved-for-Software Bits 60-59, Version 0.1
+
+If the Svrsw60t59b extension is implemented, the bits 60-59 of the page table
```
aswaterman, line 2290: "the bits" => "then bits"?

src/supervisor.adoc
```diff
@@ -2284,6 +2284,27 @@ Invalid PTEs using a bounded timer, or making address-translation caches
...
+entries (PTEs) are reserved for use by supervisor software and are ignored by
+the implementation.
+
+If the Hypervisor (H) extension is also implemented, then the bits 60-59 of the
```
aswaterman, line 2294: "then the bits" => "then bits"?

aswaterman (commented):
LGTM, but we'll leave this as an open PR for now.

AlexGhiti:
Here is the implementation for qemu: https://lists.gnu.org/archive/html/qemu-riscv/2025-03/msg00218.html

src/supervisor.adoc
ved-rivos: updated.

src/supervisor.adoc
ved-rivos: update

jrtc27:
Can this please move to not clash with the PTE bits the stable CHERI spec already is using as I mentioned a month ago?

ved-rivos:
@jrtc27 please see the discussion on Cheri TG list on this topic - https://lists.riscv.org/g/tech-tg-cheri/message/51

src/supervisor.adoc
```diff
@@ -2284,6 +2284,27 @@ Invalid PTEs using a bounded timer, or making address-translation caches
...
+implementation.
+
+The Svrsw60t59b extension depends on Sv39.
+
+[NOTE]
+====
+Operating systems frequently use reserved bits within PTEs to store metadata for
+advanced memory management features. Embedding these metadata bits directly within
+the PTEs allows for fast access with minimal overhead, avoiding costly lookups in
+auxiliary data structures.
```
ved-rivos, line 2305: By default Sv39/Sv39x4 requires a (guest)page-fault exception to be raised if the bits 60-59 are not zero.

src/supervisor.adoc
```diff
@@ -2290,6 +2290,29 @@ Invalid PTEs using a bounded timer, or making address-translation caches
...
 
+[[sec:svrsw60t59b]]
+== "Svrsw60t59b" Extension for PTE Reserved-for-Software Bits 60-59, Version 0.9
+
+If the Svrsw60t59b extension is implemented, then bits 60-59 of the page table
+entries (PTEs) are reserved for use by supervisor software and are ignored by
+the implementation.
+
+If the Hypervisor (H) extension is also implemented, then bits 60-59 of the
+G-stage PTEs are reserved for use by supervisor software and are ignored by the
```
riking, line 2301: Should this specify Virtualized Supervisor software, or is it okay being ambiguous?

aswaterman: "Supervisor" is correct here ("hypervisor" would also have been correct), but "virtualized supervisor" would misleadingly suggest that G-stage PTEs were under the control of a guest.  I think it's OK as-is.

PR #1911, opened 2025-03-17
add missing underscore in extensions example

mengzhuo:


aswaterman:
The text is correct as-is.  Multi-letter extensions only need to be separated from other multi-letter extensions with underscores; they do not need to be separated from single-letter extensions with underscores.

I agree it is easier to read with an underscore, but this example is making a specific point, and adding the underscore weakens that point.

PR #1917, opened 2025-03-19
Deploy a snapshot of the specification to GitHub pages

arichardson:
This makes it easier to view the latest snapshot without having to download a
PDF from the releases page.

I've deployed this to my github fork, so it will look as follows once deployed:
- [Unprivileged spec](https://arichardson.github.io/riscv-isa-manual/snapshot/unprivileged/).
- [Privileged spec](https://arichardson.github.io/riscv-isa-manual/snapshot/privileged/).

The HTML is deployed to a snapshot/ subdirectory to make it possible to also
deploy versioned releases in the future.

This has been very useful for the CHERI spec so far (https://riscv.github.io/riscv-cheri/). In the CHERI spec I added a warning at the top of the document that it is not a valid release:

> This document is a specification snapshot built from github.com/riscv/riscv-cheri/commit/cf51732e0bf6b2775516d168642f05dcca54652d and is not a versioned release. The latest versioned PDF release can be downloaded from github.com/riscv/riscv-cheri/releases.

If you are happy to accept this change I could also add a similar warning to the start of the document for these pages?

### DIFF
--- a/.github/workflows/isa-build.yml
+++ b/.github/workflows/isa-build.yml
@@ -129,3 +129,36 @@ jobs:
       env:
         GITHUB_TOKEN: ${{ secrets.GHTOKEN }}
 
+    # Create directory structure pages sites for privileged and unprivileged specs.
+    - name: Make GitHub pages directory
+      run: |
+        mkdir -p dist/snapshot/unprivileged dist/snapshot/privileged
+        cp build/riscv-unprivileged.html dist/snapshot/unprivileged/index.html
+        cp build/riscv-privileged.html dist/snapshot/privileged/index.html
+      if: github.event_name == 'workflow_dispatch' || (github.event_name == 'push' && github.ref == 'refs/heads/main')
+
+    - name: Upload pages artifact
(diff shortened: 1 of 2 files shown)


aswaterman:
I'll let @wmat review and decide.

trdthg:
Online version can interact well with browser extensions, hope this can be merged soon

wmat:
I don't have a problem with this short term but it will eventually be replaced with an antora based static site.

arichardson:
@wmat thanks for merging this! It appears github pages deployment via GitHub actions needs to be enabled in the settings. Here is what the settings look like for [riscv/riscv-cheri](riscv/riscv-cheri):
image

wmat:
Hmmm, deply looks like it worked but the page 404s.

arichardson:
Thanks, I can now see it under https://riscv.github.io/riscv-isa-manual/snapshot/unprivileged/ and https://riscv.github.io/riscv-isa-manual/snapshot/privileged/. The root doesn't have any contents, so maybe we should add a toplevel index.md that links to the two specs?

PR #1918, opened 2025-03-20
Add pre-commit hooks to CI

arichardson:
This adds the pre-commit hooks from the docs-spec-template repository and fixes all issues they find.
Can be enabled using `pre-commit install`.


wmat:
The build is failing.  Can you check this?

wmat:
Looks like the deploy step is failing.

arichardson:
> Looks like the deploy step is failing.
> […](#)
It sounds like this is a temporary github issue and should work again with the next commit: https://github.com/orgs/community/discussions/49948

PR #1920, opened 2025-03-21
Fix bug on wavedrom C.SRLI/C.SRAI

### LINKED ISSUE #1900: Bug in C.SRLI wavedrom
There is an error with the rendering of the rd/rs1 field  in c.srli as displayed in

Image

AFOliveira:
Fixes #1900

### DIFF
--- a/src/images/wavedrom/c-srli-srai.edn
+++ b/src/images/wavedrom/c-srli-srai.edn
@@ -5,7 +5,7 @@
 {reg: [
   {bits: 2, name: 'op', attr: ['2','C1', 'C1'],},
   {bits: 5, name: 'shamt[4:0]',    attr: ['5','shamt[4:0]', 'shamt[4:0]'],},
-  {bits: 3, name: 'rd′/rs1′',     attr: ['3','dest', 'dest'],},
+  {bits: 3, name: "rd'/rs1'",     attr: ['3','dest', 'dest'],},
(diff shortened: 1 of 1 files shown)


wmat:
I don't think we want the single quote character in this scenario, we want the prime mark character.

AFOliveira:
> I don't think we want the single quote character in this scenario, we want the prime mark character.

what would the unicode of that be?

wmat:
I believe it's U-0289 or U-2032.  I believe it appears in many other wavedrom files.  Perhaps this is one that wasn't updated correctly.

AFOliveira:
@wmat  I tried those and none of them render in Asciidoc. Is there really much difference between the two symbols from a reader standpoint that we need to match them exactly?

aswaterman:
In the asciidoc source itself, we wrote things like `rs1&#x2032;` i.e. using HTML syntax to express the prime -- does that not work here?

AFOliveira:
> In the asciidoc source itself, we wrote things like `rs1&#x2032;` i.e. using HTML syntax to express the prime -- does that not work here?

Not really, unfortunately, I don't think there are many other options other then changing the character
image

aswaterman:
OK, I'll accept whatever you / @wmat / @drom collectively decide.

wmat:
We don't need this change anymore as the font change in docs-resources was reverted and the wavedrom now displays prime mark correctly again.

PR #1921, opened 2025-03-21
Fix a step reference for Sv32 address translation.

pmundkur:
This got out of sync due to steps added in #1742.

Also fix an unrelated typo.


aswaterman (commented):
Thanks for this.

PR #1923, opened 2025-03-21
Zish4add (required by CHERI)

tariqkurd-repo:
@arichardson


src/zish4add.adoc
```diff
@@ -0,0 +1,36 @@
+== "Zish4add" Extension for 16-byte pointer addition, Version 0.9
+
+The B-extension adds `SH[1|2|3]ADD` instructions, and for RV64 all have
+an unsigned word .UW version for RV64.
```
ved-rivos, line 4: One of the "RV64" in this sentence can be removed.

src/zish4add.adoc
```diff
@@ -0,0 +1,36 @@
...
+
+Zish4add adds analogous instructions two new instructions to RV64: SH4ADD and SH4ADD.UW.
```
arichardson, line 6:
```suggestion
Zish4add adds two new analogous instructions to RV64: SH4ADD and SH4ADD.UW.
```

src/zish4add.adoc
```diff
@@ -0,0 +1,36 @@
...
+]}
+....
+
+Increment `rs2` by `rs1` shifted left by 4 bit positions and write the result to `rd`.
+
+
+[wavedrom, sh4adduw-wavedrom-reg,svg]
+....
+{reg:[
+    { bits:  7, name: 0x3b, attr: ['OP'] },
```
topperc, line 26: OP->OP-32

src/zish4add.adoc
arichardson, line 4:
```suggestion
an unsigned word .UW version.
```

arichardson:
The motivation here is that CHERI capabilities for RV64 are 16 bytes, so we end up needing quite a few multiply-by-16 indexing operations.
I would expect this to also be beneficial for non-CHERI code (e.g. `long double` is 128 bits).

src/zish4add.adoc
topperc: I think 0x3b is OP-32. sh4add says 0x33 and sh4add.uw says 0x3b, but they both label it as OP.

tariqkurd-repo: ok - got it - sorry I thought that SH4ADD was wrong not SH4ADD.UW - I'll fix that now

src/zish4add.adoc
```diff
@@ -0,0 +1,36 @@
...
```
tariqkurd-repo, line 26:
```suggestion
    { bits:  7, name: 0x3b, attr: ['OP-32'] },
```

src/zish4add.adoc
```diff
@@ -0,0 +1,36 @@
+== "Zish4add" Extension for 16-byte pointer addition, Version 0.9
+
+The B-extension adds `SH[1|2|3]ADD` instructions, and for RV64 all have
+an unsigned word .UW version.
```
arichardson, line 4:
```suggestion
an unsigned word `*.UW` version.
```

PR #1934, opened 2025-03-27
Remove unused mseccfg.edn bytefield

arichardson:
It uses wavedrom since https://github.com/riscv/riscv-isa-manual/pull/1380


arichardson:
Noticed this while integrating the CHERI spec into the privileged document and was surprised my edits did not show up in the document :)

aswaterman (approved):

aswaterman:
There are few things more satisfying than deleting code.

PR #1937, opened 2025-03-28
Add a flag to allow faster (but unreliable) incremental builds

arichardson:
Setting UNRELIABLE_BUT_FASTER_INCREMENTAL_BUILDS=1 on the make command line
will avoid wiping the entire build directory before builds. Since each
output file uses a different build directory there should be no
need to wipe the whole build directory. The Makefile already triggers a
rebuild if any file under the source directory is modified so this should
be safe.
This significantly speeds up the build for me since we no longer regenerate
all the diagrams for each build. For the unprivileged manual the build time
goes from about 4:30 to 3:40.

However, it has been shown that non-clean builds don't always produce the
same output PDF, so this is gated by an opt-in flag.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,6 +58,10 @@ else
         cd $@.workdir &&
 endif
 
+ifdef UNRELIABLE_BUT_FASTER_INCREMENTAL_BUILDS
+WORKDIR_SETUP = mkdir -p $@.workdir && ln -sfn ../../src ../../docs-resources $@.workdir/
+WORKDIR_TEARDOWN = mv $@.workdir/$@ $@
+else
 WORKDIR_SETUP = \
     rm -rf $@.workdir && \
     mkdir -p $@.workdir && \
@@ -66,6 +70,7 @@ WORKDIR_SETUP = \
 WORKDIR_TEARDOWN = \
     mv $@.workdir/$@ $@ && \
     rm -rf $@.workdir
+endif
 
 SRC_DIR := src
 BUILD_DIR := build


arichardson:
Timings with this change 4m33s -> 3m42s, so almost one minute saved by not re-rendering diagrams.
```
$ make clean && time make build/riscv-unprivileged.pdf
Input file: src/riscv-unprivileged.adoc
  Time to read and parse source: 81.73687
  Time to convert document: 186.63846
  Total time (read, parse and convert): 268.37533
mv build/riscv-unprivileged.pdf.workdir/build/riscv-unprivileged.pdf build/riscv-unprivileged.pdf
make build/riscv-unprivileged.pdf  0.20s user 0.17s system 0% cpu 4:33.45 total
```

Incremental build after adding one line to one of the files:
```
$ time make build/riscv-unprivileged.pdf
Input file: src/riscv-unprivileged.adoc
  Time to read and parse source: 27.76435
  Time to convert document: 189.06771
  Total time (read, parse and convert): 216.83206
mv build/riscv-unprivileged.pdf.workdir/build/riscv-unprivileged.pdf build/riscv-unprivileged.pdf
make build/riscv-unprivileged.pdf  0.16s user 0.16s system 0% cpu 3:42.16 total

```

Looks like asciidoctor treats the diagram generation as part of the `Time to read and parse source` (added --timings to the build).

aswaterman (changes requested):
Alas, I can't trust this.  I implemented the brute-force scheme I did because, in the past, I was getting subtly different ultimate PDFs when compiling incrementally versus not.  I'm willing to take the performance hit to not have that risk.

Probably makes more sense to adequately parallelize figure generation.  That will recoup most of the gain from this PR, while also speeding up from-scratch builds, including CI.

arichardson:
> Alas, I can't trust this. I implemented the brute-force scheme I did because, in the past, I was getting subtly different ultimate PDFs when compiling incrementally versus not. I'm willing to take the performance hit to not have that risk.
> 
> Probably makes more sense to adequately parallelize figure generation. That will recoup most of the gain from this PR, while also speeding up from-scratch builds, including CI.

Ah that's unfortunate and a very good reason not to make this change. Was not aware of this asciidoc problem.

Would you be okay with an opt-in flag that can be used for slightly faster local development? For example something like `make UNRELIABLE_INCREMENTAL_BUILDS=1 build-pdf` with an awkward name to ensure no one accidentally enables it.

aswaterman:
Yeah, SGTM.

arichardson:
Updated the PR description and the logic.

wmat:
This is fine with me as well.  Few people do local builds.

aswaterman (approved):

PR #1948, opened 2025-04-02
Specify behaviour of explicit minstreth writes

### LINKED ISSUE #1255: Unspecified behaviour with explicit writes to minstret on RV32
On RV32 if `minstret` the lower 32 bits of `instret` is 0xFFFFFFFF, and `minstreth` (the upper 32 bits) is 0. And then you do

```
csrw minstret, zero
```

What value would you expect `minstreth` to get? 0 or 1?

The two possibilities are:

0: You have explicitly written to `instret`, therefore the automatic increment is suppressed. As sort of indicated by this text:

> Some CSRs, such as the instructions-retired counter, instret, may be modified as side effects of instruction execution. In these cases, if a CSR access instruction reads a CSR, it reads the value prior to the execution of the instruction. If a CSR access instruction writes such a CSR, the write is done instead of the increment. In particular, a value written to instret by one instruction will be the value read by the following instruction.
> ...
> Furthermore, an explicit CSR read returns the CSR state before the execution of the instruction, while an explicit CSR write suppresses and overrides any implicit writes or modifications to the same CSR by the same instruction.

1: The instruction completes as usual, incrementing the counter, and *then* the write takes effect. As sort of indicated by this text:

> Any CSR write takes effect after the writing instruction has otherwise completed.

I suspect the answer is 0, but it's not really clear because it talks about `instret` (which isn't actually the name of a CSR), and "the same CSR" - but is `minstret` the same CSR as `minstreth` on RV32?

Timmmm:
This is the behaviour that is expected by Valtrix SPIKE. However Spike and the Sail model currently implement the alternate behaviour.

Fixes #1255

### DIFF
--- a/src/zicsr.adoc
+++ b/src/zicsr.adoc
@@ -162,6 +162,27 @@ In particular, a value
 written to `instret` by one instruction will be the value read by the
 following instruction.
 
+On RV32, 64-bit counters such as `instret` are split over multiple CSRs
+(`instret` and `instreth`). If a CSR access instruction writes to one
+of the CSRs then the other is not updated by any side effects.
+
+----
+li t0, 0xFFFFFFFF
+csrw minstret, t0
+# minstret = 0x????????FFFFFFFF
+csrw minstreth, t0
+# minstret = 0xFFFFFFFFFFFFFFFF (the lower 32 bits did not increment
+# because we explicitly wrote to the upper 32 bits)
+
+nop
+
+# minstret = 0x0000000000000000
+csrr t1, minstret
+# t1 = 0x00000000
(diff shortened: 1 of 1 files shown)


Timmmm:
Alternative shorter wording:

> On RV32, 64-bit counters are accessible through two 32-bit CSRs (e.g. `instret` and `instreth`). If one of these is explicitly written then the other is not updated by any side effects.

tariqkurd-repo:
@aswaterman are there any objections to merging this? Once merged we can get the simulators aligned with the behaviour.

aswaterman:
Yes, there are objections (see discussion on #1255).

PR #1950, opened 2025-04-02
[Zvknhk] initial draft for Zvknhk (vector Keccak acceleration)

nibrunieAtSi5:
Creating mostly dummy shell for PQC TG proposal for Keccak instruction.

Not ready for internal nor public review yet.

cc @mjosaarinen / @lfiolhais

Changelog:
- 2-April-2025: initial draft


src/zvknhk.adoc
```diff
@@ -0,0 +1,435 @@
...
+=== Extension Overview
+
+This section introduces `Zvknhk`.
+
+`Zvknhk` depends on Zve64x.
+
+
+As the instructions defined in this specification might be used to implement cryptographic primitives,
+they must be implemented with data-independent execution latencies as defined in 
+link:https://github.com/riscv/riscv-isa-manual/blob/main/src/scalar-crypto.adoc#data-independent-execution-latency-subset-zkt.
```
lfiolhais, line 24: Zvkt right?

src/zvknhk.adoc
```diff
@@ -0,0 +1,435 @@
+
+[[zvknhk,zvknhk]]
+== "Zvknhk" Extension for Vector Keccak 
+
+This chapter describes the `Zvknhk`  Vector Cryptography extension, which adds vector instructions to compute the Keccak sponge function using in NIST SHA3 and numerous Post Quantum Cryptography (PQC) algorithms.
```
lfiolhais, line 5: The FIPS-202 spec formally defines the Keccak construction. Can we put a reference to the publication?

src/zvknhk.adoc
```diff
@@ -0,0 +1,435 @@
...
+
+`Zvknhk` defines instructions to enable the efficient implementation of the Keccak sponge function.
+
+The instructions defined in this specification share the following constraints:
+
+- element group size (EGS) is 32
+- data independent execution timing
+- `vl`/`vstart` must be multiples of EGS=32 (else behavior is _reserved_)
+
+All of these instructions work on 2048-bit element groups comprised of 32 64-bit elements, in element group parlance `EGS=32`, `EGW=2048` and the instructions are only defined for `SEW=64`.
```
lfiolhais, line 44:
```suggestion
All of these instructions work on 2048-bit element groups comprised of 32 64-bit elements. In element group parlance `EGS=32`, `EGW=2048` and the instructions are only defined for `SEW=64`.
```

lfiolhais (commented):
Hey @nibrunieAtSi5, thank you so much for putting this together! I'm unfortunately swamped with other work and can't immediately devote a lot of time to the specification writing. I hope to change that in the near future.

From a high-level, I don't have a lot of feedback to provide. I guess my biggest comment is in regards to the Zvkt note which we should definitely discuss at a Crypto SiG meeting.

The other point of contention, that I'm sure we all agree, is the definition of the Zvknhk for VLEN < 256. The only idea that comes to my mind right now is to define specific instructions which output partial results of state.

Besides technical comments, I think we should also discuss some counter-arguments to provide when we get push back due to the large EGW. I can speak from experience, which granted I don't think is representative, that most implementation won't want to implement an instruction that requires dumping a large part or the whole register file into a single FU. Maybe an implementation guide would aquiesce some concerns?

src/zvknhk.adoc
nibrunieAtSi5: We use Zkt as the reference for the data independent execution latency definition (this is basically a copy/paste from the vector crypto spec, I kept the same reference).

PR #1953, opened 2025-04-04
Remove [5:4] after spimm in cm.push/pop wavedrom and assembly syntax explanation.

topperc:
The text refers to spimm as "the number of additional 16-byte address increments allocated for the stack frame." and "The total stack adjustment represents the total size of the stack frame, which is stack_adj_base added to spimm scaled by 16"

The "Example RV32I PUSH/POP sequences" section shows spimm with values of 0-3.

Based on the text and those examples, I think spimm is a 2-bit value rather than a 6-bit value.

### DIFF
--- a/src/zc.adoc
+++ b/src/zc.adoc
@@ -1406,11 +1406,11 @@ Encoding (RV32, RV64):
 [wavedrom, , svg]
 ....
 {reg:[
-    { bits:  2, name: 0x2,             attr: ['C2'] },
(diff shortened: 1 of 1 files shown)


aswaterman (approved):
It would make sense for the figure to say `spimm[5:4]` if the text referred to `spimm` without scaling it by 16.  The meaning would be that this 2-bit field supplies bits 5:4 of the immediate operand, with the other bits defined to be 0.  But since the text explicitly scales by 16, I agree it seems a little inconsistent.

We can try your version and see how people feel about it.

PR #1960, opened 2025-04-08
Clarify that vaadd intermediate results aren't truncated

### LINKED ISSUE #1957: question of Subtraction Wrap-Around in Vector vasubu.vv Instruction
I am currently verifying the vasubu.vv instruction, with SEW = 8,  no inactive element. One element of vs2 is 0x6F, and the corresponding element in vs1 is 0xD9. I have a question. According to the vector spec, the subtraction result will ignore overflow and wrap around. My understanding is that 0x6F - 0xD9 = 0xFFFFFF96. After ignoring the overflow in the subtraction, the result is 0x96, which is an unsigned number. Then, shifting it results in 0b01001011, which is 0x4b. However, according to the calculation by Spike, the shifting includes the sign bit in front, resulting in 0b11001011, which is 0xcb. I think this does not conform to the meaning of the spec. Could you please explain which understanding is correct?

The relevant description in the spec is as follows:

Image

Image

Image

aswaterman:
Resolves #1957

### DIFF
--- a/src/v-st-ext.adoc
+++ b/src/v-st-ext.adoc
@@ -3096,6 +3096,7 @@ vssub.vx vd, vs2, rs1, vm   # vector-scalar
 
 The averaging add and subtract instructions right shift the result by
 one bit and round off the result according to the setting in `vxrm`.
+Computation is performed in infinite precision before rounding and truncating.
 Both unsigned and signed versions are provided.
 For `vaaddu` and `vaadd` there can be no overflow in the result.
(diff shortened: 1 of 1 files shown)


src/v-st-ext.adoc
kasanovic, line 1: Looks OK, but I think it would read better to put the "infinite precision" sentence before the "both signed and unsigned versions are provided" sentence, as that then flows more naturally into last two sentences.

aswaterman: Agreed.

PR #1965, opened 2025-04-09
Add missing SSE bit in senvcfg for RV32

### LINKED ISSUE #1963: senvcfg.SSE bit missing in RV32 figure
Looking at section 12.1.10, we see bit 3 labelled as SSE in figure 61 for RV64, but the same bit is still labeled WPRI in figure 62 for RV32.

ved-rivos:
closes #1963


aswaterman (approved):
Thanks, @ved-rivos!

PR #1967, opened 2025-04-10
hypervisor: add HUPMM to hstatus

radimkrcmar:
HUPMM was forgotten when the pointer masking extension was added.

Tranlate the definitions to wavedrom while at it, to stay sane.

The manual looked like this before:
image

The manual will look like this after:
image


aswaterman:
Can you please break this PR up into two commits, the first of which translates to wavedrom with no change?  It is impractical to review as-is.

aswaterman (approved):
Thanks for humoring me, @radimkrcmar!

Issue #1849, opened 2025-02-10
Unclear sentence in "C" extension


pfusik:
https://github.com/riscv/riscv-isa-manual/blob/c164f059166da2c71e8971d69e8ea5a26c96ceb6/src/c-st-ext.adoc?plain=1#L390

pfusik:
LGTM, thanks!

aswaterman:
Resolved by #1974.

PR #1974, opened 2025-04-14
Three minor documentation tweaks

bgrayson-mips:
1. Invoking *ret from a too-low priv level results in illegal exception. This is otherwise not clearly stated.
2. Added clarifications around MISA.C and the D extension.
3. Rewrote a sentence discussing why we don't do load/store multiples in the C extension. See https://github.com/riscv/riscv-isa-manual/issues/1849 -- I think this will resolve that.

### DIFF
--- a/src/machine.adoc
+++ b/src/machine.adoc
@@ -2403,9 +2403,9 @@ also raise an illegal-instruction exception when TSR=1 in `mstatus`, as
 described in <<virt-control>>. An __x__RET instruction
 can be executed in privilege mode _x_ or higher, where executing a
 lower-privilege __x__RET instruction will pop the relevant lower-privilege
-interrupt enable and privilege mode stack. Invoking MRET in a lower privilege
-level than M, or SRET in a lower privilege level than S, will raise an
-illegal-instruction exception.  In addition to manipulating
+interrupt enable and privilege mode stack. Attempting to execute an __x__RET
+instruction in a mode less privileged than _x_ will raise an
+illegal-instruction exception. In addition to manipulating
 the privilege stack as described in <<privstack>>,
 __x__RET sets the `pc` to the value stored in the `__x__epc` register.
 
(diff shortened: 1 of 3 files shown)


src/zc.adoc
```diff
@@ -83,8 +83,8 @@ Therefore common ISA strings can be updated as follows to include the relevant Z
 
 MISA.C is set if the following extensions are selected:
 
-* Zca and not F
-* Zca, Zcf and F is specified (RV32 only)
+* Zca and not F (or D)
```
christian-herber-nxp, line 86: This could be interpreted in multiple ways.
D implies F, so it is not really necessary to mention D, then things are clearer.

bgrayson-mips: Agreed. I was thinking of "(and thus not D)" or something. But I can yank that part out.

bgrayson-mips: Fixed.

src/machine.adoc
```diff
@@ -2403,7 +2403,9 @@ also raise an illegal-instruction exception when TSR=1 in `mstatus`, as
 described in <<virt-control>>. An __x__RET instruction
 can be executed in privilege mode _x_ or higher, where executing a
 lower-privilege __x__RET instruction will pop the relevant lower-privilege
-interrupt enable and privilege mode stack. In addition to manipulating
+interrupt enable and privilege mode stack. Invoking MRET in a lower privilege
+level than M, or SRET in a lower privilege level than S, will raise an
+illegal-instruction exception.  In addition to manipulating
```
aswaterman, line 2408: For stylistic consistency, I would prefer "Attempting to execute an __x__RET instruction in a mode less privileged than _x_ will raise an illegal-instruction exception".

aswaterman (commented):
LGTM aside from my one comment; please ping me when it's time to merge.

src/machine.adoc
bgrayson-mips: Updated with your version.

aswaterman (approved):
Thanks!

PR #1975, opened 2025-04-17
Fix HRM=>HRW typo for henvcfgh in priv-csrs.adoc

adam-greenblatt:
Hi, I'm new to RISC-V, and was learning about the architecture by skimming over the instruction set manual.
I noticed what looks like a trivial typo in Table 6 in section 2.2 of Volume II: the henvcfgh CSR is listed with a privilege of of "HRM" which appears to be a typo for "HRW".


aswaterman (approved):
Thanks, clearly a typo.

PR #1987, opened 2025-04-24
Updated JVT and pseudocode 

bgrayson-mips:
The existing document and pseudocode did not mention that the base field is shifted left 6 bits before use.

### DIFF
--- a/src/hypervisor.adoc
+++ b/src/hypervisor.adoc
@@ -1142,7 +1142,9 @@ read-only zeros. Else, `vsip`.SSIP and `vsie`.SSIE are aliases of
 
 The `vstvec` register is a VSXLEN-bit read/write register that is
 VS-mode’s version of supervisor register `stvec`, formatted as shown in
-<<vstvecreg>>. When V=1, `vstvec` substitutes for
+<<vstvecreg>>. Note that, similar to `mtvec` and `stvec`, the
+least-significant two bits are omitted in the field, and will be replaced
+with zeroes when used as an address, thus enforcing 4-byte alignment. When V=1, `vstvec` substitutes for
 the usual `stvec`, so instructions that normally read or modify `stvec`
 actually access `vstvec` instead. When V=0, `vstvec` does not directly
 affect the behavior of the machine.
--- a/src/machine.adoc
+++ b/src/machine.adoc
@@ -1182,9 +1182,11 @@ include::images/bytefield/mtvec.edn[]
 
 The `mtvec` register must always be implemented, but can contain a
 read-only value. If `mtvec` is writable, the set of values the register
-may hold can vary by implementation. The value in the BASE field must
-always be aligned on a 4-byte boundary, and the MODE setting may impose
-additional alignment constraints on the value in the BASE field.
+may hold can vary by implementation. Note that the CSR contains only bits
+XLEN-1 through 2 of the address BASE. When used as an address, the unspecified
+lower 2 bits are filled with zeroes to obtain a full XLEN-bit address that is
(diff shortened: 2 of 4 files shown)


aswaterman:
I think the spec says the right thing as-is, just perhaps not to your stylistic preference.  See the image that represents the `jvt` CSR:

image

`base[XLEN-1:6]` occupies `jvt[XLEN-1:6]`.  In other words, the right CSR bits land in the right address positions without any additional shifting.

If had we instead said that `base` (without any indexing) occupies `jvt[XLEN-1:6]`, then we would need the explicit left shift.

bgrayson-mips:
Wow, that's subtle. Especially since `6` and `0` are both round and look similar at a glance on the screen, and most folks' monitors are in landscape mode which makes the fonts smaller than if they were in portrait mode, and everyone knows bitfields go to bit 0 unless the bitfield is split, except for this one instance.

I don't know if I've ever noticed a bitfield with some bits specified but other bits undefined, just hanging in the breeze, out of three architectures' worth of specs through the decades. So yeah, that's **definitely** not to my stylistic preference. 😄

I think if the collective hive-mind wants to stick to this style, then it's worth a note or something saying:
"Note that the lower bits `base[5:0]` are not specified directly in the JVT CSR, and will be filled with zeroes for jump-table address computations."

Whatcha think?

bgrayson-mips:
This also means that the pseudo-code reference to `jvt.base` and `jvt.mode` refer to a union, not a packed struct, and code that updates just the base field would be likely to clobber the mode field, at least if the pseudo-code is like C.

ved-rivos:
`c.addi16sp` is perhaps a good example.

bgrayson-mips:
> > This also means that the pseudo-code reference to jvt.base and jvt.mode refer to a union, not a packed struct, and code that updates just the base field would be likely to clobber the mode field, at least if the pseudo-code is like C.
> 
> I don't agree with this part. Nothing in the spec justifies the notion that `jvt.base[5:0]` overlaps `jvt.mode`.

I find the use of the `.` operator to refer to both a specified bitfield and the unspecified bits of the bitfield as a larger 32-bit quantity to be a syntactic stretch. One can't write in C (or any other language I'm aware of) like that.

(And I was wrong, this is like both a packed struct in that the bitfields don't overlap and are independently addressable, but also at the same time like a union in that referencing `jvt.base` provides the whole value, not just the bitfield contents. So this pseudo-language is clearly not based on C! 😄 )

But regardless, I've spent way too long writing and rewriting and rewriting and rewriting that one sentence, to try to make it as clear as possible from all stylistic preferences. What do you guys think about the latest version? It covers the unspecified bits, clarifies that _jvt.base_ refers to all 32 bits and not just the specified 26 bits, etc.

tariqkurd-repo:
I'm in agreement with @aswaterman and @pdonahue-ventana here - it's reusing the same scheme as Xtvec - it's an XLEN address where the lower bits aren't read from the CSR field, they're zero, because the lower bits are a different field.

So I guess any clarification should also be applied to all Xtvec registers, or is the description of those less confusing?

bgrayson-mips:
I've added clarification to the *tvec descriptions as well.
There's a capitalization inconsistency in that in the JVT CSR, it's the _base_ field, but in the *tvec CSRs, it's the _BASE_ field.
I can fix that in this PR as well, if desired.

aswaterman:
Manually merged with a few tweaks. I don't think the hypervisor change is necessary, since the text is just saying that the register substitutes for `stvec`; the detailed semantics are left to the definition of `stvec`.

See #1988 1988

PR #1997, opened 2025-05-02
Fix incorrect instruction names

ThinkOpenly:
The spec says:
> [...] assembler pseudoinstructions are defined to set and clear bits in the
> CSR when the old value is not required: CSRS/CSRC csr, rs1; CSRSI/CSRCI csr,
> uimm.

Neither CSRSI nor CSRCI are actual instruction names.

Correct these references to CSRRSI and CSRRCI, respetively.

### DIFF
--- a/src/zicsr.adoc
+++ b/src/zicsr.adoc
@@ -168,8 +168,8 @@ a CSR, CSRW _csr, rs1_, is encoded as CSRRW _x0, csr, rs1_, while CSRWI
 _csr, uimm_, is encoded as CSRRWI _x0, csr, uimm_.
 
 Further assembler pseudoinstructions are defined to set and clear bits
-in the CSR when the old value is not required: CSRS/CSRC _csr, rs1_;
-CSRSI/CSRCI _csr, uimm_.
+in the CSR when the old value is not required: CSRS/CSRC _csr, uimm_;
+CSRRSI/CSRRCI _x0, csr, uimm_.
 
 ==== CSR Access Ordering
 


ThinkOpenly:
More problems here...

CSRRSI needs a (missing) third parameter, `xd`, but the pseudoinstructions don't
perform a register write, so that added `xd` should be `zero`/`x0`.

topperc:
CSRSI and CSRCI are assembler pseudo instructions that omit the x0 destination.

src/zicsr.adoc
```diff
@@ -168,8 +168,8 @@ a CSR, CSRW _csr, rs1_, is encoded as CSRRW _x0, csr, rs1_, while CSRWI
 _csr, uimm_, is encoded as CSRRWI _x0, csr, uimm_.
 
 Further assembler pseudoinstructions are defined to set and clear bits
-in the CSR when the old value is not required: CSRS/CSRC _csr, rs1_;
-CSRSI/CSRCI _csr, uimm_.
+in the CSR when the old value is not required: CSRS/CSRC _csr, uimm_;
```
topperc, line 171: The last argument to csrs/csrc should be a register so rs1 was correct.

ThinkOpenly: Oh! That sentence is defining 4 pseudoinstructions. I presumed the same pattern as the previous paragraph: [pseudo] encoded as [base instruction]. My bad.

ThinkOpenly: This paragraph doesn't explicitly state what base instructions the pseudoinstructions map to. Maybe I'll add that, in case anyone else falls into the trap I did.

Issue #2000, opened 2025-05-04
`addi` is supposed to be "constant time", including the immediate value, does this mean `mv` can't be optimized?


programmerjake:
as described below, `addi` is in the short list of instructions that aren't supposed to have execution latency depend on immediate values (to avoid leaking `pc` values), this means the cpu can't do anything special when the immediate is zero, which is a `mv` pseudo-instruction. register renaming cpus very commonly treat move instructions specially, by handling them in the rename stage so they effectively take 0 cycles to execute. this contradicts the requirement that `addi` can't have different latencies for different immediate values:
https://github.com/riscv/riscv-isa-manual/blob/242d17c4f62c4726e45b915df7d1332106e5a712/src/scalar-crypto.adoc?plain=1#L3954-L3959

I think there should be text added that an `addi` with a zero immediate is allowed to have a different latency than one with a non-zero immediate, but that its latency must still not depend on the values in the registers or on which particular non-zero immediate it has if it has one.

ved-rivos:
An addi with immediate value of zero implemented as a zero cycle move will still be a constant time instruction if it always takes 0 cycles for that instruction. If the implementation however had differing latency for the same instruction depending on data value then it would not qualify as constant time.

programmerjake:
the issue isn't that `mv` isn't constant-time, but that `addi`'s latency would depend on the immediate, which is not allowed according to the quoted text.

ved-rivos:
`addi` is not one instruction - each encoding of the immediate is a different instruction and each of them need to be constant time.

programmerjake:
> `addi` is not one instruction - each encoding of the immediate is a different instruction and each of them need to be constant time.

that renders meaningless the last part of the text I quoted from the isa manual above about not leaking the PC, so afaict that's not what the authors of that text implied.

programmerjake:
afaict that text was originally added in https://github.com/rvkrypto/riscv-zkt-list/commit/6207e14a4afe51706fa0ad06a15fbe7cb1b93e78 by @mjosaarinen

mjosaarinen:
In DIEL (daya-independent execution latency, i.e.. Zkt and Zvkt extensions) execution time can always depend on immediate values, register choices etc. But not *data*, i.e. contents of registers. Execution time can also depend on PC.

programmerjake:
[...]

PR #2001, opened 2025-05-04
Remove a confusing sentences about hiding PC https://github.com/riscv…

mjosaarinen:
…/riscv-isa-manual/issues/2000
This segment is ambiguous and can be viewed to contradict other statements about how DIEL works.

### DIFF
--- a/src/scalar-crypto.adoc
+++ b/src/scalar-crypto.adoc
@@ -3953,10 +3953,7 @@ instructions composed of micro-ops) in way that avoids data-dependent latency.
 techniques such as those described in ISO/IEC 17825 cite:[IS16].
 * Binary executables should not contain secrets in the instruction encodings
 (Kerckhoffs's principle), so instruction timing may leak information about
-immediates, ordering of input registers, etc. There may be an exception to this
-in systems where a binary loader modifies the executable for purposes of
-relocation -- and it is desirable to keep the execution location (PC) secret.
-This is why instructions such as LUI, AUIPC, and ADDI are on the list.
+immediates, ordering of input registers, etc.
 * The rules used by audit tools are relatively simple to understand.
 Very briefly; we call the plaintext, secret keys, expanded keys, nonces,
 and other such variables "secrets". A secret variable (arithmetically)


wmat (approved):
I agree that this was confusing.

gfavor:
The text being removed was explaining why LUI, AUIPC, and ADDI are included in the list.  So what does removal of the text mean with regards to the inclusion of LUI, AUIPC, and ADDI?  Do they remain included for seemingly no good reason, or do they also get removed, or what??

wmat:
I think the confusing text was the "There may be.." bit.  There either is an exception or there isn't, no?  Happy to revert though if the change doesn't make sense.

gfavor:
But what is also being deleted is "This is why instructions such as LUI, AUIPC, and ADDI are on the list."  In other words this text and the preceding sentence that explains why LUI, AUIPC, and ADDI are on the list, is all being deleted.

wmat:
What about reverting and adding back that text, but without the "There may be..." bit?

"An exception to this in systems where a binary loader modifies the executable for purposes of relocation -- and it is desirable to keep the execution location (PC) secret. This is why instructions such as LUI, AUIPC, and ADDI are on the list."

aswaterman:
In the RISC-V ABIs, binaries aren't modified to facilitate relocation, so the observation is more of a curiosity, anyway.  I'm OK with whatever y'all decide; just pointing out that fact.

radimkrcmar:
Would it be sufficient to say `Instructions such as LUI, AUIPC, and ADDI are on the list to allow the execution location (PC) to remain secret.`?

(Talking about an exception to a rule that we adhere to still seems confusing.)

ved-rivos:
I interpret the statement “That is why…” as related to the first part of the paragraph which states that secrets must not be embedded in immediate. So this sentence can be retained as is. The sentence about keeping PC secret can be removed.

radimkrcmar:
@wmat, I even prepared the pull request, but then I read again https://github.com/riscv/riscv-isa-manual/issues/2000#issuecomment-2849353815 where the secret PC argument was shown infeasible, because each encoding is considered as a separate instruction wrt. timing.

So I think the current version is sufficient -- listing of these instruction seems explained later by:
```
Some instructions are on the list simply because we see no harm in including them in testing scope.
```

wmat:
@radimkrcmar the current version as with this change applied, or the original version that this PR changed?

mjosaarinen:
Well, at least ADDI needs to be on the list for security reasons unrelated to "PC remaining secret" so that was definitely misleading.

radimkrcmar:
Right, if ADDI/LUI wasn't on the list, timing minded software would have to do the following instead of `addi z, y, i`:
```
xor x, x, x  // zero x to make sure that data in x doesn't affect the timing of lui
lui x, i
add z, y, x
```
I am all in favor of including the immediate versions in Zkt, because they add basically zero implementation cost.

aswaterman:
Not sure why we're debating the spec's semantics; it's already ratified and can't be changed.  The discussion here should only be about how best to describe the thing that we ratified long ago.

PR #2005, opened 2025-05-05
pre-commit forbid tabs

### LINKED ISSUE #1982: add pre-commit hook to checks to flag tabs (should be spaces)
Brought up on doc-sig meeting Apr 22, 2025 by @allenjbaum

kbroch-rivosinc:
fixes #1982

now passes:
```
kbroch@kbroch-mac.local:~/rvi/repos/riscv/riscv-isa-manual on  dev/kbroch/pre-commit-no-tabs:main [!⇡] took 2s
❯ pre-commit run --all-files forbid-tabs
No-tabs checker..........................................................Passed
```


aswaterman (commented):
This is a nice improvement, but can you fix up the assembly code examples so that they render as intended?  e.g. in the B extension examples, it is clear that the intent was for the register specifiers to line up with each other.

kbroch-rivosinc:
> This is a nice improvement, but can you fix up the assembly code examples so that they render as intended? e.g. in the B extension examples, it is clear that the intent was for the register specifiers to line up with each other.

Ok, I think I have I have the reg's column aligned now.

aswaterman (approved):

Issue #2037, opened 2025-05-21
Extraneous Table in Chapter 35


redyoshi49q:
In Chapter 35 of document version 20250508, there's an extraneous instruction type table split across the end of page 615 and the start of page 616 (just after the table for RV64D instructions).

PR #2038, opened 2025-05-21
Update rv-32-64g.adoc to remove extraneous table

redyoshi49q:
This should fix issue #2037 , though I haven't set up a build environment for the documentation, and thus haven't tested the change.


aswaterman (approved):
> and thus haven't tested the change

That's what continuous integration is for :-)

Issue #2033, opened 2025-05-21
Control Transfer Record CSRs missing from priv-csrs.adoc


topperc:
Original table from the CTR repo

CSR Number | Name | Description
-- | -- | --
0x14E | sctrctl | Supervisor Control Transfer Records Control Register
0x14F | sctrstatus | Supervisor Control Transfer Records Status Register
0x15F | sctrdepth | Supervisor Control Transfer Records Depth Register
0x24E | vsctrctl | Virtual Supervisor Control Transfer Records Control Register
0x34E | mctrctl | Machine Control Transfer Records Control Register

CC: @aswaterman

aswaterman:
@bcstrongx Can you add these to the CSR tables in `priv-csrs.adoc`?

bcstrongx:
Oops.  Sure.

aswaterman:
Resolved by #2043

PR #2043, opened 2025-05-22
add CTR CSRs to priv CSRs section, per issue #2033

bcstrongx:


aswaterman (approved):
Thank you for the rapid turnaround, @bcstrongx.  I fixed the trailing whitespace that was causing CI to fail.

PR #2056, opened 2025-05-30
Change reference in rnmi.adoc from MRET to MNRET

Peter-Herrmann:
mnepc is used by the MNRET instruction, not the MRET instruction.


aswaterman (approved):
Thanks for finding this typo.

PR #2065, opened 2025-06-03
Correct the table for RVWMO CSR instruction dependencies

terpstra:
CSR instructions which set/clear bits must still carry the previous dependencies for the target CSR.
Conversely, instructions which update a CSR from an immediate clear dependencies.

### DIFF
--- a/src/rvwmo.adoc
+++ b/src/rvwmo.adoc
@@ -503,19 +503,23 @@ register(s) to destination register(s) as specified
 
 |CSRRW‡ |_rs1_, _csr_^*^ | _rd_, _csr_ | |^*^unless _rd_=`x0`
 
+5+| ‡ carries a dependency from _rs1_ to _csr_ and from _csr_ to _rd_
+
 |CSRRS‡ |_rs1_, _csr_ |_rd_ ^*^, _csr_ | |^*^unless _rs1_=`x0`
 
 |CSRRC‡ |_rs1_, _csr_  |_rd_ ^*^, _csr_ | |^*^unless _rs1_=`x0`
 
-5+| ‡ carries a dependency from _rs1_ to _csr_ and from _csr_ to _rd_
+5+| ‡ carries a dependency from _csr_ and _rs1_ to _csr_ and from _csr_ to _rd_
 
 |CSRRWI ‡ |_csr_ ^*^ |_rd_, _csr_  | |^*^unless _rd_=_x0_
+
+5+| ‡ carries a dependency from _csr_ to _rd_ and clears the dependencies of _csr_
 
 |CSRRSI ‡ |_csr_ |_rd_, _csr_^*^  | |^*^unless uimm[4:0]=0
 
 |CSRRCI ‡ |_csr_ |_rd_, _csr_^*^  | |^*^unless uimm[4:0]=0
 
-5+| ‡ carries a dependency from _csr_ to _rd_
+5+| ‡ carries a dependency from _csr_ to _rd_ and _csr_
 |===
 
 .RV64I Base Integer Instruction Set


src/rvwmo.adoc
```diff
@@ -503,19 +503,23 @@ register(s) to destination register(s) as specified
...
 |CSRRS‡ |_rs1_, _csr_ |_rd_ ^*^, _csr_ | |^*^unless _rs1_=`x0`
 
 |CSRRC‡ |_rs1_, _csr_  |_rd_ ^*^, _csr_ | |^*^unless _rs1_=`x0`
 
-5+| ‡ carries a dependency from _rs1_ to _csr_ and from _csr_ to _rd_
+5+| ‡ carries a dependency from _csr_ and _rs1_ to _csr_ and from _csr_ to _rd_
 
 |CSRRWI ‡ |_csr_ ^*^ |_rd_, _csr_  | |^*^unless _rd_=_x0_
 
+5+| ‡ carries a dependency from _csr_ to _rd_ and clears the dependencies of _csr_
```
terpstra: In other places, "clearing" dependencies is never explicitly stated.
Perhaps this should just be "carries a dependency from _csr_ to _rd_" without the "and ..." to match that style.

aswaterman:
It is true that CSRRS[I]/CSRRC[I] (with rs1/uimm != x0) both read and write the CSR, and they carry a dependency from the CSR to the CSR.  It is also true that CSRRW[I] writes the CSR but does not carry a dependency from the CSR to the CSR.

So I'm just tagging @daniellustig to help with the wording.  I'd rather not introduce the new terminology "clears the dependencies", in part because it's not a new concept: e.g. `lui x1, 0` does the same thing to `x1` that `csrrwi x0, sscratch, 0` does to `sscratch`.

I think the way to describe this using our existing terminology is to simply say that CSRRW[I] does not carry a dependency from the CSR to the CSR.  (Then, the fact that the CSR is written implies the clearing of dependencies on past values of the CSR.)  And for CSRRS[I]/CSRRC[I] (with rs1/uimm != x0), we say that the dependency from CSR to CSR is carried.

src/rvwmo.adoc
aswaterman: Didn't see your comment before I wrote mine.  I think we came to similar conclusions.

terpstra:
Yeah. I think the same thing. By adding "from csr to csr" to the set/clear cases, but not the swap case, I think it makes the clearing ... clear.

terpstra:
FYI, @dtzSiFive is the one who noticed this mistake while implementing the RVWMO rules for SiFive's memory model checking tool.

aswaterman:
@terpstra It looks like there is a separate bug: for CSRRS/CSRRC, there's an asterisk `* unless rs1=x0` that is applied to `rd`, indicating that `rd` is not written if `rs1=x0`.  The asterisk should actually be on `csr`.  Can you confirm you agree and fix it in this PR?

src/rvwmo.adoc
```diff
@@ -503,19 +503,23 @@ register(s) to destination register(s) as specified
...
 |CSRRW‡ |_rs1_, _csr_^*^ | _rd_, _csr_ | |^*^unless _rd_=`x0`
 
+5+| ‡ carries a dependency from _rs1_ to _csr_ and from _csr_ to _rd_
+
 |CSRRS‡ |_rs1_, _csr_ |_rd_ ^*^, _csr_ | |^*^unless _rs1_=`x0`
 
 |CSRRC‡ |_rs1_, _csr_  |_rd_ ^*^, _csr_ | |^*^unless _rs1_=`x0`
 
-5+| ‡ carries a dependency from _rs1_ to _csr_ and from _csr_ to _rd_
+5+| ‡ carries a dependency from _csr_ and _rs1_ to _csr_ and from _csr_ to _rd_
```
aswaterman, line 512: Alternatively, it might be clearer to handle the `rs1=x0` caveat here, rather than with the asterisk: "carries a dependency from csr to rd, and, if rs1 != x0, from _csr_ and _rs1_ to _csr_"... WDYT?

terpstra: I agree the asterisk is wrong, and I think it's fine to just move it rather than handle it in the text.

daniellustig:
LGTM, I agree with the earlier comments.  Thanks for catching and fixing!

aswaterman:
Thanks, y'all!

Issue #1966, opened 2025-04-10
Zicsr chapter needs clarification on CSR registers versus CSR instructions/addresses


kasanovic:
Though raised in #1255, the confusion around the distinction between CSR registers and CSR addresses/instructions should be clarified in the Zicsr chapter.

ThinkOpenly:
Could this be expanded a bit to cover the cases where there are apparently 64-bit CSRs that also have "upper half" `*h` CSR addresses? From the discussion in #1255, it seems that there is general agreement that these CSRs are architecturally 64-bit and the `*h` addresses are the means for accessing the upper half of those registers in RV32 mode.

One specific concern is in Chapter 4 "Smstateen/Sstateen Extensions", 4.1 "State Enable Extensions", which says:
> For RV32, the registers listed above are 32-bit, and for the machine-level and hypervisor CSRs there is a corresponding set of high-half CSRs for the upper 32 bits of each register: `mstateen0h`, `mstateen1h`, `mstateen2h`, `mstateen3h`, `hstateen0h`, `hstateen1h`, `hstateen2h`, and `hstateen3h`.

(The "registers listed above" are `mstateen[0-3]`, `stateen[0-3]`, and `hstateen[0-3]`.)

This statement seems self-contradictory. How can you access the "upper 32 bits" of a 32-bit register?

It seems like better text would read:
> For RV32, there is a corresponding set of high-half CSR addresses to access the upper 32 bits of each machine-level and hypervisor CSR listed above: [...]

PR #2066, opened 2025-06-04
fix: mstateen*h and hstateen*h access upper half of 64-bit registers

ThinkOpenly:
For `mstateen*` and `hstateen*`, the spec states:
> For RV32, the registers listed above are 32-bit, and for the machine-level and
> hypervisor CSRs there is a corresponding set of high-half CSRs for the upper 32
> bits of each register

This is self-contradictory. If "the registers listed above are 32-bit", there are no "upper 32 bits" to access.

From the discussion in issue #1255, there is general agreement that CSRs that are defined as 64-bit and have "upper-half" analogues to be used for RV32 are indeed architecturally 64-bit.

Remove the assertion that the 64-bit registers are 32-bit for RV32.

(Also related to issue #1966.)

### DIFF
--- a/src/smstateen.adoc
+++ b/src/smstateen.adoc
@@ -59,9 +59,8 @@ level:
 And if the hypervisor extension is implemented, another set of CSRs is added:
 `hstateen0`, `hstateen1`, `hstateen2`, and `hstateen3`.
 
-For RV32, the registers listed above are 32-bit, and for the machine-level and
-hypervisor CSRs there is a corresponding set of high-half CSRs for the upper 32
-bits of each register:
+For RV32, there are CSR addresses for accessing the upper 32 bits of
+corresponding machine-level and hypervisor CSRs:
 `mstateen0h`, `mstateen1h`, `mstateen2h`, `mstateen3h`,
 `hstateen0h`, `hstateen1h`, `hstateen2h`, and `hstateen3h`.
 


gfavor:
The current arch text is trying to say that the "above" CSRs are 64-bit CSRs under RV64 and are 32-bit CSRs under RV32.  And then trying to say that under RV32 there are additional CSRs for acessing the upper 32 bits of the 64 bits of CSR state.

There may be need for a general clarification about the distinction between "CSRs" as representing state versus "CSR" numbers as ways to access CSR state (which came up a month or so ago and may be in the works).  The current text tries to succintly describe a "CSR" with 64 bits of state that is accessed via the CSR number for that CSR mnemonic, and then points out that in RV32 that CSR number (naturally) only accesses the low 32 bits of CSR state and there is an additional CSR number (and associated mnemonic) for accessing the high 32 bits of CSR state.

Maybe a more accurate PR would add the following bold text:

For RV32, the registers listed above are 32-bit **and only access the lower 32 bits of each register**, and for the machine-level and
hypervisor CSRs there is a corresponding set of high-half CSRs for the upper 32 bits of each register

ThinkOpenly:
> There may be need for a general clarification about the distinction between "CSRs" as representing state versus "CSR" numbers as ways to access CSR state (which came up a month or so ago and may be in the works).

I agree that this is needed. I presume this is covered by #1966, but there does not appear to be any activity on that?

> The current text tries to succintly describe a "CSR" with 64 bits of state that is accessed via the CSR number for that CSR mnemonic, and then points out that in RV32 that CSR number (naturally) only accesses the low 32 bits of CSR state and there is an additional CSR number (and associated mnemonic) for accessing the high 32 bits of CSR state.
> 
> Maybe a more accurate PR would add the following bold text:
> 
> For RV32, the registers listed above are 32-bit **and only access the lower 32 bits of each register**, and for the machine-level and hypervisor CSRs there is a corresponding set of high-half CSRs for the upper 32 bits of each register

For me, at least, that text does not help clarify what entity is 64-bits. It refers to "the _registers_ are 32-bit" and that they "only access the lower 32 bits of each _register_". Since there is no clear distinction between the overloaded term "register", it's still unclear.

Perhaps:
> For RV32, the CSR addresses listed above only access the lower 32 bits of the CSR state. For machine-level and hypervisor CSRs there is a corresponding set of high-half CSR addresses for the upper 32 bits of each CSR state.

This makes the distinction between CSR _addresses_ and CSR _state_. If that looks good, I'll update this PR.

gfavor:
To me, at least, your latest PR suggestion seems good.  Go for it and let's see if Andrew feels the same.

aswaterman:
I'm punting to @jhauser-us, who I believe wrote this text and was probably intentional with his choice of wording.

ThinkOpenly:
New text pushed to this branch, slightly re-worded to:
> For RV32, the CSR addresses listed above only access the lower 32 bits of the CSR state. For machine-level and hypervisor CSRs there is a corresponding set of "high-half" CSR addresses for accessing the upper 32 bits of each CSR state:

jhauser-us:
I agree there's an error.  Registers `mstateen*` and `hstateen*` should be described as always 64-bit.

@ThinkOpenly, your first sentence about *only accessing the lower 32 bits* isn't needed, as this case is already covered by Section 2.7, "Explicit Accesses to CSRs Wider than XLEN".

I believe the original Smstateen draft was written before Section 2.7 was added to the Privileged ISA to simplify these situations.

ThinkOpenly:
Updated text has been pushed here, copied here for convenience:
> For RV32, there are CSR addresses for accessing the upper 32 bits of corresponding machine-level and hypervisor CSRs:

PR #2088, opened 2025-06-18
Clarify compressed instruction hints

Timmmm:
The text around these was a little confusing so I've changed it to be more consistent and fixed a few minor mistakes.

1. There's a lot of text that's like "C.FOO is only valid when dest != 0. Oh but actually you can also have dest=0, then it's a HINT." I changed those so that C.FOO is valid for all dests, but *also* the instances of C.FOO where dest=0 are HINTs. This is more consistent with the `RVC HINT instructions` table which lists the HINTs as being part of the C.FOO instructions, and also explicitly states that "RVC HINTs are encoded as computational instructions that do not modify the architectural state".

2. Removed a lot of redundant constraints from the `RVC Hint Instructions` table. E.g. there's no need to say `rd != x0` for C.ADDI because that's already a condition for an instruction being C.ADDI at all. So the table now only lists the *additional* constraints that make instructions into HINTs.

3. Move the NOP instruction definition closer to `C.ADDI` where it is referenced (it's basically just a specific `C.ADDI` hint).

4. Remove the `64` suffix and `RV32 and RV64 only` from the hint table. `C.SLLI64` isn't an instruction and RV32 and RV64 are the only options currently.

5. Fix a `≠` that should have been `=` for `C.ADD`.

### DIFF
--- a/src/c-st-ext.adoc
+++ b/src/c-st-ext.adoc
@@ -543,18 +543,16 @@ include::images/wavedrom/c-integer-const-gen.edn[]
 
 
 C.LI loads the sign-extended 6-bit immediate, _imm_, into register _rd_.
-C.LI expands into `addi rd, x0, imm`. C.LI is valid only when
-_rd_≠`x0`; the code points with _rd_=`x0` encode HINTs.
+C.LI expands into `addi rd, x0, imm`. C.LI with _rd_=`x0` are HINTs.
 
 C.LUI loads the non-zero 6-bit immediate field into bits 17–12 of the
 destination register, clears the bottom 12 bits, and sign-extends bit 17
 into all higher bits of the destination. C.LUI expands into
 `lui rd, imm`. C.LUI is valid only when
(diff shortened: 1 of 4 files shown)


aswaterman:
Thanks.  Going forward, please separate distinct changes into separate commits to make review and partial incorporation easier.  Each item in the PR message's numbered list could be its own commit.

I've manually merged this PR via #2093.

I agree with the sentiment of (1).  I changed the writing a little, though.

I discarded (2) because, while it is redundant, I couldn't persuade myself that the less redundant version improved clarity.

I discarded (3) as the current organization is analogous to the base ISA description, where NOP is in a separate section, but admittedly this is a judgment call.

I incorporated (4).

I couldn't locate (5) amongst the other changes.  If the bug persists after #2093 is merged, please follow up with another PR.

Timmmm:
Thanks! I'll split them up more in future.

Here's the PR for 5: #2095

PR #2093, opened 2025-06-18
Improve C extension HINT descriptions

aswaterman:
Redux of #2088.

### DIFF
--- a/src/c-st-ext.adoc
+++ b/src/c-st-ext.adoc
@@ -569,10 +569,10 @@ include::images/wavedrom/c-int-reg-immed.edn[]
 
 C.ADDI adds the non-zero sign-extended 6-bit immediate to the value in
 register _rd_ then writes the result to _rd_. C.ADDI expands into
-`addi rd, rd, imm`. C.ADDI is valid only when
-_rd_≠`x0` and _imm_≠0. The code
-points with _rd_=`x0` encode the C.NOP instruction; the remaining code
-points with _imm_=0 encode HINTs.
+`addi rd, rd, imm`.
+The code point with _rd_=`x0` and _imm_=0 encodes the C.NOP instruction.
+The code points with _rd_≠0 and _imm_=0 are HINTs.
+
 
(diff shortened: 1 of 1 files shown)


topperc:
The binutils and llvm assemblers both accept c.slli64/srai64/srli64 in RV32/RV64 to allow the hint code points to be named. Seems like we really should been accepting a shift amount of 0 for c.slli/srai/srli instead. Should we change this in the tools?

Timmmm:
> Seems like we really should been accepting a shift amount of 0 for c.slli/srai/srli instead. Should we change this in the tools?

That makes sense to me and is consistent with some of the other instructions - I only checked `c.nop` and `c.li` but both allow you to encode hints (`c.nop 1`, `c.li x0, 5`). I don't really understand where the `64` came from anyway...

topperc:
> > Seems like we really should been accepting a shift amount of 0 for c.slli/srai/srli instead. Should we change this in the tools?
> 
> That makes sense to me and is consistent with some of the other instructions - I only checked `c.nop` and `c.li` but both allow you to encode hints (`c.nop 1`, `c.li x0, 5`). I don't really understand where the `64` came from anyway...

c.slli64 was an rv128 instruction that uses the c.slli encoding with a shift of 0 to represent a shift by 64. I guess being able to shift by 64 with a 128-bit register was deemed to be an important operation. I don't know why you would want to spell that as `c.slli64` instead of `c.slli rd, rd, 64`.

aswaterman:
> Seems like we really should been accepting a shift amount of 0 for c.slli/srai/srli instead.

I think that's probably right, since slli64 etc. were described as RV128-only instructions.  Of course, leaving support for the old scheme for compatibility is probably the right thing to do.

aswaterman:
> I don't know why you would want to spell that as c.slli64 instead of c.slli rd, rd, 64.

I think it was to draw attention to the fact that 64 was encoded unnaturally.  But you're probably right.

PR #2095, opened 2025-06-19
Fix constraint for assigned C.ADD hints

Timmmm:
The C.ADD hints are split into unassigned ones where rs2≠x2-x5, and assigned ones where rs2=x2-x5, but the constraint was ≠ for both.


aswaterman (approved):
Ah, that's where it was.  Thanks.

PR #2098, opened 2025-06-20
Improve c.nop hint description

jordancarlin:
`c.nop` with a non-zero immediate is valid; it's just a hint. For example, assemblers accept `c.nop 7`.

Follow up to #2088  and #2093.

### DIFF
--- a/src/c-st-ext.adoc
+++ b/src/c-st-ext.adoc
@@ -762,8 +762,8 @@ include::images/wavedrom/c-nop-instr.edn[]
 
 `C.NOP` is a CI-format instruction that does not change any user-visible
 state, except for advancing the `pc` and incrementing any applicable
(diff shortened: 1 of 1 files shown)


aswaterman:
Getting rid of the problematic "valid" language is good, but the rest of the change is problematic because those other code points definitionally do not belong to C.NOP.  They belong to C.ADDI.  I'll make a PR that charts a middle course.

aswaterman:
My mistake; it looks like the spec became internally inconsistent after #2093, which confused me.  I think the right fix is to incorporate your change and also fix the description of C.ADDI to be consistent with it.  I'll take care of it.

aswaterman:
Handled via #2100

jordancarlin:
#2100 Looks great. Thanks!

aswaterman:
You're welcome!

PR #2104, opened 2025-06-26
Update smcdeleg.adoc

bcstrongx:
Combine non-normative and normative explanations of what event selector / counter configuration state is accessible in S-mode.


gfavor:
Looks good.

PR #2106, opened 2025-06-27
Fix vector example appendix as separate from previous.

dtzSiFive:
Presently, what should be Appendix C is included in Appendix B.

Fix this so have Appendices A, B, C, D.


dtzSiFive:
Before:

image


After:

image

aswaterman (approved):
Thanks for including the screenshots.

PR #2111, opened 2025-07-03
Mark GitHub-generated releases as "Intermediate Releases"

aswaterman:
We now have three kinds of releases:

- Draft releases, which are the product of generating a copy of the spec from this repo on one's own machine.  These releases will include the DRAFT watermark and a note on the front page that says "DRAFT---NOT AN OFFICIAL RELEASE".

- Intermediate releases, which GitHub builds automatically when PRs are merged into the `main` branch of this repo.  The intent is that the `main` branch contains only ratified material, but since they are auto-generated, they aren't necessarily carefully vetted.  These releases will omit the DRAFT watermark but will include a note on the front page indicating they're "Intermediate" releases.

- Official releases, which are posted on `riscv.org` from time to time.  These omit the DRAFT watermark, obviously, and include a note on the front page indicating they're "Official" releases.

### DIFF
--- a/.github/workflows/merge-and-release.yml
+++ b/.github/workflows/merge-and-release.yml
@@ -37,7 +37,7 @@ jobs:
         if: steps.pull_container_image.outcome == 'success'
         run: |
           docker run --rm -v ${{ github.workspace }}:/build riscvintl/riscv-docs-base-container-image:latest \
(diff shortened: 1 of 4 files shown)


aswaterman:
Merging after offline discussions with Andrea, Bill, and the IC chairs.

PR #2112, opened 2025-07-03
Fix copy/paste error, correcting hideleg to 'interrupt' rather than 'exception'.

monocasa:
Following both the abbreviation itself, and the naming convention for medeleg/mideleg, hideleg should refer to interrupt delegation, not exception delegation which is already covered by hedeleg.


monocasa:
Didn't originally add a sign-off required per contribution doc, amended and pushed with that signoff added.

PR #2125, opened 2025-07-10
Improve code highlighting

Timmmm:
Always use `rouge` for syntax highlighting and install `asciidoctor-sail` - this means the Sail code gets syntax highlighted.

I also tagged a couple of C snippets so they get highlighted.

Unfortunately rouge does not support highlighting RISC-V assembly (neither does pygments).


aswaterman:
@wmat can you review the effects and merge if it looks good to you?

wmat:
LGTM.  Will merge.

aswaterman:
Thanks, y'all!

PR #2126, opened 2025-07-10
Simplify make command in CI workflows

Timmmm:
`make` already automatically uses Docker and downloads the image. I don't think it is necessary to do it explicitly.


wmat:
Doesn't this remove the use of the Docker container for the build?

Timmmm:
No, the Makefile already automatically uses Docker if it is available - see [the build output](https://github.com/riscv/riscv-isa-manual/actions/runs/16194091183/job/45716291089?pr=2126#step:5:14).

wmat:
Ah OK, I thought I saw that logic removed, my bad.

aswaterman (approved):

PR #2136, opened 2025-07-15
Solidify description of how tval is populated on EBREAK

aswaterman:
We tried to fix this in #601, but the text doesn't quite cover the intended case, because the breakpoint exception does not occur on a fetch, load, or store: it occurs as a result of executing a non-memory instruction.  Fix by listing this case explicitly.


aswaterman:
Approved by JohnH.

PR #2138, opened 2025-07-15
List hardware-error exceptions in relevant tval text

aswaterman:
The behavior of a hardware-error exception is already described elsewhere in the text, but for clarity, we should mirror this subset of information in the mtval CSR description.


aswaterman:
Approved by JohnH.

PR #2141, opened 2025-07-17
Simplify Makefile and remove recursive Make calls

Timmmm:
1. The submodule check is done essentially every time you call Make, and it's very fast so we can just do it directly rather than in a rule.
2. Remove the 3-level recursive `make` to handle Docker, which as far as I can tell was completely unnecessary because it is already handled by the `SKIP_DOCKER` flag on line 40.


wmat:
LGTM

PR #2142, opened 2025-07-17
Fix unterminated example block

Timmmm:
This note was missing its closing `====` which causes rendering issues for all content after this point. For some reason Asciidoctor ignored this mistake until we enabled highlighting for the C source in the note.


wmat:
Thanks for this.

PR #2204, opened 2025-08-12
Treat asciidoctor warnings as errors

### LINKED ISSUE #2202: About one typo in Privilege Specification
Hi,

In RISC-V Privilege Specification PDF version, about Ssnpm, it says:

Image

The "Table 2" should be Table 55. It seems something wrong here.

Please check.

Thanks
Hao Sun

aswaterman:
Broken references keep creeping back into the document.  Use CI to enforce that can't happen.

Strangely, the treat-warnings-as-errors option only seems to work when generating the `epub` variants, but that's good enough for CI purposes.

Resolves #2202

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -109,7 +109,7 @@ OPTIONS := --trace \
            -a revremark='$(RELEASE_DESCRIPTION)' \
            $(XTRA_ADOC_OPTS) \
            -D build \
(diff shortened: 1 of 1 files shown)


aswaterman:
@wmat drawing your attention to this one, since it might cause some friction in the short run.

Issue #1163, opened 2023-11-17
Clarification needed on interrupt delegation


vishmohan:
Trying to reconcile the two statements below - 
What is the behavior when _Current Privilege Mode = S , sstatus.SIE = 1 , sip[i] = 1 and sie[i] = 1_?

_Does the machine enter M mode and then delegate the interrupt to S mode if mideleg[i] = 1?_ 
[or]
_Does the machine stay in S mode as indicated in section 4.1.3 ?_

Thank you.

```
From 3.1.8: Machine Trap Delegation registers: (medeleg and mideleg)
By default, all traps at any privilege level are handled in machine mode, though a machine-mode
handler can redirect traps back to the appropriate level with the MRET instruction 

```
```
From 4.1.3 of privileged spec:
An interrupt i will trap to S-mode if both of the following are true: (a) either the current privilege
mode is S and the SIE bit in the sstatus register is set, or the current privilege mode has less
privilege than S-mode; and (b) bit i is set in both sip and sie.
```

aswaterman:
There isn't actually a contradiction.  Note the "By default" at the beginning of the first sentence you quoted.   Reading on in 3.1.8, it becomes clear that the "non-default" case is delegation.

The answer to your question is given by that (a) clause in 4.1.3.

poemonsense:
I'm having the similar issue here. Thanks in advance if you could help clarify this case @aswaterman 

I assume what this issue concerns is that,

1) According to 3.1.8, this S interrupt (e.g., supervisor timer) is delegated to S-mode ***if the corresponding mideleg bit is set***. This is to say, whether the trap is to M- or S- mode depends on the delegation bit.

2) According to 4.1.3, this S interrupt ***always traps into S-mode regardless of the delegation bit*** when it meets some CSR and priv-mode requirements (a) and (b).

So it seems there is some contradiction here regarding **the dependence on the delegtion bit**.

BTW, I would like to also note:

3) This trap-*to-S-mode* behavior is newly introduced in https://github.com/riscv/riscv-isa-manual/pull/590 and https://github.com/riscv/riscv-isa-manual/pull/701. Before #701, the trap is not explicitly required taken into S-mode (thus by default should depend on the delagtion bit). We are confused about this case because we were jus reading two versions of the spec.

[...]

PR #2214, opened 2025-08-17
Update supervisor.adoc

davidharrishmc:
Nonnormative text clarifying why interrupts to S-mode only occur if mideleg = 1 even though this criteria is not explicit in the rule. See https://github.com/riscv/riscv-isa-manual/issues/1163

### DIFF
--- a/src/supervisor.adoc
+++ b/src/supervisor.adoc
@@ -362,6 +362,13 @@ An interrupt _i_ will trap to S-mode if both of the following are true:
 `sstatus` register is set, or the current privilege mode has less
 privilege than S-mode; and (b) bit _i_ is set in both `sip` and `sie`.
 
+NOTE
+====
+Trapping to S-mode only occurs if bit _i_ of `mideleg` is set. Otherwise,
+bit _i_ of `sip` and `sie` are read-only zero so condition (b) cannot be
+be satisfied.
+====
+
 These conditions for an interrupt trap to occur must be evaluated in a
 bounded amount of time from when an interrupt becomes, or ceases to be,
 pending in `sip`, and must also be evaluated immediately following the


src/supervisor.adoc
```diff
@@ -362,6 +362,13 @@ An interrupt _i_ will trap to S-mode if both of the following are true:
 `sstatus` register is set, or the current privilege mode has less
 privilege than S-mode; and (b) bit _i_ is set in both `sip` and `sie`.
 
+NOTE
+====
+Trapping to S-mode only occurs if bit _i_ of `mideleg` is set. Otherwise,
+bit _i_ of `sip` and `sie` are read-only zero so condition (b) cannot be
+be satisfied.
```
ved-rivos: Not always true. 
1. When mideleg[i] is 1, sip[i] and sie[i] are aliases of mip[i] and mie[i] respectively.
2. When mideleg[i] is 0 and mvien[i] is 1, sip[i] is an alias of mvip[i] and sie[i] is writable. 
3. When mideleg[i] and mvien[i] are both 0, sip[i] and sie[i] are read-only zero.

ved-rivos: Constraining this note to the major interrupts 0--12 as defined by base supervisor-level ISA would make this note hold.

davidharrishmc: Thank you.  Updated.

ved-rivos: This note however breaks the layering of the ISA. S level ISA should be orthogonal to how sip gets set. Otherwise every future  addition like mvien would need to be updated/repeated in multiple places

davidharrishmc:
@ved-rivos trying a different wording.  Could drop "such as AIA" if you feel it is limiting rather than exemplary.

ved-rivos:
Saying “if bit i of sip and/or sie is read-only zero then condition (b) cannot occur” is redundant since this is already implied by the normative register definitions. There are many possible conditions besides `mideleg` (e.g., `mvien`) that can affect `sip`/`sie`. Attempting to enumerate them would require restating portions of the machine ISA, which breaks layering. An implementation may validly support the Supervisor-level ISA without implementing any part of the Machine-level ISA.

davidharrishmc:
Trying a minimal wording.  It still mentions mideleg, but Chapter 12 mentions machine CSRs in many places.  Using the word "may" to accommodate AIA and future extensions.  

@ved-rivos if you have a better way to express this, I'd appreciate some language.  If you think this doesn't belong here, tell me you want it closed.  But I think the reader would find some note here would reduce confusion.

ved-rivos:
@davidharrishmc i don’t see the need for such a note . I would be good with closing this

PR #2215, opened 2025-08-18
Zihintntl compressed hints are predicated on Zca, not C

### LINKED ISSUE #2212: Compressed Zihintntl instructions with Zca extension
Currently, the Zihintntl extension says the following:

> If the C extension is provided, compressed variants of these HINTs are also provided: C.NTL.P1 is encoded as C.ADD x0, x2; C.NTL.PALL is encoded as C.ADD x0, x3; C.NTL.S1 is encoded as C.ADD x0, x4; and C.NTL.ALL is encoded as C.ADD x0, x5.

This specifically calls out the C extension, which seems to imply that the compressed forms of the NTL hints are not provided on an implementation that only implements Zca but not the full C extension. Is this intentional or is it just a result of Zca being added later, and if it is the latter, should this be updated to say "If the Zca extension is provided..."?

aswaterman:
This fixes an editing error that arose because the Zc* extensions were merged after Zihintntl, but Zihintntl was not updated accordingly.

Ordinarily, this kind of change would be incompatible, on the basis that we would be shoehorning instructions into Zca after the fact.  But these are HINTs, and so functionally they must be present on all implementations of Zca, regardless of the implementer's intent to implement Zihintntl. Hence, no HW implementation is invalidated by making this change.

Resolves #2212

### DIFF
--- a/src/a-st-ext.adoc
+++ b/src/a-st-ext.adoc
@@ -272,8 +272,8 @@ instruction. The dynamic code executed between the LR and SC
 instructions can only contain instructions from the base ''I''
 instruction set, excluding loads, stores, backward jumps, taken backward
 branches, JALR, FENCE, and SYSTEM instructions.
-Compressed forms of the aforementioned ''I'' instructions in the Zca and Zcb
-extensions are also permitted.
+Compressed forms of the aforementioned ''I'' instructions in the
+C (hence Zca) and Zcb extensions are also permitted.
 * The code to retry a failing LR/SC sequence can contain backwards jumps
 and/or branches to repeat the LR/SC sequence, but otherwise has the same
 constraint as the code between the LR and SC.
--- a/src/zihintntl.adoc
+++ b/src/zihintntl.adoc
@@ -89,7 +89,7 @@ system should use NTL.P1 to indicate a lack of temporal locality
 exploitable by the L1, or should use NTL.ALL indicate a lack of temporal
 locality exploitable by the L2.
 
-If the Zca extension is provided, compressed variants of these HINTs are
+If the C or Zca extension is provided, compressed variants of these HINTs are
 also provided: C.NTL.P1 is encoded as C.ADD _x0, x2_; C.NTL.PALL is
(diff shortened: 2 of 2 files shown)


gfavor:
My only question is whether it should be left as an "exercise for the reader" to recognize that C support implies Zca support (especially if the reader has to go hunt down and understand what Zca is), and hence the compressed variants of these HINTs are provided if C is implemented.  Are there a few words to include that would avoid this "exercise for the reader)?

aswaterman:
It isn't a direct answer to your question, but we already use Zca in this same fashion in the A chapter, in the context of which instructions are allowed within a constrained LR/SC loop.

If you think it would be better, we could change these cases to "C (or Zca)" or "Zca (or C)".  It seems verbose to me and risks metastasizing, but it is pretty borderline.

What's your preference?

gfavor:
The text in the A chapter says "Compressed forms of the aforementioned ''I'' instructions in the Zca and Zcb extensions are also permitted."  Which doesn't allow the compressed forms of non-Zca C instructions in an LR/SC sequence (or are there no such compressed forms to begin with - an exercise for the writer that I haven't undertaken?).

My mild leaning is towards adding the extra two words (and parentheses) in either form.  If I had to choose, I'd go with "C (or Zca)" since the C extension is what most people are readily familiar with and would be implementing (at least for RVA/RVB processors).

aswaterman:
> The text in the A chapter says "Compressed forms of the aforementioned ''I'' instructions in the Zca and Zcb extensions are also permitted." Which doesn't allow the compressed forms of non-Zca C instructions in an LR/SC sequence (or are there no such compressed forms to begin with - an exercise for the writer that I haven't undertaken?).

It's the latter.  The intersection of the sets compressed-forms-of-I and Zca, and the intersection of the sets compressed-forms-of-I and C, are identical.  This is because the only things in C that aren't in Zca are floating-point loads and stores.  I guess if we go with "C (or Zca)", we relieve the reader of the need to figure that out.

> My mild leaning is towards adding the extra two words (and parentheses) in either form. If I had to choose, I'd go with "C (or Zca)" since the C extension is what most people are readily familiar with and would be implementing (at least for RVA/RVB processors).

SGTM.

PR #2228, opened 2025-08-25
Fix Makefile warning message for submodule checkout

jordancarlin:
The warning uses the `--recurse` flag but the actual command uses the `--recursive` flag.


wmat:
Thanks!

PR #2235, opened 2025-08-27
Antora refactor 2

wmat:
Adds partials directory with appropriate symlinks back to diagram files in src/images.
Updates diagram include in rnmi.adoc.


wmat:
Not required.

PR #2261, opened 2025-09-09
Add Zqinx extension and quad-precision details

Liaoshihua:
Add Zqinx extension and quad-precision details


aswaterman:
Zqinx was never actually proposed in the standards process.  The new policy for this repo is that PRs should be created when extensions have been proposed as standards and are in the stable state, and then they can be merged upon ratification.  So, we should close this PR unless RVIA eventually decides to pursue Zqinx.

PR #2262, opened 2025-09-09
Add support for normative rule definitions to ISA manual

james-ball-qualcomm:
The ISA manual already has some of the normative rule tags in the adoc files (look for "norm:") but now I've added the ability to take those normative rule tags and then create a machine-readable list of normative rules. For each ISA manual adoc file, there will be a corresponding YAML file (in the top-level normative_rules_defs directory) that creates normative rules from the tags. The docs-resources repo has the required Ruby script to take all these input files (adoc and norm rule def files) and create a norm-rules.json output file. The docs-resources also has JSON schema files that define the format of the def files and the output file so you can get live schema checking of the defs YAML files if you use vscode to edit them. There are also documentation files (preliminary) that explain how to add normative rules and our naming convention.

I've already got UDB consuming this norm-rules.json file to generate a test plan for the CSC (called a CTP - Certification Test Plan). I've attached a preliminary version of it if you are curious where these normative rules are getting used.
[MockProcessor-CTP.pdf](https://github.com/user-attachments/files/22224095/MockProcessor-CTP.pdf)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,6 +64,7 @@ ifneq ($(SKIP_DOCKER),true)
         ${DOCKER_BIN} run --rm \
             -v ${PWD}/$@.workdir:/build${DOCKER_VOL_SUFFIX} \
             -v ${PWD}/src:/src:ro${DOCKER_EXTRA_VOL_SUFFIX} \
+            -v ${PWD}/normative_rule_defs:/normative_rule_defs:ro${DOCKER_EXTRA_VOL_SUFFIX} \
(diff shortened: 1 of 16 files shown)


james-ball-qualcomm:
Need to add --preserve-quotes to yamlfmt in the pre-commit config file and then rollback my commit that removed the quotes.

james-ball-qualcomm:
@rpsene How do I get this PR to use the updated pre-commit check for yamlfmt?

james-ball-qualcomm:
Going to make a fresh PR.

PR #2289, opened 2025-09-18
Fixup typo of RL for PPO6 & 7

guoren83:
Current riscv AQ & RL (Zalasr) instructions are RCsc. This typo affects PPO6 and PPO7.


aswaterman:
Thanks for the PR.  I'd like one of the memory-model experts to review this one.

PR #2305, opened 2025-09-30
RVWMO: Fixup litmus_ppoca sequence description

guoren83:
The (b) "fence w,w" is not "executes and retires". Correct it with "executes, guarantee the first hart's write order".


gfavor:
Closing since this question appears to be resolved.

PR #2316, opened 2025-10-03
fix inconsistent use of hyphens in modes

### LINKED ISSUE #2315: Use of hyphenation in "mode" names is inconsistent
```
riscv-isa-manual$ grep -i -E 'machine mode|supervisor mode|hypervisor mode' src/* 2>/dev/null | wc -l
57
riscv-isa-manual$ grep -i -E 'machine-mode|supervisor-mode|hypervisor-mode' * 2>/dev/null | wc -l
55
```

The same problem exists with "<single character> mode", but is a little more difficult to search, so these could be inaccurate:
```
riscv-isa-manual$ grep -E 'M mode|S mode|H mode' * 2>/dev/null | wc -l
16
riscv-isa-manual$ grep -E 'M-mode|S-mode|H-mode' * 2>/dev/null | wc -l
512
```

kbroch-rivosinc:
fixes #2315

### DIFF
--- a/src/f-st-ext.adoc
+++ b/src/f-st-ext.adoc
@@ -208,7 +208,7 @@ intervention on the part of user-level software (unlike the Alpha ISA
 floating-point trap barriers). We believe full hardware handling of
 exceptional cases will become more common, and so wish to avoid
 complicating the user-level ISA to optimize other approaches.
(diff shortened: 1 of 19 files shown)


aswaterman:
I think most of these are fixed, but reopen if we missed any.

PR #2319, opened 2025-10-06
Clarify: PTE A/D bits must be updated in memory after setting

### LINKED ISSUE #2308: Unclear interaction between address translation algorithm and translation cache, and potential infinite loop
Note: all excerpts and section references come from `Version 20250508` (current as of 2025-10-01) of the privileged architecture ISA manual

In the address translation algorithm step 2 is:
> 2. Let `pte` be the value of the PTE at address `a+va.vpn[i]×PTESIZE`

The document also says
> The results of implicit address-translation reads in step 2 may be held in a read-only, incoherent address-translation cache but not shared with other harts. [ ... ] Entries in the address-translation cache may then satisfy subsequent step 2 reads if the ASID associated with the entry matches the ASID loaded in step 0 or if the entry is associated with a global mapping. To ensure that implicit reads observe writes to the same memory locations, an `SFENCE.VMA` instruction must be executed after the writes to flush the relevant cached translations.

Step 9 of the address translation algorithm is:
> 9. If `pte.a=0`, or if the original memory access is a store and `pte.d=0`:
> - If the Svade extension is implemented, stop and raise a page-fault exception corresponding to the original access type.
> - If a store to pte would violate a PMA or PMP check, raise an access-fault exception corresponding to the original access type.
> - Perform the following steps atomically:
>   - Compare `pte` to the value of the PTE at address `a+va.vpn[i]×PTESIZE`.
>   - If the values match, set `pte.a` to 1 and, if the original memory access is a store, also set `pte.d` to 1.
>   - If the comparison fails, return to step 2.

Additionally:
> The address-translation cache cannot be used in step 9; accessed and dirty bits may only be updated in memory directly.

I see a few issues with the current wording of the address translation algorithm and its accompanying information.

### 1
The document is unclear as to which parts of step 9 should not interact with the translation cache. Most notably, whether the variable reference `pte` comes from the cache or if it should be re-read from memory.

I believe that it's meant to come from the cache, as it would otherwise be possible to set A or D bits in pages that would have otherwise failed steps 3-8 and caused faults. But as written it is unclear.

Since `pte` is probably meant to come from the cache, "set `pte.a` to 1" is ambiguous in isolation. As written that seems to imply that it should update the value read from the cache, but that value is not used again. Additionally, the cache is meant to be read-only, so I assume the intended wording is "set the A bit of the pte at `a+va.vpn[i]×PTESIZE` to 1" and similar wording for the D bit case.

"A store to `pte`" is similarly ambiguous, it probably should be written as "a store to `a+va.vpn[i]×PTESIZE`".

### 2
The algorithm does not require the translation cache to be invalidated for any reason during the execution of the algorithm itself. This seems like it could create an infinite loop in the algorithm in a conformant implementation. Consider the following case:

- suppose there exists a page `p` with its leaf pte at address `addr` with the `v`, `r`, and `w` bits set and the `d` bit unset and that this pte has been loaded into the translation cache.
- the pte at `addr` gets updated to set the `x` bit, but `sfence.vma` is not executed to clear the translation cache
- a write to page `p` is executed.
- steps 1-4 of the translation algorithm are executed and eventually the leaf pte at `addr` is loaded from the cache into variable `pte`.
- steps 5-8 pass, a write is permitted to this page.
- the `d` bit in `pte` is clear, and this is a store, so step 9 should be executed.
- assume that the Svade extension is not in use and that a store to `addr` is valid.
- `pte` from the cache is compared with the value at `addr`.
- the value has changed, the `x` bit is now set, so the algorithm should go back to step 2.
- in step 2, the pte at `addr` is read *again from the cache* into `pte`.
- everything through step 8 passes again, because `pte` from the cache is unchanged.
- step 9 fails the comparison again, because `pte` from the cache is unchanged.
- repeat steps 2 through 9 forever

The following small sequence of instructions performs a **valid store** to the page but after the in-memory structure has been updated, but in theory may never finish executing the store.
```
// update the page to be executable
ld a0, (pte)
or a0, PTE_X
sd a0, (pte)

// store some value to the page
sd a2, (p)

// maybe the sfence.vma is later
sfence.vma x0, x0
```

I belive the simplest way to resolve this is to REQUIRE implementations to clear their translation cache for a given `addr` when the jump from step 9 to step 2 happens.

ved-rivos:
closes #2308

### DIFF
--- a/src/supervisor.adoc
+++ b/src/supervisor.adoc
@@ -1577,13 +1577,13 @@ __a__=__pte__.__ppn__×PAGESIZE and go to step 2.
 . If _pte_._a_=0, or if the original memory access is a store and _pte_._d_=0:
 
 * If the Svade extension is implemented, stop and raise a page-fault exception corresponding to the original access type.
-* If a store to _pte_ would violate a PMA or PMP check,
+* If a store to the PTE at address __a__+__va.vpn__[__i__]×PTESIZE would violate a PMA or PMP check,
 raise an access-fault exception corresponding to the original access
 type.
 * Perform the following steps atomically:
 ** Compare _pte_ to the value of the PTE at address __a__+__va.vpn__[__i__]×PTESIZE.
 ** If the values match, set _pte_._a_ to 1 and, if the
-original memory access is a store, also set _pte_._d_ to 1.
+original memory access is a store, also set _pte_._d_ to 1. Then store _pte_ to the PTE at address __a__+__va.vpn__[__i__]×PTESIZE.
 ** If the comparison fails, return to step 2.
 
 . The translation is successful. The translated physical address is


src/supervisor.adoc
```diff
@@ -1577,13 +1577,13 @@ __a__=__pte__.__ppn__×PAGESIZE and go to step 2.
 . If _pte_._a_=0, or if the original memory access is a store and _pte_._d_=0:
 
 * If the Svade extension is implemented, stop and raise a page-fault exception corresponding to the original access type.
-* If a store to _pte_ would violate a PMA or PMP check,
+* If a store to PTE would violate a PMA or PMP check,
```
asquared31415:
```suggestion
* If a store to to the PTE at address __a__+__va.vpn__[__i__]×PTESIZE would violate a PMA or PMP check,
```

to mirror the language below and be more clear

ved-rivos: I think this seems fine. It mirrors the language used in step 2.

ved-rivos: But can remove any ambiguity about which PTE. Updated.

PR #2322, opened 2025-10-07
Enable section links

Timmmm:
This turns the section headings in the generated HTML into links to themselves. It doesn't change the styling at all but it means you can easily right-click->copy-link-address on a heading to get a direct link to it. Previously you could only do that from the TOC which is very long and difficult to navigate.


wmat:
This is fine with me.

PR #2330, opened 2025-10-10
Making titles consistent

wmat:
Updating both titles to have a colon between Manual and Volume and a dash between Volume and book name.

### DIFF
--- a/src/riscv-privileged.adoc
+++ b/src/riscv-privileged.adoc
@@ -1,4 +1,4 @@
-[[manual:priv,RISC-V ISA Manual: Volume II - Privileged Architecture]]
+[[manual:priv,RISC-V ISA Manual, Volume II: Privileged Architecture]]
 = The RISC-V Instruction Set Manual: Volume II: Privileged Architecture
 include::../docs-resources/global-config.adoc[]
 :description: Volume II - Privileged Architecture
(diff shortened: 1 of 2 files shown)


omasanori:
Thanks! Also unifying the level 1 section titles would be nice IMHO.

wmat:
Fixed to use , and :

aswaterman (approved):

src/riscv-privileged.adoc
```diff
@@ -1,4 +1,4 @@
-[[manual:priv,RISC-V ISA Manual Volume II: Privileged Architecture]]
+[[manual:priv,RISC-V ISA Manual, Volume II: Privileged Architecture]]
 = The RISC-V Instruction Set Manual: Volume II: Privileged Architecture
```
aswaterman, line 2: Did you mean to fix this line to be consistent, too? "Manual:" to "Manual,"

src/riscv-unprivileged.adoc
```diff
@@ -1,4 +1,4 @@
-[[manual:unpriv,RISC-V ISA Manual Volume I: Unprivileged Architecture]]
+[[manual:unpriv,RISC-V ISA Manual, Volume I: Unprivileged Architecture]]
 = The RISC-V Instruction Set Manual Volume I: Unprivileged Architecture
```
aswaterman, line 2: Same question here.

aswaterman (approved):
Feel free to merge once my comments are addressed.

src/riscv-privileged.adoc
wmat: Correct.  I was taking your advice from earlier.

src/riscv-unprivileged.adoc
wmat: Yep, just making them the same.

PR #2366, opened 2025-10-30
Mention STCE when describing STIP readability

Timmmm:
This text caused some confusion on the mailing list a while ago because it didn't take STCE into account.


Timmmm:
All good points. This was just on my TODO list after the mailing list discussion, but I'm happy to close it.

> implementation behaves as though the extension is not implemented

In my experience this causes quite a lot of confusion. I wonder if there could be a different word to "implemented" because people *really* want it to mean that the device you are running on has hardware for that extension (even if it is currently disabled). Maybe something like "available"? I'm not volunteering to make that change though!

aswaterman:
@Timmmm FWIW, the ARC tussled with this topic.  After much deliberation, we settled on the parsimonious approach, to avoid admitting yet another architectural concept: an extension that could be turned on but hasn't been.  Our experience has been that the specs are simplified by having a binary universe, rather than a ternary one.

Timmmm:
Yeah I meant keeping the semantics as they are but just using a different word instead of "implemented". Basically s/implemented/available/. Because "implemented" has connotations that are different to what RISC-V uses it to mean.

But... that's a pretty big change and maybe not worth the effort (and I'm not sure "available" is the best word either).

Anyway I'm going to close this.

kasanovic:
I think "available" is actually worse for this particular confusion (think of it as available but not turned on).  One thing that will help the confusion is having the intro clearly spell out that we are specifying the behavior of an abstract execution environment not a piece of hardware.  In that context, if something is implemented it can be used. otherwise not.

gfavor:
I agree with Andrew's comment, including the note that "implementation behaves as though the extension is not implemented" is an established pattern in RISC-V architecture.  It would be bad to not follow the pattern here while following it elsewhere.  And that would still leave the same people struggling to understand the meaning of this pattern in other places it is used.  (I also agree that "available" would be a poor alternative".)

Lastly I'll note that - at least to me - "as though not implemented" seems pretty clear, i.e. in all possible ways the extension would appear to be not implemented - the same as if it was actually not implemented.  So, if anything, I would ask for a concrete example of someone's confusion.

Lastly, if there was any spec text to be added, it would need to be in some relatively generic place (since this issue isn't specific to Sstc).  But again, what would that text need to say instead of saying "as though not implemented"?

PR #2397, opened 2025-11-17
Update PMP rule note in machine.adoc

davidharrishmc:
Clarify PMP rule regarding M-mode access to unlocked regions.

This topic has been the source of a lot of confusion, such as 

https://github.com/riscv-non-isa/riscv-arch-test/issues/740#issuecomment-3534216177

In particular, it is unclear from the spec whether the paragraph above takes priority over the earlier paragraph that states application to M-mode is optional.  I think it is helpful to be explicit.

```
PMP checks are applied to all accesses whose effective privilege mode is S or U, including instruction
fetches and data accesses in S and U mode, and data accesses in M-mode when the MPRV bit in
mstatus is set and the MPP field in mstatus contains S or U. PMP checks are also applied to page-table
accesses for virtual-address translation, for which the effective privilege mode is S. Optionally, PMP
checks may additionally apply to M-mode accesses, in which case the PMP registers themselves are
locked, so that even M-mode software cannot change them until the hart is reset. 
```

### DIFF
--- a/src/machine.adoc
+++ b/src/machine.adoc
@@ -3275,7 +3275,7 @@ addresses.
 
 [NOTE]
 ====
-This rule superseds the earlier rule about PMP checks optionally applying to M-mode. The example above will fail even for a M-mode access to an unlocked region. 
+This rule superseds the earlier rule about PMP checks optionally applying to M-mode. The example above will fail even for a M-mode access to an unlocked region.
 ====
 
 If a PMP entry matches all bytes of an access, then the L, R, W, and X


allenjbaum:
I was going to rewrite paragraphs 1-4 of 3.7.1./3 like this:
PMP priority and matching
PMP checks are made on a memory access basis, not an instruction request basis.
Memory requests are not guaranteed to be atomic unless explicitly required, and may otherwise be broken up into multiple accesses under certain conditions (e.g. misaligned accesses, accesses that span multiple virtual memory pages, accesses that load or store from multiple registers)
 
PMP entries are statically prioritized. The lowest-numbered PMP entry that matches any byte of an access has the highest priority and that determines whether that access succeeds or fails.
For each access:
1.     IF the access crosses a misaligned address granularity boundary, and the access is an instruction fetch,
      or misaligned access exceptions have higher priority than page fault exception
     then the access fails with a misaligned address exception trap of the appropriate type
(NOTE: if the PMP granularity<=10, then this can be performed even before address translation)
(NOTE: for instruction fetches, the granularity is 2bytes if C extension is enabled and 4B bytes otherwise.)
2.     Else if no byte of an entry is contained within any PMP entry and at least one entry is implemented then
    a.     If the effective privilege mode is not  M, then the access fails
    b.     If the effective privilege mode is M, the access succeeds
3.     Else if all accessed bytes aren't contained within the highest-priority/ lowest-numbered containing PMP entry 
      then the access fails
4.     Else if  ePMP is not implemented, then
    a.     If the effective privilege mode=M and the matching entry’s pmpcfg.L=0, then the access succeeds
    b.     Else if the matching entry’s pmpcfg.R, .W, or.X bit is set for the corresponding access type, then the access succeeds, elses it fails
5.     Else the ePMP truth table of section 6.2.1 determines if the access succeeds or fails.

davidharrishmc:
Closing this PR to replace with @allenjbaum text

Issue #2405, opened 2025-11-20
Remove latexmath invocations where suitable due to interference with norm rule extraction and general searching in HTML/PDF ISA manuals


james-ball-qualcomm:
See https://github.com/riscv/docs-resources/issues/126 for a thorough discussion of the problem and the proposed solution.

james-ball-qualcomm:
There are currently 389 instances of latexmath in the ISA manuals.

james-ball-qualcomm:
I'm trying GitHub copilot to make the changes. I just told it "Replace calls to latexmath in AsciiDoc source file with regular characters wherever suitable." and it is chugging away.

james-ball-qualcomm:
@wmat, @aswaterman, @davidharrishmc, @jordancarlin

james-ball-qualcomm:
Then I told it to "Replace <= with {le} and => with {ge}" and then "Replace right arrow with -> and left arrow with <-". That got it down to 303 instances of latexmath. I'll commit that and then keep going.

james-ball-qualcomm:
Now I asked it to "What about remaining uses of latexmath?" and it found more it could safely replace and then showed me examples of those it doesn't think it should replace which I agreed with. Down to 241 remaining calls to latexmath.

```
Complex mathematical expressions (should keep):
Division equation: latexmath:[$\textrm{dividend} = \textrm{divisor} \times \textrm{quotient} + \textrm{remainder}$] in [m-st-ext.adoc](vscode-file://vscode-app/c:/Users/jameball/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)
Log expressions: latexmath:[$\log_2n$] in [machine.adoc](vscode-file://vscode-app/c:/Users/jameball/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)
Typewriter font: latexmath:[${\tt pmpcfg}_i$] and latexmath:[${\tt pmpaddr}_i$] in PMP sections
Formal model notation: latexmath:[$\circ$], latexmath:[$\bullet$], arrows in [mm-formal.adoc](vscode-file://vscode-app/c:/Users/jameball/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)
[...]

PR #2410, opened 2025-11-21
Replaced all latexmath calls with built-in AsciiDoc substitutes.

james-ball-qualcomm:
Done to make normative rule extraction provide better results (i.e., rules without images, just normal characters).

See the extensive details in issue #2405. I used a combination of Copilot AI and hand changes. I reviewed all the Copilot changes and built both PDF and HTML versions and sampled the results to look for any problems.

### DIFF
--- a/src/bfloat16.adoc
+++ b/src/bfloat16.adoc
@@ -181,7 +181,7 @@ instruction will ever produce an sNaN; These are only provided as input data. Op
 an invalid operation exception. Operating on a Quiet NaN usually does not cause an exception.
 
 QNaNs are provided as the result of an operation when it cannot be represented
-as a number or {inf}. For example, performing the square root of -1 will result in a qNaN because
+as a number or infinity. For example, performing the square root of -1 will result in a qNaN because
 there is no real number that can represent the result. NaNs can also be used as inputs.
 
 NaNs include a sign bit, but the bit has no meaning.
@@ -536,7 +536,7 @@ floating-point-to-floating-point conversion instructions.
 
 [NOTE]
 ====
-If the input is normal or {inf}, the BF16 encoded value is shifted
+If the input is normal or infinity, the BF16 encoded value is shifted
 to the left by 16 places and the
 least significant 16 bits are written with 0s.
 
--- a/src/f-st-ext.adoc
+++ b/src/f-st-ext.adoc
@@ -127,7 +127,7 @@ particular, with regard to decoding legal vs. reserved encodings).
 |Rounding Mode |Mnemonic |Meaning
 |000 |RNE |Round to Nearest, ties to Even
 |001 |RTZ |Round towards Zero
-|010 |RDN |Round Down (towards −infinity)
+|010 |RDN |Round Down (towards −{inf})
 |011 |RUP |Round Up (towards +{inf})
 |100 |RMM |Round to Nearest, ties to Max Magnitude
 |101 | |_Reserved for future use._
--- a/src/machine.adoc
+++ b/src/machine.adoc
@@ -1626,7 +1626,7 @@ of this register.
 
 When the CY, IR, or HPM__n__ bit in the `mcountinhibit` register is clear,
 the `mcycle`, `minstret`, or `mhpmcountern` register increments as usual.
-When the CY, IR, or HPM_n_ bit is set, the corresponding counter does
+When the CY, IR, or HPM__n__ bit is set, the corresponding counter does
 not increment.
 
 The `mcycle` CSR may be shared between harts on the same core, in which
@@ -2101,7 +2101,7 @@ include::images/bytefield/mconfigptrreg.edn[]
 
 The pointer alignment in bits must be no smaller than MXLEN:
 i.e., if MXLEN is
-8{times}_n_, then `mconfigptr`[log~2n~-1:0]
+8{times}__n__, then `mconfigptr`[log~2n~-1:0]
 must be zero.
 
 The `mconfigptr` register must be implemented, but it may be zero to indicate the
--- a/src/mm-formal.adoc
+++ b/src/mm-formal.adoc
@@ -920,13 +920,13 @@ An instruction instance _i_ in state Plain(Load_mem(_kind_,
 _address_, _size_, _load_continuation_)) can always initiate the
 corresponding memory load operations. Action:
 
-. Construct the appropriate memory load operations *mlos*:
-* if _address_ is aligned to _size_ then *mlos* is a single
+. Construct the appropriate memory load operations _mlos_:
+* if _address_ is aligned to _size_ then _mlos_ is a single
 memory load operation of _size_ bytes from _address_;
-* otherwise, *mlos* is a set of _size_ memory load
+* otherwise, _mlos_ is a set of _size_ memory load
 operations, each of one byte, from the addresses
 _address_..._address_+_size_−1.
-. set _mem_loads_ of _i_ to *mlos*; and
+. set _mem_loads_ of _i_ to _mlos_; and
 . update the state of _i_ to
 Pending_mem_loads(_load_continuation_).
 [NOTE]
@@ -941,7 +941,7 @@ others.
 
 For a non-AMO load instruction instance _i_ in state
 Pending_mem_loads(_load_continuation_), and a memory load operation
-*mlo* in _i.mem_loads_ that has
+_mlo_ in _i.mem_loads_ that has
 unsatisfied slices, the memory load operation can be partially or
 entirely satisfied by forwarding from unpropagated memory store
 operations by store instruction instances that are program-order-before
@@ -965,19 +965,19 @@ entirely satisfied; and
 . all program-order-previous store-acquire-release instructions are
 finished;
 
-Let *msoss* be the set of all unpropagated memory store
+Let _msoss_ be the set of all unpropagated memory store
 operation slices from non-`sc` store instruction instances that are
 program-order-before _i_ and have already calculated the
 value to be stored, that overlap with the unsatisfied slices of
-*mlo*, and which are not superseded by intervening store
+_mlo_, and which are not superseded by intervening store
 operations or store operations that are read from by an intervening
 load. The last condition requires, for each memory store operation slice
-*msos* in *msoss* from instruction
+_msos_ in _msoss_ from instruction
 _i'_:
 
 * that there is no store instruction program-order-between _i_
 and _i'_ with a memory store operation overlapping
-*msos*; and
+_msos_; and
 * that there is no load instruction program-order-between _i_
 and _i'_ that was satisfied from an overlapping memory store
 operation slice from a different hart.
@@ -985,15 +985,15 @@ operation slice from a different hart.
 Action:
 
 . update _i.mem_loads_ to indicate that
-*mlo* was satisfied by *msoss*; and
+_mlo_ was satisfied by _msoss_; and
 . restart any speculative instructions which have violated coherence as
 a result of this, i.e., for every non-finished instruction
 _i'_ that is a program-order-successor of _i_,
-and every memory load operation *mlo'* of _i'_
-that was satisfied from *msoss'*, if there exists a memory
-store operation slice *msos'* in *msoss'*, and
+and every memory load operation _mlo'_ of _i'_
+that was satisfied from _msoss'_, if there exists a memory
+store operation slice _msos'_ in _msoss'_, and
 an overlapping memory store operation slice from a different memory
-store operation in *msoss*, and *msos'* is not
+store operation in _msoss_, and _msos'_ is not
 from an instruction that is a program-order-successor of
 _i_, restart _i'_ and its _restart-dependents_.
 
@@ -1021,14 +1021,14 @@ Forwarding memory store operations to a memory load might satisfy only
 some slices of the load, leaving other slices unsatisfied.
 
 A program-order-previous store operation that was not available when
-taking the transition above might make *msoss* provisionally
+taking the transition above might make _msoss_ provisionally
 unsound (violating coherence) when it becomes available. That store will
 prevent the load from being finished (see <<finish, Finish instruction>>), and will cause it to
 restart when that store operation is propagated (see <<prop_store, Propagate store operation>>).
 
 A consequence of the transition condition above is that
 store-release-RCsc memory store operations cannot be forwarded to
-load-acquire-RCsc instructions: *msoss* does not include
+load-acquire-RCsc instructions: _msoss_ does not include
 memory store operations from finished stores (as those must be
 propagated memory store operations), and the condition above requires
 all program-order-previous store-releases-RCsc to be finished when the
@@ -1039,11 +1039,11 @@ load is acquire-RCsc.
 
 For an instruction instance _i_ of a non-AMO load
 instruction or an AMO instruction in the context of the <<do_amo, Satisfy, commit and propagate operations of an AMO>> transition,
-any memory load operation *mlo* in
+any memory load operation _mlo_ in
 _i.mem_loads_ that has unsatisfied slices, can be
 satisfied from memory if all the conditions of <sat_by_forwarding, Satisfy memory load operation by forwarding from unpropagated stores>> are satisfied. Action:
-let *msoss* be the memory store operation slices from memory
-covering the unsatisfied slices of *mlo*, and apply the
+let _msoss_ be the memory store operation slices from memory
+covering the unsatisfied slices of _mlo_, and apply the
 action of <<do_amo, Satisfy memory operation by forwarding from unpropagates stores>>.
 [NOTE]
 ====
@@ -1087,14 +1087,14 @@ An instruction instance _i_ in state Plain(Store_ea(_kind_,
 _address_, _size_, _next_state_)) can always announce its pending memory
 store operation footprint. Action:
 
-. construct the appropriate memory store operations *msos*
+. construct the appropriate memory store operations _msos_
 (without the store value):
-* if _address_ is aligned to _size_ then *msos* is a single
+* if _address_ is aligned to _size_ then _msos_ is a single
 memory store operation of _size_ bytes to _address_;
-* otherwise, *msos* is a set of _size_ memory store
+* otherwise, _msos_ is a set of _size_ memory store
 operations, each of one-byte size, to the addresses
 _address_..._address_+_size_−1.
-. set _i.mem_stores_ to *msos*; and
+. set _i.mem_stores_ to _msos_; and
 . update the state of _i_ to Plain(_next_state_).
 [NOTE]
 ====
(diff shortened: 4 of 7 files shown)


james-ball-qualcomm:
BTW, with this PR, there are no more calls to latexmath in the ISA manuals. Does this mean we can remove --require=asciidoctor-mathematical?

src/f-st-ext.adoc
```diff
@@ -127,8 +127,8 @@ particular, with regard to decoding legal vs. reserved encodings).
 |Rounding Mode |Mnemonic |Meaning
 |000 |RNE |Round to Nearest, ties to Even
 |001 |RTZ |Round towards Zero
-|010 |RDN |Round Down (towards latexmath:[$-\infty$])
-|011 |RUP |Round Up (towards latexmath:[$+\infty$])
+|010 |RDN |Round Down (towards −infinity)
```
aswaterman: This one should be `-{inf}` rather than `-infinity`

src/m-st-ext.adoc
```diff
@@ -66,8 +66,7 @@ the sign of a nonzero result equals the sign of the dividend.#
 [NOTE]
 ====
 For both signed and unsigned division, except in the case of overflow, it holds
-that
-latexmath:[$\textrm{dividend} = \textrm{divisor} \times \textrm{quotient} + \textrm{remainder}$].
+that dividend = divisor {times} quotient + remainder.
```
aswaterman, line 69: I see sometimes you are using `×` and other times you are using `{times}`.  Can you see which one looks better and use them consistently, unless there's some good reason to use both?

src/machine.adoc
```diff
@@ -2101,7 +2101,7 @@ include::images/bytefield/mconfigptrreg.edn[]
 
 The pointer alignment in bits must be no smaller than MXLEN:
 i.e., if MXLEN is
-latexmath:[$8\times n$], then `mconfigptr`[latexmath:[$\log_2n$]-1:0]
+8{times}_n_, then `mconfigptr`[log~2n~-1:0]
```
aswaterman: Same comment about `×` vs. `{times}`

src/zfh.adoc
```diff
@@ -19,9 +19,9 @@ intermediate precision. Although this extension provides explicit
 conversion instructions that suffice to implement that pattern, future
 extensions might further accelerate such computation with additional
 instructions that implicitly widen their operands—e.g.,
-halflatexmath:[$\times$]halflatexmath:[$+$]singlelatexmath:[$\rightarrow$]single—or
+half{times}half+single->single—or
```
aswaterman, line 22: Same comment about `×` vs. `{times}`

src/machine.adoc
```diff
@@ -3225,20 +3225,19 @@ If  `pmpaddr~i-1~&#8805;``pmpaddr~i~``` and `pmpcfg~i~.A`=TOR, then PMP entry _i
...
+then bits pmpaddr~i~[G-1:0] read as all zeros. Bits
+pmpaddr~i~[G-1:0] do not affect the TOR address-matching logic.
+Although changing pmpcfg~i~.A[1] affects the value
+read from pmpaddr~i~, it does not affect the
 underlying value stored in that register—in particular,
-latexmath:[${\tt pmpaddr}_i$][G-1] retains its original value when
-latexmath:[${\tt pmpcfg}_i$].A is changed from NAPOT to TOR/OFF then
+pmpaddr~i~[G-1] retains its original value when
+pmpcfg~i~.A is changed from NAPOT to TOR/OFF then
 back to NAPOT.
```
aswaterman, line 3241: This looks fine at a glance, but can you double-check that this looks OK in the rendered version?

src/mm-formal.adoc
```diff
@@ -882,29 +872,29 @@ generate the new system state.
...
 _loc_; and
-. if latexmath:[$i$]’s pseudocode has already written an address to
+. if _i_’s pseudocode has already written an address to
 _pc_, then _loc_ must be that address, otherwise _loc_ is:
 * for a conditional branch, the successor address or the branch target
 address;
 * for a (direct) jump and link instruction (`jal`), the target address;
 * for an indirect jump instruction (`jalr`), any address; and
-* for any other instruction, latexmath:[$i.\textit{program\_loc}+4$].
+* for any other instruction, _i.program_loc_+4.
```
aswaterman, line 886: The underscore inside the italics looks fishy.  You might need to double them up on the outside, e.g. `__i.program_loc__+4`

src/mm-formal.adoc
```diff
@@ -926,18 +916,18 @@ addresses are detected.
...
 memory load operation of _size_ bytes from _address_;
-* otherwise, latexmath:[$mlos$] is a set of _size_ memory load
+* otherwise, *mlos* is a set of _size_ memory load
 operations, each of one byte, from the addresses
-latexmath:[$\textit{address}\ldots\textit{address}+\textit{size}-1$].
-. set _mem_loads_ of latexmath:[$i$] to latexmath:[$mlos$]; and
-. update the state of latexmath:[$i$] to
+_address_..._address_+_size_−1.
+. set _mem_loads_ of _i_ to *mlos*; and
+. update the state of _i_ to
```
aswaterman: Isn't `*mlos*` bold or something?  Shouldn't it be `_mlos_`?

aswaterman: There are several similar ones that follow; please audit all of the uses of `*`

src/riscv-unprivileged.adoc
```diff
@@ -49,6 +49,9 @@ endif::[]
 :ne: &#8800;
 :approx: &#8776;
 :inf: &#8734;
+:circ: &#x2218;
+:bullet: &#x2219;
+:times: &#x00D7;
```
aswaterman, line 54: If we do end up getting rid of `{times}`, remove this

src/riscv-privileged.adoc
```diff
@@ -52,6 +52,9 @@ endif::[]
 :ne: &#8800;
 :approx: &#8776;
 :inf: &#8734;
+:circ: &#x25CB;
+:bullet: &#x2219;
+:times: &#x00D7;
```
aswaterman, line 57: If we do end up getting rid of `{times}`, remove this

src/rv64.adoc
```diff
@@ -171,23 +171,23 @@ defined in this subspace.
...
+|XORI |_rd_=_x0_ |2^17^
 
-|ADDIW |_rd_=_x0_ |latexmath:[$2^{17}$]
+|ADDIW |_rd_=_x0_ |2^17^
 
-|ADD |_rd_=_x0_, _rs1_≠_x0_ |latexmath:[$2^{10}-32$]
+|ADD |_rd_=_x0_, _rs1_≠_x0_ |2^10^−32
 
-|ADD |_rd_=_x0_, _rs1_=_x0_, _rs2_≠_x2_-_x5_| 28
+|ADD |_rd_=_x0_, _rs1_=_x0_, _rs2_≠_x2-x5_ | 28
```
aswaterman: I think this change should be reverted; the minus sign shouldn't be italicized

src/bfloat16.adoc
```diff
@@ -181,7 +181,7 @@ instruction will ever produce an sNaN; These are only provided as input data. Op
 an invalid operation exception. Operating on a Quiet NaN usually does not cause an exception.
 
 QNaNs are provided as the result of an operation when it cannot be represented
-as a number or infinity. For example, performing the square root of -1 will result in a qNaN because
+as a number or {inf}. For example, performing the square root of -1 will result in a qNaN because
```
aswaterman: Undo this one; the intent was for the word infinity to be spelled out

src/bfloat16.adoc
```diff
@@ -536,7 +536,7 @@ floating-point-to-floating-point conversion instructions.
 
 [NOTE]
 ====
-If the input is normal or infinity, the BF16 encoded value is shifted
+If the input is normal or {inf}, the BF16 encoded value is shifted
```
aswaterman: Undo this one; the intent was for the word infinity to be spelled out

src/bfloat16.adoc
```diff
@@ -658,7 +658,7 @@ However, here the SEW-width format is limited to BF16.
...
```
aswaterman, line 661: Undo this one; the intent was for the word infinity to be spelled out

src/zfa.adoc
```diff
@@ -90,13 +90,13 @@ If the Zfh or Zvfh extension is implemented, FLI.H performs the
 analogous operation, but loads a half-precision floating-point value
 into register _rd_. Note that entry 1 (corresponding to the minimum
 positive normal value) has a numerically different value for
-half-precision. Furthermore, since latexmath:[$2^{16}$] is not
+half-precision. Furthermore, since 2^16^ is not
 representable in half-precision floating-point, entry 29 in the table
-instead loads positive infinity—i.e., it is redundant with entry 30.
+instead loads positive {inf}—i.e., it is redundant with entry 30.
```
aswaterman: Undo this one; the intent was for the word infinity to be spelled out

src/zfa.adoc
```diff
@@ -161,7 +161,7 @@ The FCVTMOD.W.D instruction is defined similarly to the FCVT.W.D
 instruction, with the following differences. FCVTMOD.W.D always rounds
 towards zero. Bits 31:0 are taken from the rounded, unbounded two's
 complement result, then sign-extended to XLEN bits and written to
-integer register _rd_. latexmath:[$\pm\infty$] and NaN are converted to
+integer register _rd_. ±{inf} and NaN are converted to
```
aswaterman, line 164: Did you double-check that `±` renders correctly in both HTML and PDF?

src/zfh.adoc
```diff
@@ -19,9 +19,9 @@ intermediate precision. Although this extension provides explicit
...
 implicitly narrow their results—e.g.,
-halflatexmath:[$+$]singlelatexmath:[$\rightarrow$]half.
+half+single->half.
```
aswaterman, line 24: This should be a right arrow rather than the two characters `->`.  You can try using `→` directly, or defining a `{rightarrow}` macro (U+2192)

src/zfh.adoc
aswaterman: Also, same comment about the right arrow as https://github.com/riscv/riscv-isa-manual/pull/2410/files#r2548409115

aswaterman:
> BTW, with this PR, there are no more calls to latexmath in the ISA manuals. Does this mean we can remove --require=asciidoctor-mathematical?

Yeah, we should.

aswaterman:
Careful with the Copilot approach!  There were a bunch of subtle errors, and I'm not confident I caught them all.

src/f-st-ext.adoc
```diff
@@ -270,8 +270,8 @@ rounding mode using the _rm_ field with the encoding shown in
 <<rm>>.
 
 Floating-point minimum-number and maximum-number instructions FMIN.S and FMAX.S write, respectively, the smaller or larger of _rs1_ and _rs2_ to _rd_. For the purposes of these instructions only, the value
-latexmath:[$-0.0$] is considered to be less than the value
-latexmath:[$+0.0$]. If both inputs are NaNs, the result is the canonical NaN. If only one operand is a NaN, the result is the non-NaN operand.
+−0.0 is considered to be less than the value
++0.0. If both inputs are NaNs, the result is the canonical NaN. If only one operand is a NaN, the result is the non-NaN operand.
```
Timmmm, line 274: I would double check these are rendered correctly because `+` has special meaning in Asciidoc.

src/m-st-ext.adoc
```diff
@@ -100,17 +99,17 @@ overflow cannot occur.
...
- - |latexmath:[$-1$] +
- latexmath:[$-2^{L-1}$] +
-  |latexmath:[$x$] +
+Overflow (signed only) |_x_ +
+-2^L-1^ |0 +
+−1 |2^L^-1 +
+ - |_x_ +
+ - |−1 +
+ -2^L-1^ +
+  |_x_ +
```
Timmmm, line 108: Should `x` be italicised? `L` isn't...

src/machine.adoc
Timmmm: `n` only italicised in the first equation here.

james-ball-qualcomm:
[Uploading riscv-unprivileged.html…]()
[Uploading riscv-privileged.html…]()

src/m-st-ext.adoc
james-ball-qualcomm: It's not me. I found all uses of latexmath to get a multiply sign and replaced them with {times}. So, what you are seeing is places in the ISA manual where the Unicode character for multiply was already present. I didn't change those because I'm trying to just remove the calls to latexmath. Is there a policy about using Unicode characters (e.g., ×) vs. its Unicode character value (e.g., 00D7) or is equivalent macro that we define (e.g., {times})? To make such a change is outside the scope of what I'm trying to do here (i.e., remove latexmath calls) so seems like a candidate for a different issue/PR. Since it isn't needed for certification and normative rule extraction (which is a huge effort and taking most of my time these days), I'm reluctant to sign up for other ISA manual cleanup tasks just because we happened to notice them while I'm working on this PR. Sound fair? If you agree, could you please create a new GitHub issue against the ISA manual and ask someone like @wmat to work on it?

src/machine.adoc
james-ball-qualcomm: Yes, someone should clean these up.

src/zfh.adoc
james-ball-qualcomm: Actually, adoc has built-in support for arrows and so I'm just using that facility (no need for a Unicode character). See https://docs.asciidoctor.org/asciidoc/latest/subs/replacements.

Here's a screenshot of the relevant doc:

image

And here's how it looks (the same as when latexmath was used):

image

src/machine.adoc
james-ball-qualcomm: Looks good to me:

image

james-ball-qualcomm: BTW, I really appreciate the time you are taking to review this. I did look at many of the rendered outputs (and found some problems like overline in PDF, apparently a known limitation of AsciiDoctor) but not all rendered outputs.

src/mm-formal.adoc
james-ball-qualcomm: This one I checked in the rendered output for the same reason as you. I actually looks correct.

image

src/mm-formal.adoc
james-ball-qualcomm: I could go either way on this so with your feedback, will just use italics instead of bold.

src/riscv-unprivileged.adoc
james-ball-qualcomm: For another issue/PR ...

src/riscv-privileged.adoc
james-ball-qualcomm: For another issue/PR ...

src/rv64.adoc
james-ball-qualcomm: Agreed. BTW, here's another example of inconsistent use of embedded Unicode characters (≠) since I also found places that actually put the equivalent 8800 in the adoc source (didn't even use the ne macro). So, once again, I think this is a separate cleanup job for another issue/PR.

src/bfloat16.adoc
james-ball-qualcomm: Will do.

src/bfloat16.adoc
james-ball-qualcomm: Done

src/zfa.adoc
james-ball-qualcomm: done

src/zfa.adoc
james-ball-qualcomm: I'll do that now. This change was made by Copilot so its new to me. Looks good.

HTML

image

PDF

image

src/zfh.adoc
james-ball-qualcomm: Duplicate that I've already explained the use of ->

src/f-st-ext.adoc
james-ball-qualcomm: Yes, this one I checked the PDF/HTML and it looks good.

james-ball-qualcomm:
> Careful with the Copilot approach! There were a bunch of subtle errors, and I'm not confident I caught them all.

I did review what Copilot did and to be honest, it probably made less mistakes than I would have. So, I'm actually pretty impressed with what it did.

src/m-st-ext.adoc
james-ball-qualcomm: L is a constant but x is a variable so I assume that is why the original author italicized x but not L. Note that I didn't make any changes to this adoc in this PR so your comment is outside the scope of this PR.

src/machine.adoc
james-ball-qualcomm: Actually, I think this change is an improvement since 8 is a constant (so no italics) but "n" is a variable (so italics). However, I checked the PDF and HTML and by coincidence there just happens to be a problem (see screenshots below). For some reason AsciiDoctor isn't italicizing "n" and is passing through the underscores. I'll have to look into why this is the case unless one of you know what's going on here.

PDF

image

HTML

image

james-ball-qualcomm: I Googled this (in AI mode) and got:

AsciiDoctor is not italicizing the letter "i" because single-underscore markup (_i_) is designed for constrained formatting (whole words or phrases), not for a single character within a word boundary. To italicize a single character (unconstrained formatting), you must use double underscores (__i__).

james-ball-qualcomm: And from the adoc manual:

You can emphasize (aka italicize) a word or phrase by enclosing it in a single pair of underscores (e.g., _word_) (constrained). You can emphasize bounded characters (i.e., characters within a word) by enclosing them in a pair of double underscores (e.g., char__act__ers) (unconstrained).

Example 1. Italic inline formatting
An italic _word_, and an italic _phrase of text_.

Italic c__hara__cter__s__ within a word.

james-ball-qualcomm: I'm going to look for these and fix them.

james-ball-qualcomm: Trying to have Copilot find and fix these (I'll review carefully all its changes). I told it "Look in all AsciiDoc source files for incorrect usage of italics within a word. Need to put two underscores around italicized text if inside a word instead of just one underscore around italicized text."

james-ball-qualcomm: Copilot is finding this italics issue in adoc that I didn't even change in this PR. For example, "When the CY, IR, or HPM_n_ bit is set,". This HPM_n_ has the underscores passed through to PDF/HTML. So, it should be changed to HPM__n__. This isn't really related to the scope of this PR but I think there are cases where Copilot put in the wrong single underscores for this PR so I'll fix it for the whole ISA manual (since Copilot is doing the heavy lifting).

james-ball-qualcomm: Copilot found a few of these problems. Here's its summary:

```
Fixed in [machine.adoc](vscode-file://vscode-app/c:/Users/jameball/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html):

Line 1629: HPM_n_ → HPM__n__ (2 occurrences in lines 1577, 1627, 1629)
Line 2104: 8{times}_n_ → 8{times}__n__
Fixed in [hypervisor.adoc](vscode-file://vscode-app/c:/Users/jameball/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html):

Line 732: HPM_n_ → HPM__n__
```

james-ball-qualcomm:
After making the agreed fixes, I rebuilt the ISA manuals. Here they are for reference:

[riscv-unprivileged.html](https://github.com/user-attachments/files/23681844/riscv-unprivileged.html)
[riscv-unprivileged.pdf](https://github.com/user-attachments/files/23681850/riscv-unprivileged.pdf)
[riscv-privileged.html](https://github.com/user-attachments/files/23681852/riscv-privileged.html)
[riscv-privileged.pdf](https://github.com/user-attachments/files/23681856/riscv-privileged.pdf)

src/m-st-ext.adoc
aswaterman: Gotcha.  We can leave it as-is for now.

src/machine.adoc
aswaterman: Deferring to another PR.

src/zfh.adoc
aswaterman: Oh, cool!

src/riscv-privileged.adoc
aswaterman: Deferring to another PR.

src/bfloat16.adoc
aswaterman: Looks like there were two of these and I didn't catch the other, but I'll handle it myself as part of merging this.

src/machine.adoc
aswaterman: @james-ball-qualcomm yeah, you're welcome!

PR #2413, opened 2025-11-23
Tagged priv-intro

davidharrishmc:
Added one normative rule to priv-intro.  This appears to be the only rule in the chapter, so the chapter can be marked done if the PR is accepted.


davidharrishmc:
Should be made into CSC repo, not riscv repo.

PR #2418, opened 2025-11-25
Use unicode symbols instead of HTML entities

Timmmm:
Instead of HTML entities for the special symbol macros, use their unicode characters directly. This doesn't have any effect on the PDF or HTML output, but it makes the tags JSON output easier to read, and also makes the Asciidoc source easier to read (you don't have to look up any weird hex codes to see what `{circ}` looks like).

### DIFF
--- a/src/riscv-privileged.adoc
+++ b/src/riscv-privileged.adoc
@@ -47,15 +47,9 @@ endif::[]
 :hide-uri-scheme:
 :stem: latexmath
 :footnote:
-:le: ≤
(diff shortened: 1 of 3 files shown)


Timmmm:
Fixed and rebased. Note that it doesn't use exactly the same circle character as 37c336612f4c4b71ff6815c1ce3d07561c05236f chose but I don't think it really matters because I think the only text that uses it is a bit broken - see #2448.

PR #2419, opened 2025-11-25
Fix non-normative description of RV32V_Zdinx

aswaterman:
EEW=64 FP operands are supplied by register pairs.


kasanovic:
Looks good to me.

aswaterman:
Verbally approved by ARC

Issue #2436, opened 2025-12-02
Still 2 calls to latexmath in mm_formal.adoc


james-ball-qualcomm:
In PR #2411 I thought I removed all calls to latexmath. There are still 2 in mm_format.adoc that I should remove:

These 2 can just be replaced with _dep_ to get the equivalent italics that the latexmath provides.

```
. add _reg_name_ to _i.reg_writes_ with
latexmath:[$deps$] and _reg_value_; and
. update the state of _i_ to Plain(_next_state_).

where latexmath:[$deps$] is a pair of the set of all _read_sources_ from
_i.reg_reads_, and a flag that is true iff
_i_ is a load instruction instance that has already been
entirely satisfied.
```

aswaterman:
Resolved by #2437

PR #2437, opened 2025-12-02
Removing the last 2 calls to latexmath

james-ball-qualcomm:
Fixes issue #2436 

In PR #2411 I thought I removed all calls to latexmath. There are still 2 in mm_format.adoc that I should remove:

These 2 can just be replaced with _dep_ to get the equivalent italics that the latexmath provides.

```
. add _reg_name_ to _i.reg_writes_ with
latexmath:[$deps$] and _reg_value_; and
. update the state of _i_ to Plain(_next_state_).

where latexmath:[$deps$] is a pair of the set of all _read_sources_ from
_i.reg_reads_, and a flag that is true iff
_i_ is a load instruction instance that has already been
entirely satisfied.
```

### DIFF
--- a/src/bfloat16.adoc
+++ b/src/bfloat16.adoc
@@ -658,7 +658,7 @@ However, here the SEW-width format is limited to BF16.
 
 [NOTE]
 ====
-If the input is normal or {inf}, the BF16 encoded value is shifted
+If the input is normal or infinity, the BF16 encoded value is shifted
 to the left by 16 places and the
 least significant 16 bits are written with 0s.
 ====


james-ball-qualcomm:
Somehow a change to bfloat16 changing the word "infinity" was changed to "{inf}" which gives us an infinity symbol. Based on the content of the sentence, I'm assuming we want the word infinity and not the symbol so I'll fix that and update this PR.

james-ball-qualcomm:
Okay, only the mm_formal.adoc file is changed by this PR which was the original intent.

james-ball-qualcomm:
@wmat, you should be interested in this PR. After it is merged in, do you want to remove the --request to asciidoctor in our Makefile that supports the latexmath calls?

james-ball-qualcomm:
Adding @rpsene

PR #2440, opened 2025-12-03
Clarify that instruction addresses store to *tval are not pointer masked

ved-rivos:


ved-rivos:
@aswaterman @gfavor

gfavor:
Applying pointer masking to a pointer-masked address sounds rather circular.  It seems like what wants to be said is that the address associated with the exception, whether pointer masked or not, is what is written into stval.  

Or that the address associated with the exception (i.e. the address after any ignore transform has been applied), is what is written into stval.

Or that the address associated with the exception, after any ignore transform has been applied (if applicable), is what is written into stval.

gfavor:
In any case, it seems desirable to add something to the sentence that caused this problem, e.g. "In contrast to software writes, pointer masking, when applicable, is applied for hardware writes to a CSR".

aswaterman:
Ah, should've given you the chance to weigh in before merging.  If you and @ved-rivos agree on updated language, I can review another PR.

gfavor:
I'll let Ved take my comments and various suggestions, and create an updated PR as he sees fit.

ved-rivos:
@gfavor @aswaterman - please see #2442

PR #2441, opened 2025-12-03
Generalize RVWMO rule for atomic memory access

ved-rivos:
@aswaterman @gfavor

### DIFF
--- a/src/rvwmo.adoc
+++ b/src/rvwmo.adoc
@@ -89,7 +89,7 @@ operation:
 
 * An unsuccessful SC instruction does not give rise to any memory operations.
 * Floating-point load and store instructions that access more than XLEN bits
-  (i.e., FLD/FSD in RV32) may each give rise to multiple memory operations.
+  (e.g., FLD/FSD in RV32) may each give rise to multiple memory operations.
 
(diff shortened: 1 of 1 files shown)


src/rvwmo.adoc
```diff
@@ -76,22 +76,24 @@ simultaneously. All memory operations are single-copy atomic: they can
...
-extensions such as *V* (vector) and *P* (SIMD) may give rise to multiple
-memory operations. However, the memory model for these extensions has
-not yet been formalized.
+Among instructions in RV32GC and RV64GC, the following are exceptions to the
+rule that an aligned memory instruction gives rise to exactly one memory
+operation:
+
+* An unsuccessful SC instruction does not give rise to any memory operations.
+* Floating-point load and store instructions that access more than XLEN bits
+  (i.e., FLD/FSD in RV32) may each give rise to multiple memory operations.
```
aswaterman: I'd either change this to "FLD/FSD in RV32 and FLQ/FSQ in RV64", or change the "i.e." back to "e.g.".  I guess I prefer just changing it to "e.g." and keeping it short.

ved-rivos: Updated.

aswaterman (approved):

PR #2442, opened 2025-12-03
Clarify pointer masking is applied only if applicable on HW writes to *tval

ved-rivos:


gfavor:
This LGTM.

Ideally something like the change I suggested to the original problematic sentence would also be made.  But I'm fine if you all feel that this change is sufficient.

ved-rivos:
@gfavor - I revised it a bit - please see the updated version

gfavor:
LGTM.

ved-rivos:
@gfavor @aswaterman please take a final look. Sorry about the churn.

gfavor:
Overall looks good.  Thanks.

PR #2444, opened 2025-12-04
Refactor Makefile

rpsene:
This update refactors the documentation build pipeline to improve performance while preserving full functionality. The new workflow significantly reduces runtime and container overhead.

Key Changes:

- Adopt riscvintl/riscv-docs-base-container-image:small as the default build container.
- Introduce FAST=1 mode for tight development loops (PDF-only builds).
- Add DEBUG=1 switch to enable verbose Asciidoctor flags only when needed.
- Streamline DOCKER_RUN_CMD and ensure compatibility with both Podman and Docker.
- Improve build target selection logic and reduce unnecessary rebuilds.
- Normalize tag/EPUB/HTML/PDF generation rules and unify Asciidoctor options.
- Enhance maintainability with clearer variables, consolidated requires, and predictable paths.

### DIFF
--- a/src/machine.adoc
+++ b/src/machine.adoc
@@ -2358,8 +2358,9 @@ not exist.
 [[sec:mseccfg]]
 ==== Machine Security Configuration (`mseccfg`) Register
 
-`mseccfg` is an optional 64-bit read/write register, formatted as
-shown in <<mseccfg>>, that controls security features.
+`mseccfg` is a 64-bit read/write register, formatted as shown in <<mseccfg>>,
+that controls security features. It exists if any extension that adds a field
+to `mseccfg` is implemented. Otherwise, it is reserved.
 
 [[mseccfg]]
 .Machine security configuration (`mseccfg`) register.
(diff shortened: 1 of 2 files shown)


rpsene:
New container base image for reference: https://github.com/riscv/riscv-docs-base-container-image/blob/main/Dockerfiles/debian

james-ball-qualcomm:
I've been told by Tim Hutt (@Timmmm) that building HTML is much faster than building PDF. So, I've switched to building HTML only for my "fast" build. Given that you are refactoring the Makefile mainly to improve build performance, should we make the FAST=1 mode use HTML instead of PDF?

james-ball-qualcomm:
How much does having a "small" docker image improve build time? Is it worth it to have it?

aswaterman:
> Given that you are refactoring the Makefile mainly to improve build performance, should we make the FAST=1 mode use HTML instead of PDF?

I was going to suggest this, too.

aswaterman:
@rpsene are we moving forward with these changes?

PR #2452, opened 2025-12-04
Add workflow to allow /clarification label via comment

rpsene:
This adds a GitHub Action that listens for issue comments starting with "/clarification" and applies the "Clarification" label, allowing users without triage permissions to categorize issues.

- Adds .github/workflows/label-clarification.yml
- Triggers on issue comments starting with /clarification
- Uses github-script to apply the label safely

### DIFF
--- a/.github/workflows/clarification.yml
+++ b/.github/workflows/clarification.yml
@@ -1,0 +1,23 @@
+---
+name: Label "Clarification" on Comment
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  label_issue:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+    if: startsWith(github.event.comment.body, '/clarification')
+    steps:
+      - name: Add Clarification Label
+        uses: actions/github-script@v6
+        with:
+          script: |
+            github.rest.issues.addLabels({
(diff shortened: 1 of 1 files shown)


james-ball-qualcomm:
So here?

image

rpsene:
> Oh no. Does this mean someone creating an issue can't just select a "clarification" label in the normal place on a GitHub issue page?

It is possible and will be available. I'm just adding one more option.

james-ball-qualcomm:
Just sent you an email about using GitHub issue templates as another way to set a specific label. We could also create a label with each chapter name so people can identify which chapter they are filing the issue on (whether or not it is a clarification).

james-ball-qualcomm:
How would people to know to start the issue with /clarification? That's not obvious to me. My recommendation is to drop this idea in favor of the issue template approach in another PR you created today.

rpsene:
@james-ball-qualcomm we have both approaches, this one here will be handy when an issue should be converted to clarification, for instance.

james-ball-qualcomm:
> @james-ball-qualcomm we have both approaches, this one here will be handy when an issue should be converted to clarification, for instance.

Good point!

PR #2454, opened 2025-12-04
Add RISC-V ISA clarification issue template

rpsene:


aswaterman:
It should be a GitHub permalink rather than any ephemeral thing. We should include instructions on how to obtain the permalink (i.e. click on the line in GitHub, press the Y key, copy the URL).

james-ball-qualcomm:
Yes, a GitHub permalink from a GitHub issue sounds good to me. I had to read up on permalink and got the following Google AI summary so I'll give it a try.

```
A GitHub permalink is a permanent, immutable URL that points to a specific version of a file or a specific line (or range of lines) within a file in a GitHub repository, tied to a particular commit. Unlike standard links to files on a branch, which can become outdated as the branch evolves, a permalink will always refer to the exact content at the time it was created.
How to create a permalink:

    Navigate to the file: Open the desired file within your GitHub repository.
    Select the line(s):
        For a single line: Click on the line number in the gutter to the left of the code.
        For a range of lines: Click on the first line number, then hold down the Shift key and click on the last line number of the desired range. 
    Copy the permalink:
        Click on the three dots (...) that appear next to the selected line(s).
        Select "Copy permalink" from the dropdown menu. 

Alternatively, for the entire file at a specific commit:

    Navigate to the file: in the repository.
    Press the 'y' key: on your keyboard while viewing the file. This will update the URL in your browser's address bar to include the commit hash, making it a permalink to that specific version of the file.
```

rpsene:
@aswaterman @james-ball-qualcomm permalink reference added.

james-ball-qualcomm:
If the text isn't too long (e.g., paragraph or two), would copying the adoc text into the GitHub issue be an acceptable alternative to a permalink?

PR #2480, opened 2025-12-09
Add workflow to tag and untag stalled issues and PRs

rpsene:


aswaterman:
I agree with the concept, but I didn't actually perform a code review.

jordancarlin:
Why not just use the official GitHub Action for tagging stale issues/PRs: https://github.com/actions/stale.

aswaterman:
@rpsene can you consider @jordancarlin's suggestion?

PR #2487, opened 2025-12-10
Fix bit-width for RoundKeyB in vaeskf2.vi Sail pseudocode

GH4169:
In the Sail pseudocode for vaeskf2.vi, the definition of RoundKeyB appears to have an incorrect bit-width specification.
The code currently uses:
`let RoundKeyB[3:0] : bits(32) = get_velem(vd, EGW=128, i); // Previous round key`
However, get_velem(vd, EGW=128, i) returns a 128-bit element group. It should instead be:
`let RoundKeyB[3:0] : bits(128) = get_velem(vd, EGW=128, i); `

Suggestion to address [riscv/riscv-crypto#403](https://github.com/riscv/riscv-crypto/issues/403#issuecomment-3637250308)

### DIFF
--- a/src/vector-crypto.adoc
+++ b/src/vector-crypto.adoc
@@ -1812,7 +1812,7 @@ function clause execute (VAESKF2(rnd, vd, vs2)) = {
 
   foreach (i from eg_start to eg_len-1) {
       let CurrentRoundKey[3:0]  : bits(128)  = get_velem(vs2, EGW=128, i);
-      let RoundKeyB[3:0] : bits(32)  = get_velem(vd, EGW=128, i); // Previous round key
+      let RoundKeyB[3:0] : bits(128)  = get_velem(vd, EGW=128, i); // Previous round key
 
       let w[0] : bits(32) = if (rnd[0]==1) then
         aes_subword_fwd(CurrentRoundKey[3]) XOR RoundKeyB[0];


nibrunieAtSi5:
As I said on the issue, I believe the SAIL pseudo code is intrinsically inconsistent, the variable name implies an array of 4 values, the `bits(32)` was certainly meant to say that each element is 32-bit wide, but as you pointed out `get_velem(vd, EGW=128, i)` can be interpreted as returning a single 128-bit value. 

I think we should consider importing a snippet of the real SAIL implementation rather than trying to fix the pseudo-code.

aswaterman:
@nibrunieAtSi5 please advise as to how we should handle this PR in the meantime.

nibrunieAtSi5:
the current sail version relies on an helper function
```
    let current_round_key = get_velem_quad_vec(vs2_val, i);
    let round_key_b       = get_velem_quad_vec(vd_val, i);
```
https://github.com/riscv/sail-riscv/blob/a882640962cb9e21650633fd5f07a0fcb8c9f390/model/extensions/vector_crypto/zvkned_insts.sail#L305-L306
But the function itself is pretty large and does not seem suitable for insertion in the ISA spec as is.


This PR would at least unify the pseudo-code with `vaeskf1.vi` 
```
      let CurrentRoundKey[3:0]  : bits(128)  = get_velem(vs2, EGW=128, i);
```
https://github.com/riscv/riscv-isa-manual/blob/2efad6f194bf5eb4308fbc42481e2851683d54d4/src/vector-crypto.adoc#vaeskf1vi

and with the first round key for `vaeskf2.vi`
```
      let CurrentRoundKey[3:0]  : bits(128)  = get_velem(vs2, EGW=128, i);
```
https://github.com/riscv/riscv-isa-manual/blob/2efad6f194bf5eb4308fbc42481e2851683d54d4/src/vector-crypto.adoc#vaeskf2vi

They both use `bits(128)` as `type` although a 4x 32-bit word split is expected.

So, upon reflection I believe we should merged it and revisit once a proper process for integrating SAIL code into the ISA spec has been devised.

nibrunieAtSi5:
@wmat, I am curious: is there any plan to streamline / automate link between SAIL description and the ISA specifications ?

wmat:
@nibrunieAtSi5 there's no documented plan in place yet, however, there has been discussion about integrating SAIL into the ISAs at the same time we do the reorganization of them.  The SAIL team has already completed an integration POC, we just need to work out how that applies to the ISAs and ensure we get versioning of SAIL correct.

aswaterman (approved):

PR #2493, opened 2025-12-10
Integrate Smepmp chapter into Priv spec

ved-rivos:
1. Move description of fields introduced in mseccfg to corresponding section in machine mode chapter
2. Move rationale in mseccfg to a rationale chapter.

@aswaterman


ved-rivos:
@aswaterman thanks. done.

PR #2494, opened 2025-12-10
Move Smmpm and Smnpm to machine and Ssnpm to machine/supervisor/hypervisor chapter

ved-rivos:
@aswaterman @martinmaas


aswaterman (approved):
LGTM.

PR #2501, opened 2025-12-12
Move Zicboz/Zicbom *envcfg controls to their corresponding section

ved-rivos:
May need a careful review as these fields dont follow the normal pattern.


src/machine.adoc
```diff
@@ -2220,9 +2220,29 @@ If the Smcdeleg extension is implemented, the CDE (Counter Delegation Enable) bi
 
 The definition of the STCE field is furnished by the Sstc extension.
 
-The definition of the CBZE field is furnished by the Zicboz extension.
-
-The definitions of the CBCFE and CBIE fields are furnished by the Zicbom extension.
+The Zicboz extension adds the `CBZE` (Cache Block Zero instruction enable) field
+to `menvcfg`. When the `CBZE` field is set to 1, it enables execution of the
+cache block zero instruction, `CBO.ZERO`, in privilege modes less than M.
```
aswaterman: Same comment here as on #2502.

src/machine.adoc
```diff
@@ -2220,9 +2220,29 @@ If the Smcdeleg extension is implemented, the CDE (Counter Delegation Enable) bi
...
+Otherwise, the instruction raises an illegal-instruction exception and is not
+HS-qualified.
```
aswaterman: I think we should delete the "and is not HS-qualified" (also for CBCFE and CBIE).  We don't mention this term anywhere else in the M-mode chapter, and in any case, the current definition of HS-qualified covers these cases.

aswaterman: I think the phrasing should actually be "Otherwise, the instruction raises an illegal-instruction exception in modes less privileged than M."  Without the last part, it sounds like CBZE affects M-mode.

src/machine.adoc
```diff
@@ -2220,9 +2220,29 @@ If the Smcdeleg extension is implemented, the CDE (Counter Delegation Enable) bi
...
+
+The Zicbom extension adds the `CBCFE` (Cache Block Clean and Flush instruction
+Enable) field and the `CBIE` (Cache Block Invalidate instruction Enable) WARL
+field to `menvcfg`.
+
```
aswaterman: Merge these paragraphs?

src/machine.adoc
```diff
@@ -2220,9 +2220,29 @@ If the Smcdeleg extension is implemented, the CDE (Counter Delegation Enable) bi
...
+an illegal-instruction exception and are not HS-qualified.
+
+The `CBIE` field controls execution of the cache block invalidate instruction,
+`CBO.INVAL`, in privilege modes less than M. When `CBIE` is set to `00b`, the
+instruction raises an illegal-instruction exception and is not HS-qualified.
+The encoding `10b` is reserved. When `CBIE` is set to `01b` or `11b`, and when
+enabled for execution at privilege modes less than M, it behaves as follows:
+
+* `01b` -- The instruction is executed and performs a flush operation.
+* `11b` -- The instruction is executed and performs an invalidate operation.
```
aswaterman: I realize this would be redundant, but since the composition rules are more complicated than usual, maybe try something like this:

```
* `01b` -- The instruction is executed and performs a flush operation (even if configured by a mode less privileged than M to perform an invalidate operation).
* `11b` -- The instruction is executed and performs an invalidate operation (unless configured by a mode less privileged than M to perform a flush operation).
```

src/hypervisor.adoc
```diff
@@ -685,9 +685,39 @@ If Svadu is not implemented, ADUE is read-only zero.
...
+`CBIE` field is set to `01b` or `11b`, the instruction is enabled for execution;
+otherwise, it raises a virtual-instruction exception.
+
+If `CBO.INVAL` is enabled in HS-mode to perform a flush operation, then when the
+instruction is enabled in VS- or VU-mode it performs a flush operation, even if
+`CBIE` is set to `11b`. Otherwise, when the instruction is enabled for
+execution, its behavior depends on the `CBIE` encoding, as follows:
+
+* `01b` -- The instruction is executed and performs a flush operation.
+* `11b` -- The instruction is executed and performs an invalidate operation.
```
aswaterman: Similar to below, see what you think about this more explicit version:

```
* `01b` -- The instruction is executed and performs a flush operation (even if configured by VS-mode to perform an invalidate operation).
* `11b` -- The instruction is executed and performs an invalidate operation (unless configured by VS-mode to perform a flush operation).
```

src/supervisor.adoc
```diff
@@ -829,10 +829,59 @@ to implement that we consider it worth supporting even if only rarely
...
+Execution of `CBO.ZERO` in U-mode is enabled only if execution of the
+instruction is enabled for use in S-mode and `CBZE` is set to 1; otherwise, an
+illegal-instruction exception is raised.
+
+Execution of `CBO.ZERO` in VU-mode is permitted only if the instruction is
+HS-qualified and execution of the instruction is enabled for use in VS-mode.
+If the instruction is not HS-qualified, an illegal-instruction exception is
+raised. If the instruction is HS-qualified, execution in VU-mode is enabled
+only if the instruction is enabled for execution in VS-mode and `CBZE` is
+set to 1; otherwise, a virtual-instruction exception is raised.
```
aswaterman: The Supervisor chapter shouldn't have any mention of VS or VU modes, or of the term HS-qualified.  I think you can delete this whole paragraph.

ved-rivos: @aswaterman agree that the supervisor chapter should not have a reference to VU or VS mode. But to avoid that we may need to make a copy of senvcfg section in hypervisor chapter. Otherwise we dont have a place to specify behavior for VU mode.

src/machine.adoc
ved-rivos: updated

src/machine.adoc
ved-rivos: updated.

src/machine.adoc
ved-rivos: done

src/machine.adoc
ved-rivos: updated

src/hypervisor.adoc
ved-rivos: updated

src/supervisor.adoc
ved-rivos: Removed it but specfication for VU mode would have to assume its same as U but with the HS-qualified being an additional constraint. That should perhaps be okay.

ved-rivos:
@aswaterman please take a look.

aswaterman (approved):
This version LGTM.  Thanks!

Issue #2462, opened 2025-12-05
Need additional clarification on mstatus.tw documentation due to use of term "lower privilege" without explaining lower than what?


james-ball-qualcomm:
I can't create a normative rule for mstatus.tw that is clear enough for certification needs with the text available in the ISA manual so I'm creating this "clarification" GitHub issue.

In the M-mode chapter, there is this text:
   https://github.com/riscv/riscv-isa-manual/blob/761b5284a2ac0b0e2907048473e3f7dff432a0f8/src/machine.adoc?plain=1#L777C1-L787C19

```
The TW (Timeout Wait) bit is a *WARL* field that supports intercepting the WFI
instruction (see <<wfi>>). When TW=0, the WFI
instruction may execute in lower privilege modes when not prevented for
some other reason. When TW=1, then if WFI is executed in any
less-privileged mode, and it does not complete within an
implementation-specific, bounded time limit, the WFI instruction causes
an illegal-instruction exception. An implementation may have WFI always
raise an illegal-instruction exception in less-privileged modes when
TW=1, even if there are pending globally-disabled interrupts when the
instruction is executed. TW is read-only 0 when there are no modes less
privileged than M.
```

My question is privilege lower than what level? Lower than the level of the hart when the WFI instruction is executed or lower than M-mode (since this is the M-mode chapter and mstatus CSR)? Or lower than some other level definition?

rpsene:
@james-ball-qualcomm done!

ved-rivos:
In any less-privileged mode than M such the M can intercept WFI.

gfavor:
Adding on to Ved's response:  Since this arch text is in a Machine-level ISA chapter and is clearly describing M-mode arch functionality, it should be understood or taken by default that all this functionality is with respect to M-mode.  And hence "any less-privileged mode" refers to any privilege mode below M-mode.

james-ball-qualcomm:
Ved and Greg, thanks for the clarification.

Greg, is there normative text in the ISA manual that states M-mode CSRs can only affect behavior in M-mode? Without this, the normative rule I can extract for mstatus.tw isn't clear enough (IMHO). In fact, aren't there lots of M-mode CSRs that affect behaviors in other privilege modes so there can't be normative text to this effect?

ved-rivos:
> Lower than the level of the hart when the WFI instruction is executed

[...]

PR #2502, opened 2025-12-12
Clarify that mstatus.TW applies to privilege modes less than M

ved-rivos:
Addresses #2462


ved-rivos:
@aswaterman done

james-ball-qualcomm:
Ved, thanks for making this change.

PR #2503, opened 2025-12-12
Move SSEED and USEED description to mseccfg section

ved-rivos:


ved-rivos:
@aswaterman updated.

src/machine.adoc
```diff
@@ -2286,8 +2286,49 @@ shown in <<mseccfg>>, that controls security features.
 .Machine security configuration (`mseccfg`) register.
 include::images/wavedrom/mseccfg.edn[]
 
-The definitions of the SSEED and USEED fields are furnished by the
-entropy-source extension, Zkr.
+The Zkr extension adds the `SSEED` and `USEED` fields to the `mseccfg` CSR to
+control access to the `seed` CSR from privilege modes less than M.
```
aswaterman: Same comment here as on #2502.

ved-rivos: updated.

PR #2504, opened 2025-12-12
Move Sstc text to relevant sections in Machine, Supervisor, and Hypervisor chapters

ved-rivos:


src/hypervisor.adoc
```diff
@@ -546,10 +546,14 @@ read-only in `hip`, and is the logical-OR of these interrupt sources:
...
 VS-level.
 
-Bits `hip`.VSTIP and `hie`.VSTIE are the interrupt-pending and
-interrupt-enable bits for VS-level timer interrupts. VSTIP is read-only
-in `hip`, and is the logical-OR of `hvip`.VSTIP and any other
-platform-specific timer interrupt signal directed to VS-level.
+Bits `hip`.VSTIP and `hie`.VSTIE are the interrupt-pending and interrupt-enable
+bits for VS-level timer interrupts. VSTIP is read-only in `hip`, and is the
+logical-OR of `hvip`.VSTIP and the timer interrupt signal resulting from
+`vstimecmp` (if `vstimecmp` is implemented). The `hip`.VSTIP bit, in response to
```
aswaterman: "if the Sstc extension is implemented" instead of "if vstimecmp is implemented"?  I guess they're equivalent in this context, so whatever you prefer.

src/hypervisor.adoc
```diff
@@ -546,10 +546,14 @@ read-only in `hip`, and is the logical-OR of these interrupt sources:
...
+timer interrupts generated by `vstimecmp`, is set and cleared by writing
+`vstimecmp` with a value that respectively is less than or equal to, or greater
```
aswaterman: I realize this is old text, but it is hard to parse, and we might as well fix it up while we're here.  Maybe try something like:

```
The `hip`.VSTIP bit, in response to timer interrupts generated by `vstimecmp`, is set by writing `vstimecmp` with a value that is less than or equal to the sum of `time` and `htimedelta`, truncated to 64 bits; it is cleared by writing `vstimecmp` with a greater value.
```

src/hypervisor.adoc
ved-rivos: I copied this verbatim from sstc.adoc but I prefer "if the Sstc extension is implemented" . Will update in a bit.

src/supervisor.adoc
```diff
@@ -1050,6 +1054,47 @@ upon `satp` writes reduces the cost of context switches, provided a
...
+[[stimecmp]]
+==== Supervisor Timer (`stimecmp`) Register
+
+The `stimecmp` CSR is a 64-bit register and has 64-bit precision on all RV32 and
+RV64 systems. In RV32 only, accesses to the `stimecmp` CSR access the low 32
+bits, while accesses to the `stimecmph` CSR access the high 32 bits of
+`stimecmp`.
+
+The CSR numbers for `stimecmp` / `stimecmph` are 0x14D / 0x15D (within the
+Supervisor Trap Setup block of CSRs).
```
aswaterman: I don't think we need this, as it is covered by the `priv-csrs` chapter.

src/supervisor.adoc
```diff
@@ -401,10 +401,14 @@ implemented, SEIP is read-only in `sip`, and is set and cleared by the
...
-implemented, STIP is read-only in `sip`, and is set and cleared by the
-execution environment.
+Bits `sip`.STIP and `sie`.STIE are the interrupt-pending and interrupt-enable
+bits for supervisor-level timer interrupts. If implemented, STIP is read-only in
+sip, and is either set and cleared by the execution environment (if `stimecmp`
+is not implemented), or reflects the timer interrupt signal resulting from
+`stimecmp` (if `stimecmp` is implemented). The `sip`.STIP bit, in response to
+timer interrupts generated by `stimecmp`, is set and cleared by writing
+`stimecmp` with a value that respectively is less than or equal to, or greater
+than, the current time value.
```
aswaterman: Same two comments here as up above.

ved-rivos: Updated

src/supervisor.adoc
ved-rivos: yes. Removed

src/hypervisor.adoc
ved-rivos: updated.

src/hypervisor.adoc
ved-rivos: updated.

src/hypervisor.adoc
```diff
@@ -701,7 +705,12 @@ When ADUE=0, the implementation behaves as though Svade were implemented for
 VS-stage address translation.
 If Svadu is not implemented, ADUE is read-only zero.
 
-The definition of the STCE field is furnished by the Sstc extension.
+The Sstc extension adds the `STCE` (STimecmp Enable) bit to `henvcfg` CSR. When
+the Sstc extension is not implemented, `STCE` is read-only zero. The `STCE` bit
+enables `vstimecmp` for VS-mode when set to one. When `STCE` bit is `henvcfg` is
+zero, an attempt to access `stimecmp` (really `vstimecmp`) when V = 1 raises a
```
aswaterman, line 711: `V = 1` -> `V=1` (I realize this is more old text...)

aswaterman: Actually, I'll handle this one myself.

aswaterman (approved):

PR #2506, opened 2025-12-15
Better integrate Zacas and Zabha into ISA manual

ved-rivos:
1. Move Zacas PMA specification into Machine chapter
2. Move Zacas and Zabha rationale to Rationale chapter


aswaterman (approved):
LGTM, thanks.

PR #2512, opened 2025-12-18
Added pre-commit check for norm tagged text changes and ref files to compare against.

james-ball-qualcomm:
Uses new detect_tag_changes.rb script from docs-resources.

### DIFF
--- a/src/bfloat16.adoc
+++ b/src/bfloat16.adoc
@@ -155,7 +155,7 @@ in the trailing fractional bits are called subnormal numbers. Basically, the ide
 a trade off of precision to support _gradual underflow_.
 
 [[norm:bf16_subnorm]]
-All XXX of the BF16 instructions in the extensions defined in this specification (i.e., Zfbfmin, Zvfbfmin,
+All of the BF16 instructions in the extensions defined in this specification (i.e., Zfbfmin, Zvfbfmin,
 and Zvfbfwma) fully support subnormal numbers. That is, instructions are able to accept subnormal values as
 inputs and they can produce subnormal results.
 


james-ball-qualcomm:
@rpsene and @wmat and @aswaterman and @Timmmm, can I get your reviews/feedback?

aswaterman (changes requested):
This is basically an anti-refactoring provision.  I think it should be on the CSC folks to track the spec, rather than forcing this effort onto the spec maintainers.

Timmmm:
> I think it should be on the CSC folks to track the spec, rather than forcing this effort onto the spec maintainers.

I agree, and that's how I was expecting this would be used.

One benefit to this is that it will guarantee that there's an up-to-date `tags.json` easily available for every commit, which would be very useful for some uses. But for that it's probably better if we just change the Github releases CI job to include the tags as well as the PDF/epubs.

james-ball-qualcomm:
I think we are now seeing that the normatively tagged text in the ISA manuals will need to be maintained as changes are made to the ISA manual. We definitely want to encourage improvements to the ISA manual to make it clearer, merge new extensions into chapters they affect (e.g., machine, supervisor, hypervisor), and other things like adding Sail code.

However, here's the situation with the normative rules and the normatively tagged text created for the ISA manual. The CSC has been overseeing the work to go through the ISA manual and capture normative rules which means editing adoc files to add anchors to normative text and creating a separate YAML file to assemble normative rules from the tagged text. This work was done largely by university students who are no longer available to maintain the normatively tagged text in the ISA manuals.

Andrew has proposed that the CSC folks track the spec (so presumably that means maintaining the normatively tagged text). It is true that the CSC has been overseeing the initial effort to extract normative rules from the ISA manuals. This happened because the CSC has an immediate need for normative rules. The normative rules should be benefiting the larger RVI community since it should provide clarity to core designers in their development (RTL coding and verification tests) and the software community to know how a RISC-V core should behave. So, the normative rules are part of the standard and we will be asking TSC TGs creating new extensions to provide the normative rules in their adoc chapters.

How would it work for the CSC to maintain the normatively tagged text in the ISA manual? For example, would getting a review from a CSC member to approve ISA manual PRs be a new requirement? Or would the CSC be expected to notice that the ISA manual was changed in a way that affect the normatively tagged text and then go in a make changes to the ISA manual to accommodate the changes? Either way, who's the resource doing this? The CSC was able to drive the initial effort to extract normative rules but now our resources are working very hard to create test plans and certification tests and aren't available to maintain normatively tagged text. They didn't sign up to maintain these but then of course, neither did the spec maintainers.

So, what do we do going forward?

james-ball-qualcomm:
Forgot to respond to Andrew's statement about this pre-commit check being an "anti-refactoring provision". If you move normatively tagged text around (e.g., moving from one chapter to another) your PR would still pass this pre-commit check. It's just if you delete or modify the normatively tagged text that your PR would fail this pre-commit check. And if you are doing that, the anyone in the RVI community using normative rules including but not limited to the CSC would need to be aware of the deletion/change since it could impact their implementation, test plans, or tests.

The spec authors are free to change any text not normatively tagged and its pretty easy to see what text is tagged as normative in the adoc. And spec authors are free to make changes like formatting (bold, italics, etc.) and whitespace changes since my comparison tool filters out those changes. It isn't currently smart enough to understand that grammer changes like changing "isn't" to "is not" should be ignored. Maybe we could employ AI to provide a smarter comparison but I have no idea how much work that would be to setup and run in a PR pre-commit check and how good it would be but it's an idea. The point is trying to minimize the burden on spec maintainers.

james-ball-qualcomm:
This PR build is failing because it uses the latest version of docs-resources which won't work until changes from PR #2518 is merged. So, if we merge #2518 then we can merge this PR.

Timmmm:
> And if you are doing that, the anyone in the RVI community using normative rules including but not limited to the CSC would need to be aware of the deletion/change since it could impact their implementation, test plans, or tests.

Right but the whole point is that they can now do that easily and automatically by diffing the extracted rules between two spec versions.

For example in the Sail model we would use it like this:

1. Add attributes to functions, bitfield definitions, etc. to link them to a normative rule *and spec version* (either with a version per link, or maybe with a global version).
2. When a new spec is published/ratified (tbh I'm still not sure how that works exactly), we diff the normative rules between our current spec version and the new one.
3. We then manually audit the changes, verify that they don't affect the model (or update it if necessary), and finally update the spec version in model.

james-ball-qualcomm:
@Timmmm, your scenario makes complete sense to me.  The norm rules are in the ISA manual repo so a consumer of the manual using specific builds/releases will expect to be able to diff normative rules. The tool in this PR is used by the ISA manual producers to ensure that the normative rules keep being available as the manual changes (i.e., the rules can still build).

A consumer should only expect to see new rules being added over time. Rule names or their contents shouldn't be changing except in rare cases. We probably should provide something like errata or release notes that list any rules that have changed or have been deleted. We can ask our VTM Group consultants such as Jamie Grimm how normative rule evolution is handled by other standards like WiFi, USB, etc.

Timmmm:
> A consumer should only expect to see new rules being added over time. Rule names or their contents shouldn't be changing except in rare cases.

Sure, but they *do* change occasionally. Sometimes due to new features, e.g. "Encodings other than X are reserved" might change when the reserved encodings are used. But also due to simple rewordings/clarification. I think this is what @aswaterman meant by "anti-refactoring".

So given that they are allowed to change, I think the only real value of adding another step to change them is if people have been *accidentally* changing them... @aswaterman would know if this is actually a problem but I haven't noticed it.

I think something informative, like having CI add a comment to PRs saying "these normative points have changed; ensure this is intended" might be just as effective at preventing accidental changes without requiring people to manually rebuild them.

james-ball-qualcomm:
@Timmmm, let me make sure I understand your proposal. We'd still have the pre-commit check as proposed in this PR do a comparison of the normatively tagged text in the PR vs. what is expected. However, instead of a difference triggering a pre-commit failure, the pre-commit check would add an informative comment to the PR to flag the PR as one that changes normatively tagged text.

Assuming my understanding is correct, what are downstream consumers of the normatively tagged text such as the CSC Test Plan Workgroup and the TSC ACT SIG (owner of certification test suite) supposed to do? To better understand this, let me break down the various cases for changes to the normatively tagged text:

> Case 1. Moving normatively tagged text
> Case 2. Non-normative changes to normatively tagged text
> Case 3. Normative changes to normatively tagged text
> Case 4. Deleting normatively tagged text
 
Case 1 won't be flagged as a pre-commit change since our tooling is insensitive to location (just has a unique adoc anchor name).
 
Case 2 will be flagged as a pre-commit change and the PR author can just update the expected text (located in the new "ref" directory) to avoid a pre-commit failure again.

Case 3 and 4 will be flagged as a pre-commit change but in this case the PR author shouldn't update the expected text to avoid the pre-commit failure. Instead, my recommendation is that downstream consumers of the normatively tagged text be in the loop **before** the PR is merged. Otherwise it creates an undue burden on these consumers to maintain the normative rules.

Timmmm:
> We'd still have the pre-commit check as proposed in this PR do a comparison of the normatively tagged text in the PR vs. what is expected. However, instead of a difference triggering a pre-commit failure, the pre-commit check would add an informative comment to the PR to flag the PR as one that changes normatively tagged text.

I would do it as a Github CI workflow that runs a script rather than a pre-commit check. But otherwise, yes.

> what are downstream consumers of the normatively tagged text such as the CSC Test Plan Workgroup and the TSC ACT SIG (owner of certification test suite) supposed to do?

When they upgrade the version of the spec they use, they can diff the normative text using these tags and ensure that they've handled all of the changes.

> Case 2 will be flagged as a pre-commit change and the PR author can just update the expected text (located in the new "ref" directory) to avoid a pre-commit failure again.

The script wouldn't have "expected" text; it would just build the tags for `master` and for the PR and then diff them.

> Instead, my recommendation is that downstream consumers of the normatively tagged text be in the loop before the PR is merged.

I guess it doesn't hurt to tell them where possible, but note that "downstream consumers" is unbounded. Every company that has a RISC-V chip that they verify is (hopefully) going to be using these tags.

james-ball-qualcomm:
So let me make sure I understand this correctly. All the various downstream consumers of the normative rules from the ISA manual repo would do a git pull to update their local workspace and then build the normative rules and compare them against what they last saw in their last snapshot. Is that the suggestion?

If this is your suggestion, it sounds chaos and a lot of duplicate work for people trying to consume the normative rules. The make target to build them in the latest version probably won't even work because of changes to normatively tagged text as a result of refactoring.

Are you expecting every consumer to look through all the ISA PRs between the latest and their previous snapsot of the manual and then reverse engineer all that and guess at what the normatives rule should be. There is no guarantee that the consumers will actually come up with the same answer so there goes having one set of golden normative rules that all consumers use.

The normative rules need to stay up to date and accurate as the ISA manual changes. That's why I put the files required to generate the normal rules in the same repo as the ISA manual. We also want one golden list of them based on that commit of the ISA manual repo.

Timmmm:
> So let me make sure I understand this correctly. All the various downstream consumers of the normative rules from the ISA manual repo would do a git pull to update their local workspace and then build the normative rules and compare them against what they last saw in their last snapshot. Is that the suggestion?

Yeah exactly (or something similar to that; I think it would make sense to attach the tags JSON to Github Releases of the ISA manual so people don't have to actually build it - it's quite slow).

> The make target to build them in the latest version probably won't even work because of changes to normatively tagged text as a result of refactoring.

If there's anything that should be expected to always work then it should really be checked in CI. I'm not totally sure I've understood what you meant here.

> Are you expecting every consumer to look through all the ISA PRs between the latest and their previous snapsot of the manual and then reverse engineer all that and guess at what the normatives rule should be. 

Honestly, I wasn't really expecting the rules to be a thing - I'm not sure they add much value. But I think if we just check them in CI that's sufficient to make sure they always work?

aswaterman:
> > So let me make sure I understand this correctly. All the various downstream consumers of the normative rules from the ISA manual repo would do a git pull to update their local workspace and then build the normative rules and compare them against what they last saw in their last snapshot. Is that the suggestion?
> 
> Yeah exactly (or something similar to that; I think it would make sense to attach the tags JSON to Github Releases of the ISA manual so people don't have to actually build it - it's quite slow).
> 

I think this is the right answer: publish the new JSON with the github releases, but don't check them against a reference as part of CI.  Folks downstream will chase the spec.

> If this is your suggestion, it sounds chaos and a lot of duplicate work for people trying to consume the normative rules.

I don't agree with the assertion this will inherently cause chaos or duplicated work, but of course it moves the onus of that coordination problem downstream.

It's a good tradeoff for RISC-V overall to shift the burden away from the maintainers of this repository, even if it's locally suboptimal for some parties.

james-ball-qualcomm:
What do you want to have in the ISA manual repo related to normative rules from this list?
1. Edited adoc files tagged with normative text
2. One YAML file per chapter that defines normative rules from normatively tagged text
3. Makefile changes to use a tool created by Tim in docs-resources repo to extract normatively tagged text from adoc into JSON files
4. Makefile changes to use a tool created by me in docs-resources repo to use #2 and #3 above to build JSON and HTML versions of the normative rules.

Timmmm:
I would say 1 and 3 are required. 2 and 4 are probably not strictly necessary but probably a good idea too.

james-ball-qualcomm:
Thanks @Timmmm. So, your suggestion is #1 (adoc files with normatively tagged text) and #3 (tags JSON) would be in the ISA manual repo . Presumably that means #2 (norm rule def YAML files) and the makefile to build the norm rules JSON and HTML would live in a completely different repo (yet to be created). Let's call is the "norm rules repo" for now.

@Timmmm, with your suggestion, does that mean the ISA manual maintainers would maintain the normatively tagged text in the adoc files and some other resource would maintain the "norm rules repo"?

james-ball-qualcomm:
@aswaterman and @Timmmm, any response to my previous comment? I'm working today and tomorrow and on holiday the rest of the week but will be checking work emails and GitHub PR comments.

james-ball-qualcomm:
I'm reading over the comments in this PR and I'm seeing a potential gap in understanding. The tagged text that is in the adoc files is not the normative rules. The rules are defined in the per chapter yaml files and they make references to the normatively tagged text in the adoc files. Often there's a one-to-one mapping from a normative rule to normatively tagged text but there's many cases where it's one to many or many to one mapping.

So if you just compare the normatively take text from one release to another, you're not getting the actual diff of the normative rules. You need to run the tool that I created in the docs-resources repo to take the yaml files and the normatively tagged text as inputs and actually generate normative rules.

This is why I'm recommending running this tool in CI during a PR pre-commit check so we can ensure that generating the normative rules doesn't break.

james-ball-qualcomm:
Another response to a comment from @Timmmm:
   "Honestly, I wasn't really expecting the rules to be a thing - I'm not sure they add much value."

We can't have a successful certification program without normative rules. Given that the BoD has stated that getting a certification program in place in 2026 is a top goal, then the normative rules are very important to RISC-V. Our standards certification consultants from VTM Group have made it clear that a certification program needs to have its tests and test plans linked to normative rules in the standards. This is to ensure that the tests are "correct" and that when a given test fails on a certification customer's implementation, we can clearly defend the test behavior based on the normative rule(s) it links to in the standard. Given that the ISA manual isn't written with normative rules, the CSC could have asked the ISA manual maintainers to create them for us. Instead, we've mustered together a small army of university students, consultants, and RVI members like me to extract them on the ISA manual maintainers behalf.

BTW, for a good example of how standards are written with normative rules, look no further than the [Server SoC Spec](https://github.com/riscv-non-isa/server-soc/blob/main/src/server_soc_requirements.adoc) written by people such as @ved-rivos.

aswaterman:
Including the normative rules within the adoc is logical.  Making sure that the build is not broken is also logical.  Diffing them against a defined ruleset is a bridge too far.  I think this roughly means that @Timmmm and I are in agreement on principles 1 and 3, but I am not in concordance on the others.

james-ball-qualcomm:
> Including the normative rules within the adoc is logical. Making sure that the build is not broken is also logical. Diffing them against a defined ruleset is a bridge too far. I think this roughly means that @Timmmm and I are in agreement on principles 1 and 3, but I am not in concordance on the others.

If you want a way to make sure the build of the normative rules isn't broken, you'll need 2 (norm rule def YAML files) in the ISA manual repo and some convenient way to invoke the tool that builds the normative rules. This is exactly what the other PR (#2518) offers you so I recommend you merge it.

I hear you that you don't want what this PR (#2512) offers which is a pre-commit check that does a diff against a defined ruleset and fails. Would it be acceptable to change this to diff to a warning instead of a failure? If yes, would you want a build of the norm rules (not a diff, just a build) to run during pre-commit and create a failure or warning?

james-ball-qualcomm:
@aswaterman, do you want me to make this pre-commit check a warning instead of a error?

aswaterman:
Warning instead of error is OK with me, as long as the social convention is that the spec authors are allowed to merge PRs that have warnings, and CSC folks will come along and fix the problems later.

james-ball-qualcomm:
I'm working on having a GitHub issue created on a PR if a tag diff is detected. Then these won't fall through the cracks.

aswaterman:
OK, make sure those issues are given a distinct label.

jordancarlin:
Looks like right now this creates an issue if there is a mismatch in the PR. It probably makes more sense for it to just create a comment on the PR and only create an issue if/when the PR is actually merged. Otherwise it's going to create a lot of unnecessary noise.

aswaterman:
Agreed with @jordancarlin.  If the PR comment is written in a searchable way, then we get the benefit without the additional github issue noise.

rpsene:
@aswaterman @james-ball-qualcomm @jordancarlin this is how the issue looks like when opened: https://github.com/rpsene/riscv-isa-manual/issues/5

james-ball-qualcomm:
I agree with everyone that the GitHub issue should only be created when the PR is actually merged and will have a unique GitHub label to be able to find/filter them.

james-ball-qualcomm:
@rpsene, I just checked the GitHub issue you created (https://github.com/rpsene/riscv-isa-manual/issues/5) and it looks really good. One key thing to add is the output of scripts/check-tag-changes.sh to the GitHub issue since it provides a comprehensive report of the changes it detected. It runs quickly so you can either capture the stdout from when it is first invoked when the PR is merged (and just its exit status is used to trigger creating an issue) or just run it again when creating the GitHub issue.

Output looks like this:

```

|===================================|
| Checking unprivileged tag changes |
|===================================|
Reference file: ref/riscv-unprivileged-norm-tags.json
Current file: build/riscv-unprivileged-norm-tags.json
Deleted 1 tag:
  * "norm:fcvt-bf16-s_op": "Narrowing convert FP32 value to a BF16 value. Round according to the RM field."

Modified 1 tag:
  * "norm:bf16_subnorm":
      Reference: "All of the BF16 instructions in the extensions defined in this specification (i.e., Zfbfmin, Zvfbfmi..."
      Current:   "All XXX of the BF16 instructions in the extensions defined in this specification (i.e., Zfbfmin, Zvf..."

No additions to merge into ref/riscv-unprivileged-norm-tags.json
```

james-ball-qualcomm:
@rpsene, note that scripts/check-tag-changes.sh will automatically add any new normatively tagged text to the files in the ref directory. This should happen when the PR is merged. So, how would that work with this solution? It's similar to how things like the standard GitHub YAML format checker updates the YAML files if needed.

james-ball-qualcomm:
@rpsene, can we close this now? I think you created a different PR to handle this and that is closed.

PR #2517, opened 2025-12-18
Explicitly clarify that WFI in U-mode is unaffected by mstatus.TW

ved-rivos:


src/machine.adoc
```diff
@@ -793,10 +793,11 @@ OS, rather than wastefully idling in the current guest.
 ====
 
 When S-mode is implemented, then executing WFI in U-mode causes an
-illegal-instruction exception, unless it completes within an
-implementation-specific, bounded time limit. A future revision of this
-specification might add a feature that allows S-mode to selectively
-permit WFI in U-mode. Such a feature would only be active when TW=0.
+illegal-instruction exception, regardless of the value of TW bit, unless the
```
aswaterman: "the" TW bit

src/machine.adoc
```diff
@@ -793,10 +793,11 @@ OS, rather than wastefully idling in the current guest.
...
+instruction completes within an implementation-specific, bounded time limit.
+A future revision of this specification might add a feature that allows S-mode
+to selectively permit WFI in U-mode. Such a feature would only be active when
+TW=0.
```
aswaterman: While we are here, we should either delete this sentence or make it non-normative, since it has no normative meaning at this time.

src/machine.adoc
ved-rivos: fixed

src/machine.adoc
ved-rivos: I think it makes sense to delete it. Updated.

PR #2524, opened 2025-12-19
Issue #2523 fixes

### LINKED ISSUE #2523: Zbkc is a subset of Zbc, and should say so in its description
Zbkc is zbc minus the clmulr instruction. The Zbkc extension description does not make this clear, nor does Zbc mention the existence of a subset of itself. This had lead to confusion for folks.
I can create a PR with my idea for how to improve this at some point soon.

bgrayson-mips:
This rewrites Zbkc as a subset of Zbc, and refers to that section for the instruction details, to avoid duplicate entries for clmul and clmulh when searching. It also adds a column to the table in Zbc to specify which of the Zbc instructions are part of Zbkc.

Let me know if any further tweaks are desired.

Resolves #2523

### DIFF
--- a/src/b-st-ext.adoc
+++ b/src/b-st-ext.adoc
@@ -750,17 +750,13 @@ Carry-less multiplication is the multiplication in the polynomial ring over GF(2
 
 *clmulr* produces bits 2{times}XLEN−2:XLEN-1 of the 2{times}XLEN carry-less product.
 
-Note that the Zbkc <<zbkc>> extension is a subset of Zbc, and removes the *clmulr* instruction to reduce implementation costs for cryptography-focused implementations.
-
-[%header,cols="^1,^1,^1,4,8"]
+[%header,cols="^1,^1,4,8"]
 |===
 |RV32
 |RV64
-|Part of Zkbc?
 |Mnemonic
 |Instruction
 
-|{check}
 |{check}
 |{check}
 |clmul _rd_, _rs1_, _rs2_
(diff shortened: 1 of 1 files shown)


src/b-st-ext.adoc
```diff
@@ -750,25 +750,31 @@ Carry-less multiplication is the multiplication in the polynomial ring over GF(2
...
 
+|{check}
 |{check}
 |{check}
 |clmulh _rd_, _rs1_, _rs2_
 |<<#insns-clmulh>>
 
 |{check}
 |{check}
+|
```
aswaterman: Let's drop these changes and keep the changes constrained to the description of Zbkc.  The spec scales better when we avoid bidirectional references.

src/b-st-ext.adoc
```diff
@@ -929,29 +935,7 @@ common operations in cryptographic workloads.
...
-|{check}
-|clmul _rd_, _rs1_, _rs2_
-|<<#insns-clmul>>
-
-|{check}
-|{check}
-|clmulh _rd_, _rs1_, _rs2_
-|<<#insns-clmulh>>
-
-|===
```
aswaterman, line 954: Dropping this table is an improvement, even with my above suggestion.  It's redundant, regardless as to how we choose to describe this extension.

topperc:
@bgrayson-mips are you still working on this? I would like to also fix LLVM to make sure `-march=` with Zbc sets the `__riscv_zbkc` preprocessor define for software.

bgrayson-mips:
@topperc thanks for the poke, I got too busy with my day job lately! I had to redo this PR, since commit signature verification has been enabled in the meantime, so track https://github.com/riscv/riscv-isa-manual/pull/2739 for the actual PR.

@aswaterman if you could look at this new PR, I made your suggested changes of NOT adding that column to the table, and making Zbc unchanged, and removed the table from Zkbc.

I'll close this PR.

Issue #2386, opened 2025-11-09
[Spelling] Use of both carry-less and carryless spelling


nibrunieAtSi5:
Both forms exists in the specification with "carry-less" being the predominant and historical form (e.g. in scalar crypto https://github.com/riscv/riscv-isa-manual/blob/98ea4b5a409456ee28749748e1eafa8533c463bd/src/scalar-crypto.adoc#L199 or the B extension) and "carryless" having been introduced with the **vector-crypto** specification. https://github.com/riscv/riscv-isa-manual/blob/98ea4b5a409456ee28749748e1eafa8533c463bd/src/vector-crypto.adoc#L543

I believe we should unify the spelling (easier for indexing, ...). As a non-native speaker, I don't feel knowledgeable enough to say which form should be preferred though.

ved-rivos:
Carry-less would be preferred as it’s a correct compound and aligns with language in NIST specifications and other architectures

nibrunieAtSi5:
> Carry-less would be preferred as it’s a correct compound and aligns with language in NIST specifications and other architectures

Thank you @ved-rivos 

I will enforce this in the upcoming PR for Zvbc32e and Zvkgs (https://github.com/riscv/riscv-isa-manual/pull/2369) and open a different PR to update the other instances of "carryless" to "carry-less".

wmat:
Note that we can enforce this rule using Vale as well. @kersten1

aswaterman:
@nibrunieAtSi5 please submit a PR that rectifies the spelling.

nibrunieAtSi5:
https://github.com/riscv/riscv-isa-manual/pull/2527 should correct all carryless forms in the vector crypto chapter to carry-less.

aswaterman:
Resolved by #2527

PR #2527, opened 2025-12-20
[#2386] unifying carry-less spelling in vector crypto chapter

nibrunieAtSi5:
Vector crypto chapter was using a different spelling, carryless, compared to the usually accepted form, carry-less, which is also the spelling used in the scalar crypto chapter.  This patch unifies the spelling to carry-less.

On the branch, it seems there is no longer any instance of "carry-less"
```
$  git grep -i "carryless" | wc -l                                                                               
0
```


nibrunieAtSi5:
@ved-rivos, whenever you get a change could you take a look at that PR that should solve #2386. (cc @wmat )

PR #2529, opened 2025-12-21
Add Zvzip extension

ved-rivos:
Zvzip standard extension provides instructions for reordering structured data in vector registers. These instruction address usages such as packing and unpacking data structures such as color components of a pixel, real and imaginary components of complex numbers, transposing small matrices, among others.


aswaterman:
Thanks!  As usual, we'll leave this open until ratification.

src/zvzip.adoc
```diff
@@ -0,0 +1,126 @@
...
+constraints, the instruction encoding is reserved.
+
+----
+vzip.vv vd, vs2, vs1, vm   # for i in 0 to 2*VL
+                           #   vd[i] = (i % 2 == 0) ? vs2[i/2] : vs1[i/2]
+----
+
+=== Vector Unzip Even and Unzip Odd Instructions
+
+The vector unzip-even instruction (VNZIPE) extracts the VL number of
```
camel-cdr: `VNZIPE` -> `VUNZIPE`

ved-rivos: Thanks.

nadime15:
Is it intentional that this chapter is not included in the compilation flow? Since it’s not embedded anywhere (e.g., in `/src/riscv-unprivileged.adoc`), it won’t appear in the final artifacts.

ved-rivos:
> Is it intentional that this chapter is not included in the compilation flow? Since it’s not embedded anywhere (e.g., in /src/riscv-unprivileged.adoc), it won’t appear in the final artifacts.

I have included it in the build but it may not be the best place to include the chapter.

src/zvzip.adoc
```diff
@@ -0,0 +1,128 @@
...
+vpairo.vv vd, vs2, vs1, vm  # vd[i] = (i % 2) == 0 ? vs2[i + 1] : vs1[i + 0]
+----
+
+[NOTE]
+====
+VPAIRO may read one element past VL in vs2 if VL is odd. The general policy for
+such cases is to return the value 0 when the index is greater than VLMAX in the
+source vector register group.
+
+The folloing example illustrates use of the vector pair-even and pair-odd to
```
nibrunieAtSi5:
```suggestion
The following example illustrates use of the vector pair-even and pair-odd to
```

src/zvzip.adoc
```diff
@@ -0,0 +1,128 @@
...
+  {bits: 5, name: 'vd'},
+  {bits: 3, name: 'funct3', attr:['OPIVV','OPMVV']},
+  {bits: 5, name: 'vs1'},
+  {bits: 5, name: 'vs2'},
+  {bits: 1, name: 'vm'},
+  {bits: 6, name: '001111', attr:['VPAIRE','VPAIRO']},
+], config:{lanes: 1, hspace:1024}}
+....
+
+The destination register cannot overlap the source registers and, if masked,
```
nibrunieAtSi5: to be consistent across the document:
```suggestion
The destination vector register group cannot overlap the source registers and, if masked,
```

src/zvzip.adoc
ved-rivos: Thanks.

src/zvzip.adoc
ved-rivos: Thanks.

nibrunie (commented):
Although it seems  all those instructions support optional masking, there are very few information about the masking behavior (in particular which index of mask should be used for each result element in `vzip.vv`). I am wondering if it could be integrated in the pseudo code or described somewhat.

src/zvzip.adoc
```diff
@@ -0,0 +1,128 @@
...
+vpaire.vv vd, vs2, vs1, vm  # vd[i] = (i % 2) == 0 ? vs2[i + 0] : vs1[i - 1]
+vpairo.vv vd, vs2, vs1, vm  # vd[i] = (i % 2) == 0 ? vs2[i + 1] : vs1[i + 0]
+----
+
+[NOTE]
+====
+VPAIRO may read one element past VL in vs2 if VL is odd. The general policy for
+such cases is to return the value 0 when the index is greater than VLMAX in the
+source vector register group.
+
```
nibrunie, line 360: the Note paragraph and the example appear grouped in github asciidoc rendering

image

If this affect the standard rendering flow, it might be good to find a way to split them (I don't believe we want the example to appear in the note).

Furthermore, I believe that notes are generally non-normative and this note is rather uncommitted, "The general policy", would'nt it be clearer to say whether this is the only accepted behavior (preferred) of if there is leniency ?

ved-rivos: The first paragraph was pulled into the note but was not intended to be part of the note. It was to be normative text.

ved-rivos:
> Although it seems all those instructions support optional masking, there are very few information about the masking behavior (in particular which index of mask should be used for each result element in vzip.vv). I am wondering if it could be integrated in the pseudo code or described somewhat.

@nibrunie The pseudocode is updated to include the masking behavior. Please take a look.

nibrunieAtSi5:
> > Although it seems all those instructions support optional masking, there are very few information about the masking behavior (in particular which index of mask should be used for each result element in vzip.vv). I am wondering if it could be integrated in the pseudo code or described somewhat.
> 
> @nibrunie The pseudocode is updated to include the masking behavior. Please take a look.

Looks good to me, thank you for making the changes @ved-rivos.

nibrunieAtSi5 (approved):

src/zvzip.adoc
```diff
@@ -0,0 +1,379 @@
+== "Zvzip" Extension for Reordering Structured Data, Version 0.1
+
+This chapter describes the Zvzip standard extension for reordering structured
+data in vector registers. These instruction address usages such as packing and
```
pmundkur:
```suggestion
data in vector registers. These instructions address usages such as packing and
```

src/zvzip.adoc
```diff
@@ -0,0 +1,379 @@
...
+                                                                               +
+The destination vector register group cannot overlap the source vector register
+groups and, if masked, cannot overlap the mask register. If the overlap
+violates these constraints, the instruction encoding is reserved.              +
+                                                                               +
+Prestart, inactive, and tail element handling follow the standard vector rules.
+                                                                               +
+VPAIRO may read one element past `VL` in `vs2` when `VL` is odd. If an element
+index is greater than or equal to VLMAX in the source vector register group,
+the value 0 is returned for that element.
```
pmundkur: I'm wondering why the limit is VLMAX and not VL: i.e., why not 
```
If an element index is greater than or equal to VL in the ...
```
Wouldn't any access beyond VL be out-of-bounds?

ved-rivos: As noted, the instruction is allowed to access one element beyond VL. However, if that index is VLMAX or beyond - i.e. reading beyond the bounds of hardware - then the behavior is to return 0. This follows behavior for such instructions in vector extension such as vrgather.

src/zvzip.adoc
ved-rivos: Thanks.

src/zvzip.adoc
pmundkur: I guess I'm confused: Accessing one element beyond VL happens when VL is odd, which means that that out-of-bounds access will be at an even boundary.  Since VLMAX is even, shouldn't that out-of-bounds (with respect to VL) still always be within bounds for an always even VLMAX?

nibrunieAtSi5: isn't it possible for VLMAX to be =1  ? (e.g. LMUL=1, VLEN=64, SEW=64)

In that case `vpairo.vv` with `vl`=1 would access past VLMAX from `vs2`

pmundkur: That makes sense, thanks!

nibrunieAtSi5: You are welcome

nadime15:
@ved-rivos I guess the extension depends on Zve32x, right? If so, I think this should be mentioned.

ved-rivos:
> I guess the extension depends on Zve32x, right? If so, I think this should be mentioned.

The extension depends on the V Extension for Application Processors or the Zve32x Extension for Embedded Processors. I will mention it.

nadime15:
Zve32x is a subset of V, so mentioning Zve32x should be enough.

nibrunieAtSi5:
> Zve32x is a subset of V, so mentioning Zve32x should be enough.

Agreed with @nadime15 , V implies Zve32x, so mentioning Zve32x should be enough (Zve64x could be added to extend to SEW=64 case).
No mention of **v** should be required IMHO (we have done that in the past for other extensions and I think it was a mistake).

camel-cdr:
Would it be appropriate to add notes to each instruction for how they can be implemented by software in base RVV?

Because I suspect people will search the spec for things like "zip" or "transpose", when needing those things in code, and there are relatively efficient patterns for all of them (much better than vrgather), which are non-trivial to come up with:

```
vzip.vv vd, vs1, vs2 -> vwaddu.vv vd, vs1, vs2; vwmaccu.vx vd, -1, vs2

vunzipe.v vd, vs1 -> vnsrl.vx vd, vs1, 0
vunzipo.v vd, vs1 -> vnsrl.vx vd, vs1, \SEW/8

vpaire.v vd, vs1, vd -> vslide1up.vx   vd, vs1, 0, vm=odd-bits
vpairo.v vd, vs1, vd -> vslide1down.vx vd, vs1, 0, vm=even-bits
```

nibrunieAtSi5:
> Would it be appropriate to add notes to each instruction for how they can be implemented by software in base RVV?
> 
> Because I suspect people will search the spec for things like "zip" or "transpose", when needing those things in code, and there are relatively efficient patterns for all of them (much better than vrgather), which are non-trivial to come up with:
> 
> ```
> vzip.vv vd, vs1, vs2 -> vwaddu.vv vd, vs1, vs2; vwmaccu.vx vd, -1, vs2
> 
> vunzipe.v vd, vs1 -> vnsrl.vx vd, vs1, 0
> vunzipo.v vd, vs1 -> vnsrl.vx vd, vs1, \SEW/8
> 
> vpaire.v vd, vs1, vd -> vslide1up.vx   vd, vs1, 0, vm=odd-bits
> vpairo.v vd, vs1, vd -> vslide1down.vx vd, vs1, 0, vm=even-bits
> ```

Does the first pattern work for any SEW (including ELEN) ?

camel-cdr:
The unzip/zip pattern doesn't work for SEW=ELEN

Xu-Dsus4:
Hi all, I would like to clarify whether there should be a constraint of "if masked, cannot overlap the mask register" for all instructions in the zvzip extension. Currently, the specification only mentions that this check needs to be performed for vpairo/vpaire. Thanks.

aswaterman:
@Xu-Dsus4 All vector instructions have a standard set of overlap rules, unless otherwise relaxed.  Those rules include the one you mention.  The spec authors can comment, but my guess is that the vpairo/vpaire mask-overlap rules are redundant.

ptomsich:
After taking this over from Ved and a set of review cycles on our end, the updated draft is in https://github.com/riscv/riscv-isa-manual/pull/3233 (to build the version for Internal Review).

PR #2538, opened 2025-12-24
Fix bitmanip extensions integration

### LINKED ISSUE #2019: Confusing content of bitmanip extensions
This paragraph confuses me:

https://github.com/riscv/riscv-isa-manual/blob/584390d9d872835b1bc45ab9ea946b56db3ad485/src/b-st-ext.adoc?plain=1#L34

- Why does it mention "Public Review" if these extensions are long ratified?
- It gives an impression that Zbc is part of B. According to https://github.com/riscv/riscv-isa-manual/blob/584390d9d872835b1bc45ab9ea946b56db3ad485/src/b-st-ext.adoc?plain=1#L4 it is not.

Then comes a table. Why does it include Zbc, but not Zbkb, Zbkc, Zbkx ?

Last thing: Zbkb, Zbkc, Zbkx instructions are also specified in [scalar-crypto.adoc](https://github.com/riscv/riscv-isa-manual/blob/584390d9d872835b1bc45ab9ea946b56db3ad485/src/scalar-crypto.adoc). Is this intended? There are differences, for instance:
- `xnor` is Zbb 0.93 in `b-st-ext.adoc` but Zbb 1.0.0 in `scalar-crypto.adoc`
- `xperm` shows 7 low opcode bits in `b-st-ext.adoc` but 2+5 in `scalar-crypto.adoc`

ved-rivos:
Resolves #2019

### DIFF
--- a/src/b-st-ext.adoc
+++ b/src/b-st-ext.adoc
@@ -1,448 +1,100 @@
 [[bits]]
-== "B" Extension for Bit Manipulation, Version 1.0.0
+== Bit Manipulation Extensions
 
-The B standard extension comprises instructions provided by the Zba, Zbb, and
-Zbs extensions.
-
-[[preface]]
-=== Zb* Overview
-
-The bit-manipulation (bitmanip) extension collection is comprised of several component extensions to the base RISC-V architecture that are intended to provide some combination of code size reduction, performance improvement, and energy reduction.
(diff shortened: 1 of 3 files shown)


aswaterman (approved):
Thanks @ved-rivos, this is an improvement.  I think there's still work to be done, but that will be at least partially subsumed by the one-page-per-instruction appendices.

PR #2567, opened 2026-01-14
Fix 'Reserved for counter-overflow interrupt' text

Timmmm:
In the `mcause` and `scause` this is no longer listed as reserved - looks like it was just missed in they hypervisor chapter.


aswaterman (approved):
Indeed, thanks @Timmmm.

PR #2568, opened 2026-01-15
Clarify counter/timer behavior in critical-error state

ved-rivos:


aswaterman (approved):
LGTM.

allenjbaum:
Does this imply that the instret counter is unspecified? (or that instret
is a timer/counter under this definition.
Because if you're halting, no further instructions are executed so it
should be frozen, not unspecified.
That sounds pretty useful to narrow down what went wrong.

ved-rivos:
The previous rule states that the hart stops execution and the pc is not updated. So the instret is frozen by that rule.

allenjbaum:
Yes, but it also says that counter/timer values are unspecified. That's a
contradiction.

ved-rivos:
whether the counters and timers update in critical-error state is unspecified by the ISA since there is no architectural mechanism to examine any state in the hart once it has entered critical-error state. The set of actions that occur in critical-error state are  platform specific.

allenjbaum:
Ah, correct, This doesn't prohibit JTAG from examining what is going on and
reading the instret counter, but JTAG is platform specific.
Thanks

PR #2576, opened 2026-01-17
[Zvdot4a8i] Extension for vector dot product (4-element, 8-bit integer) Version 0.1

nibrunieAtSi5:
Specification of a new **vector** ISA extension for constant size (4D) vector dot product of integers (for now limited to 8-bit values).

This PR imports changes made by @kdockser and others in https://github.com/riscv/riscv-dot-product from the standalone repo to **riscv-isa-manual**. It also implements ARC's feedback.

- Tracking ticket: https://riscv.atlassian.net/browse/RVS-1971
- **riscv-opcodes** associated change: https://github.com/riscv/riscv-opcodes/pull/402
- **riscv-isa-sim** (spike) associated change: https://github.com/riscv-software-src/riscv-isa-sim/pull/2212
- **sail-riscv** (formal model) associated change: https://github.com/riscv/sail-riscv/pull/1495
- **rvv-intrinsic-doc** Intrinsics associated change: https://github.com/riscv-non-isa/rvv-intrinsic-doc/pull/418


src/vector-dot.adoc
```diff
@@ -0,0 +1,470 @@
...
+wrapping on signed overflow.
++
+The vector-scalar (`.vx`) variant reuses a single `SEW`-bit element,
+read from integer register `rs1`, instead of a vector (`vs1`) of such elements.
+
+
+Operation::
+[source,sail]
+--
+function clause execute (VQDOT(vs2, vs1, vd, suffix)) = {
```
nibrunieAtSi5: function clause name should be updated too.

src/vector-dot.adoc
```diff
@@ -0,0 +1,470 @@
...
+
+[[Zvdot4a8i,Zvdot4a8i]]
+=== `Zvdot4a8i` - Vector quad widening 4D Dot Product of 8-bit Integers
+
+
+`Zvdot4a8i` defines 8-bit Integer dot-product instructions performing the dot product between two 4-element vectors of 8-bit integer elements and accumulating it into a 32-bit integer accumulator.
+
+`Zvdot4a8i` depends on `Zve32x`.
+
+`SEW` is used to indicate both the size of the accumulator elements and the size of the 4-element byte vectors.
```
nibrunieAtSi5:
```suggestion
`SEW` is used to indicate both the size of the accumulator elements and the size of the 4-element vectors.
```

src/vector-dot.adoc
```diff
@@ -0,0 +1,470 @@
...
+These instructions are only defined for `SEW`=32.
```
nibrunieAtSi5:
```suggestion
These instructions are only defined for `SEW`=32. Other values of SEW are _reserved_.
```

aswaterman (commented):
At least superficially this looks good.  I think it's time to push it to the next stage in the process.

src/vector-dot.adoc
```diff
@@ -0,0 +1,470 @@
...
+extensions for RISC-V.
+
+Convolution operations have seen considerable speedup on other microprocessor architectures with the adoption of specialized dot-product instructions. Such convolutions are used in many AI/ML, image-processing, and DSP algorithms. These algorithms can be found in benchmark suites which are used to measure the relative performance of microprocessor cores with different ISAs against each other. This suite of _vector dot product_ extensions is designed to bring RISC-V vector performance up to or to exceed the level of the competition, depending on the implementation vector-length and number of vector ALUs.
+
+The dot product operation is defined for four `SEW/4`-bit values held in a `SEW`-bit element group with an accumulation back into a single `SEW`-bit value with quad-widening.
+The dot-product operation takes the `SEW/4`-bit values at the same index from each source and multiplies them together.
+The four products are summed together and accumulated into the `SEW`-bit element of the destination register.
+If the instruction variant has any signed source data then the accumulation is performed into a signed value.
```
topperc:
```suggestion
If the instruction variant has any signed source data, then the accumulation is performed into a signed value.
```

src/vector-dot.adoc
```diff
@@ -0,0 +1,470 @@
...
+=== Extensions Overview
+
+The section introduces all of the extensions in the Vector Dot Product Instruction Set Extension Specification.
+
+All the Vector Dot Product Extensions can be built on _any_ embedded (Zve*) or application ("V") base Vector Extension.
+
+[[Zvdot4a8i,Zvdot4a8i]]
+=== `Zvdot4a8i` - Vector quad widening 4D Dot Product of 8-bit Integers
+
+
```
topperc, line 30: Unnecessary blank line?

src/vector-dot.adoc
```diff
@@ -0,0 +1,470 @@
...
+// ------------------------------------------------------------
+
+[[vector_dot_product_insns, reftext="Vector Dot Product Instructions"]]
+=== Vector Dot Product Instructions
+
+[[insns-vdota4, Vector 8-bit Signed-Signed Dot Product]]
+==== vdota4.[vv,vx]
+
+Synopsis::
+Vector 8-bit Signed-Signed Dot Product
```
topperc: Should this mention that 4 elements are accmulated together? That would be more important if more SEWs become supported.

src/vector-dot.adoc
```diff
@@ -0,0 +1,470 @@
...
+Description::
+`vdota4` performs a partial dot product of two vectors of `SEW/4`-bit signed integers,
+accumulating the partial sums into a vector of `SEW`-bit signed integers.
++
+In the vector-vector (`.vv`) variant,
+each `SEW`-bit element of `vs1` and `vs2` is interpreted as a list of four `SEW/4`-bit signed integer sub-elements,
+whereas each `SEW`-bit element of `vd` is interpreted as a single `SEW`-bit signed integer.
+For each pair of elements taken consecutively from `vs1` and `vs2`,
+the associated sub-element lists are multiplied componentwise,
+widening (with sign-extension) to  `SEW` bits.
```
topperc:
```suggestion
widening (with sign-extension) to `SEW` bits.
```

src/vector-dot.adoc
topperc: Should this say SEW/4-bit to match the vdota4u synopsis?

src/vector-dot.adoc
```diff
@@ -0,0 +1,470 @@
...
+Description::
+`vdota4su` performs a partial dot product of a vector of `SEW/4`-bit signed integers by a vector of `SEW/4`-bit unsigned integers,
+accumulating the partial sums into a vector of `SEW`-bit signed integers.
++
+In the vector-vector (`.vv`) variant,
+each `SEW`-bit element of `vs1` (respectively `vs2`) is interpreted as a list of four `SEW/4`-bit unsigned (respectively signed) integer sub-elements,
+whereas each `SEW`-bit element of `vd` is interpreted as a single `SEW`-bit signed integer.
+For each pair of elements taken consecutively from `vs1` and `vs2`,
+the associated sub-element lists are multiplied componentwise,
+widening (with sign-extension) to `SEW` bits.
```
topperc: Should this sentence be updated to say that vs1 is zero extended and vs2 is signed extended. Currently, the stence is identical to `vdota4`.

src/vector-dot.adoc
```diff
@@ -0,0 +1,470 @@
...
+Description::
+`vdota4us` performs a partial dot product of a vector of `SEW/4`-bit unsigned integers by a vector of `SEW/4`-bit signed integers,
+accumulating the partial sums into a vector of `SEW`-bit signed integers.
++
+Each `SEW`-bit element of `vs2` is interpreted as a list of four `SEW/4`-bit unsigned integer sub-elements;
+the `SEW`-bit element read from general purpose register `rs1` is interpreted as a list of four `SEW/4`-bit signed integer sub-elements;
+whereas each `SEW`-bit element of `vd` is interpreted as a single `SEW`-bit signed integer.
+For each element taken consecutively from `vs2`,
+the associated sub-element list is multiplied componentwise by the element list from `rs1`,
+widening (with sign-extension) to 32 bits.
```
topperc:
```suggestion
widening (with sign-extension) to `SEW` bits.
```

src/vector-dot.adoc
nibrunieAtSi5:
```suggestion
```

src/vector-dot.adoc
nibrunieAtSi5: I believe we always perform sign extension on the products (since this is signed-unsigned, the products are signed and must be sign extended). 
The difference with respect to `vdot4a` is actually on the signedness of the product **operands**, so if we need to make a change it would be on the previous sentence (or mention that the products are signed values).

src/vector-dot.adoc
nibrunieAtSi5: Good idea, I think we could actually get rid of the `SEW/4` mentions in the Title and Synopsys (the appear) later and focus on the fact that those are 4-element Dot Product Accumulates

src/vector-dot.adoc
topperc: Ok. Re-reading that sentence it is just talking about widening the products before adding them together. It's not stated that the products are SEW/2 bits wide.

4vtomat:
A question about naming, is there a specific reason tha we don't use `zvdot32a8i` as we have 32 bits accumulator? We have  same scanerio in `zvmm` or `xsfmm` that number before `a`(accumulator) is accumulator size, e.g. `xsfmm32a8i`.

aswaterman:
I would parse the name as "4-input dot product, with accumulation, of 8-bit integers".  It's not that it's inconsistent with the matrix extensions; it's that the fixed-input dot-product instructions have a different naming scheme than the matrix extensions for legitimate reasons.  I would argue that for a dot product with a fixed number of inputs, including that number in the name is more important.

(The accumulator being 32 bits is implied by the fact that this is a non-widening instruction, so the output elements are the same size as the input elements, i.e. 4x8.)

topperc:
Is there a version number be we can use in toolchains? The 0.0.1 version in the existing spec for Zvqdotq has been giving us trouble in the toolchain https://github.com/riscv/riscv-dot-product/issues/11

src/vector-dot.adoc
```diff
@@ -0,0 +1,470 @@
...
+
+[[dot_product_extensions]]
+=== Extensions Overview
+
+The section introduces all of the extensions in the Vector Dot Product Instruction Set Extension Specification.
+
+All the Vector Dot Product Extensions can be built on _any_ embedded (Zve*) or application ("V") base Vector Extension.
+
+[[Zvdot4a8i,Zvdot4a8i]]
+=== `Zvdot4a8i` - Vector quad widening 4D Dot Product of 8-bit Integers
```
nibrunieAtSi5:
```suggestion
=== `Zvdot4a8i` - Vector quad widening 4D Dot Product of 8-bit Integers, Version 0.1
```

nibrunieAtSi5:
> Is there a version number be we can use in toolchains? The 0.0.1 version in the existing spec for Zvqdotq has been giving us trouble in the toolchain [riscv/riscv-dot-product#11](https://github.com/riscv/riscv-dot-product/issues/11)

I have added an implicit version number https://github.com/riscv/riscv-isa-manual/pull/2576/commits/6b0d5efb2b139f8607f6c1f7dd2575c37325507a

This bumps it up from the original 0.0.1 (@aswaterman that the renaming mandated such a change anyway)

src/vector-dot.adoc
```diff
@@ -0,0 +1,469 @@
...
+--
+function clause execute (VDOT4A(vs2, vs1, vd, suffix)) = {
+  foreach (i from vstart to vl-1) {
+    // TODO sign extension
+    let {a0 @ a1 @ a2 @ a3} : bits (4*SEW/4) = get_velem(vs2, i)
+    let {b0 @ b1 @ b2 @ b3} : bits (4*SEW/4) = if suffix =="vv" then get_velem(vs1, i)
+                          else zext_or_truncate_to_sew(X(rs1));
+    let acc : bits (SEW) = get_velem(vd, i);
+    let res : bits (SEW) = a0 * b0 + a1 * b1 + a2 * b2 + a3 * b3 + acc;
+    set_velem(vd, i, product);
```
Xu-Dsus4: Hi all, for this code, do you meaning "set_velem(vd, i, res);" ? There is no vairable named "product". Please help to confirm this, thanks!

nibrunieAtSi5: You are right @Xu-Dsus4 , it should be `res`
```suggestion
    set_velem(vd, i, res);
```

src/vector-dot.adoc
```diff
@@ -0,0 +1,469 @@
...
+--
+function clause execute (VDOT4AU(vs2, vs1, vd, suffix)) = {
+  foreach (i from vstart to vl-1) {
+    // TODO sign extension
+    let {a0 @ a1 @ a2 @ a3} : bits (4*8) = get_velem(vs2, i)
+    let {b0 @ b1 @ b2 @ b3} : bits (4*8) = if suffix =="vv" then get_velem(vs1, i)
+                          else zext_or_truncate_to_sew(X(rs1));
+    let acc : bits (SEW) = get_velem(vd, i);
+    let res : bits (SEW) = a0 * b0 + a1 * b1 + a2 * b2 + a3 * b3 + acc;
+    set_velem(vd, i, product);
```
nibrunieAtSi5:
```suggestion
    set_velem(vd, i, res);
```

src/vector-dot.adoc
```diff
@@ -0,0 +1,469 @@
...
+--
+function clause execute (VDOT4ASU(vs2, vs1, vd, suffix)) = {
+  foreach (i from vstart to vl-1) {
+    // TODO sign extension
+    let {a0 @ a1 @ a2 @ a3} : bits (4*8) = get_velem(vs2, i)
+    let {b0 @ b1 @ b2 @ b3} : bits (4*8) = if suffix =="vv" then get_velem(vs1, i)
+                          else zext_or_truncate_to_sew(X(rs1));
+    let acc : bits (SEW) = get_velem(vd, i);
+    let res : bits (SEW) = a0 * b0 + a1 * b1 + a2 * b2 + a3 * b3 + acc;
+    set_velem(vd, i, product);
```
nibrunieAtSi5:
```suggestion
    set_velem(vd, i, res);
```

src/vector-dot.adoc
```diff
@@ -0,0 +1,469 @@
...
+[source,sail]
+--
+function clause execute (VDOT4AUS(vs2, vs1, vd)) = {
+  foreach (i from vstart to vl-1) {
+    // TODO sign extension
+    let {a0 @ a1 @ a2 @ a3} : bits (4*SEW/4) = get_velem(vs2, i)
+    let {b0 @ b1 @ b2 @ b3} : bits (4*SEW/4) = zext_or_truncate_to_sew(X(rs1));
+    let acc : bits (SEW) = get_velem(vd, i);
+    let res : bits (SEW) = a0 * b0 + a1 * b1 + a2 * b2 + a3 * b3 + acc;
+    set_velem(vd, i, product);
```
nibrunieAtSi5:
```suggestion
    set_velem(vd, i, res);
```

ubc-guy:
A few comments:

a) this doesn't appear to 'accumulate' the past results of vd (ie, using vd as both a src and dst operand), which the long/batched version of these instructions can do (https://github.com/aswaterman/riscv-misc/blob/main/isa/ldot-bdot/ldot-bdot.adoc). if you add this feature, then this instruction overlaps almost 100% with the long 8b batched integer version, and that version also has specs for 16b etc. i think these two specifications can be better unified, perhaps?

b) there should also be scalar equivalent instructions, for non-vector architectures to "mimic" the behaviour. not sure this falls under this "fast track" to the "vector" spec, but we should discuss how/where to add it. (most likely, it should be added to the P spec, but I think it should be stand-alone from the P spec so as to avoid inheriting too much baggage and because I don't think P spec has any ternary operators that use vd as a src)

topperc:
> A few comments:
> 
> a) this doesn't appear to 'accumulate' the past results of vd (ie, using vd as both a src and dst operand), which the long/batched version of these instructions can do (https://github.com/aswaterman/riscv-misc/blob/main/isa/ldot-bdot/ldot-bdot.adoc). if you add this feature, then this instruction overlaps almost 100% with the long 8b batched integer version, and that version also has specs for 16b etc. i think these two specifications can be better unified, perhaps?

These instructions do use `vd` as an source.

> 
> b) there should also be scalar equivalent instructions, for non-vector architectures to "mimic" the behaviour. not sure this falls under this "fast track" to the "vector" spec, but we should discuss how/where to add it. (most likely, it should be added to the P spec, but I think it should be stand-alone from the P spec so as to avoid inheriting too much baggage and because I don't think P spec has any ternary operators that use vd as a src)

The P extension has many flavors of multiply accumulate instructions that use `rd` as a source. There other instructions like `slx`, `srx`, `merge`, that also use `rd` as a source.

nibrunieAtSi5:
> A few comments:
> 
> a) this doesn't appear to 'accumulate' the past results of vd (ie, using vd as both a src and dst operand), which the long/batched version of these instructions can do (https://github.com/aswaterman/riscv-misc/blob/main/isa/ldot-bdot/ldot-bdot.adoc). if you add this feature, then this instruction overlaps almost 100% with the long 8b batched integer version, and that version also has specs for 16b etc. i think these two specifications can be better unified, perhaps?
> 

As @topperc mentioned, all the 7 instructions specified by this document are dot product accumulate. I have clarified the text and the instruction section titles to make this clearer. I still believe the extension name should not mention "Accumulate" explicitly for a few reasons: 

1. most dot product extensions (e.g. **Zvdota** and **Zvbdota**
 families) also implement some form of dot product accumulate and skip the explicit mention of **accumulate**.
2. This makes the extension title very verbose, and is not required because the textual description and the instruction descriptions are already explicit enough (IMO)

I agree, that we could argue for those specifications to be unified, but since the plans were already approved as separate fast track efforts, I believe we should continue this way and simply make sure we cross review the efforts.

> b) there should also be scalar equivalent instructions, for non-vector architectures to "mimic" the behaviour. not sure this falls under this "fast track" to the "vector" spec, but we should discuss how/where to add it. (most likely, it should be added to the P spec, but I think it should be stand-alone from the P spec so as to avoid inheriting too much baggage and because I don't think P spec has any ternary operators that use vd as a src)

This is an interesting suggestion but since it goes beyond the "charter" for the fast track, I think it is outside the scope of this pull request.
I believe a scalar/packed extension would address a different need.

@ubc-guy , note that P's extension `PM4ADD*.B` could already be considered as a P-extension equivalent of `vdot4a`:
https://github.com/riscv/riscv-p-spec/blob/master/P-ext-proposal.adoc#pm4add-b

> The PM4ADD.B instruction multiplies corresponding packed bytes within each 32-bit lane as signed values and sums the four 32-bit products to produce one packed 32-bit result per lane in rd.

ubc-guy:
> > A few comments:
> > a) this doesn't appear to 'accumulate' the past results of vd (ie, using vd as both a src and dst operand), which the long/batched version of these instructions can do

> I have clarified the text and the instruction section titles to make this clearer. 

I apologize -- when I looked at this spec, I did not see the accumulate aspect in the specification. I should learn to read better... but I appreciate changes made to enhance this clarity, thank you.

> >  (https://github.com/aswaterman/riscv-misc/blob/main/isa/ldot-bdot/ldot-bdot.adoc). if you add this feature, then this instruction overlaps almost 100% with the long 8b batched integer version, and that version also has specs for 16b etc. i think these two specifications can be better unified, perhaps?
> I agree, that we could argue for those specifications to be unified, but since the plans were already approved as separate fast track efforts, I believe we should continue this way and simply make sure we cross review the efforts.

I didn't intend to suggest merging the fast tracks or the specifications.

I meant make sure the features in one are also reflected in the other. In particular, this current extension (not the Waterman one) lacks a quad-16b integer version (summing into a 64b integer).

I'm not sure why this current extension only sums to a 32b integer; is it intended only for RV32? Why? corresponding versions for RV64 should be planned, which would be a quad-16b-to-64b, an octo-8b-to-64b, and a version that does two quad8b-to-32b within the two 32b lanes of a 64b datapath.

> > b) there should also be scalar equivalent instructions, for non-vector architectures to "mimic" the behaviour. not sure this falls under this "fast track" to the "vector" spec, but we should discuss how/where to add it. (most likely, it should be added to the P spec, but I think it should be stand-alone from the P spec so as to avoid inheriting too much baggage and because I don't think P spec has any ternary operators that use vd as a src)
> 
> This is an interesting suggestion but since it goes beyond the "charter" for the fast track, I think it is outside the scope of this pull request. I believe a scalar/packed extension would address a different need.

Yes, technically, it is out of scope for this TG to produce such an instruction.

However, technically, communication between TGs must be constant. If there is a demand for this vector version of this instruction, there should also be a demand for a scalar version (for improving RVB/RVM profiles, for example). So, the vector versions being added should talk with the relevant scalar TGs (Scalar Efficiency, P, etc) to consider adding a non-vector version of this instruction. (if it's in demand for vector, why wouldn't it be in demand for scalar?)

> @ubc-guy , note that P's extension `PM4ADD*.B` could already be considered as a P-extension equivalent of `vdot4a`: https://github.com/riscv/riscv-p-spec/blob/master/P-ext-proposal.adoc#pm4add-b

Thanks -- I couldn't find this myself.

> > The PM4ADD.B instruction multiplies corresponding packed bytes within each 32-bit lane as signed values and sums the four 32-bit products to produce one packed 32-bit result per lane in rd.

I note that the P version does not accumulate into Rd. I would expect better consistency, if it doesn't change the scope of P too much (ie, are there any other P instructions where Rd is both source and dest? if so, then an accumulating version should exist).

topperc:
> > > The PM4ADD.B instruction multiplies corresponding packed bytes within each 32-bit lane as signed values and sums the four 32-bit products to produce one packed 32-bit result per lane in rd.
> 
> I note that the P version does not accumulate into Rd. I would expect better consistency, if it doesn't change the scope of P too much (ie, are there any other P instructions where Rd is both source and dest? if so, then an accumulating version should exist).

The instruction you're looking for is PM4ADDA.B

keyon1992:
Any update for this extension?  It seems the internal review stage haven't start yet.
https://riscv.atlassian.net/browse/RVS-1971

PR #2582, opened 2026-01-19
Clarify misaligned address exception text

### LINKED ISSUE #2578: [Clarification] - access faults vs misaligned address exceptions for misaligned accesses
When you have a misaligned access to memory that doesn't support it, my understanding is that it should work like this:

1. If you CPU doesn't support misaligned access to *any* address, you can raise a misaligned address exception before address translation.  ([The high priority option.](https://riscv-specs.timhutt.co.uk/spec/riscv-isa-release-57ff123-2026-01-17/riscv-privileged.html#:~:text=Optionally%3A-,Load/store/AMO%20address%20misaligned,-13%2C%2015%2C%205))
2. Otherwise do address translation and then...
3. If the physical memory doesn't support misaligned access then
    a) If it allows *emulated* access raise a misaligned exception
    b) If it doesn't raise an access fault.

That is consistent with [this bit of the spec](https://riscv-specs.timhutt.co.uk/spec/riscv-isa-release-57ff123-2026-01-17/riscv-privileged.html#:~:text=Non%2Didempotent%20regions%20might,cause%20unexpected%20side%20effects.):

> Non-idempotent regions might not support misaligned accesses. Misaligned accesses to such regions should raise access-fault exceptions rather than address-misaligned exceptions, indicating that software should not emulate the misaligned access using multiple smaller accesses, which could cause unexpected side effects.

However there's [this other bit of the spec](Implementations that support misaligned accesses only to some physical addresses must translate and check the address before determining whether the misaligned access may proceed, in which case raising the page-fault exception or access is more appropriate.) that seems to contradict it:

> Implementations that support misaligned accesses only to some physical addresses must translate and check the address before determining whether the misaligned access may proceed, in which case raising the page-fault exception or access is more appropriate.

I think there's a word missing around `access` (`access fault`?). But anyway:

1. Should you really raise a *page fault* due to misaligned access to a page that doesn't support misaligned accesses? That seems weird and it would also make emulation impossible, or at least super awkward.
2. Same question for access faults.

I think it should read something like this:

> Implementations that support misaligned accesses only to some physical addresses must translate and check the address before determining whether the misaligned access may proceed and if not whether it can be emulated. If emulation is allowed raise a misaligned address exception, otherwise an access fault.

One final point. In the case that the CPU doesn't support misaligned access at all, and it always raises a misaligned address exception, you lose the information about whether the access can be emulated. I guess that's unavoidable though and the execution environment would just have to magically know what to do.

Timmmm:
The old text was a little confusingly worded and made it sound like you should raise a page fault or access fault exception in response to misaligned accesses. It was really just saying that they take priority.

Fixes #2578

### DIFF
--- a/src/machine.adoc
+++ b/src/machine.adoc
@@ -1985,8 +1985,8 @@ can unconditionally raise the misaligned-address exception without
 performing address translation or protection checks. Implementations
 that support misaligned accesses only to some physical addresses must
 translate and check the address before determining whether the
-misaligned access may proceed. In this case any page-fault or access-fault
-exception raised during the address translation process would take priority
+misaligned access may proceed. In this case, any page-fault or access-fault
+exception raised during the address translation process may take priority
 over misaligned-address exceptions.
 
 ***


allenjbaum:
I think you need to replace the word "would" with "may". The spec allows misaligned to have either higher or lower priority than page/access fault.
So the response could be one or the other, depending on whether that option is implemented or not.

Timmmm:
That clause is specifically talking about the case where it *isn't* higher priority though.

src/machine.adoc
```diff
@@ -1985,8 +1985,9 @@ can unconditionally raise the misaligned-address exception without
 performing address translation or protection checks. Implementations
 that support misaligned accesses only to some physical addresses must
 translate and check the address before determining whether the
-misaligned access may proceed, in which case raising the page-fault
-exception or access is more appropriate.
+misaligned access may proceed. In this case any page-fault or access-fault
+exception raised during the address translation process would take priority
+over misaligned-address exceptions.
```
aswaterman, line 1990: The original text has slightly different connotation.  It wasn't stating a definitive consequence, but rather remarking that raising PF/AF would be the more straightforward choice in that style of microarchitecture.

Timmmm: I don't think that's what it was trying to say though. You surely shouldn't raise a page fault for misaligned accesses an aligned access to the same page would succeed?

aswaterman: OK, I now see that you are validly reading something different from the text than I am.

I tend to agree with Allen, though.  "may" seems more appropriate than "would" since this is not a mandate.  Also, comma after "In this case".

PR #2593, opened 2026-01-22
Improve manual formating of Chapter CSR in Privilige Manual

cebarobot:
This PR provides 4 patches which may improve the formating of Chapter CSR in Privilige Manual.

* Move table "csrrwpriv" to correct Chapters

  Table Allocation of RISC-V CSR address ranges (csrrwpriv) was located in Section CSR Listing (Section 2.2), which is strange because all references to it is in Section CSR Address Mapping Conventions (Section 2.1).

  This patch moves the table to correct place.

* Merge table mcsrnames0 and mcsrnames1

  Table "Currently allocated RISC-V machine-level CSR addresses" was divided into two tables mcsrnames0 and mcsrnames1. This might because long table crosses pages and makes problems.

  However, current asciidoctor could render the long tables properly, and that division only make confusion.

  This patch merges the two table into one. 

* Merge two "Machine Counter Setup" groups into one

  Previously, there is two CSR groups in Table "Currently allocated RISC-V machine-level CSR addresses": one is "Machine Counter Configuration" (from #2045), and the other is "Machine Counter Setup".

  It is confusing to divide these CSRs with similar themes and continuous addresses into two groups. This patch put them together.

* Add subsections for CSR Listing

  This patch adds subsection for Section CSR Listing of Chapter CSR, which may enhance the navigation of such long table.

All patches do not have any substantive modification. If some patches is not inappropriate or unnecessary, please let me know.

Best wishes.

### DIFF
--- a/normative_rule_defs/machine.yaml
+++ b/normative_rule_defs/machine.yaml
@@ -11,8 +11,6 @@ normative_rule_definitions:
   # M-mode general
   - name: M_highest_priv_mode
     tag: "norm:M_highest_priv_mode"
-  - name: M_mode_at_reset
-    tag: "norm:M_mode_at_reset"
   - name: M_access_all_lower_priv_CSRs
     tag: "norm:M_access_all_lower_priv_CSRs"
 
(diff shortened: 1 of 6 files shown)


aswaterman:
Please fix the following problem with the PR.  (Sorry, I don't have any advice on how to do so.)

Screenshot 2026-01-22 at 2 30 19 PM

aswaterman:
This is what Rafael sent me.  See if it works for you (I haven't tried it yet).

[github_signed_commits_step_by_step.pdf](https://github.com/user-attachments/files/24811883/github_signed_commits_step_by_step.pdf)

aswaterman:
Great, thanks.

PR #2597, opened 2026-01-22
Clarify that Zicfiss depends on the Zaamo extension.

### LINKED ISSUE #2596: Should `ssamoswap.{d,w}` depend on Zaamo?
The Spike implementation requires Zaamo (https://github.com/riscv-software-src/riscv-isa-sim/pull/2134), and QEMU requires A or Zaamo (https://github.com/qemu/qemu/blob/2339d0a1cfac6ecc667e6e062a593865c1541c35/target/riscv/insn_trans/trans_rvzicfiss.c.inc#L87).  But the spec does not seem to explicitly require this.  
This comment https://github.com/riscv-software-src/riscv-isa-sim/pull/2134#pullrequestreview-3447361293 seems to indicate it should?

pmundkur:
Fixes #2596.

### DIFF
--- a/src/unpriv-cfi.adoc
+++ b/src/unpriv-cfi.adoc
@@ -376,7 +376,7 @@ system or the administrator) configuration, could either deny the request or
 deactivate the Zicfiss extension for the application. It is strongly recommended
 that the policy enforces a strict security posture and denies the request.
 
-The Zicfiss extension depends on the Zicsr and Zimop extensions. Furthermore,
+The Zicfiss extension depends on the Zicsr, Zimop and Zaamo extensions. Furthermore,
 if the Zcmop extension is implemented, the Zicfiss extension also provides the
 `C.SSPUSH` and `C.SSPOPCHK` instructions. Moreover, use of Zicfiss in U-mode
 requires S-mode to be implemented. Use of Zicfiss in M-mode is not supported.


aswaterman:
@ved-rivos I think this is right, but can you make sure you agree, and merge it if so?

src/unpriv-cfi.adoc
```diff
@@ -378,8 +378,10 @@ that the policy enforces a strict security posture and denies the request.
 
 The Zicfiss extension depends on the Zicsr and Zimop extensions. Furthermore,
```
ved-rivos: The Zicfiss extension depends on the Zicsr, Zimop, and Zaamo extensions.

src/unpriv-cfi.adoc
```diff
@@ -378,8 +378,10 @@ that the policy enforces a strict security posture and denies the request.
...
 if the Zcmop extension is implemented, the Zicfiss extension also provides the
-`C.SSPUSH` and `C.SSPOPCHK` instructions. Moreover, use of Zicfiss in U-mode
-requires S-mode to be implemented. Use of Zicfiss in M-mode is not supported.
+`C.SSPUSH` and `C.SSPOPCHK` instructions, and if the Zaamo extension is implemented,
+the Zicfiss extension also provides the `SSAMOSWAP.W` and the `SSAMOSWAP.D` instructions.
```
ved-rivos: Implementing SSAMOSWAP is not optional when implementing ZIcfiss. Suggest removing this change and replace with the change suggested above.

pmundkur: Done.

pmundkur:
Hmm, not sure what I need to do to satisfy the verified signature check.  My commits to `sail-riscv` show up as verified.

aswaterman:
I don't know, either.  I've asked the RISC-V TPMs for help.

aswaterman:
This is what Rafael sent me.  See if it works for you (I haven't tried it yet).

[github_signed_commits_step_by_step.pdf](https://github.com/user-attachments/files/24811883/github_signed_commits_step_by_step.pdf)

pmundkur:
Should be fixed now.

PR #2599, opened 2026-01-22
Add note that Svnapot is usable in G-stage PTEs

### LINKED ISSUE #2594: [Clarification] - Should G-stage translation support NAPOT when Svnapot enabled?
I've looked at both the latest descriptions of both the Svnapot extension and the H extension and can see no explicit mention of the impact of the Svnapot extension on G-stage translation. There is the line:

> Except as noted, all other aspects of Sv32, Sv39, Sv48, or Sv57 are adopted unchanged for G-stage translation.

And also this comment:

> G-stage address translation uses the identical format for PTEs as regular address 
> translation, even including the U bit, due to the possibility of sharing some (or all)
> page tables between G-stage translation and regular HS-level address translation.
> Regardless of whether this usage will ever become common, we chose not to
> preclude it.

Which implies that PTE.N should be supported in G-stage pages to allow table sharing but an explicit note to this effect would aid the clarity here.

aswaterman:
Fixes #2594

### DIFF
--- a/normative_rule_defs/machine.yaml
+++ b/normative_rule_defs/machine.yaml
@@ -11,8 +11,6 @@ normative_rule_definitions:
   # M-mode general
   - name: M_highest_priv_mode
     tag: "norm:M_highest_priv_mode"
-  - name: M_mode_at_reset
-    tag: "norm:M_mode_at_reset"
   - name: M_access_all_lower_priv_CSRs
     tag: "norm:M_access_all_lower_priv_CSRs"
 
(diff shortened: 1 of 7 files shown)


src/supervisor.adoc
```diff
@@ -2219,6 +2219,9 @@ to be useful, we have chosen to standardize only 64 KiB support as a
 first step.
 ====
 
+If the hypervisor extension is implemented, Svnapot is also supported in
```
ved-rivos:
```suggestion
If the hypervisor extension is also implemented, Svnapot is also supported in
```

aswaterman: Yeah, that's better.

PR #2609, opened 2026-01-23
Draft of Smpmpind, Smcfiss, Smucfiss, and Sspmpss extensions

ved-rivos:
Draft specification for internal review

### DIFF
--- a/src/smcfiss.adoc
+++ b/src/smcfiss.adoc
@@ -1,0 +1,155 @@
+[[smcfiss]]
+
+== Contributors
+
+Allen Baum, Ruud Derwig, Brian Grayson, Deepak Gupta, Freddie Huang, Yann
+Loisel, Ved Shanbhogue, Geoffrey Thorpe, Joe Xie, Adam Zabrocki.
+
+== "Smpmpind" Extension for Indirect PMP CSR Access, Version 0.1
+
+The Smpmpind extension defines a range of selector values for accessing PMP
+registers using the CSRs provided by the Smcsrind extension. The Smpmpind
+extension depends on the Smcsrind extension.
+
+For indirect access to PMP registers, `miselect` selects the target PMP
+entry; `mireg` accesses its `pmpaddr` register, and `mireg2` accesses its
+`pmpcfg` register. The `mireg3` through `mireg6` CSRs are reserved.
+
+The `pmpcfg` is a MXLEN wide register. The low 8 bits of this register are an
+alias of the 8-bit field of the PMP configuration register for this entry
+that can be accessed using a direct CSR access.
+
+The behavior of PMP entry locking is unchanged by this extension.
+
+Bit MXLEN-1 of the `pmpcfg` register is defined as the extended-attributes (`E`)
+field. When the `E` field is 0, the `XWR` bits in the `pmpcfg` register are
+interpreted as if the `Smpmpind` extension were not implemented.
+
+When the `E` field is 1, all encodings of the `XWR` bits are reserved unless
+they are defined by another standard extension. Attempting to access a PMP
+region when `E` is 1 and the `XWR` encoding is reserved raises an access fault
+exception corresponding to the original access type.
+
+== "Smcfiss" Extension for Machine-mode Shadow Stack, Version 0.1
+
+The Smcfiss extension enables the use of a shadow stack to enforce
+backward-edge control-flow integrity in machine mode.
+
+The Smcfiss extension depends on the Zicfiss and Smpmpind extensions.
+
+The Smcfiss extension introduces an `MSSE` field in `mseccfg` to enable the use
+of Zicfiss in machine mode. When the Smcfiss extension is implemented, the
+shadow-stack-enabled (SSE) state, denoted by the term `xSSE`, is set to the
+value of `mseccfg.MSSE`. When Smcfiss is not implemented, the `MSSE` field is
+read-only zero.
+
+When the Smepmp extension is also implemented, if the `mseccfg.RLB` field is 0
+while `pmpcfg.L` is 1 in any PMP entry, further writes to the `mseccfg.MSSE`
+field are ignored until a PMP reset.
+
+When `mseccfg.MSSE` is 1, a PMP entry with `E=1` and `XWR=010` denotes an M-mode
+shadow stack memory region.
+
+When `E=1` and `mseccfg.MSSE` is 0, the `XWR=010` encoding remains reserved.
+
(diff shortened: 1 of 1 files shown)


aswaterman:
@ved-rivos RVIA instituted a requirement that commits be signed; see this guide.

[github_signed_commits_step_by_step.pdf](https://github.com/user-attachments/files/24832236/github_signed_commits_step_by_step.pdf)

ved-rivos:
@aswaterman Thanks. I have signed the commit now.

src/smcfiss.adoc
```diff
@@ -0,0 +1,154 @@
...
+
+If `satp.MODE` is not `Bare` or if `menvcfg.SSE` is 0, the implementation
+behaves as if Sspmpss is not implemented. When Sspmpss is implemented, the
+`XWR=010` encoding in `spmpcfg[i]` with `spmpcfg.SHARED=0` denotes a shadow
+stack memory region. When Sspmpss is not implemented, the `XWR=010` encoding
+with `spmpcfg.SHARED=0` remains reserved.
+
+When Sspmpss is implemented, the following additional rules apply to memory
+accesses with effective privilege mode S/HS or U:
+
```
eckhard-delfs-qualcomm, line 128: Why is HS listed here, if the H-extension specifics are discussed further down?

src/smcfiss.adoc
```diff
@@ -0,0 +1,154 @@
...
+  or by shadow stack instructions.
+
+The access faults caused by these additional rules occur at lower priority than
+a page fault due to the memory access being disallowed by the U bit, given the
+effective privilege mode of the access and the state of the `SUM` and `MXR`
+fields of the `mstatus` register.
+
+When the H extension is also implemented, the following additional rules apply
+to memory accesses with effective privilege mode VS or VU:
+
```
eckhard-delfs-qualcomm, line 149: Should there be a statement that applicability on VS/VU requires vsatp.MODE set to Bare?

ved-rivos: If vsatp.MODE is Bare then shadow stack accesses lead to an access fault. To support shadow stack with Bare mode in VS/VU we need the upcoming vSPMP extension.

src/smcfiss.adoc
ved-rivos: The discussion further down is about VS and VU mode. I can update this statement to say "S or U, and when H extension is also implemented, then HS". However, I think that would be very distracting. Please let me know if you feel strongly about this.

eckhard-delfs-qualcomm: Since the draft separates H-extension specifics in a dedicated section, introducing HS in a common, preceding section may cause some confusion to the reader. I would suggest to be more explicit.

```suggestion
When Sspmpss is implemented, the following rules apply to memory accesses with effective privilege S or U.  When the H extension is implemented, these rules also apply to HS, and additional requirements apply for VS/VU as described below.
```

src/smcfiss.adoc
eckhard-delfs-qualcomm: Makes sense. Thanks for the clarification.

src/smcfiss.adoc
ved-rivos: Thanks. I have updated the text to call out dependence of HS on H.

eckhard-delfs-qualcomm: LGTM. Thanks.

src/smcfiss.adoc
```diff
@@ -0,0 +1,155 @@
...
+* Shadow stack instructions that access memory require the SPMP to grant both
+  read and write permission; otherwise, a store/AMO guest-page fault exception
+  is raised.
+* Any attempt to access a shadow stack memory region, including by shadow stack
+  instructions or implicit accesses, raises a guest-page fault exception
+  corresponding to the original access type.
```
eckhard-delfs-qualcomm: @ved-rivos : we assume the last bullet 

> Any attempt to access a shadow stack memory region, including by shadow stack
  instructions or implicit accesses, raises a guest-page fault exception
  corresponding to the original access type

should be read to mean shadow stacks are not supported from VS/VU mode if Sspmpss is implemented. Is that correct? If so, what is the purpose of the statement before: 

> Shadow stack instructions that access memory require the SPMP to grant both
  read and write permission; otherwise, a store/AMO guest-page fault exception
  is raised.

It reads as if in VS/VU shadow stack instructions do _not_ fault unconditionally. This appears contradictory.

ved-rivos: Shadow stacks are supported in VS/VU mode when Sspmpss is implemented. When Sspmpss is implemented, the SPMP stand-in for the G-stage page tables when V=1 and are hypervisor managed. Similar to how shadow stack instructions require read and write permissions in G-stage page tables when effective privilege mode of the access is VS or VU, they require read and write permissions from the SPMP. 

The last quoted text refers to shadow stack encoding in the SPMP. A shadow stack encoding in SPMP can only be used when effective privilege mode of the access is HS/U as they are hypervisors shadow stacks.

Supporting SPMP based shadow stacks in VS/VU requires the upcoming Ssvspmp extension that introduces vSPMP to support two-stage SPMP. When Ssvspmp is implemented, the VS/VU mode shadow stacks can be protected using VS level SPMPs.

eckhard-delfs-qualcomm: Thanks for the detailed explanation, @ved-rivos. 

May I propose the following change:

> Shadow stack instructions that access memory **that is not a shadow stack memory region** 
require the SPMP to grant both read and write permission; otherwise, a store/AMO 
guest-page fault exception is raised.

This does not introduce any semantic change; it simply clarifies the intended distinction that, in VS/VU, shadow stack instructions may access ordinary memory subject to SPMP read/write permissions, whereas shadow stack memory regions identified by the special encoding are not accessible and always fault.

ved-rivos: > in VS/VU, shadow stack instructions may access ordinary memory subject to SPMP read/write permissions, whereas shadow stack memory regions identified by the special encoding are not accessible and always fault.
@eckhard-delfs-qualcomm 
That is not quite right. The encoding XWR=010 (intended to be used as shadow stack encoding when V=0) is reserved for accesses made with effective privilege mode of VS/VU - they are not accessible by memory accesses originated by any instruction or by implicit accesses. I have reworded the text to remove the use of "shadow stack memory region" which is likely leading to the confusion.

eckhard-delfs-qualcomm: Looks good to me. Thanks for the clarification!

eckhard-delfs-qualcomm:
Resolved after clarification.

PR #2618, opened 2026-01-29
Add Zvdota/Zvbdota draft

aswaterman:


nibrunieAtSi5 (approved):
LGTM


regarding the suggested RTO behavior

> Rounding to Odd behavior in dot product mode
>
> Rounding to odd (RTO) is not part of the IEEE-754 standard (at least not until and including revision 2019).
>
> The version used for the dot product operation admits two divergences with the generally accepted definition:
>
> * When overflowing, an infinity result is returned (rather than the largest magnitude normal number) see [Overflow](https://github.com/riscv/riscv-isa-manual/pull/2618/changes#RVBNAOverflow)
>
> * A zero result is always positive (+0) whatever the sign of the actual zero term(s) of the dot product sum


Note on the suggested definition of RTO for the RVBNA: the IEEE-754 has started some discussion to possibly introduce a binary roundTowardsOdd mode in the next revision. One of the proposal rounds overflows to the maxValue (aligned with the generally accepted definition of RTO) with the argument that this allows to distinguish arithmetic on infinities from overflow (this argument would apply more generally to any saturating arithmetic but applies to RTO because maxValue is generally odd while infinity is "even").
I am not sure the argument holds for a large sum of products (as is the case BDOT). In particular during the rounding of the products: since this is done before the accumulation, rounding very large products to maxValue could lead to very large cancellation during the accumulation and loss of information about the overflow altogether (while rounding products which overflow to infinities would lead most likely to a NaN result when overflowing products of opposite signs are summed, or to a NaN if all overflow share the same sign).

I believe this is a desirable behavior, but wouldn't mind other people sharing their opinion.

The definition of RTO used here should make no difference (with the generally accepted definition) when the product exponent is much smaller than the accumulation (e.g. accumulation of [O]FP8 or FP16 products with a FP32 accumulator). The difference becomes visible when there is a possibility for the products to overflow in the accumulation format (e.g. BF16 products with FP32 accumulation).

aswaterman:
It seems like an anti-feature that a negative overflow result and a positive overflow result sum to zero. The behavior we’ve defined hews more closely to what you’d get if you performed the arithmetic without bulk normalization, using more conventional rounding like nearest-even. Following the principle of least surprise is a good thing. Our deliberate divergence remains sound, IMO.

The 754 committee’s potential choice to make round-to-odd saturate at the maximum normal value makes sense to me, since it helps with successive conversion steps. E.g. if you’re trying to convert from double to half, using the dynamic rounding mode, using a single-precision intermediate, then the first step mustn’t generate an infinity, since the dynamic rounding mode could have been towards the opposite infinity.

src/zvdota.adoc
```diff
@@ -0,0 +1,840 @@
...
+
+The Zvfwdota16bf extension adds one instruction, `vfwdota.vv`, which computes
+a dot product between two BF16 vectors, accumulating into an FP32 scalar.
+The instruction is reserved unless `vtype`.`altfmt`=1 and SEW=16.
+The `vd` operand has EEW=2*SEW.
+
+----
+# Zvfwdota16bf BF16 dot-product instruction
+# altfmt=1
+vfwdota.vv vd, vs2, vs1, vm   # vd[0] += vs2 dot vs1
```
Xu-Dsus4, line 100: Hi @aswaterman,
I’d like to enquire whether, for this instruction, the intermediate result format is always BF16, except for the last step where the accumulated intermediate result is added to the FP32 scalar value from vd[0]. Is that correct?
In other words, based on my understanding, this instruction can be described with the following code:
```
function clause execute (VFWDOTA_VV(vs2, vs1, vd, vm)) = {
  foreach (i from vstart to VL-1) {
    if (vm == 0b1) | (v0[i] == 0b1) then
      let (bfloat16) op1 = get_velem(vs1, SEW=16, i);
      let (bfloat16) op2 = get_velem(vs2, SEW=16, i);
      let (bfloat16) tmp_result += bf16_mul(op1, op2);
  }
  let (float32) fcvt_result = bf16_to_f32(tmp_result);
  let (float32) op3 = get_velem(vd, 32, 0);
  let (float32) result = f32_add(fcvt_result, op3);
  set_velem(vd, EEW=32, 0, result);
  // tail element handling follows VTA
  RETIRE_SUCCESS
}
```
I’m working on the Spike implementation and would appreciate your confirmation. Thanks!

aswaterman: Hi @Xu-Dsus4,

There's already a Spike implementation of these instructions that has been merged into the master branch.  It is just that the names are out of date, since the ARC requested a rename from bdot to bdota.  If you want to contribute a PR to riscv-opcodes and then to riscv-isa-sim to fix the extension and instruction names, I would be very grateful.

To answer your semantics question, the intermediate result has full precision, and it is then either fed to RVBNA or accumulated in FP32 (or a combination of both).  An implementation that rounds the products back to BF16 is not correct.

Xu-Dsus4: Thanks for your reply, I noticed the implementation with old instruction names. I am glad to contribute to Spike. I will review and updated the Spike code related to the dot product instruction.

src/zvdota.adoc
```diff
@@ -0,0 +1,840 @@
...
+        let A_i_isInf = (A_i_exp == maskExp) && (A_i_mant == 0)
+        let B_i_isInf = (B_i_exp == maskExp) && (B_i_mant == 0)
+        let A_i_isNaN = (A_i_exp == maskExp) && (A_i_mant != 0)
+        let B_i_isNaN = (B_i_exp == maskExp) && (B_i_mant != 0)
+        let A_i_isSNaN = A_i_isNaN && (A_i_mant & (1 << (m_l - 1))) == 0
+        let B_i_isSNaN = B_i_isNaN && (B_i_mant & (1 << (m_r - 1))) == 0
+
+        let invalidProd = (A_i_isInf && B_i_isZero) || (B_i_isInf && A_i_isZero)
+        let infiniteProdLHS = (A_i_isInf && !B_i_isNaN  && !B_i_isZero)
+        let infiniteProdRHS = (B_i_isInf && !A_i_isNaN  && !A_i_isZero)
```
nibrunieAtSi5: @aswaterman , it looks like the pseudo code is correct (not affected by https://github.com/riscv-software-src/riscv-isa-sim/issues/2273)

src/zvdota.adoc
```diff
@@ -0,0 +1,840 @@
...
+// g = o = log2(n)
+//
+// A[i] are IEEE-encoded floating point numbers on (e_l+p_l) bits
+// B[i] are IEEE-encoded floating point numbers on (e_r+p_r) bits
+// MSB is sign, next e_l (resp. e_r) bits are biased exponent,
+// last m_l (resp. m_r) bits are the mantissa.
+// exponent bias is lhs_bias (resp. rhs_bias)
+// p_l = m_l + 1
+// p_r = m_r + 1
+// the output is an IEEE-encoded floating-point number on (f+q) bits
```
omasanori: s/IEEE-encoded/IEEE 754-encoded binary/

src/zvdota.adoc
```diff
@@ -0,0 +1,840 @@
...
+        }
+    }
+}
+```
+
+Note:: In the current specification the most significant bit of the max product has a weight of `max_exp + 1`, which means that there are `q-2+g` fractional bits in the max product. `g` could be increased to `ceil(log2(n)) + 1` to ensure the number of fractional bits is at least `q-1` (identical to the output format). Numerical impacts of the value of `g` have not been evaluated.
+
+Note:: Any one of the `p_l + p_r` bits of the max product can be its leading bit due to leading zeros in subnormal inputs. It is also possible for the maximum product to have more leading zeros than the other products (including when product alignment is taken into account). This is discussed in more detailed in the next section <<#RVBNAReferenceExponent>>.
+
+Note:: The RVBNA pseudocode assumes that operands are IEEE encoded, but this is not actually a requirement of RVBNA.
```
omasanori: s/IEEE encoded/represented in the IEEE 754-2008 binary encoding/

src/zvdota.adoc
```diff
@@ -0,0 +1,840 @@
...
+
+
+
+
+
+==== Rounding modes
+
+For floating-point dot product operations, RVBNA only supports rounding-to-odd (RTO) with some specificites (see <<#RVBNARTO>>).
+
+For floating-point multiply operations, RVBNA supports all the rounding modes mandated by RISC-V **F** extension.
```
omasanori: s/RISC-V/the/

src/zvdota.adoc
```diff
@@ -0,0 +1,840 @@
...
+
+For integer operations, no rounding is required.
+
+[#RVBNARTO]
+==== Rounding to Odd behavior in dot product mode
+
+Rounding to odd (RTO) is not part of the IEEE-754 standard (at least not until and including revision 2019).
```
omasanori: s/IEEE-754/IEEE 754 arithmetic/

src/zvdota.adoc
```diff
@@ -0,0 +1,840 @@
...
+// comment to force next notes to be left aligned
+Note:: A product is said to be infinite when it is the product between an infinity and a non-zero finite number.
+
+Note:: The invalid exception flag can be raised even if one of the operands is a quiet NaN (for example with `inf - inf + qNaN`, or `inf * 0 + qNaN`)
+
+
+[#RVBNAOverflow]
+===== Overflow
+
+The overflow flag is raised according to the IEEE-754 definition:
```
omasanori: s/IEEE-754/IEEE 754-2008/

omasanori (commented):
See <https://github.com/riscv/riscv-isa-manual/issues/2906> also for data types.

src/zvdota.adoc
```diff
@@ -0,0 +1,830 @@
...
+with 0 indicating E4M3 and 1 indicating E5M2.
+These instructions are reserved unless SEW=8.
+The `vd` operand has EEW=4*SEW.
+
+----
+# Zvfqwdota8f OCP FP8 dot-product instructions
+# format of vs1 supplied by altfmt (0 E4M3, 1 E5M2)
+vfqwdota.vv     vd, vs2, vs1, vm   # vd[0] += E4M3(vs2) dot vs1
+vfqwdota.alt.vv vd, vs2, vs1, vm   # vd[0] += E5M2(vs2) dot vs1
+----
```
omasanori: While the both OCP FP8 and OFP8 are defined in the scalar FP chapter now, OFP8 seems preferable to me as it is shorter and OCP here does not add information.

src/zvdota.adoc
```diff
@@ -0,0 +1,830 @@
...
+These instructions are reserved unless SEW=8 and LMUL=1.
+The `vs1` and `vs2` operands have EEW=SEW, and the `vd` operand has
+EEW=4*SEW.
+
+----
+# Zvfqwbdota8f OCP FP8 batched dot-product instructions
+# format of vs1 supplied by altfmt (0 E4M3, 1 E5M2)
+vfqwbdota.vv     vd, vs2, vs1, ci, vm   # vs2 E4M3
+vfqwbdota.alt.vv vd, vs2, vs1, ci, vm   # vs2 E5M2
+----
```
omasanori: Ditto.

src/zvdota.adoc
aswaterman: Agreed.

src/zvdota.adoc
```diff
@@ -0,0 +1,840 @@
...
+        let A_i_isZero = (A_i_isSub && A_i_mant == 0)
+        let B_i_isZero = (B_i_isSub && B_i_mant == 0)
+        let prod_isZero = A_i_isZero || B_i_isZero
+
+        // detecting corner cases
+        let A_i_isInf = (A_i_exp == maskExp) && (A_i_mant == 0)
+        let B_i_isInf = (B_i_exp == maskExp) && (B_i_mant == 0)
+        let A_i_isNaN = (A_i_exp == maskExp) && (A_i_mant != 0)
+        let B_i_isNaN = (B_i_exp == maskExp) && (B_i_mant != 0)
+        let A_i_isSNaN = A_i_isNaN && (A_i_mant & (1 << (m_l - 1))) == 0
```
rnax: I have a question regarding the signaling NaN checking flow in RVBNA.

Because the previous comment in this functional description of RVBNA
> // A[i] and B[i] are IEEE 754-encoded binary floating point numbers on (e+p) bits

I think the OCP FP8 (E4M3/E5M2) types won’t use the same approach to define the signaling NaN in RVBNA.
If this is correct, will the specification define the interpretation of the mantissa values for all OCP FP8 NaNs?

aswaterman: RVBNA is written to describe IEEE 754-like formats. OFP8 formats don’t quite match IEEE 754. You should just do the obvious thing in those cases (implement isNaN etc. as appropriate for the given format) rather than literally following the RVBNA text.

cc @nibrunieAtSi5

nibrunieAtSi5: I agree,
we have already a non-normative note which mentions some way to generalize (mostly for the reference exponent part)

> The RVBNA pseudocode assumes that operands are represented in the IEEE 754-2008 binary encoding, but this is not actually a requirement of RVBNA. The pseudocode can easily be generalized to other floating-point encodings (e.g. Open Compute OFP formats). The actual requirement is to used biased exponent to compute the reference exponent and not real exponent (in particular because we want to avoid normalizing subnormal operands before computing the reference exponent). The significand also need to be materialized correctly for the product (including possible implicit digit(s)).

we could refine the note a bit to make suggestions, but it is going to be difficult to cover every existing and future formats.

rnax: Thanks for the reply @aswaterman @nibrunieAtSi5 

My purpose here is to make sure whether the Zvdota/Zvbdota/RNBNA specifications define the OCP FP8 sNaN/qNaN encoding (for instance, do all OCP FP8 NaNs signal or remain silent, or adhere to IEEE 754 standards?).

In current Spike implementations (branch: `master` commit: [c567a19](https://github.com/riscv-software-src/riscv-isa-sim/tree/c567a199a7dbd9c0f5689e0567415c76d9a21ade)), all of the OCP FP8 NaNs are quiet NaNs.
https://github.com/riscv-software-src/riscv-isa-sim/blob/c567a19/riscv/bulknormdot.h#L135-L136

So I am curious to understand whether this encoding is defined by the specification or is an implementation-defined.

aswaterman: The Zvfofp8min spec defines the canonical OFP8 NaN.  Maybe it should also remark that all OFP8 NaNs are quiet NaNs?  @nibrunieAtSi5 if you agree (or if you have a better idea) can you make a PR against the zvfofp8min branch?

pz9115:
Would it make sense to define family-level roll-up names in this ISA spec,
similar to the B and Zk extension model?

For example:

  Zvdota  -> Zvqwdota8i + Zvqwdota16i + Zvfwdota16bf + Zvfqwdota8f

  Zvbdota -> Zvqwbdota8i + Zvqwbdota16i + Zvfwbdota16bf + Zvfqwbdota8f + Zvfbdota32f

The motivation is not to replace the individual subextensions.  Those
remain important because implementations may reasonably support only part
of the family, such as int8 dot-product for quantized inference, int16
dot-product for DSP/fixed-point workloads, or BF16/FP8 dot-product for
AI workloads.  The roll-up names would just provide a convenient way to
request the full family when that is what the user or library wants.

PR #2627, opened 2026-01-30
Fixed spelling of Zicfilp in priv-cfi.adoc

davidharrishmc:


UmerShahidengr:
@davidharrishmc update in .yaml script will also be required. It is failing CI due to that

UmerShahidengr:
Correction: PR #2626 has updated .yaml script. Either we have to merge both at the same time or merge them into a single PR to pass the CI

PR #2631, opened 2026-01-30
Clarify CSR privilege encoding convention for hypervisor CSRs.

### LINKED ISSUE #2630: [Clarification] - Privilege level encoding for hypervisor CSR access
The privilege level encoding described in the table here
https://docs.riscv.org/reference/isa/priv/priv-intro.html#privlevels has `10` as a reserved encoding, and `01` as encoding the Supervisor (S) privilege.

The Hypervisor privilege modes (https://docs.riscv.org/reference/isa/priv/hypervisor.html#HPrivModes) do not discuss encoding, but use the `S` as the nominal privilege for HS and VS modes.

The CSR Address Mapping Conventions (https://docs.riscv.org/reference/isa/priv/priv-csrs.html#2-1-csr-address-mapping-conventions) talk about the privilege level encoding for CSR access control
```
The next two bits (csr[9:8]) encode the lowest privilege level that can access the CSR.
```
The CSR listing (https://docs.riscv.org/reference/isa/priv/priv-csrs.html#2-2-csr-listing) use a value of `10` for `csr[9:8]` for Hypervisor and VS CSRs.  

Taken literally, the lowest privilege level that can access the Hypervisor CSRs is either a reserved privilege (10) or Machine mode (11), making them inaccessible to the HS and VS privilege levels that are (presumably) encoded with `01`.

Clearly, this reading is incorrect.  Perhaps this has been clarified somewhere, but I couldn't find it.

pmundkur:
Fixes #2630.

### DIFF
--- a/src/priv-csrs.adoc
+++ b/src/priv-csrs.adoc
@@ -28,7 +28,7 @@ The standard RISC-V ISA sets aside a 12-bit encoding space (csr[11:0])
 for up to 4,096 CSRs. By convention, the upper 4 bits of the CSR address
 (csr[11:8]) are used to encode the read and write accessibility of the
 CSRs according to privilege level as shown in <<csrrwpriv>>. [#norm:Zicsr_rw]#The top two bits (csr[11:10]) indicate whether the register is read/write (`00`,`01`, or `10`) or read-only (`11`).#
-[#norm:Zicsr_access]#The next two bits (csr[9:8]) encode the lowest privilege level that can access the CSR.#
+[#norm:Zicsr_access]#The next two bits (csr[9:8]) encode the lowest privilege level that can access the CSR, with the pattern `10` representing hypervisor CSRs.#
 
 [NOTE]
 ====


Timmmm:
I don't know if this is actually any clearer - it's obvious that `0b10` means hypervisor CSRs from the table. What I actually meant was that it isn't totally obvious which permission modes "hypervisor CSRs" can be accessed from - there isn't a "hypervisor" privilege mode. I'd expect something like

> Hypervisor CSRs can only be accessed from HS mode or M mode.

Or maybe just a full table of which CSRs can be accessed from which privilege modes (including HS, VS, etc.) would be most clear. Like

| CSR address [9:8] | M | S or HS | VS | U or VU | 
|-|-|-|-|-|
| 0b00 | Ok | Ok | Ok | Ok |
| 0b01 | Ok | Ok | Ok | - |
| 0b10 | Ok | Ok | - | - |
| 0b11 | Ok | - | - | - |

PR #2518, opened 2025-12-19
Adding ~2000 normative rules from CSC to main repo

james-ball-qualcomm:
I've just merged in all upstream changes and updated our normative rules to track the ISA manual changes (e.g., moving Sstc into the M/S/H chapters).

I'll do smaller PRs in the future as we get more chapters tagged by CSC folks and UET in Pakistan.

PR #2635, opened 2026-02-02
[vector-crypto] Clarifying relationship between Zvknhb and Zvknha

nibrunieAtSi5:
The current vector crypto specification does not mention any specific link between the two extensions. Their respective description implies that Zvknhb is a subset of Zvknha. This patch attempts at clarifying this relationship as simply as possible.

I have not modified any of the definition (**Zvknhb** still defines support for both SEW=32 and 64, and **Zvknha** defines support for SEW=32, other values being reserved encodings).


aswaterman (commented):
Do normative tags for extension dependence rules really make sense?   How could you write a test that tests this property?

nibrunieAtSi5:
> Do normative tags for extension dependence rules really make sense? How could you write a test that tests this property?

I will have to consult @james-ball-qualcomm and the authors of https://github.com/riscv/riscv-isa-manual/pull/2518, as I simply reproduced what was done for other extensions in vector-crypto: 

https://github.com/riscv/riscv-isa-manual/blob/b9f1438191c25c95e877e521431f949990f8aa85/normative_rule_defs/vector-crypto.yaml#L40-L48

radimkrcmar:
I think this would be better as a non-normative clarification, because Zvknha is already defined as subset of Zvknhb.
(The subset of Zvknhb is behaviorally indistinguishable from Zvknha.)

The change wouldn't have any effect on the ISA string requirements either... Normative rules should have a meaning.

topperc:
> I think this would be better as a non-normative clarification, because Zvknha is already defined as subset of Zvknhb. (The subset of Zvknhb is behaviorally indistinguishable from Zvknha.)
> 
> The change wouldn't have any effect on the ISA string requirements either... Normative rules should have a meaning.

It does have a meaning to the toolchain. We need to know if -march=zvknhb should set the `__riscv_zvknha` preprocessor define for software. Currently it does not, this PR would change that.

radimkrcmar:
> It does have a meaning to the toolchain. We need to know if -march=zvknhb should set the __riscv_zvknha preprocessor define for software. Currently it does not, this PR would change that.

Zvknhb already implies Zvknha, so the normative rule would change nothing.  The behavior defined by Zvknha is a subset of the behavior defined by Zvknhb, therefore any implementation with Zvknhb always has Zvknha.

I think -march=zvknhb should set __riscv_zvknha as I see very little reason to be stricter than the ISA mandates.

topperc:
> > It does have a meaning to the toolchain. We need to know if -march=zvknhb should set the __riscv_zvknha preprocessor define for software. Currently it does not, this PR would change that.
> 
> Zvknhb already implies Zvknha, so the normative rule would change nothing. The behavior defined by Zvknha is a subset of the behavior defined by Zvknhb, therefore any implementation with Zvknhb always has Zvknha.
> 
> I think -march=zvknhb should set __riscv_zvknha as I see very little reason to be stricter than the ISA mandates.

I agree it should set `__riscv_zvkha`. Having an explicit normative rule makes that much more obvious and makes it easier to autogenerate or automatically verify the toolchains do this correctly. Needing to expand the instruction list for Zvknhb and figure out that Zvknha is a subset of it is more complicated.

topperc (approved):
Oops I accidentally put a comment I meant for the equivalent LLVM PR here, but I can't delete it.

PR #2638, opened 2026-02-02
Added pre-commit check for norm tagged text changes and ref files to compare against.

rpsene:
Changes initially made by James Ball from Qualcomm via https://github.com/riscv/riscv-isa-manual/pull/2512/changes/7291cabb22f8ea93875c01201d0584ebe31d63a6. 
                                                                        
  Discussed Requirement: Warning not error - spec authors can merge PRs                                                            
  Status: ✅ Covered                                                                                                               
  Implementation: Workflow uses continue-on-error: true on check step. Only modifications/deletions fail the check, additions pass.
                                                                                                                                   
    Bypass label (normative-change-approved) allows approved changes.                                                              
  ────────────────────────────────────────                                                                                         
  Discussed Requirement: GitHub issues created on merge (not on PR open)                                                           
  Status: ✅ Covered                                                                                                               
  Implementation: create-tag-change-issue.js runs only on push events to configured branches (main, ci-dev)                        
  ────────────────────────────────────────                                                                                         
  Discussed Requirement: Issues have distinct label                                                                                
  Status: ✅ Covered                                                                                                               
  Implementation: Issues use NormRules label (line 5 in create-tag-change-issue.js)                                                
  ────────────────────────────────────────                                                                                         
  Discussed Requirement: Comment added to PR if mismatch detected                                                                  
  Status: ✅ Covered                                                                                                               
  Implementation: comment-pr-changes.js posts detailed change report to PRs                                                        
  ────────────────────────────────────────                                                                                         
  Discussed Requirement: Include check script output in issue/comment                                                              
  Status: ✅ Covered                                                                                                               
  Implementation: Both PR comments and issues include the full $TAG_CHANGES output from detect_tag_changes.rb                      
  ────────────────────────────────────────                                                                                         
  Discussed Requirement: New tags auto-added to ref files on merge                                                                 
  Status: ✅ Covered                                                                                                               
  Implementation: Lines 123-144 in workflow update ref files and commit changes automatically                                      
  ────────────────────────────────────────                                                                                         
  Discussed Requirement: CSC folks track changes, not spec maintainers                                                             
  Status: ✅ Covered                                                                                                               
  Implementation: Issues are assigned to rpsene (CSC), spec maintainers can merge with warnings                                    
  Additional Features Implemented Beyond Discussion:                                                                               
                                                                                                                                   
  1. Bypass mechanism - The normative-change-approved label allows intentional normative changes to pass                           
  2. Duplicate prevention - Both PR comments and issues check for existing entries before creating new ones                        
  3. Configurable trigger branches - ISSUE_TRIGGER_BRANCHES env var allows easy configuration                                      
  4. Local script - scripts/check-tag-changes.sh for local pre-commit validation                                                   
  5. Detailed documentation - ref/README.md explains the entire workflow


.github/scripts/create-tag-change-issue.js
```diff
@@ -0,0 +1,66 @@
+module.exports = async ({ github, context, tagChanges }) => {
+    const owner = context.repo.owner;
+    const repo = context.repo.repo;
+    const assignees = ['james-ball-qualcomm', 'jordancarlin'];
```
james-ball-qualcomm: Should we assign this to some CSC email list that we setup just for this purpose? Could use the same facility you have for help@riscv.org or I guess a new groups.io list for this purpose. Then we could add/remove people to the email list over time without having to update this script.

.github/scripts/create-tag-change-issue.js
```diff
@@ -0,0 +1,66 @@
...
+    const labels = ['NormRules'];
```
james-ball-qualcomm: Since multiple labels are allowed, could we keep this NormRules label and also add one that uniquely identifies that this script created the issue such as NormRulesPRChange or something similar/better?

.github/workflows/check-normative-tags.yml
```diff
@@ -0,0 +1,178 @@
...
+      - name: Check for tag changes
+        id: check
+        continue-on-error: true
+        run: |
+          set +e  # Don't exit on error, we want to capture the output
+
+          echo "## Normative Tag Change Report" > tag_changes.txt
+          echo "" >> tag_changes.txt
+
+          echo "### Unprivileged Specification" >> tag_changes.txt
```
james-ball-qualcomm: Instead of duplicating for unpriv and priv, can you make it a function and call it twice or keep the existing code but run it in a loop with unpriv and priv in an array? I'm thinking ahead to the plans to add more volumes to the ISA manual and am thinking we don't want to have to copy & paste this code for new volumes.

.github/workflows/check-normative-tags.yml
```diff
@@ -0,0 +1,178 @@
...
+          echo "$PRIV_OUTPUT" | tee -a tag_changes.txt
+          echo '```' >> tag_changes.txt
+
+          # Save output for later steps
+          echo "TAG_CHANGES<<EOF" >> $GITHUB_ENV
+          cat tag_changes.txt >> $GITHUB_ENV
+          echo "EOF" >> $GITHUB_ENV
+
+          # Check if ANY changes were detected (additions, modifications, or deletions)
+          HAS_CHANGES=false
```
james-ball-qualcomm: Instead of grepping for messages, should we change the detect_tags_changes.rb script to provide a more informative exit status? Here's the code that creates it now and it would be easy to change/enhance:

Image

ref/README.md
```diff
@@ -0,0 +1,84 @@
+# Normative Tag Reference Files
```
james-ball-qualcomm, line 1: Sweet!

scripts/check-tag-changes.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
```
james-ball-qualcomm, line 1: Is this script still being used? If yes, no problem. If no, do we want to keep it?

james-ball-qualcomm (changes requested):

.github/scripts/create-tag-change-issue.js
rpsene: We can add any label, I will add one like "Script Generated".

.github/scripts/create-tag-change-issue.js
rpsene: We can't allocate to a team or mailing list. We need users.

p.s: GH does not support allocating an issue to a team, this is feature request. https://github.com/dear-github/dear-github/issues/170

.github/workflows/check-normative-tags.yml
rpsene: Yes! Will adjust it.

scripts/check-tag-changes.sh
rpsene: if contributors just run make build-tags and the CI does the real checking, the script is redundant, I will remove it.

.github/scripts/create-tag-change-issue.js
jordancarlin: Instead of adding assignees, we could mention a GitHub team that is created for this purpose using something like “@riscv/csc” in the issue body.

james-ball-qualcomm: A GitHub team sounds like a good idea.

james-ball-qualcomm: In the issue body.

.github/scripts/create-tag-change-issue.js
james-ball-qualcomm: Yes

.github/workflows/check-normative-tags.yml
james-ball-qualcomm: Thx

scripts/check-tag-changes.sh
james-ball-qualcomm: I want to be able to have a makefile target that does the same checking at the CI. So, I think this means we keep the script. I just want to ensure that a "make" or "make all" won't run the script. Ideally a "make build-norm-rules" will run the script.

rpsene: done!

.github/workflows/check-normative-tags.yml
rpsene: Done!

.github/workflows/check-normative-tags.yml
rpsene: Done!

rpsene: Done!

.github/scripts/create-tag-change-issue.js
rpsene: Done!

rpsene:
@james-ball-qualcomm please, review.

Makefile
```diff
@@ -146,6 +146,9 @@ build-pdf: $(DOCS_PDF)
 build-html: $(DOCS_HTML)
 build-epub: $(DOCS_EPUB)
 build-tags: $(DOCS_NORM_TAGS)
+check-tags:
+	@./scripts/check-tag-changes.sh
+
 build-norm-rules-json: $(NORM_RULES_JSON)
 build-norm-rules-html: $(NORM_RULES_HTML)
 build-norm-rules: build-norm-rules-json build-norm-rules-html
```
james-ball-qualcomm: Please add the check-tags dependency to the build-norm-rules rule. This way if someone does a "make" or "make all" it won't run the tag check and potentially cause the make to fail (and block them). However, if someone explicitly says "make build-norm-rules" then it will run the tag check which seems the right choice.

rpsene: Done!

PR #2639, opened 2026-02-03
Add weekly stale-branch monitor workflow

rpsene:
A weekly GitHub Action scans all branches via the GitHub GraphQL API, calculates the age of the last commit, and opens an issue for any branch older than 30 days. 

Issues are labeled stale-branch, assigned to aswaterman and wmat, and include deletion steps. 

To avoid duplicates, each issue embeds a hidden marker with the branch name and commit SHA; if an issue already exists for that exact branch+SHA it’s reused (reopened if needed). 

The workflow also supports an exclude list (with wildcards) and defaults to the repo’s default branch for deletion instructions.

### DIFF
--- a/.github/workflows/stale-branches.yml
+++ b/.github/workflows/stale-branches.yml
@@ -1,0 +1,288 @@
+---
+name: Stale Branch Monitor
+on:
+  schedule:
+    - cron: '0 2 * * 1'
+  workflow_dispatch:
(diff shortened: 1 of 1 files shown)


wmat:
Will this remove branches? Just on my phone so can't test but we beed our release branches to persist on riscv-isa-manual repo.

PR #2646, opened 2026-02-04
Zicfiss SSE clarifications and corrections

Timmmm:
1. The spec says when `henvcfg.SSE` is 0 then `senvcfg.SSE` is read-only and reads as 0. However it also says when `menvcfg.SSE` is 0, then `henvcfg.SSE` and `senvcfg.SSE` "are read-only zero". That is both very unclear (does it actually change their underlying values?) and inconsistent with `henvcfg.SSE`'s behaviour. I suspect it was a mistake. This changes them both to be "read-only and read as 0".

2. The `ssp` CSR access conditions redundently say either `henvcfg.SSE=0` OR `senvcfg.SSE=0` raise an exception in VU mode. This is redundant becahse `henvcfg.SSE=0` causes `senvcfg.SSE` to read as 0 (see above), so you only need to check `senvcfg.SSE`. Simplifying this case also makes the link between this check and xSSE slightly clearer.

3. Execution of the Zicfiss instructions (including SSAMOSWAP) depends on S-mode being implemented, but access to the CSR doesn't. That isn't very obvious so I made this explicit.

4. The "When S-mode is not implemented then `xSSE` is 0 at both M and U privilege modes." sentence doesn't make any mention of VS or VU modes. I assume they were added later and this sentence was overlooked. In any case `xSSE` is simply 0 in all modes if S-mode is not implemented so it's clearer to say that. Same for the change in the NOTE.

I think the behaviour of `SSE` is still not described clearly (I would expect many implementation bugs here!) but I believe this at least makes it consistent. For reference I believe the simplest description of the behaviour would be this mostly valid Sail:

```
// "Raw" value of xenvcfg.SSE according to the current privilege.
// This is not the same as xSSE from the ISA manual becuase it is 1 for
// machine mode and it is not 0 if supervisor is not implemented.
function zicfiss_xSSE_raw(priv : Privilege) -> bool =
  bool_bits(match priv {
    Machine           => 0b1
    Supervisor        => menvcfg[SSE],
    VirtualSupervisor => read_henvcfg()[SSE], // The `read_` takes care of the "reads as 0".
    User              => read_senvcfg()[SSE],
    VirtualUser       => read_senvcfg()[SSE],
  })

// xSSE from the ISA manual. 0 in machine mode or if S-mode is not implemented.
function zicfiss_xSSE(priv : Privilege) -> bool =
  priv != Machine & extensionEnabled(Ext_S) & zicfiss_xSSE_raw()

// Return true if we should raise a virtual instruction exception instead
// of an illegal instruction.
function zicfiss_exception_is_virtual(priv : Privilege) -> bool =
  menvcfg[SSE] == 0b1 & (priv == VirtualSupervisor | priv == VirtualUser)

function access_ssp(priv : Privilege) = {
    if not(zicfiss_xSSE_raw(priv))
    then return (if zicfiss_exception_is_virtual(priv) then Virtual_Instruction else Illegal_Instruction);
    // ... ok
}

function execute_sspush() = {
    if not(zicfiss_xSSE(cur_privilege)) then return RETIRE_SUCCESS;

    // ... normal execution
}

function execute_ssamoswap() = {
    if not(zicfiss_xSSE(priv))
    then return (if zicfiss_exception_is_virtual(priv) then Virtual_Instruction else Illegal_Instruction);

    // ... normal execuion
}
```

### DIFF
--- a/normative_rule_defs/priv-cfi.yaml
+++ b/normative_rule_defs/priv-cfi.yaml
@@ -30,7 +30,7 @@ normative_rule_definitions:
   - name: Zicfiss_ssp_vs_mode
     tags: ["norm:zicfiss_vs_henvcfg_sse", "norm:zicfiss_ssp_csr"]
   - name: Zicfiss_ssp_vu_mode
-    tags: ["norm:zicfiss_vu_henvcfg_senvcfg_sse", "norm:zicfiss_ssp_csr"]
+    tags: ["norm:zicfiss_vu_senvcfg_sse", "norm:zicfiss_ssp_csr"]
   - name: Zicfiss_ssp_hs_modes
     tags: ["norm:zicfiss_sse_access", "norm:zicfiss_ssp_csr"]
   - name: Zicfiss_smode_xsse
--- a/ref/riscv-spec-norm-tags.json
+++ b/ref/riscv-spec-norm-tags.json
@@ -2772,9 +2772,9 @@
     "norm:zicfiss_m_menvcfg_sse": "If the privilege mode is less than M and csr:menvcfg[sse] is 0, an\nillegal-instruction exception is raised.",
     "norm:zicfiss_u_senvcfg_sse": "Otherwise, if in U-mode and csr:senvcfg[sse] is 0, an illegal-instruction\nexception is raised.",
     "norm:zicfiss_vs_henvcfg_sse": "Otherwise, if in VS-mode and csr:henvcfg[sse] is 0, a virtual-instruction\nexception is raised.",
-    "norm:zicfiss_vu_henvcfg_senvcfg_sse": "Otherwise, if in VU-mode and either csr:henvcfg[sse] or csr:senvcfg[sse] is 0,\na virtual-instruction exception is raised.",
+    "norm:zicfiss_vu_senvcfg_sse": "Otherwise, if in VU-mode and csr:senvcfg[sse] is 0,\na virtual-instruction exception is raised.",
     "norm:zicfiss_sse_access": "Otherwise, the access is allowed.",
-    "norm:zicfiss_smode_xsse": "When S-mode is not implemented, then xSSE is 0 at both M and U privilege modes.",
+    "norm:zicfiss_smode_xsse": "When S-mode is not implemented, then xSSE is 0.",
     "norm:ss_page_enc": "The\nencoding R=0, W=1, and X=0, is defined to represent an SS page.",
     "norm:ssmp_menvcfg_sse": "When\ncsr:menvcfg[sse]=0, this encoding remains reserved.",
     "norm:ssmp_henvcfg_sse": "Similarly, when V=1 and\ncsr:henvcfg[sse]=0, this encoding remains reserved at VS and VU levels.",
@@ -12372,7 +12372,7 @@
                           "norm:zicfiss_m_menvcfg_sse",
                           "norm:zicfiss_u_senvcfg_sse",
                           "norm:zicfiss_vs_henvcfg_sse",
-                          "norm:zicfiss_vu_henvcfg_senvcfg_sse",
+                          "norm:zicfiss_vu_envcfg_sse",
                           "norm:zicfiss_sse_access"
                         ]
                       },
--- a/src/priv/cfi.adoc
+++ b/src/priv/cfi.adoc
@@ -138,9 +138,11 @@ csr:envcfg[sse] fields. The conditions are specified as follows:
   exception is raised.#
 * [#norm:zicfiss_vs_henvcfg_sse]#Otherwise, if in VS-mode and csr:henvcfg[sse] is 0, a virtual-instruction
   exception is raised.#
-* [#norm:zicfiss_vu_henvcfg_senvcfg_sse]#Otherwise, if in VU-mode and either csr:henvcfg[sse] or csr:senvcfg[sse] is 0,
+* [#norm:zicfiss_vu_senvcfg_sse]#Otherwise, if in VU-mode and csr:senvcfg[sse] is 0,
   a virtual-instruction exception is raised.#
 * [#norm:zicfiss_sse_access]#Otherwise, the access is allowed.#
+
+Access to the `ssp` CSR is not contingent on S-mode being implemented.
 
 ===== Shadow-Stack-Enabled (SSE) State
 
@@ -161,7 +163,7 @@ When S-mode is implemented, it is determined as follows:
 |===
 
 [[norm:zicfiss_smode_xsse]]
-When S-mode is not implemented, then `xSSE` is 0 at both M and U privilege modes.
+When S-mode is not implemented, then `xSSE` is 0.
 
 [NOTE]
 ====
@@ -194,7 +196,7 @@ illegal-instruction exception. Execution of programs that use these instructions
 on such machines is not supported.
 
 Activating ext:zicfiss[] in M-mode is currently not supported. Additionally, when
-S-mode is not implemented, activation in U-mode is also not supported. These
+S-mode is not implemented, activation is not supported in any privilege mode. These
 functionalities may be introduced in a future standard extension.
 ====
 
--- a/src/priv/hypervisor.adoc
+++ b/src/priv/hypervisor.adoc
@@ -828,7 +828,7 @@ apply when `V=1`:
 * 32-bit Zicfiss instructions will revert to their behavior as defined by Zimop.
 * 16-bit Zicfiss instructions will revert to their behavior as defined by Zcmop.
 * The `pte.xwr=010b` encoding in VS-stage page tables becomes reserved.
-* The `senvcfg.SSE` field will read as zero and is read-only.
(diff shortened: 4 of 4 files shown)


Timmmm:
@aswaterman for context @pmundkur is adding Zicfiss support to the Sail model: https://github.com/riscv/sail-riscv/pull/1408

ved-rivos:
1. The wording means the same. Its read only and on read returns a value of 0. Whether the implementation masks on write or masks on read is implementation specific. The CSR field modulation rules allow both forms of implementation. When menvcfg.SSE is cleared the henvcfg.SSE and senvcfg.SSE can assume a legal value - either 0 or 1. 

2. The OR is catering to the case where henvcfg.SSE is 1 and senvcfg.SSE is 0. 

3. When S mode is not implemented, the Zicfiss extension cannot be implemented. When Zicfiss extension is not implemented, there is no ssp CSR implemented.  Likewise, ssamoswap is not implemented when Zicfiss is not implemented.

4. When S mode is not implemented, the VS and VU modes do not exist and hence mentioning VS and VU modes is not required.

Timmmm:
> 1. The wording means the same.

I don't know I think saying a value "is" 0 and it "reads as" 0 seem quite different. I get the intention, but in any case I think it's better to use consistent wording and the wording from the hypervisor part of the spec is clearer.

> 2. The OR is catering to the case where henvcfg.SSE is 1 and senvcfg.SSE is 0.

It says it's illegal when `henvcfg.SSE` is 0, OR `senvcfg.SSE` is 0. However the former implies the latter so the condition is equivalent to only checking `senvcfg.SSE`.

> 3. When S mode is not implemented, the Zicfiss extension cannot be implemented. When Zicfiss extension is not implemented, there is no ssp CSR implemented. Likewise, ssamoswap is not implemented when Zicfiss is not implemented.

I don't think the spec says that anywhere. It says Zicfiss can't be "activated" (via xSSE) if S-mode is not implemented. Also the "use" of Zicfiss in some modes isn't supported:

> Moreover, use of Zicfiss in U-mode requires S-mode to be implemented. Use of Zicfiss in M-mode is not supported.

But I think that's talking about executing instructions.

> 4. When S mode is not implemented, the VS and VU modes do not exist and hence mentioning VS and VU modes is not required.

Good point! Still simpler just to not mention any modes though.

ved-rivos:
1. It's okay to reword it as read-only-zero. It does not change the
allowance made by the architecture to do either write masking or read
masking.

2. While the former implies the latter, the latter does not imply the
former. But its fine to only test senvcfg.SSE as the behavior is the same.

3. Yes, I misspoke about that. Zicfiss cannot be used when S-mode is not
implemented. There is no use case for implementing Zicfiss for M or M+U,
but it can be implemented. However, if it is implemented, xSSE is always
0 in such cases, and all shadow stack instructions besides ssamoswap - 
including ssrdp, which accesses the ssp--revert to being Zimop or Zcmop. 
ssamoswap will raise an illegal instruction exception. This is because a 
virtual memory system is required for shadow stack instructions to have
any sensible behavior, and virtual memory is not supported unless S-mode
is implemented. Access to the ssp CSR is, however, not gated on S-mode
being implemented. If Zicfiss is implemented and S-mode is not 
implemented, then access from U-mode will be either allowed or disallowed
as determined by menvcfg.SSE.

4. It's fine to reword it that way. I hope your confusion about VS and
VU modes not being mentioned was cleared up.

ved-rivos (approved):

src/machine.adoc
```diff
@@ -2343,7 +2343,7 @@ the following rules apply to privilege modes that are less than M:#
 * `SSAMOSWAP.W/D` raises an illegal-instruction exception.
 
 [#norm:menvcfg_sse_rdonly0]#When `menvcfg.SSE` is 0, the `henvcfg.SSE` and `senvcfg.SSE` fields are
-read-only zero.#
+read-only and read as zero.
```
aswaterman: "read-only zero" is standard terminology in this manual (it shows up 102 times) and it means exactly this.  If it's confusing here, then presumably it's confusing everywhere, and the right thing to do would seem to define it somewhere central.  But that should be a matter for a different PR.

allenjbaum:
There are 4 different ways of handling how write of non-zero values can be
done to "RdOnly0" state whose behavior can change to be writable dependent
on the value of a different field:
 - writes set the underlying state, but can't be seen by reads, until some
kind of CSR value update unmasks the bit and allows must recent write (read
masking)
- writes are ignored, and unmasking will find the original value is retained
- writes are ignored, and the underlying state is zeroed
- writes are ignored, and the underlying state is UNSPECIFIED when unmasked.

I think the latter is most prevalent in the spec (and I think is default
unless otherwise speced), though there are other cases (but I can't point
to them easily)

src/machine.adoc
Timmmm: Fair point. I think only a subset of those are the sort of "read-only zero" where it "is read-only and reads as 0". A lot of them are *always* read-only zero, e.g.

> medeleg[11] is read-only zero

> Bits of mie that are not writable must be read-only zero.

I think it *is* subtle and somewhat confusing and probably could be improved (maybe hyperlink them to the CSR modulation section and add an explanation of what exactly it means?). Probably something for another PR.

In this PR I think we should at least use the same wording for both `menvcfg.SSE` and `henvcfg.SSE` though. Should I change the `henvcfg.SSE` one to `read-only zero` instead?

allenjbaum:
Yes, unless there is a situation where a field value (including a single
bit) can change from legal to illegal and vice versa,
then read-only is (as Andrew said) just as it says, with no state behind
the read.

src/machine.adoc
ved-rivos: > Should I change the henvcfg.SSE one to read-only zero instead?

Yes

Timmmm: Done.

Timmmm:
Updated & I updated the commit message to reflect the discussion. CI isn't passing because of an issue with the normative tag infrastructure: https://github.com/riscv/docs-resources/issues/199

Timmmm:
@aswaterman could this be merged if I fix the conflicts?

aswaterman (commented):
@Timmmm Yeah, ping me when it is ready to merge.

Timmmm:
@aswaterman Ok this is ready to go. Sorry for the delay; I forgot about it.

Also I'm not quite sure what is going on with `ref/riscv-spec-norm-tags.json` - if I run `make update-ref` it updates a load of things, but if I just cherry-pick the changes I've made (which is what I've committed in this PR) then the build passes, which suggests that the check to verify that it is up-to-date isn't working. @james-ball-qualcomm @rpsene

Timmmm:
Oh wait I think I see what has happened. The `detect_tag_changes.py` script strips asciidoc formatting before comparison. I think that probably shouldn't be done because it means `ref/riscv-spec-norm-tags.json` can fall out of date. If you do change the formatting of tags it should be pretty easy to review that. Maybe we can have a `MODIFIED (formatting only):` output. I'll make a separate PR for that.

aswaterman:
Don't forget that you need to go back and sign this commit (both with the `Signed-off-by` line in the commit message and by having uploaded your public key to github as a signing key).

I guess the PR is old enough at this point that both of those requirements were added after its creation.

Timmmm:
Ah I thought that check was talking about the `signed-off-by` thing. It seems like it has an error posting a comment to say that I haven't signed the commit: `HttpError: Resource not accessible by integration`.

Anyway I'll sign it.

aswaterman:
> Ah I thought that check was talking about the `signed-off-by` thing. It seems like it has an error posting a comment to say that I haven't signed the commit: `HttpError: Resource not accessible by integration`.

@wmat can you look into this?

> Anyway I'll sign it.

Thanks, yes, I agree it is confusing that there are two kinds of signatures required and the built-in github message does not indicate the signature it is talking about is the SSH key.

rpsene:
> Also I'm not quite sure what is going on with `ref/riscv-spec-norm-tags.json` - if I run `make update-ref` it updates a load of things, but if I just cherry-pick the changes I've made (which is what I've committed in this PR) then the build passes, which suggests that the check to verify that it is up-to-date isn't working. @james-ball-qualcomm @rpsene

I will check what is going on. If need be we can force-push the commit to bypass the normative rules change.

wmat (approved):
LGTM

wmat:
@Timmmm can you resign the original commits and force push?  GitHub doesn't want to let me ignore those failures.

Timmmm:
Success! @wmat

wmat:
Not sure if this matters here, but:  https://github.com/riscv/riscv-isa-manual/actions/runs/27022639674

Do normative tags require regeneration?

Timmmm:
Yeah they do (but not because of this PR). See [this comment](https://github.com/riscv/riscv-isa-manual/pull/2646#issuecomment-4506465059). I'll open an issue for that.

Timmmm:
https://github.com/riscv/riscv-isa-manual/issues/3186

PR #2648, opened 2026-02-05
Fix typos, grammar, and make a few things more consistent.

eak:
This a few small things I noticed with some greps and skimming. Since I haven't generated an ISA manual PR before, I wanted to try this first before doing on to other things (e.g., indirect-call).


aswaterman:
@wmat I asked Earl to push these even if he hadn't gotten the signature verification set up.  I'm going to manually incorporate them into another PR.

PR #2673, opened 2026-02-15
[Norm Tags] Updated Normative rules for scalar-crypto.adoc

UmerShahidengr:
scalar-crypto instructions normative rules are added, other rules are updated in scalar-crypto.adoc file.


mmhus:
@UmerShahidengr Please make sure to add the NormRules label as well, LGTM otherwise. I didn't check the rules themselves.

PR #2677, opened 2026-02-17
Fix broken character in RVC quadrant diagrams

arichardson:
The C.NOP (HINT, imm≠0) not-equal symbol was not being rendered correctly since asciidoctor-pdf falls back to only CP1252 characters if no explicit font is set and we end up with a logic not glyph (¬). See https://docs.asciidoctor.org/pdf-converter/latest/theme/font-support/

Explicitly set the font family for the :math attributes to fix this.

Found by building with asciidoctor-pdf with --verbose.

### DIFF
--- a/src/images/bytefield/rvc-instr-quad0.edn
+++ b/src/images/bytefield/rvc-instr-quad0.edn
@@ -1,6 +1,7 @@
 [bytefield]
 ----
 (defattrs :plain [:plain {:font-family "M+ 1p Fallback"}])
+(defattrs :math [:math {:font-family "M+ 1p Fallback"}])
 (def row-height 35 )
 (def row-header-fn nil)
 (def left-margin 0)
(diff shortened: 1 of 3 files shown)


arichardson:
After this change I see one more WARNING: that is printed with --verbose, which can be fixed by updating docs-resources (PR https://github.com/riscv/docs-resources/pull/189 / issue https://github.com/riscv/docs-resources/issues/187).

arichardson:
> ## Normative Rule Changes Detected
> 
> > **Note:** New tags (additions) are automatically added to the reference files when PRs are merged to main. Only modifications and deletions require manual review.
> 
> _This comment was automatically generated by the normative tag check workflow._

@rpsene this seems like a false-positive to me?

rpsene:
> > ## Normative Rule Changes Detected
> > > **Note:** New tags (additions) are automatically added to the reference files when PRs are merged to main. Only modifications and deletions require manual review.
> > 
> > 
> > _This comment was automatically generated by the normative tag check workflow._
> 
> @rpsene this seems like a false-positive to me?

@arichardson yes, a clear false positive. I will investigate and fix.

wmat:
@arichardson can you verify your signature for merges to happen?

arichardson:
> @arichardson can you verify your signature for merges to happen?

Sure let me push an update in a few hours!

arichardson:
> @arichardson can you verify your signature for merges to happen?

Should be done now.

Issue #2711, opened 2026-02-27
[NormRules] - Granularity of 4 misaligned load/store impldefs should be combined into 1 (too fine-grained)


james-ball-qualcomm:
<img width="2752" height="632" alt="Image" src="https://github.com/user-attachments/assets/4616df4a-323c-46fb-80eb-5af11d230bcc" />

(cc @riscv/csc)

james-ball-qualcomm:
CC: @dhower-qc, @jordancarlin, @davidharrishmc, @mmhus 

Since changing the normative rules potentially breaks downstream tools, I'm wondering if we really need to change these 4 impldefs into 1 just because we want one parameter for all 4 of them. The parameter can point to all 4 norm rules. How about I just leave this the way it is and put in all 4 norm rule names into the RVI20 CRD's MISALIGNED_LDST parameter.

I'm creating an adoc include file in the crd directory (in the ACT repo) that contains a table mapping the parameter names I'm using across all CRDs to their impldefs. I am making links to the latest-greatest norm rules in https://riscv.github.io/riscv-isa-manual/snapshot/norm-rules/norm-rules.html but am not thrilled with that solution since it is pretty fragile (i.e., changes to norm rules and names will break the CRD).

<img width="2007" height="705" alt="Image" src="https://github.com/user-attachments/assets/c1d3a824-6797-41c1-b721-66fe74ef0e44" />

jordancarlin:
I think that the norm rules are still early enough in their lifecycle that it is worth changing them to simplify things where we can. Downstream users (currently just the CTP as far as I know) can be updated accordingly. As much as possible, I want to push for one-to-one mappings between impdefs and parameters, and this seems like a natural spot for it to be a single impdef.

dhower-qc:
> I think that the norm rules are still early enough in their lifecycle that it is worth changing them to simplify things where we can. 

+1

PR #2712, opened 2026-02-28
LR_SC_ALIGNMENT now identified as a impldef and combined for misaligned load/store impldefs into one

### LINKED ISSUE #2710: [NormRules] - lr_sc_alignment should be tagged as an impldef
In the Parameters SIG meeting today, we discovered that the lr_sc_alignment norm rule allows choices so should be an impldef. Here's its text with **bold** showing the choice allowed.

```
For LR and SC, the Zalrsc extension requires that the address held in rs1
be naturally aligned to the size of the operand (i.e., eight-byte
aligned for doublewords and four-byte aligned for words). **If the
address is not naturally aligned, an address-misaligned exception or an
access-fault exception will be generated**. The access-fault exception can
be generated for a memory access that would otherwise be able to
complete except for the misalignment, if the misaligned access should
not be emulated.
```

(cc @riscv/csc)
CC: @jordancarlin, @davidharrishmc, @dhower-qc

james-ball-qualcomm:
Fixes #2710 and #2711


james-ball-qualcomm:
Copied from issues:

CC: @dhower-qc, @jordancarlin, @davidharrishmc, @mmhus

Since changing the normative rules potentially breaks downstream tools, I'm wondering if we really need to change these 4 impldefs into 1 just because we want one parameter for all 4 of them. The parameter can point to all 4 norm rules. How about I just leave this the way it is and put in all 4 norm rule names into the RVI20 CRD's MISALIGNED_LDST parameter.

I'm creating an adoc include file in the crd directory (in the ACT repo) that contains a table mapping the parameter names I'm using across all CRDs to their impldefs. I am making links to the latest-greatest norm rules in https://riscv.github.io/riscv-isa-manual/snapshot/norm-rules/norm-rules.html but am not thrilled with that solution since it is pretty fragile (i.e., changes to norm rules and names will break the CRD).

image

james-ball-qualcomm:
Replacing MISALIGNED_LDST_FULLY_HW_SUPPORTED, MISALIGNED_LDST_INVISIBLE_TRAP, and MISALIGNED_LDST_HW_OR_INVISIBLE_TRAP_FUNC_OF_ADDR into MISALIGNED_LDST_HW_SUPPORT as an enum with 3 values: full, none, and func_of_addr.

Combining MISALIGNED_LDST_FULLY_HW_SUPPORTED_OR_VISIBLE_TRAP and LDST_ADDR_MISALIGNED_OR_ACCESS_FAULT_EXC into MISALIGNED_LDST_EXC_TYPE.

@jordancarlin, can you please make the corresponding changes in the ACT repo for combining the 4 misaligned load/store norm rules into 1? 

CC @dhower-qc

james-ball-qualcomm:
After using this parameter in the CRDs, I'm thinking we change it to this:

```
MISALIGNED_LDST_HW_SUPPORT as a Boolean

Description: Hardware support for misaligned load/store instructions. Excludes atomic instructions which are covered by the MISALIGNED_ATOMIC parameter.

Values:
  - False: No hardware support for misaligned load/store instructions
  - True: Misaligned load/store instructions are supported in hardware (potentially with restrictions expressed by other parameters).
```

jordancarlin:
> After using this parameter in the CRDs, I'm thinking we change it to this:
> 
> ```
> MISALIGNED_LDST_HW_SUPPORT as a Boolean
> 
> Description: Hardware support for misaligned load/store instructions. Excludes atomic instructions which are covered by the MISALIGNED_ATOMIC parameter.
> 
> Values:
>   - False: No hardware support for misaligned load/store instructions
>   - True: Misaligned load/store instructions are supported in hardware (potentially with restrictions expressed by other parameters).
> ```

I don't think that matters for this PR. The values that the parameter can have will be decided in the Parameters SIG. This PR just updates the normative rule, which feeds into the parameter.

james-ball-qualcomm:
Yes, in this PR we just need to agree on the name of the impldef norm rule since we want the parameter name to match. So, this is why I'm thinking about the type and allowed values of the parameter since it will influence the parameter's name and hence the impldef norm rule's name. So, are we okay with the name MISALIGNED_LDST_HW_SUPPORT? I think MISALIGNED_LDST isn't descriptive enough.

jordancarlin:
We used `HW_MISALIGNED_LDST_SUPPORTED` in https://riscv.atlassian.net/wiki/spaces/PRMT/pages/1368784897/RVI20+Parameters.

james-ball-qualcomm:
HW_MISALIGNED_LDST_SUPPORTED  is inconsistent with all the other misaligned load/store parameter names (MISALIGNED_LDST_*) in UDB. So, I think we should be using MISALIGNED_LDST_HW_SUPPORT

james-ball-qualcomm:
Just noticed that lr_reservation_set_size should also be flagged as an impldef to be referenced by the LRSC_RESERVATION_STRATEGY parameter.

james-ball-qualcomm:
And now noticed that normative rule amo_misaligned_atomicity_granule is an impldef since need to know the size of the NAPOT region (0 if not supported). Here's an example of the Unpriv ISA referencing the Priv ISA M-mode chapter. Oh well. I'll change the unpriv amo_misaligned_atomicity_granule norm rule to be an impldef.

james-ball-qualcomm:
The norm rule name amo_misaligned_atomicity_granule is misleading since this isn't just for AMOs. So, renaming to MISALIGNED_ATOMICITY_GRANULE_SIZE

mmhus (approved):

PR #2721, opened 2026-03-04
Allow ELEN>VLEN in Vector extension 

DmitryUtyansky:
Change/amend wording and formulas in Vector extension to allow ELEN>VLEN configurations, along the lines agreed upon in Vector SIG. To allow e.g. ELEN=64, VLEN=32 extensions.

### DIFF
--- a/src/v-st-ext.adoc
+++ b/src/v-st-ext.adoc
@@ -4908,8 +4908,8 @@ e q r d c b v a    # v11 destination after vrgather using viota.m under mask
 [#norm:vmv-nr-r_op]#The `vmv<nr>r.v` instructions copy whole vector registers (i.e., all
 VLEN bits) and can copy whole vector register groups.  The `nr` value
 in the opcode is the number of individual vector registers, NREG, to
-copy.  The instructions operate as if EEW=min(VLEN, SEW), EMUL = NREG, and effective
-length `evl`= EMUL * VLEN/EEW.#
+copy.  The instructions operate as if EEW = min(VLEN, SEW), EMUL = NREG, and effective
+length `evl` = EMUL * VLEN/EEW.#
 
 NOTE: These instructions are intended to aid compilers to shuffle
 vector registers without needing to know or change `vl`.
--- a/src/v-st-ext.adoc
+++ b/src/v-st-ext.adoc
@@ -3930,9 +3930,9 @@ destination format is converted to the destination format's largest finite value
 === Vector Reduction Operations
 
 [#norm:vreduction_scalar_def]#Vector reduction operations take a vector register group of elements
-and a scalar held in element 0 of a vector register (or a vector register group), and perform a
+and a scalar held in element 0 of a vector register group, and perform a
 reduction using some binary operator, to produce a scalar result in
-element 0 of a vector register (or a vector register group).#  [#norm:vreduction_scalar_disregard_LMUL]#The scalar input and output operands
+element 0 of a vector register group.#  [#norm:vreduction_scalar_disregard_LMUL]#The scalar input and output operands
 are held in element 0 of a group with EMUL=ceil(EEW/VLEN) regardless of LMUL setting.#
 
 NOTE: The previous edition specified use of a single vector register (EMUL=1). EMUL = ceil(EEW/VLEN) allows for a case
--- a/src/v-st-ext.adoc
+++ b/src/v-st-ext.adoc
@@ -27,9 +27,16 @@ Standard vector extensions (<<sec-vector-extensions>>) and
 architecture profiles may set further constraints on _ELEN_ and _VLEN_.
 
 NOTE: Future extensions may allow ELEN {gt} VLEN by holding one
-element using bits from multiple vector registers. In this case ELEN/VLEN
-ratio, the number of vector registers required to hold one ELEN-sized
-element, becomes greater than 1.
+element using bits from multiple vector registers. The formulas in this
+specification have been updated to allow this. The changes only
+affect ELEN {gt} VLEN cases, when ELEN {le} VLEN nothing is changed.
+E.g.: EMUL for "scalar in vector" inputs/outputs of reductions is now
+defined as ceil(EEW/VLEN), so it is still 1 (specifying a single register,
+as before) when EEW {le} VLEN. Element width for whole vector register
+moves and loads is now defined as min(VLEN, EEW), giving unchanged
+EEW when EEW {le} VLEN. Therefore, nothing is changed for all standard
+extensions defined in this specification, all of which require
+ELEN {le} VLEN.
 
 NOTE: The upper limit on VLEN allows software to know that indices
 will fit into 16 bits (largest VLMAX of 65,536 occurs for LMUL=8 and
@@ -316,11 +323,6 @@ consider the use of this case to be __reserved__.
 
 NOTE: It is recommended that assemblers provide a warning (not an
 error) if a `vsetvli` instruction attempts to write an LMUL < SEW~MIN~/min(ELEN, VLEN).
-
-NOTE: The above formulas have been updated to use min(ELEN, VLEN)
-instead of ELEN in the previous edition of the specification, to allow
-future vector extensions to have ELEN > VLEN. For extensions where ELEN {le} VLEN,
-min(ELEN,VLEN) = ELEN and so all formulas and rules remain unchanged.
 
 [[norm:lmul]]
 LMUL is set by the signed `vlmul` field in `vtype` (i.e., LMUL =
@@ -3936,9 +3938,6 @@ reduction using some binary operator, to produce a scalar result in
 element 0 of a vector register group.#  [#norm:vreduction_scalar_disregard_LMUL]#The scalar input and output operands
 are held in element 0 of a group with EMUL=ceil(EEW/VLEN) regardless of LMUL setting.#
 
-NOTE: The previous edition specified use of a single vector register (EMUL=1). EMUL = ceil(EEW/VLEN) allows for a case
-of EEW/VLEN > 1. If EEW < VLEN, ceil(EEW/VLEN) = 1, resulting in EMUL=1, a single vector, as before.
-
 [#norm:vreduction_vd_overlap_vs]#The destination vector register can overlap the source operands,
 including the mask register.#
 
@@ -4536,10 +4535,6 @@ The integer scalar read/write instructions transfer a single
 value between a scalar `x` register and element 0 of a vector
 register group with EMUL = ceil(EEW/VLEN).  [#norm:vmv-x-s_vmv-s-x_ignoreLMUL]#The instructions ignore LMUL, EMUL is computed as ceil(EEW/VLEN).#
 
-NOTE: The previous edition specified that these instructions ignore LMUL and vector register groups.
-The change to EMUL = ceil(EEW/VLEN) allows for the case of EEW > VLEN. If EEW {le} VLEN, EMUL = ceil(EEW/VLEN) = 1, as before.
-
-
 ----
 vmv.x.s rd, vs2  # x[rd] = vs2[0] (vs1=0)
 vmv.s.x vd, rs1  # vd[0] = x[rs1] (vs2=0)
@@ -4572,9 +4567,6 @@ and `vmv.s.x` are reserved.#
 The floating-point scalar read/write instructions transfer a single
 value between a scalar `f` register and element 0 of a vector
 register group with EMUL=ceil(EEW/VLEN).  [#norm:vfmv-f-s_vfmv-s-f_ignoreLMUL]##The instructions ignore LMUL, EMUL is computed as ceil(EEW/VLEN).#
-
-NOTE: The previous edition specified that these instructions ignore LMUL and vector register groups.
-The change to EMUL=ceil(EEW/VLEN) allows for the case of EEW/VLEN > 1. If EEW {le} VLEN, EMUL = ceil(EEW/VLEN) = 1, as before.
 
 ----
 vfmv.f.s rd, vs2  # f[rd] = vs2[0] (rs1=0)
--- a/src/v-st-ext.adoc
+++ b/src/v-st-ext.adoc
@@ -21,14 +21,15 @@ Each hart supporting a vector extension defines two parameters:
 
 . [#norm:elen]#The maximum size in bits of a vector element that any operation can produce or consume, _ELEN_ {ge} 8, which
 must be a power of 2.#
-. [#norm:vlen]#The number of bits in a single vector register, _VLEN_ {ge} ELEN, which must be a power of 2, and must be no greater than 2^16^.#
+. [#norm:vlen]#The number of bits in a single vector register, _VLEN_, which must be a power of 2, and must be no greater than 2^16^.#
 
 Standard vector extensions (<<sec-vector-extensions>>) and
 architecture profiles may set further constraints on _ELEN_ and _VLEN_.
 
 NOTE: Future extensions may allow ELEN {gt} VLEN by holding one
-element using bits from multiple vector registers, but this
-extension does not include this option.
+element using bits from multiple vector registers. In this case ELEN/VLEN 
+ratio, the number of vector registers required to hold one ELEN-sized 
+element, becomes greater than 1. 
 
 NOTE: The upper limit on VLEN allows software to know that indices
 will fit into 16 bits (largest VLMAX of 65,536 occurs for LMUL=8 and
@@ -280,15 +281,19 @@ register-resident vectors.
 Implementations must provide fractional LMUL settings that allow the
 narrowest supported type to occupy a fraction of a vector register
 corresponding to the ratio of the narrowest supported type's width to
-that of the largest supported type's width.  In general, the
-requirement is to support LMUL {ge} SEW~MIN~/ELEN, where SEW~MIN~ is
-the narrowest supported SEW value and ELEN is the widest supported SEW
-value.  In the standard extensions, SEW~MIN~=8.  For
+that of the largest supported type's width or to the vector register length, 
+if the latter is smaller than the largest supported type (ELEN/VLEN > 1).  
+In general, the requirement is to support LMUL {ge} SEW~MIN~/min(ELEN, VLEN), 
+where SEW~MIN~ is the narrowest supported SEW value, ELEN is the widest 
+supported SEW value and VLEN is the number of bits in a vector register.  
+In the standard extensions, SEW~MIN~=8.  For
 standard vector extensions with ELEN=32, fractional LMULs of 1/2 and
-1/4 must be supported.  For standard vector extensions with ELEN=64,
+1/4 must be supported.  For standard vector extensions with ELEN=64 and ELEN/VLEN {le} 1,
 fractional LMULs of 1/2, 1/4, and 1/8 must be supported.
+For a vector extensions with SEW~MIN~=8, ELEN=64 and VLEN=32, fractional LMULs of 1/2 and 1/4 
+must be supported.
 
-NOTE: When LMUL < SEW~MIN~/ELEN, there is no guarantee
+NOTE: When LMUL < SEW~MIN~/min(ELEN, VLEN), there is no guarantee
 an implementation would have enough bits in the fractional vector
 register to store at least one element, as VLEN=ELEN is a
 valid implementation choice. For example, with VLEN=ELEN=32,
@@ -297,20 +302,25 @@ storage in a vector register.
 
 [[norm:vtype_sew_val]]
 For a given supported fractional LMUL setting, implementations must support
-SEW settings between SEW~MIN~ and LMUL * ELEN, inclusive.
+SEW settings between SEW~MIN~ and LMUL * min(ELEN, VLEN), inclusive. 
 
 [[norm:vtype_lmul_fval_rsv]]
-The use of `vtype` encodings with LMUL < SEW~MIN~/ELEN is
+The use of `vtype` encodings with LMUL < SEW~MIN~/min(ELEN, VLEN) is
 __reserved__, but implementations can set `vill` if they do not
 support these configurations.
 
 NOTE: Requiring all implementations to set `vill` in this case would
 prohibit future use of this case in an extension, so to allow for a
-future definition of LMUL<SEW~MIN~/ELEN behavior, we
+future definition of LMUL<SEW~MIN~/min(ELEN, VLEN) behavior, we
 consider the use of this case to be __reserved__.
 
 NOTE: It is recommended that assemblers provide a warning (not an
-error) if a `vsetvli` instruction attempts to write an LMUL < SEW~MIN~/ELEN.
+error) if a `vsetvli` instruction attempts to write an LMUL < SEW~MIN~/min(ELEN, VLEN).
+
+NOTE: The above formula have been updated to use min(ELEN, VLEN) 
+instead of ELEN in the previous edition of the specification, to allow 
+future vector extensions to have ELEN > VLEN. For extensions where ELEN {le} VLEN, 
+min(ELEN,VLEN) = ELEN and so all formula and rules remain unchanged.
 
 [[norm:lmul]]
 LMUL is set by the signed `vlmul` field in `vtype` (i.e., LMUL =
@@ -776,6 +786,12 @@ lowest-numbered vector register and moving to the
 next-highest-numbered vector register in the group once each vector
 register is filled.
 
+If a vector extension supports EEW/VLEN > 1, one element can span multiple 
+vector registers and to have EEW bits to hold one element the requirement is that 
+the LMUL {ge} EEW/VLEN. In this case 
+each element occupies multiple consecutive vector registers, the least 
+significant bits of each element go to the lower-numbered vector register.
+
 ----
  LMUL > 1 examples
 
@@ -834,6 +850,14 @@ register is filled.
  v4*n+1              7       6       5       4
  v4*n+2              B       A       9       8
  v4*n+3              F       E       D       C
+
+ VLEN=32b, SEW=64b, LMUL=4
+
+ Byte         3 2 1 0
+ v4*n               0
+ v4*n+1             
+ v4*n+2             1
+ v4*n+3             
 ----
 
 [[sec-mapping-mixed]]
@@ -1033,6 +1057,8 @@ LMUL=8 is reserved as this would imply a result EMUL=16.
 Widened scalar values, e.g., input and output to a widening reduction
 operation, are held in the first element of a vector register and
 have EMUL=1.
+If a vector extension supports EEW/VLEN > 1, EEW-wide widened scalar 
+values are held in a vector register group with EMUL = EEW/VLEN
 
 ==== Vector Masking
 
@@ -2079,6 +2105,12 @@ Hence, it would have sufficed to provide only EEW=8 variants.
 The full set of EEW variants is provided so that the encoded EEW can be used
 as a hint to indicate the destination register group will next be accessed
 with this EEW, which aids implementations that rearrange data internally.
+
+In vector extensions supporting ELEN/VLEN {gt} 1 element size can exceed 
+the number of bits available in a vector register or a vector register 
+group. In that case whole vector register load and store instructions 
+still operate on the specified number of vector register(s), using the 
+least significant bits of the element.
 
 The vector whole register store instructions are encoded similar to
 unmasked unit-stride store of elements with EEW=8.
@@ -3898,12 +3930,13 @@ destination format is converted to the destination format's largest finite value
 === Vector Reduction Operations
 
 [#norm:vreduction_scalar_def]#Vector reduction operations take a vector register group of elements
-and a scalar held in element 0 of a vector register, and perform a
+and a scalar held in element 0 of a vector register (or a vector register group), and perform a
 reduction using some binary operator, to produce a scalar result in
-element 0 of a vector register.#  [#norm:vreduction_scalar_disregard_LMUL]#The scalar input and output operands
-are held in element 0 of a single vector register, not a vector
-register group, so any vector register can be the scalar source or
-destination of a vector reduction regardless of LMUL setting.#
+element 0 of a vector register (or a vector register group).#  [#norm:vreduction_scalar_disregard_LMUL]#The scalar input and output operands
+are held in element 0 of a group with EMUL=ceil(EEW/VLEN) regardless of LMUL setting.#
+
+NOTE: The previous edition specified use of a single vector register (EMUL=1). EMUL = ceil(EEW/VLEN) allows for a case 
+of EEW/VLEN > 1. If EEW < VLEN, ceil(EEW/VLEN) = 1, resulting in EMUL=1, a single vector, as before.
 
 [#norm:vreduction_vd_overlap_vs]#The destination vector register can overlap the source operands,
 including the mask register.#
@@ -4500,7 +4533,11 @@ around within the vector registers.
 
 The integer scalar read/write instructions transfer a single
 value between a scalar `x` register and element 0 of a vector
-register.  [#norm:vmv-x-s_vmv-s-x_ignoreLMUL]#The instructions ignore LMUL and vector register groups.#
+register group with EMUL = ceil(EEW/VLEN).  [#norm:vmv-x-s_vmv-s-x_ignoreLMUL]#The instructions ignore LMUL, EMUL is computed as ceil(EEW/VLEN).#
+
+NOTE: The previous edition specified that these instructions ignore LMUL and vector register groups. 
+The change to EMUL = ceil(EEW/VLEN) allows for the case of EEW > VLEN. If EEW {le} VLEN, EMUL = ceil(EEW/VLEN) = 1, as before.
+
 
 ----
 vmv.x.s rd, vs2  # x[rd] = vs2[0] (vs1=0)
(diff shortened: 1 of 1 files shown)


omasanori:
This change seems actual revision to hardware-software contracts. Will it bump the version of V extension to 1.01, 1.1, 2.0 or whatever?

aswaterman (changes requested):
My recollection from having discussed this with @kasanovic is that it will come in the form of new extensions, not revisions of the existing ones.

DmitryUtyansky:
@aswaterman , indeed new extensions will be required to specify the configurations; existing "V" and zve* remain intact). But the baseline specification must allow that (as of now it directly prohibits). We talked about that with @kasanovic and in Vector SIG. But of course there is a question how to best put that to the spec. I am not sure that my version is the best possible, certainly this requires a thorough review, and I'd appreciate any feedback/suggestions.

aswaterman:
@DmitryUtyansky I will provide more feedback sometime down the road, but at first blush this appears to be on the right track (modulo the extensionology matter).

src/v-st-ext.adoc
```diff
@@ -2080,6 +2106,12 @@ The full set of EEW variants is provided so that the encoded EEW can be used
 as a hint to indicate the destination register group will next be accessed
 with this EEW, which aids implementations that rearrange data internally.
 
+In vector extensions supporting ELEN/VLEN {gt} 1 element size can exceed
```
topperc:
```suggestion
In vector extensions supporting ELEN/VLEN {gt} 1, element size can exceed
```

src/v-st-ext.adoc
```diff
@@ -2080,6 +2106,12 @@ The full set of EEW variants is provided so that the encoded EEW can be used
...
+the number of bits available in a vector register or a vector register
+group. In that case whole vector register load and store instructions
```
topperc: Whole register store instructions don't have an element size encoded.

src/v-st-ext.adoc
```diff
@@ -4531,8 +4569,11 @@ and `vmv.s.x` are reserved.#
 ==== Floating-Point Scalar Move Instructions
 
 The floating-point scalar read/write instructions transfer a single
-value between a scalar `f` register and element 0 of a vector
-register.  [#norm:vfmv-f-s_vfmv-s-f_ignoreLMUL]#The instructions ignore LMUL and vector register groups.#
+value between a scalar `f` register and element 0 of avector
```
topperc:
```suggestion
value between a scalar `f` register and element 0 of a vector
```

DmitryUtyansky:
@omasanori , I am not sure what's the best approach with the versioning. The extensions defined in the document (V, zvl*, zve*) are not changed (at least that's the idea, unless I messed up somewhere). Existing contracts shall stay intact. But the text is changed anyway.

Alternatively, one could think of more clustered/localized changes, giving the updated formulas only for ELEN>VLEN cases and leaving the existing paragraphs intact. But overall document is then changed all the same.

src/v-st-ext.adoc
DmitryUtyansky: Thanks for spotting! I removed "and store".

src/v-st-ext.adoc
```diff
@@ -4878,6 +4919,10 @@ in the opcode is the number of individual vector registers, NREG, to
 copy.  The instructions operate as if EEW=SEW, EMUL = NREG, effective
 length `evl`= EMUL * VLEN/SEW.#
 
+If a vector extension supports SEW/VLEN > 1, the calculated `evl` can be fractional, less than 1.
```
aswaterman: We don't need the conditionalization on "if a vector extension supports [...]"; we can just specify the behavior unconditionally.  The new semantics implicitly do not apply to implementations that don't have ELEN > VLEN.

src/v-st-ext.adoc
```diff
@@ -4878,6 +4919,10 @@ in the opcode is the number of individual vector registers, NREG, to
...
+In this case the instruction shall still copy EMUL = NREG vector registers, transferring the LSBs of the element.
+The fractional `evl` is still valid for the purposes of checking for `vstart` {ge} `evl`.
```
aswaterman: This works, but I wonder if this is the best definition.  The reason we parameterize the behavior on SEW is to facilitate implementations that internally recode vector registers based on EEW.  But if we're only transferring part of an element, it doesn't make sense microarchitecturally to characterize the destination register as having that EEW; really it has the truncated EEW.

A simpler definition might be to say that "The instructions operate as if EEW=min(VLEN, SEW), EMUL=NREG, effective length `evl`= EMUL * VLEN/EEW."

src/v-st-ext.adoc
```diff
@@ -4532,7 +4570,10 @@ and `vmv.s.x` are reserved.#
 
 The floating-point scalar read/write instructions transfer a single
 value between a scalar `f` register and element 0 of a vector
-register.  [#norm:vfmv-f-s_vfmv-s-f_ignoreLMUL]#The instructions ignore LMUL and vector register groups.#
+register group with EMUL=ceil(EEW/VLEN).  [#norm:vfmv-f-s_vfmv-s-f_ignoreLMUL]##The instructions ignore LMUL, EMUL is computed as ceil(EEW/VLEN).#
+
+NOTE: The previous edition specified that these instructions ignore LMUL and vector register groups.
+The change to EMUL=ceil(EEW/VLEN) allows for the case of EEW/VLEN > 1. If EEW {le} VLEN, EMUL = ceil(EEW/VLEN) = 1, as before.
```
aswaterman: I think we can get rid of the local non-normative notes and have a global explanation towards the beginning.

src/v-st-ext.adoc
```diff
@@ -4500,7 +4533,11 @@ around within the vector registers.
 
 The integer scalar read/write instructions transfer a single
 value between a scalar `x` register and element 0 of a vector
-register.  [#norm:vmv-x-s_vmv-s-x_ignoreLMUL]#The instructions ignore LMUL and vector register groups.#
+register group with EMUL = ceil(EEW/VLEN).  [#norm:vmv-x-s_vmv-s-x_ignoreLMUL]#The instructions ignore LMUL, EMUL is computed as ceil(EEW/VLEN).#
+
+NOTE: The previous edition specified that these instructions ignore LMUL and vector register groups.
+The change to EMUL = ceil(EEW/VLEN) allows for the case of EEW > VLEN. If EEW {le} VLEN, EMUL = ceil(EEW/VLEN) = 1, as before.
```
aswaterman: https://github.com/riscv/riscv-isa-manual/pull/2721/changes#r2887222374

src/v-st-ext.adoc
```diff
@@ -3898,12 +3930,13 @@ destination format is converted to the destination format's largest finite value
...
 reduction using some binary operator, to produce a scalar result in
-element 0 of a vector register.#  [#norm:vreduction_scalar_disregard_LMUL]#The scalar input and output operands
-are held in element 0 of a single vector register, not a vector
-register group, so any vector register can be the scalar source or
-destination of a vector reduction regardless of LMUL setting.#
+element 0 of a vector register (or a vector register group).#  [#norm:vreduction_scalar_disregard_LMUL]#The scalar input and output operands
+are held in element 0 of a group with EMUL=ceil(EEW/VLEN) regardless of LMUL setting.#
+
+NOTE: The previous edition specified use of a single vector register (EMUL=1). EMUL = ceil(EEW/VLEN) allows for a case
+of EEW/VLEN > 1. If EEW < VLEN, ceil(EEW/VLEN) = 1, resulting in EMUL=1, a single vector, as before.
```
aswaterman: https://github.com/riscv/riscv-isa-manual/pull/2721/changes#r2887222374

src/v-st-ext.adoc
```diff
@@ -3898,12 +3930,13 @@ destination format is converted to the destination format's largest finite value
 === Vector Reduction Operations
 
 [#norm:vreduction_scalar_def]#Vector reduction operations take a vector register group of elements
-and a scalar held in element 0 of a vector register, and perform a
+and a scalar held in element 0 of a vector register (or a vector register group), and perform a
```
aswaterman: "vector register (or a vector register group)" => "vector register group"

src/v-st-ext.adoc
```diff
@@ -2080,6 +2106,12 @@ The full set of EEW variants is provided so that the encoded EEW can be used
 as a hint to indicate the destination register group will next be accessed
 with this EEW, which aids implementations that rearrange data internally.
 
+In vector extensions supporting ELEN/VLEN {gt} 1, element size can exceed
+the number of bits available in a vector register or a vector register
+group. In that case whole vector register load instructions
+still operate on the specified number of vector register(s), using the
+least significant bits of the element.
```
aswaterman: If we agree with my comment on whole-register moves, the same probably applies here.

aswaterman (commented):
Seems like we're on the right track.

~Since we are globally relaxing the ELEN <= VLEN requirement, we should explicitly change the definitions of the Zve32x etc. extensions to require ELEN <= VLEN.  That keeps the existing extensions functionally unchanged.~

kasanovic:
Table 60 in Section 30.18.2 already explicitly lists the ELEN <= VLEN requirement for each extension.

aswaterman:
Ah right, thanks.

src/v-st-ext.adoc
DmitryUtyansky: I am not sure what "operate as if EEW=min(VLEN, SEW)" entails exactly for cases where these "element width hints" are really used. For VLEN<SEW, if that  internally recodes the register as if EEW=VLEN (while that was a vector of SEW) --- that will actually result in incorrect data, right?

Though I do not see a reason to use such instructions (with EEW>VLEN), it doesn't make a lot of sense to move part of an element. I would as well make this case "reserved". I do not have a strong opinion...

aswaterman: Any internal recoding is required to be ISA-invisible, so there's no risk of incorrect operation--or rather, if there were a risk of incorrect operation, it would mean the implementation doesn't conform to the ISA.

I don't think we should make this case reserved, as it risks causing headaches for compilers, which in some cases want to move vector registers without knowing or caring what the current SEW happens to be.  It is a legitimate thing to do to move a register elsewhere, then overwrite that register for some unrelated purpose, then restore its previous value--and it's fine that register happened to contain only half an element.  (This example would cause a performance hiccup on an implementation that internally recodes vector registers, but it would still be functionally correct.)

src/v-st-ext.adoc
DmitryUtyansky: This makes sense. I thought that people would see a formula changed and would wonder what's happening, and so having a local explanation with the old version of the formula should help. Still this does not look formally well-defined enough. Let me think more...

src/v-st-ext.adoc
DmitryUtyansky: Right, compilers and overall compatibility is the main argument for such moves and loads to be valid. 

Ok, probably more concise wording is better, I'll edit.

aswaterman: Cool, ping me after you've finished this round of edits and I'll re-review.

DmitryUtyansky:
@aswaterman , I moved the explanations into one big note in the beginning, and did other small edits. Now I believe the doc is good for review, please take a look.

src/v-st-ext.adoc
```diff
@@ -21,14 +21,22 @@ Each hart supporting a vector extension defines two parameters:
 
 . [#norm:elen]#The maximum size in bits of a vector element that any operation can produce or consume, _ELEN_ {ge} 8, which
 must be a power of 2.#
-. [#norm:vlen]#The number of bits in a single vector register, _VLEN_ {ge} ELEN, which must be a power of 2, and must be no greater than 2^16^.#
+. [#norm:vlen]#The number of bits in a single vector register, _VLEN_, which must be a power of 2, and must be no greater than 2^16^.#
```
aswaterman: I think we want `_VLEN_ {ge} 8` here.  I realize this constraint could be imposed by the new extensions we'll eventually define, but I suspect there will be some corner cases  in the spec that will result from 64-bit elements not being representable in an EMUL=8 register group.  (That, and a vector unit with a 16-byte regfile isn't terribly interesting.)

src/v-st-ext.adoc
```diff
@@ -21,14 +21,22 @@ Each hart supporting a vector extension defines two parameters:
...
+element using bits from multiple vector registers. The formulas in this
+specification have been updated to allow this. The changes only
+affect ELEN {gt} VLEN cases, when ELEN {le} VLEN nothing is changed.
+E.g.: EMUL for "scalar in vector" inputs/outputs of reductions is now
+defined as ceil(EEW/VLEN), so it is still 1 (specifying a single register,
+as before) when EEW {le} VLEN. Element width for whole vector register
+moves and loads is now defined as min(VLEN, EEW), giving unchanged
+EEW when EEW {le} VLEN. Therefore, nothing is changed for all standard
+extensions defined in this specification, all of which require
+ELEN {le} VLEN.
```
aswaterman: IMO, the "changelog" aspect of this is pretty redundant with the actual spec changes, and people who are interested enough can look up the git history.  Can I recommend something like

```
NOTE: Following the ratification of the V extension, this specification
has been revised to admit the possibility of future extensions that
allow ELEN > VLEN, wherein one element is held using bits from
multiple vector registers.
These relaxations have no impact on implementations with ELEN {le}
VLEN or on existing software that assumes ELEN {le} VLEN.
```

src/v-st-ext.adoc
```diff
@@ -280,15 +288,19 @@ register-resident vectors.
 Implementations must provide fractional LMUL settings that allow the
 narrowest supported type to occupy a fraction of a vector register
 corresponding to the ratio of the narrowest supported type's width to
-that of the largest supported type's width.  In general, the
-requirement is to support LMUL {ge} SEW~MIN~/ELEN, where SEW~MIN~ is
-the narrowest supported SEW value and ELEN is the widest supported SEW
-value.  In the standard extensions, SEW~MIN~=8.  For
+that of the largest supported type's width or to the vector register length,
+if the latter is smaller than the largest supported type (ELEN/VLEN > 1).
```
aswaterman: I'd rephrase this as "the smaller of VLEN and the largest supported type's width".  I realize it's basically a restatement of the formula below, but it's a bit less cumbersome to read.

src/v-st-ext.adoc
```diff
@@ -280,15 +288,19 @@ register-resident vectors.
...
+In general, the requirement is to support LMUL {ge} SEW~MIN~/min(ELEN, VLEN),
+where SEW~MIN~ is the narrowest supported SEW value, ELEN is the widest
+supported SEW value and VLEN is the number of bits in a vector register.
+In the standard extensions, SEW~MIN~=8.  For
 standard vector extensions with ELEN=32, fractional LMULs of 1/2 and
-1/4 must be supported.  For standard vector extensions with ELEN=64,
+1/4 must be supported.  For standard vector extensions with ELEN=64 and ELEN/VLEN {le} 1,
```
aswaterman: I think `VLEN {ge} ELEN` would read better here than `ELEN/VLEN {le} 1`.

src/v-st-ext.adoc
```diff
@@ -776,6 +788,12 @@ lowest-numbered vector register and moving to the
 next-highest-numbered vector register in the group once each vector
 register is filled.
 
+If a vector extension supports EEW/VLEN > 1, one element can span multiple
```
aswaterman: I'd express this as EEW > VLEN.

src/v-st-ext.adoc
aswaterman: BTW, I think in this document, you need to use the `>` and `<` symbols directly, but use `{le}` and `{ge}` for the inclusive cases.

src/v-st-ext.adoc
```diff
@@ -2080,6 +2109,12 @@ The full set of EEW variants is provided so that the encoded EEW can be used
 as a hint to indicate the destination register group will next be accessed
 with this EEW, which aids implementations that rearrange data internally.
 
+In vector extensions supporting ELEN/VLEN {gt} 1, element size can exceed
```
aswaterman: `ELEN > VLEN`

src/v-st-ext.adoc
```diff
@@ -4515,7 +4548,8 @@ ignored.  If SEW < XLEN, the value is sign-extended to XLEN bits.#
 NOTE: [#norm:vmv-x-s_vstartgevl_vl0]#`vmv.x.s` performs its operation even if `vstart` {ge} `vl` or `vl`=0.#
 
 [#norm:vmv-s-x_op]#The `vmv.s.x` instruction copies the scalar integer register to element 0 of
-the destination vector register.  If SEW < XLEN, the least-significant bits
+the destination vector register group with EMUL = ceil(EEW/VLEN)).
```
aswaterman: Not sure if having the spaces around the `=` sign looks better or worse (I guess no spaces is probably better?), but please make it consistent with your change below this one.

src/v-st-ext.adoc
```diff
@@ -4532,7 +4566,7 @@ and `vmv.s.x` are reserved.#
 
 The floating-point scalar read/write instructions transfer a single
 value between a scalar `f` register and element 0 of a vector
-register.  [#norm:vfmv-f-s_vfmv-s-f_ignoreLMUL]#The instructions ignore LMUL and vector register groups.#
+register group with EMUL=ceil(EEW/VLEN).  [#norm:vfmv-f-s_vfmv-s-f_ignoreLMUL]##The instructions ignore LMUL, EMUL is computed as ceil(EEW/VLEN).#
```
aswaterman: Semicolon instead of comma after "ignore LMUL"

src/v-st-ext.adoc
```diff
@@ -4875,8 +4909,8 @@ e q r d c b v a    # v11 destination after vrgather using viota.m under mask
 [#norm:vmv-nr-r_op]#The `vmv<nr>r.v` instructions copy whole vector registers (i.e., all
 VLEN bits) and can copy whole vector register groups.  The `nr` value
 in the opcode is the number of individual vector registers, NREG, to
-copy.  The instructions operate as if EEW=SEW, EMUL = NREG, effective
-length `evl`= EMUL * VLEN/SEW.#
+copy.  The instructions operate as if EEW=min(VLEN, SEW), EMUL = NREG, effective
```
aswaterman: "and" before "effective" (I realize this bug was present before your edits)

src/v-st-ext.adoc
```diff
@@ -3898,12 +3933,10 @@ destination format is converted to the destination format's largest finite value
...
 [#norm:vreduction_scalar_def]#Vector reduction operations take a vector register group of elements
-and a scalar held in element 0 of a vector register, and perform a
+and a scalar held in element 0 of a vector register group, and perform a
 reduction using some binary operator, to produce a scalar result in
-element 0 of a vector register.#  [#norm:vreduction_scalar_disregard_LMUL]#The scalar input and output operands
-are held in element 0 of a single vector register, not a vector
-register group, so any vector register can be the scalar source or
-destination of a vector reduction regardless of LMUL setting.#
+element 0 of a vector register group.#  [#norm:vreduction_scalar_disregard_LMUL]#The scalar input and output operands
+are held in element 0 of a group with EMUL=ceil(EEW/VLEN) regardless of LMUL setting.#
```
aswaterman: ", regardless of the LMUL setting."

(This is an old one, but it's better to fix it while we are here)

src/v-st-ext.adoc
```diff
@@ -2080,6 +2109,12 @@ The full set of EEW variants is provided so that the encoded EEW can be used
 as a hint to indicate the destination register group will next be accessed
 with this EEW, which aids implementations that rearrange data internally.
 
+In vector extensions supporting ELEN/VLEN {gt} 1, element size can exceed
+the number of bits available in a vector register or a vector register
+group. In that case whole vector register load instructions
```
aswaterman: "In that case, the whole"

src/v-st-ext.adoc
```diff
@@ -2080,6 +2109,12 @@ The full set of EEW variants is provided so that the encoded EEW can be used
...
+still operate on the specified number of vector register(s), using the
+least significant bits of the element.
```
aswaterman: Hyphenate "least-significant"

src/v-st-ext.adoc
```diff
@@ -1033,6 +1059,8 @@ LMUL=8 is reserved as this would imply a result EMUL=16.
 Widened scalar values, e.g., input and output to a widening reduction
 operation, are held in the first element of a vector register and
 have EMUL=1.
+If a vector extension supports EEW/VLEN > 1, EEW-wide widened scalar
+values are held in a vector register group with EMUL = EEW/VLEN
```
aswaterman: period at end of sentence

src/v-st-ext.adoc
```diff
@@ -1033,6 +1059,8 @@ LMUL=8 is reserved as this would imply a result EMUL=16.
 Widened scalar values, e.g., input and output to a widening reduction
 operation, are held in the first element of a vector register and
 have EMUL=1.
+If a vector extension supports EEW/VLEN > 1, EEW-wide widened scalar
```
aswaterman: I think this is another case where `EEW > VLEN` is clearer.

aswaterman (commented):
Thanks, I think this is getting close.  Ping me after this round of feedback is addressed.

src/v-st-ext.adoc
DmitryUtyansky: This aspect is arguable. I might be biased of course by being new to the process, but I think it's important for the document to be self-explanatory and more explicit regarding why the changes have no impact on ... . Maybe we can come up with more concise yet informative formulation. I will think more.

src/v-st-ext.adoc
DmitryUtyansky: Agree. Also individual register smaller than a byte seems strange.

src/v-st-ext.adoc
DmitryUtyansky: Thanks, I find this much better as well.

src/v-st-ext.adoc
DmitryUtyansky: I don't have a strong opinion, maybe without spaces is better. The existing document however uses both a lot. It appears that = with spaces is used in formulas and headers, and without spaces it's used in short "assignments", like VLEN=8. Unified to that for now...

src/v-st-ext.adoc
DmitryUtyansky: I agree. At some point I thought of introducing some variable for ELEN/VLEN to be used for parameterization, but did not find a good way, so ELEN {le} VLEN would be better.

src/v-st-ext.adoc
DmitryUtyansky: All right, I agree. The "changelog" is not complete anyway. 
@aswaterman , I believe I did all the changes, so please have one more look.

aswaterman: Sorry, I'm just getting back to this now, but I'm not going to disagree about us agreeing :)

src/v-st-ext.adoc
```diff
@@ -776,6 +782,12 @@ lowest-numbered vector register and moving to the
 next-highest-numbered vector register in the group once each vector
 register is filled.
 
+If a vector extension supports EEW > VLEN, one element can span multiple
+vector registers and to have EEW bits to hold one element the requirement is that
+the LMUL {ge} EEW/VLEN. In this case
+each element occupies multiple consecutive vector registers, the least
+significant bits of each element go to the lower-numbered vector register.
```
aswaterman: This paragraph says the right things, but I suggest reordering the consequences:

```
If a vector extension supports EEW > VLEN, one element can span multiple
vector registers, in which case the least-significant bits of the element
are held in the lowest-numbered vector register.
Instructions that access vector register groups with EMUL < EEW/VLEN are
reserved.
```

aswaterman (approved):
@DmitryUtyansky Thank you for hanging in there with my review; I think this is good to go after you address my last minor comment.  But I'd like to give @kasanovic the chance to weigh in.

src/v-st-ext.adoc
DmitryUtyansky: This looks fine to me as well (don't see much difference, maybe a bit shorter and so cleaner). I changed the text, thank you for the suggestion.

src/v-st-ext.adoc
```diff
@@ -21,14 +21,17 @@ Each hart supporting a vector extension defines two parameters:
...
 
-NOTE: Future extensions may allow ELEN {gt} VLEN by holding one
-element using bits from multiple vector registers, but this
-extension does not include this option.
+NOTE: Following the ratification of the V extension, this specification
+has been revised to admit the possibility of future extensions that
+allow ELEN > VLEN, wherein one element is held using bits from
+multiple vector registers.
+These relaxations have no impact on implementations with ELEN {le}
+VLEN or on existing software that assumes ELEN {le} VLEN.
```
kasanovic: We can make a stronger statement that "These relaxations have no impact on implementations or software using the ratified vector extensions, which require ELEN {le} VLEN".

src/v-st-ext.adoc
```diff
@@ -280,15 +283,18 @@ register-resident vectors.
...
+the smaller of VLEN and the largest supported type's width.
+In general, the requirement is to support LMUL {ge} SEW~MIN~/min(ELEN, VLEN),
+where SEW~MIN~ is the narrowest supported SEW value, ELEN is the widest
+supported SEW value and VLEN is the number of bits in a vector register.
+In the standard extensions, SEW~MIN~=8.  For
 standard vector extensions with ELEN=32, fractional LMULs of 1/2 and
-1/4 must be supported.  For standard vector extensions with ELEN=64,
+1/4 must be supported.  For standard vector extensions with ELEN=64 and ELEN {le} VLEN,
 fractional LMULs of 1/2, 1/4, and 1/8 must be supported.
+For a vector extensions with SEW~MIN~=8, ELEN=64 and VLEN=32, fractional LMULs of 1/2 and 1/4
```
kasanovic: There are no such standard vector extensions at this point, and given this follows sentences about standard vector extensions,  I would rephrase as "For a vector implementation with ..." or "For future vector extensions with..".  Alternatively, drop "standard" in preceding sentences.

src/v-st-ext.adoc
```diff
@@ -1033,6 +1053,8 @@ LMUL=8 is reserved as this would imply a result EMUL=16.
 Widened scalar values, e.g., input and output to a widening reduction
 operation, are held in the first element of a vector register and
 have EMUL=1.
+If a vector extension supports EEW > VLEN, EEW-wide widened scalar
```
kasanovic: I would say "vector implementation" not "vector extension".

src/v-st-ext.adoc
```diff
@@ -776,6 +782,12 @@ lowest-numbered vector register and moving to the
 next-highest-numbered vector register in the group once each vector
 register is filled.
 
+If a vector extension supports EEW > VLEN, one element can span multiple
```
kasanovic: "Vector extension" -> "vector implementation" throughout these changes.

src/v-st-ext.adoc
```diff
@@ -2069,8 +2091,9 @@ handlers, and OS context switches.  Software can determine the number
 of bytes transferred by reading the `vlenb` register.
 
 [[norm:vector_ls_seg_wholereg_eew]]
-The load instructions have an EEW encoded in the `mew` and `width`
+The load instructions have the element width encoded in the `mew` and `width`
 fields following the pattern of regular unit-stride loads.
+EEW is computed as EEW=min(VLEN, EEW_encoded).
```
kasanovic: EEW is used as a hint to microarchitecture and doesn't affect architectural behavior of these instructions - I don't think it should change in case some funny trick is used by micorarchitecture knowing that EEW > VLEN?

kasanovic (commented):
I think this is mostly OK but there are a few tweaks in wording , extension->implementation, and also the whole register move to update.

src/v-st-ext.adoc
DmitryUtyansky: I can see the underlying thinking.  But I find this terminology unclear a bit: isn't anything we're talking about here "vector extension" (like, certainly this is not an implementation of a geometrical vector or whatever)? And so perhaps "vector extension implementation" should be used instead of "vector implementation"?

src/v-st-ext.adoc
aswaterman: Dropping "standard" seems clean to me.

src/v-st-ext.adoc
aswaterman: How about just "If an implementation [...]"?  It's implicit we are talking about vectors in this context.

DmitryUtyansky: This sounds better to me: does not introduce the term "vector implementation". 
Strictly speaking, the existing document already uses it, once, in this Note below. I would change that to "implementation" as well then.

"The number of bits implemented in vl depends on the implementation’s maximum vector length of the smallest supported type. The smallest vector implementation with VLEN=32 and supporting SEW=8 would need at least six bits in vl to hold the values 0-32 (VLEN=32, with LMUL=8 and SEW=8, yields VLMAX=32)."

aswaterman: Let's leave the note alone.  It reads better as-is and it's not normative anyway.

src/v-st-ext.adoc
aswaterman: I think this is a clean definition that doesn't actually preclude funny tricks.  I suggested it because it's important we don't preclude vtype-unaware register spill/fill code from e.g. moving `v1` to `v2` when SEW > VLEN.  Without this definition, the reference to `v1` would make the instruction reserved.

A uarch is free to ignore this dictum in its trickery and represent the destination with whatever EEW it wants, e.g. it could compute its internal-representation EEW as a function of VLEN, EEW_encoded, and the register specifiers involved.

src/v-st-ext.adoc
DmitryUtyansky: Agree, this is better.

src/v-st-ext.adoc
DmitryUtyansky: Dropped "standard".

src/v-st-ext.adoc
DmitryUtyansky: I changed all such cases to "implementation" for now, except this note (where it remains "vector implementation").

src/v-st-ext.adoc
DmitryUtyansky: I deleted there the paragraph left by oversight after the earlier edit introducing "EEW=min(VLEN, EEW_encoded)." ("In implementations supporting ELEN > VLEN, element size can exceed the number of bits available in a vector register or a vector register group. In that case, the whole vector register load instructions
still operate on the specified number of vector register(s), using the
least-significant bits of the element.")

It's either one or the other, EEW as min(...), as @aswaterman suggested or "use LSBs of an unchanged bigger EEW" as it was originally.

The benefit of having EEW defined through min(...) is that the subsequent formulas `evl`=NFIELDS*VLEN/EEW works without giving fractional evl.

Thinking more about this, the formula with min(VLEN, EEW_encoded) for e.g. VLEN=32 EEW=64 breaks e.g. vl2re64.v: a perfectly valid "load pair of vregs with EEW=64" is now switched to EEW=32 (potentially misguiding all those "microarch hints").
I have changed the formula to "EEW=min(VLEN\*NFIELDS, EEW_encoded)", limiting EEW to whatever fits into the requested group.  Same logic applies to whole register move: I have changed the formula there as well, to EEW = min(VLEN\*EMUL, SEW), factoring in EMUL.

@aswaterman , @kasanovic , please tell me what you think of the current wording.

src/v-st-ext.adoc
DmitryUtyansky: Changed to "implementation" for now, I think it's better (not using a new term). I don't have a very strong opinion here, if you believe "vector implementation" is better -- we can use it.

src/v-st-ext.adoc
aswaterman: Good catch.  I think this scheme hangs together.  I'll chat with @kasanovic about this topic later this week and try to reach a conclusion.

DmitryUtyansky:
@aswaterman , @kasanovic , so what's your opinion, does this require any further refinement, what's the next step(s)?

aswaterman:
Sorry, I'll try to re-review this before the end of the month.  I think it is basically OK, but I want the chance to think it through carefully again.

src/v-st-ext.adoc
```diff
@@ -4500,7 +4521,7 @@ around within the vector registers.
 
 The integer scalar read/write instructions transfer a single
 value between a scalar `x` register and element 0 of a vector
-register.  [#norm:vmv-x-s_vmv-s-x_ignoreLMUL]#The instructions ignore LMUL and vector register groups.#
+register group with EMUL = ceil(EEW/VLEN).  [#norm:vmv-x-s_vmv-s-x_ignoreLMUL]#The instructions ignore LMUL, EMUL is computed as ceil(EEW/VLEN).#
```
aswaterman: semicolon instead of comma after LMUL

src/v-st-ext.adoc
```diff
@@ -4515,7 +4536,8 @@ ignored.  If SEW < XLEN, the value is sign-extended to XLEN bits.#
 NOTE: [#norm:vmv-x-s_vstartgevl_vl0]#`vmv.x.s` performs its operation even if `vstart` {ge} `vl` or `vl`=0.#
 
 [#norm:vmv-s-x_op]#The `vmv.s.x` instruction copies the scalar integer register to element 0 of
-the destination vector register.  If SEW < XLEN, the least-significant bits
+the destination vector register group with EMUL = ceil(EEW/VLEN)).
```
aswaterman: extra closing paren

aswaterman (commented):
I think this is ready to go after my minor comments are addressed.  Please ping me once you've done so.

Also, after merging, please delete the `riscv/riscv-isa-manual-dmitry` repo.  (In the normal github workflow, forks should be in your personal github area rather than the original one.)

DmitryUtyansky:
> I think this is ready to go after my minor comments are addressed. Please ping me once you've done so.

Thanks for getting back to this and spotting! I did the corrections.

PR #2739, opened 2026-03-11
Redid Zbkc section to mostly reference Zbc

bgrayson-mips:
This is for Issue #2523, but I had to redo this since signature verification has been implemented in the mean time.


bgrayson-mips:
@topperc here is the PR to follow.

PR #2743, opened 2026-03-12
Incorporate Zvfbfa spec

aswaterman:
This spec has passed internal review and ARC review; see https://lists.riscv.org/g/tech-unprivileged/message/1109


nibrunieAtSi5 (approved):
For those following at home, the spec is tracked by https://riscv.atlassian.net/browse/RVS-3748

@aswaterman what do you think about adding a ††† symbol after `vfrec7.v` to point to the end of chapter note ? It would making it easier for readers to notice that this instruction specification for BF16 requires extra details. (Maybe this was reverted for a reason)

aswaterman:
@nibrunieAtSi5 agreed, done

PR #2750, opened 2026-03-13
Replace 'low word' and 'high word' in descriptions of vector multiply with 'low half' and 'high half'

topperc:
'word' means 32 bits in RISC-V, but these product results are SEW bits wide.


aswaterman (approved):

aswaterman:
@topperc can you upload your public key to your github account as your Signing Key?  I think that should suffice to let this through, but you might need to force-push over it afterwards.  See https://github.com/riscv/riscv-isa-manual/blob/0dc93d93b6ffe987c25cf820a21ba130dd4a3297/CONTRIBUTING.md#signing-github-commits--step-by-step

PR #2751, opened 2026-03-13
FLEN not an impl-def. ENDIANNESS_LITTLE_OR_BIG renamed ENDIANNESS.

### LINKED ISSUE #2749: [NormRules] - FLEN is not an implementation defined behavior
`FLEN` is fixed by the ISA manual to be 128 if Q is supported, 64 if D is supported, or 32 if F is supported. It is not implementation defined. @james-ball-qualcomm 

(cc @riscv/csc)

james-ball-qualcomm:
Fixes #2749  

Also get latest docs-resources commit to support "base: Any" schema.


james-ball-qualcomm:
@jordancarlin, take a look.

@mmhus, CC

PR #2754, opened 2026-03-16
Unbreak bullet-list formatting that was broken by normative rules.

pmundkur:
This got broken in #2571.


pmundkur:
This is the broken HTML rendering:
Screenshot_2026-03-16_14-43-50
while it was previously
Screenshot_2026-03-16_14-28-48

aswaterman (approved):

aswaterman:
Thanks, @pmundkur!

PR #2755, opened 2026-03-16
[PRIV] Clarification to STIP behavior when STCE=1

bgrayson-mips:
See https://lists.riscv.org/g/tech-privileged/message/2422 for more details.


bgrayson-mips:
@gfavor @Timmmm one of my coworkers noticed that the discussion in https://lists.riscv.org/g/tech-privileged/message/2422 did not yet result in a spec clarification. This adds the clarification that Greg suggested, and also adds missing back-ticks on `mip`.

Timmmm:
I did make a PR but [it was decided not to change anything](https://github.com/riscv/riscv-isa-manual/pull/2366#issuecomment-3472839483).

PR #2767, opened 2026-03-19
[RVBNA] Generalizing the definition of RVBNA to non symmetric product operands

nibrunieAtSi5:
This PR builds on https://github.com/riscv/riscv-isa-manual/pull/2618

RVBNA was initially descrbed with the same format for both left hand side and right hand side products. This is not a requirement and RISC-V specifications (e.g. VME) actually require the support of non-uniform formats.

This patch generalizes the definition to different formats for left hand side and right hand side.
The RVBNA pseudo-code has also been generalized to this case, and the value of the exponent biasing for the result was clarified.

### DIFF
--- a/src/zvdota.adoc
+++ b/src/zvdota.adoc
@@ -763,11 +763,11 @@ The **reference exponent** is a proxy to the product exponent used to determine 
 
 
 The **reference exponent** of a product is evaluated as the sum of the factors' biased exponents.
-If a non-zero factor is subnormal then biased `emin_normal` (`=1`) is used as its biased exponent, for purposes of computing the **reference exponent**.
+If a non-zero factor is subnormal then biased `emin_normal` (`=1`) is used as its biased exponent, for purposes of computing the **reference exponent**. Note that in that case, the value is independent from the format. More generally, unbiasing is not needed to compute the reference exponent.
 
 Note:: the minimal reference exponent for a non-zero product is `2` (`emin_normal + emin_normal`).
 
-If a product operand is zero then the product exponent is set to a value which ensures that every non-zero product is considered greater than every zero product when determining the maximum product. A zero product should not force any loss of accuracy on non-zero products.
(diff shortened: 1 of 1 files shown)


src/zvdota.adoc
```diff
@@ -525,8 +525,8 @@ representable number.
 
 Formally, bulk normalization is characterized by multiple parameters:
 
-- `p`: the size of each factor's significand (significand product is `2p`-bit wide, signed product is `2p+1`-bit wide)
-- `e`: the size of each factor's biased exponent (the bias is `2^(e-1) - 1`)
+- `p_l`, `p_r`: the size of each left (resp. right) hand side factor's significand (significand product is `p_l + p_r`-bit wide, signed product is `p_l + p_r + 1`-bit wide)
+- `e_l`, `e_r`: the size of each left (resp. right) hand side factor's biased exponent (the bias is `2^(e_l-1) - 1` / `2^(e_r-1) - 1`)
```
nibrunieAtSi5:
```suggestion
- `e_l`, `e_r`: the size of each left (resp. right) hand side factor's biased exponent (the bias is `2^(e_l-1) - 1` resp. `2^(e_r-1) - 1`)
```

src/zvdota.adoc
```diff
@@ -546,15 +546,16 @@ This reduces the space of possible implementations. An implementation using RVBN
 Supporting multiple values of `g` is straightforward: an implementation can build a shared accumulator datapath based on the maximum `g` value: gmax.
 Behavior for smaller `g` values would be obtained by OR-reducing both the sticky bits defined by `gmax` and the extra `gmax - g` least significant bits of the aligned products into the least significant accumulator bit. This should not require more than a few extra gates for hardware implementations.
 
-Note:: In its current version, RVBNA is only defined for symmetric products (i.e., both factors have the same `e` and `p`). It could be later extended to asymmetric products (mixed formats) if the need arises.
+Note:: Initially, RVBNA was only defined for symmetric products (i.e., `e_l = e_r` and `p_l = p_r`). The current definition generalizes RVBNA to asymmetric products.
++ It is generally expected, but not necessary that `e_l + p_l = e_r + p_r`.
```
nibrunieAtSi5:
```suggestion
+ It is generally expected, but not necessary, that `e_l + p_l = e_r + p_r`. 
```

PR #2769, opened 2026-03-20
MSTATUS_UXL_LEGAL_VALS_SMODE not impl-def

### LINKED ISSUE #2768: [NormRules] - MSTATUS_UXL_LEGAL_VALS_SMODE is not an impl-def
ISA manual text is:
    If S-mode is implemented, the set of legal values that the UXL field may assume excludes those that would cause UXLEN to be greater than SXLEN.

(cc @riscv/csc)

james-ball-qualcomm:
Fixes #2768 

Just removed property identifying this as impl-def.


aswaterman (approved):
Assuming @mmhus will approve of this simple one anyway, so I'll go ahead and merge it.

PR #2770, opened 2026-03-20
Merge Volumes I and II into a single document

aswaterman:
This is the first step in actioning the spec reorg plan laid out in https://lists.riscv.org/g/sig-documentation/message/691

To get cross-references formatted sanely in the multi-volume document--printing the volume number when the reference is to another volume, or just the chapter/section number if the reference is local--I needed to add a custom script.

Appendix numbering is still not what I'd prefer.  The Asciidoc book format doesn't understand the notion of multiple appendices, and so having appendix names restart at A for each volume isn't readily doable.  The workaround is to make each volume have one appendix, which has multiple sections.  Unfortunately, Volume II's appendix is named B, Volume III's is named C, etc., but we can live with that nit.  The important thing is that the names are stable.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -65,7 +65,7 @@ Outputs land in the `build/` directory.
 
 ## Repository structure
 
-- `src/` is the canonical AsciiDoc content used by `make` builds. `src/riscv-unprivileged.adoc` and `src/riscv-privileged.adoc` are the entrypoints that include chapter files.
+- `src/` is the canonical AsciiDoc content used by `make` builds. `src/riscv-spec.adoc` is the entry point that includes volume files (e.g. `unpriv.adoc` and `priv.adoc`).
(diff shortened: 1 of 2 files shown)


src/unpriv.adoc
```diff
@@ -0,0 +1,52 @@
+[[vol:unpriv]]
+[volume=I]
+= The RISC-V Instruction Set Manual, Volume I: Unprivileged Architecture
+
+include::colophon.adoc[]
```
omasanori: How about renaming `colophon.adoc` to `preface.adoc` to align with `priv-preface.adoc`? Bikeshed thing, though.

omasanori (commented):
Looks good overall.

src/unpriv.adoc
aswaterman: Yeah, I was going to do that renaming as part of some directory restructuring in the future.  But I can go ahead and do it now.

jordancarlin:
The GitHub workflows need to be updated to upload/publish the new files (they’re still looking for the old file names).

PR #2771, opened 2026-03-23
Integrate profiles as Volume III

aswaterman:
This is just a structural integration.  It needs more cleanup and better narrative flow.

### DIFF
--- a/src/profiles/rva23.adoc
+++ b/src/profiles/rva23.adoc
@@ -121,7 +121,7 @@ regions.  The extension will be added to the PMA section of the
 privileged architecture manual.
 
 - <<zvbc,**Zvbc**>> Vector carryless multiplication.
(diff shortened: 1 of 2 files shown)


aswaterman:
@james-ball-qualcomm That's what I'd like to do, both for consistency and for easier navigation, but obviously it'll be a bit disruptive to change Volumes I and II to that scheme.  I'll probably hold off on doing that until I start working on the chapter reorg, since the two changes will be disruptive in similar ways.

james-ball-qualcomm:
Yes, we don't want to break people. However, couldn't we just adjust the Makefile to handle the change of moving vol1/vol2 into their own subdirectory?

aswaterman:
It'll be easy to fix up the build system.  I'm just referring to merge conflicts resulting from churn in the file/directory structure.  So I'd like to break everything exactly once, rather than breaking a large subset several times as I gradually restructure the document.

PR #2773, opened 2026-03-24
Sets :appendix-number: to @

wmat:
Asciidoctor increments the appendix number before setting it.  Setting it to @, the character right before A in the ASCII character set results in Asciidoctor restarting the Appendix numbering at A.

This fix assumes that the Appendix numbering should restart for each Volume.  Note that for future volumes this would be required in their appendix as well.

### DIFF
--- a/src/priv-appendices.adoc
+++ b/src/priv-appendices.adoc
@@ -1,3 +1,4 @@
+:appendix-number: @
 [appendix]
 == Privileged Architecture Explanatory Material
 


aswaterman:
Thanks @wmat.  Hopefully this is the answer.  Let me play around with this before we merge it.

aswaterman:
Thanks, @wmat!  This is great; it works with my cross-reference fixup script, too.  I'll incorporate it into a follow-on PR.

PR #2779, opened 2026-03-25
Made some minor fixes to M-mode norm rules and impl-defs

### LINKED ISSUE #2778: [NormRules] - Found some M-mode norm rules not properly tagged
* WARL_RST shouldn't be an impl-def (it is a link to norm rule describing behavior of WARL fields in general on reset)
* PMP_CSRS_WARL_ACCESS should be renamed PMPCFG_WARL to be consistent with other names
* PMP_ADDR_WARL should be renamed PMPADDR_WARL also to be consistent with other names
* The norm rule pointed to by PMP_RWX_WARL should just be included in PMPCFG_WARL since it is the same impl-def behavior.

(cc @riscv/csc)

james-ball-qualcomm:
Fixes #2778 

* WARL_RST shouldn't be an impl-def (it is a link to norm rule describing behavior of WARL fields in general on reset)
* PMP_CSRS_WARL_ACCESS should be renamed PMPCFG_WARL to be consistent with other names
* PMP_ADDR_WARL should be renamed PMPADDR_WARL also to be consistent with other names
* The norm rule pointed to by PMP_RWX_WARL should just be included in PMPCFG_WARL since it is the same impl-def behavior.


james-ball-qualcomm:
@aswaterman, looks like you have to approve this again since the branch became out of date.

PR #2780, opened 2026-03-25
Since Smstateen is ratified, link internally, not externally

bgrayson-mips:
Title says it all.... This tweak is for the hyperlink in Smcsrind.


aswaterman:
Thanks!

PR #2782, opened 2026-03-25
cfi: fix ordering of textual SSPUSH behavior, clarify ordering.

dtzSiFive:


src/unpriv-cfi.adoc
```diff
@@ -419,9 +419,9 @@ UXLEN or SXLEN may never be 32, then the bit 2 is also read-only zero.#
 [[SS_PUSH]]
 ==== Push to the Shadow Stack
 [[norm:ss_push]]
-A shadow stack push operation is defined as decrement of the `ssp` by `XLEN/8`
-followed by a store of the value in the link register to memory at the new top
-of the shadow stack.
+A shadow stack push operation is defined as a store of the value in the link
+register to memory at one level down from the top of the shadow stack, which is
+`ssp - XLEN/8`, followed by updating `ssp` to this address.
```
aswaterman, line 424: I don't think this actually changes the memory ordering implications.  Either way, the store depends on `ssp`.  It doesn't really matter whether it depends on the incremented value or the original value.

src/unpriv-cfi.adoc
```diff
@@ -469,8 +469,14 @@ if (xSSE == 1)
...
-The `ssp` is decremented by `SSPUSH` and `C.SSPUSH` only if the store to the
-shadow stack completes successfully.
+The `ssp` is decremented by `SSPUSH` and `C.SSPUSH` only if it is determined
+that the store to the shadow stack will complete successfully. This does not
+imply ordering: subsequent `SSPUSH` and `C.SSPUSH` instructions may be globally
+visible before earlier `SSPUSH` and `C.SSPUSH` instructions, despite
+potentially having an address dependency on the update to `ssp`. The decrement
+of `ssp` and the store to the shadow stack are not ordered with respect to each
+other; the store may or may not be globally visible before the decrement of
+`ssp`.
```
aswaterman, line 479: I think this is true, but if this follows from RVWMO's handling of implicit CSR accesses (which I think it does, but I haven't confirmed it), then I'd rather not add any new text.  I'll leave this open for now.

dtzSiFive: Aye, I'll check and agreed.  Thanks!

src/unpriv-cfi.adoc
dtzSiFive: The main change intent is to not indicate ssp is decremented first "followed by" the store, when the code listing has them in the other order.

But okay I see what you mean, if the update to ssp can be said to precede the store but doesn't commit until after regardless this change is unnecessary.

PR #2783, opened 2026-03-26
Restructure privileged architecture

aswaterman:
This is only a structural reorganization.

### DIFF
--- a/src/priv/svvptc.adoc
+++ b/src/priv/svvptc.adoc
@@ -1,4 +1,4 @@
-[[sec:svvptc]]
+[[svvptc]]
 === "Svvptc" Extension for Obviating Memory-Management Instructions after Marking PTEs Valid, Version 1.0
 
 [#norm:Svvptc_explicit_stores_update_valid_bit]#When the Svvptc extension is implemented, explicit stores by a hart that update
--- a/ref/riscv-spec-norm-tags.json
+++ b/ref/riscv-spec-norm-tags.json
@@ -11780,7 +11780,7 @@
           },
           {
             "title": "\"Svadu\" Extension for Hardware Updating of A/D Bits, Version 1.0",
-            "id": "sec:svadu",
+            "id": "svadu",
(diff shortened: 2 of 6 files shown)


src/priv/priv.adoc
```diff
@@ -0,0 +1,21 @@
...
+include::preface.adoc[]
+include::intro.adoc[]
+include::csrs.adoc[]
+include::machine.adoc[]
+include::supervisor.adoc[]
+include::sm.adoc[]
+include::sv.adoc[]
+include::ss.adoc[]
+include::hypervisor.adoc[]
+include::sh.adoc[]
```
omasanori, line 15: How about the `machine`, `supervisor`, `hypervisor`, `sm`, `sv`, `ss`, `sh` order, otherwise the `machine`, `sm`, `sv`, `supervisor`, `ss`, `hypervisor`, `sh` order? The current order looks inconsistent a bit.

aswaterman: It is inconsistent.  I think base ISAs first and extensions afterwards would work better, though.  We can play around with it.

src/priv/svvptc.adoc
```diff
@@ -0,0 +1,51 @@
...
+
+Depending on the microarchitecture, some possible ways to facilitate
+implementation of Svvptc include: not having any address-translation caches, not
+storing Invalid PTEs in the address-translation caches, automatically evicting
+Invalid PTEs using a bounded timer, or making address-translation caches
+coherent with store instructions that modify PTEs.
+====
+
+[[sec:svrsw60t59b]]
+=== "Svrsw60t59b" Extension for PTE Reserved-for-Software Bits 60-59, Version 1.0
```
jordancarlin: This should probably be broken out into a separate file since it is a separate extension.

aswaterman: Yes, it should be. It’s short enough that I skipped over it accidentally.

Issue #2788, opened 2026-03-26
[NormRules] - Missing norm rules for M-mode CSR defined as WARL/WLRL only in register layout diagrams


james-ball-qualcomm:
PR #2707 added WARL/WLRL norm rules for CSRs defined as WARL/WLRL only in register layout diagrams (but not the text). For some reason, CSRs in M-mode (e.g., mip, mie) didn't get these norm rules added for them too.

(cc @riscv/csc)

PR #2790, opened 2026-03-26
Added missing WARL norm rule for mip/mie CSRs and add missing impl-def for MCAUSE_EXCCODE_MISALIGNED_VS_FAULT_PRI

james-ball-qualcomm:
Simple fix for issue #2788

### DIFF
--- a/src/profiles/rvb23.adoc
+++ b/src/profiles/rvb23.adoc
@@ -19,16 +19,6 @@ of binary kernel and binary application software distributions.
 However, individual software ecosystems may build upon RVB profiles to
 produce a more targeted standard interface for a certain market.
 
-=== Profile-Defined Extensions
-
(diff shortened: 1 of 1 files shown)


james-ball-qualcomm:
Just noticed that an impl-def is missing for "Load/store/AMO address-misaligned exceptions may have either higher or lower priority than load/store/AMO page-fault and access-fault exceptions."

Does have a norm rule called "mcause_exccode_pri" but will need to create a new norm rule for just the impl-def part of the norm rule.

james-ball-qualcomm:
@aswaterman, thanks for approving this. However, it was still in draft state and was out of date with the main branch. So, I've changed it to a normal PR (no longer draft) and updated the branch. Think it needs your approval again though ...

aswaterman:
> However, it was still in draft state

Yeah, I noticed that only after approving it.

PR #2796, opened 2026-03-27
Fixes in norm rules found today (Mar 27 2026)

### LINKED ISSUE #2794: [NormRules] - Fix norm rules issues found on Mar 27 2026
So far have found these issues:

- MTVAL_VARLEN_WR shouldn't be an impl-def (just a regular norm rule)
- Should add kind/instance for impl-defs added yesterday (MIP_WARL, MIE_WARL)
- Missing WARL norm rule for menvcfg.PMM (was only declared WARL in table caption so was missed originally)
- The norm tags/rules for the smstateen.adoc chapter apply to M-mode, S-mode, HS-mode, or all modes but the naming isn't consistent with this. Should use mstateen_* for M-mode, sstateen_* for S-mode, hstateen_* for HS-mode, and stateen_* for norm rules/tags that apply to all modes. Some norm tags/rules are consistent with this but some aren't so I'm going to make them all consistent.
- Added missing norm tag and rule for mstateen_lower_priv_roz

CC: @jordancarlin @davidharrishmc 

(cc @riscv/csc)

james-ball-qualcomm:
Fixes #2794 

- MTVAL_VARLEN_WR shouldn't be an impl-def (just a regular norm rule)
- Should add kind/instance for impl-defs added yesterday (MIP_WARL, MIE_WARL)
- Missing WARL norm rule for menvcfg.PMM (was only declared WARL in table caption so was missed originally)
- The norm tags/rules for the smstateen.adoc chapter apply to M-mode, S-mode, HS-mode, or all modes but the naming isn't consistent with this. Should use mstateen_* for M-mode, sstateen_* for S-mode, hstateen_* for HS-mode, and stateen_* for norm rules/tags that apply to all modes. Some norm tags/rules are consistent with this but some aren't so I'm going to make them all consistent.
- Added missing norm tag and rule for mstateen_lower_priv_roz

Name changes in smstateen.adoc/yaml:

image


james-ball-qualcomm:
CC: @jordancarlin @davidharrishmc 
Changing some norm rule names that you might depend on.

PR #2799, opened 2026-03-28
Increase ToC level for the whole spec

inochisa:
Increase ToC level for the spec so the CSR section of each privilege mode and instruction section can be found in the ToC.

### DIFF
--- a/src/riscv-spec.adoc
+++ b/src/riscv-spec.adoc
@@ -27,7 +27,7 @@ endif::[]
 :sectnumlevels: 4
 :sectlinks:
 :toc: left
-:toclevels: 2
+:toclevels: 3
 :source-highlighter: rouge
 :table-caption: Table
 :figure-caption: Figure


inochisa:
Sorry for forgetting to claim the use case: For ISA implemtation, it is more common to lookup a single CSR or instruction, instead of starting from the section begin. Using search is quite painful as the CSR can exist many times (e.g. `mstatus`), which lead to inefficient lookup, especially for the html build.

omasanori:
While it is beyond the scope of this PR, how about making an index of CSRs (with the page numbers of definition in bold) in the end of the manual, so that we can keep the current ToC level and address the inconvenience? In long term, Unified DB would address the problem I guess, though.

inochisa:
> While it is beyond the scope of this PR, how about making an index of CSRs (with the page numbers of definition in bold) in the end of the manual, so that we can keep the current ToC level and address the inconvenience? In long term, Unified DB would address the problem I guess, though.

Hi @omasanori, it is a good suggestion I guess. But I am not sure whether it can reference the extra behavior for some existing CSR defined in another section. (e.g. the section 5.3 for the Volume II, it is for machine CSR change of the H extension).

Anyway, I think it could be a good idea to open an issue to lookup some advice?

omasanori:
> > While it is beyond the scope of this PR, how about making an index of CSRs (with the page numbers of definition in bold) in the end of the manual, so that we can keep the current ToC level and address the inconvenience? In long term, Unified DB would address the problem I guess, though.
> 
> Hi @omasanori, it is a good suggestion I guess. But I am not sure whether it can reference the extra behavior for some existing CSR defined in another section. (e.g. the section 5.3 for the Volume II, it is for machine CSR change of the H extension).

Sure, we would need something to extract Normative Rules or such for each CSR.

> Anyway, I think it could be a good idea to open an issue to lookup some advice?

Yes, please. Inconvenience around CSRs is worth exploration I think, and increasing ToC level is a solution but not definite one.

inochisa:
@omasanori Thanks, I have created the issue. And I will leave this PR opened as a candidate solution. If there is a better solution, feel free to close this if I have forgotten.

wmat:
Note that the addition of an instruction per page index is in the works.  This would be produced out of UnifiedDB and added to the fully integrated single book spec.  The creation of the single book spec is in process now.

aswaterman (changes requested):
Increasing the ToC depth likely won’t make sense once the manual reorg is complete. Let’s leave it as-is for now and make sure the ToC is doing a good enough job after the reorg is complete.

inochisa:
> Increasing the ToC depth likely won’t make sense once the manual reorg is complete. Let’s leave it as-is for now and make sure the ToC is doing a good enough job after the reorg is complete.

OK, So I think it is fine to close it for now. As I have already opened an issue to address this.

PR #2804, opened 2026-04-01
Reorganize Zi, memory model chapters

aswaterman:
This is just a structural change; there's plenty more work to be done.

### DIFF
--- a/src/unpriv/cmo.adoc
+++ b/src/unpriv/cmo.adoc
@@ -1,5 +1,5 @@
 [[cmo]]
-=== CMO Extensions for Base Cache Management Operation ISA, Version 1.0.0
+=== Cache Management Operations (CMOs)
 
 ==== Pseudocode for instruction semantics
 
--- a/src/unpriv/zi.adoc
+++ b/src/unpriv/zi.adoc
@@ -1,5 +1,5 @@
 [[zi]]
-== "I" Scalar Integer Extensions
+== Scalar Integer Extensions
 
 NOTE: This chapter is currently being restructured.
 Its contents are normative, but the presentation might appear disjoint.


src/unpriv/zicntr.adoc
```diff
@@ -1,13 +1,11 @@
-[[counters]]
-== "Zicntr" and "Zihpm" Extensions for Counters, Version 2.0
+[[zicntr]]
+=== "Zicntr" Extension for Base Counters and Timers
```
omasanori, line 2: Why was *Version 2.0* omitted?

src/unpriv/zihpm.adoc
```diff
@@ -0,0 +1,48 @@
+[[zihpm]]
+=== "Zihpm" Extension for Hardware Performance Counters
```
omasanori, line 2: Why was *Version 2.0* omitted?

aswaterman: Well, it was never there to begin with, but see above.

src/unpriv/zicntr.adoc
aswaterman: My plan is to delete version numbers from all chapters. The spec will always reflect the most recent version. The preface can indicate the version information.

omasanori: I see. Thanks.

src/unpriv/zi.adoc
```diff
@@ -0,0 +1,30 @@
+[[zi]]
+== "I" Scalar Integer Extensions
```
omasanori:
```suggestion
== Scalar Integer Extensions
```
It contains `M` and `Zm*`, and the bitmanip chapter already omits `"ABC"` part.

src/unpriv/cmo.adoc
```diff
@@ -1,13 +1,13 @@
 [[cmo]]
-== "CMO" Extensions for Base Cache Management Operation ISA, Version 1.0.0
+=== CMO Extensions for Base Cache Management Operation ISA, Version 1.0.0
```
omasanori:
```suggestion
=== Cache Management Operation (CMO)
```

omasanori:
The changes proposed by me was once applied and then dropped. Intentional?

aswaterman:
I messed up the git history; thanks for pointing that out.

Issue #2810, opened 2026-04-01
Update the reference to IEEE 754 to 2019 version


omasanori:
Currently, the specification refers to IEEE 754-2008 as the floating-point arithmetic standard, though `FMIN.*` and `FMAX.*` refer to IEEE 754-201x (sic) and Zfa refers to IEEE 754-2019. IEEE 754-2019 is not in the bibliography list nevertheless. Using IEEE 754-2019 as the single source of IEEE 754-based floating-point arithmetic is desired.

Open questions:

- Does this change increase the version number of extensions?
  - `FMIN.*` and `FMAX.*` are already amended in Version 2.2. Does this change affect the behavior of other instructions?
- Do we need to create a TG under FP SIG for this change, or is a PR by random individuals (like me) enough?

omasanori:
I found that BF16 has independent bibliography section and the both 2008 and 2019 versions are listed there. Also, that chapter uses the phrases *IEEE-754* and *'754,* while the chapter of F extension uses the phrases ~~*IEEE 754-2008 standard* consistently~~ *IEEE 754-2008 standard* and *IEEE 754-2008 arithmetic standard*.

omasanori:
The vector extension uses the phrase *IEEE-754/2008.*

omasanori:
Cleanup that should never alter the specified behavior of instructions was finished. The remaining part is updating references to IEEE 754-2008 to IEEE 754-2019, and I think this part should be led by FP SIG. I sent a [message to FP SIG](https://lists.riscv.org/g/sig-fp/message/409).

omasanori:
> Cleanup that will never alter the specified behavior of instructions was finished. The remaining part is updating references to IEEE 754-2008 to IEEE 754-2019, and I think this part should be led by FP SIG. I sent a [message to FP SIG](https://lists.riscv.org/g/sig-fp/message/409).

Al Martin (Akeana) told me that IEEE 754-2019 should be backward compatible and just a PR without TG nor fast-track will be enough. ~~I will make a PR after all cleanup PRs are merged.~~

Update: Guy Lemieux objects, and I agree with Guy: a careful review should be done before updating the references, when it actually happens.

nibrunieAtSi5:
Thank you for starting this effort @omasanori 

Can you provide arguments about 
> Using IEEE 754-2019 as the single source of IEEE 754-based floating-point arithmetic is desired.

Is this for uniformity ? 
[...]

PR #2813, opened 2026-04-02
Use the phrase "IEEE 754-2008" consistently

omasanori:
It consistently uses *the IEEE 754-2008 arithmetic standard* for the first time in the chapter, and *IEEE 754-2008* afterwards. The standard format is referred as *the IEEE 754-2008 format.* Only scalar FP extensions are updated in this PR to receive feedbacks before working on the whole text.

Reference: https://github.com/riscv/riscv-isa-manual/issues/2810

### DIFF
--- a/src/d-st-ext.adoc
+++ b/src/d-st-ext.adoc
@@ -44,9 +44,13 @@ including variadic functions, user-level threading libraries, virtual machine
 migration, and debugging.
 ====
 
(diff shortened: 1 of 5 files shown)


omasanori:
@aswaterman Thank you for your feedback! I revised the patch.

aswaterman (approved):

PR #2817, opened 2026-04-02
intro: use the phrase IEEE 754-2008

omasanori:
Note that while preface contains *IEEE 754-201x,* I decided to leave it as-is since the sentence is about the situation at that time.

Reference: https://github.com/riscv/riscv-isa-manual/issues/2810

### DIFF
--- a/src/intro.adoc
+++ b/src/intro.adoc
@@ -19,7 +19,8 @@ FPGA), but which allows efficient implementation in any of these.
 a base for customized accelerators or for educational purposes, and
 optional standard extensions, to support general-purpose software
 development.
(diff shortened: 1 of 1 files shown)


src/intro.adoc
```diff
@@ -605,8 +606,7 @@ a hart depends on the EEI.
 
 [NOTE]
 ====
-Our use of "exception" and "trap" is compatible with that in the
-IEEE-754 floating-point standard.
```
aswaterman, line 609: I think in this introductory context, the original text reads better.

omasanori: Addressed. Thanks!

PR #2818, opened 2026-04-02
Clarify that V is compatible with Zfinx

aswaterman:
The V extension is compatible with Zfinx [1], but the current spec language indicating that it depends on F contradicts that.  This PR resolves that contradiction by allowing V to depend on either F or Zfinx, etc.

The longer-term and more elegant solution is to declare ISAs with Zfinx to be distinct base ISAs. The band-aid in this PR is compatible with the long-term plan.

[1] https://github.com/riscv/riscv-isa-manual/blob/5b8f8d6c04b515e633023deecca49de855f452d2/src/v-st-ext.adoc#L941

### DIFF
--- a/src/v-st-ext.adoc
+++ b/src/v-st-ext.adoc
@@ -5138,14 +5138,15 @@ do not include those with floating-point operands, and Zve64f does not include t
 with EEW=64 floating-point operands.#
 
 [#norm:Zve32x_dependent_Zicsr]#The Zve32x extension depends on the Zicsr extension.#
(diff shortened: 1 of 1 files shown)


topperc:
Toolchains have V imply F and D today.

Do you expect assemblers to start accepting `.vf` instructions with GPR operands when Zfinx is enabled?

aswaterman:
@topperc just chatted with Krste and we agreed we'll drop this issue until we create the new Zfinx base ISA.  So the toolchains can continue having V imply F and D.

But to answer your question, yes, once we eventually allow this combination, the `.vf` forms will accept `x` registers.

PR #2819, opened 2026-04-03
Bibliography database cleanup

omasanori:
- Removed uncited entries.
- Listed IEEE as an author for a better sorting order.
- Put DOI to the `doi` field if appropriate
- Marked preprints as such if the final version is published elsewhere
- Other corrections


aswaterman (approved):
Thanks!

aswaterman:
@omasanori is this ready to merge?

omasanori:
Yeah, while I have not checked the remaining entries, let's merge the finished ones and do the rest another day.

PR #2821, opened 2026-04-04
Recover some missing numbers

int-e:
In the conversion from TeX to asciidoc, some numbers using the custom `\wunits` macro got lost. This series recovers them.

(The most useful part is the table in `zihintntl.adoc`. @b-jonas0 noticed the missing numbers there.)


int-e:
The normative tag change report looks bogus to me; I think the `main` branch is violating those checks.

PR #2823, opened 2026-04-04
Volume III: Unify styling

omasanori:


omasanori:
> I don't think there's agreement on what the unified style should be.  So let's hold off on this for now.

Ah, I thought that monospace font is, as you had used some in recent commits. OK.

aswaterman:
I'm not sure we want to use the same stylization for extension names and instructions.  We'll think about it some more.

Maybe the same font is OK since the extension names are camel-case and the instructions are capitalized, but let's call it TBD for now.

omasanori:
> I'm not sure we want to use the same stylization for extension names and instructions. We'll think about it some more.

Agreed.

> Maybe the same font is OK since the extension names are camel-case and the instructions are capitalized, but let's call it TBD for now.

So, let me send a new PR for capitalizing instruction names part only? Currently, the case of `ecall` and `fence.*` in Volume III are inconsistent.

omasanori:
Closing it. We can work on it later once a consensus is established.

PR #2824, opened 2026-04-05
Preface: add new extensions from the Profile spec

omasanori:
I put them in the tail of the extensions table. One can sort them after the reorganization by Andrew is finished.

Reference: https://github.com/riscv/riscv-isa-manual/pull/2771

### DIFF
--- a/src/priv/preface.adoc
+++ b/src/priv/preface.adoc
@@ -1,5 +1,93 @@
 [colophon]
+// Had to make the above a level 1 heading (two equals signs) to avoid error when building
+// the ISA manual as a book with other "parts". This is opposite to what the adoc says to do
+// but otherwise asciidoctor creates the error message:
+//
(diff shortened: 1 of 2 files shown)


omasanori:
I forgot to update the table in the preface of Volume II. ~~I will update this PR later.~~ I have updated.

src/priv/preface.adoc
```diff
@@ -1,10 +1,98 @@
...
+*Svvptc Extension* +
+*Pointer-Masking Extensions* +
+*Svrsw60t59b Extension* +
+*Shcounterenw* +
+*Shvstvala* +
+*Shvstvecd* +
+*Shvsatpa* +
+*Shgatpa* +
+*Sha*
+
```
omasanori: I should have add *Extension* to each entry. I will do.

PR #2825, opened 2026-04-05
profiles: capitalize instruction names

omasanori:
Reference: https://github.com/riscv/riscv-isa-manual/pull/2823#issuecomment-4188346550


kersten1:
The Authoring guide says this:

Formatting guidelines for names
⚫ Instructions are lowercase (ld, c.lw)
⚫ Extensions are Capitialized (A, C, Zicsr)
⚫ CSR short names are lowercase (misa).
⚫ All of these names are in monospace

We can definitely change if we decide to use upper.

omasanori:
Let's do cleanup after the style convention is agreed. Closing.

PR #2859, opened 2026-04-09
Zce: Fix examples

omasanori:
I could not understand why `RV32IMC` becomes `RV32IM_Zce`. If *becomes* means *is equivalent to,* the examples should be these in this change.

### DIFF
--- a/src/zc.adoc
+++ b/src/zc.adoc
@@ -75,8 +75,8 @@ The Zce extension is intended to be used for microcontrollers, and includes all 
 
 Therefore common ISA strings can be updated as follows to include the relevant Zc extensions, for example:
 
-* RV32IMC becomes RV32IM_Zce
-* RV32IMCF becomes RV32IMF_Zce
+* RV32IMC_Zcb_Zcmp_Zcmt is equivalent to RV32IM_Zce
+* RV32IMFC_Zcb_Zcmp_Zcmt is equivalent to RV32IMF_Zce
 
 [#misaC]
 === MISA.C


src/zc.adoc
omasanori, line 74: By the way, this means `RV32IMFDC_Zcb_Zcmp_Zcmt` is equivalent to `RV32IMFD_Zcd_Zce` and `RV64IMFDC_Zcb_Zcmp_Zcmt` is equivalent to `RV64IMFD_Zcd_Zce` correct? (It is unlikely that microcontrollers implements D, though.)

omasanori: Ah, OK, Zcd and Zcmp / Zcmt are incompatible so they are impossible configuration.

aswaterman (approved):
You're clearly right; not sure how this escaped notice.

src/zc.adoc
```diff
@@ -75,8 +75,8 @@ The Zce extension is intended to be used for microcontrollers, and includes all
 
 Therefore common ISA strings can be updated as follows to include the relevant Zc extensions, for example:
 
-* RV32IMC becomes RV32IM_Zce
-* RV32IMCF becomes RV32IMF_Zce
+* RV32IMC_Zcb_Zcmp_Zcmt becomes RV32IM_Zce
+* RV32IMFC_Zcb_Zcmp_Zcmt becomes RV32IMF_Zce
```
aswaterman: Can you change "becomes" to "is equivalent to"?

omasanori:
Updated wording.

aswaterman (approved):
Thanks!

PR #2860, opened 2026-04-09
unpriv: preface: Add missing extensions

omasanori:
To keep the version information after <https://github.com/riscv/riscv-isa-manual/pull/2804#discussion_r3019969748>, preface should list all ratified extensions. This change adds every missing extension I found to the table.


src/preface.adoc
```diff
@@ -47,29 +48,66 @@ h|Extension h|Version
...
+|*Zba* |*1.0*
+|*Zbb* |*1.0*
+|*Zbc* |*1.0*
+|*Zbs* |*1.0*
 |*Zbkb* |*1.0*
 |*Zbkc* |*1.0*
 |*Zbkx* |*1.0*
-|*Zk* |*1.0*
+|*V* |*1.0*
+|*Zvl&#42;* |*1.0*
```
aswaterman: Can you use `\*` instead of `&#42;`?  It is easier for readers of the source code.

omasanori: Unfortunately Asciidoctor converts `*Zvl\**` to `<strong>Zvl\</strong>*` so I use this method.

src/preface.adoc
```diff
@@ -47,29 +48,66 @@ h|Extension h|Version
...
+|*Zknd* |*1.0*
+|*Zkne* |*1.0*
+|*Zknh* |*1.0*
+|*Zksed* |*1.0*
+|*Zksh* |*1.0*
+|*Zkr* |*1.0*
+|*Zkn* |*1.0*
 |*Zks* |*1.0*
+|*Zk* |*1.0*
+|*Zkt* |*1.0*
```
omasanori, line 97: Scalar crypto extensions are not version 1.0 but version 1.0.1.

omasanori: The change in scalar crypto 1.0.1 is that *fix typos to show that `c.srli`, `c.srai`, and `c.slli` are Zkt instructions in RV64.* Then, should Zkt be treated as 1.0.1 while the others be kept in 1.0?

omasanori: 895aa382e97cd34b24079396056e7d70da81dcb0 adds Zcb to Zkt. Does this change make Zkt version 1.0.2 or 1.1?

omasanori: Making `Zkt` 1.1 and the other `Zk*` 1.0 makes sense to me but I'd ask others' opinion.

omasanori:
Zawrs is not 1.0 but 1.01 now. I will send a follow-up PR to update tables for older versions for this specific case. It is time-consuming with a little benefit to add missing extensions to every table.

src/preface.adoc
aswaterman: OK, how about trying to add `:asterisk: *` or `:asterisk: &#42;` to `src/symbols.adoc` and using `:asterisk:` here?

src/preface.adoc
aswaterman: Keep it as 1.0.  We have stopped using subminor version numbers.

As new extensions are added, Zkt will be expanded as a property of those new extensions--not because Zkt has changed with respect to extensions that already existed.  So the version number should stay constant.

aswaterman:
> Zawrs is not 1.0 but 1.01 now. 

1.0 here as well.  No subminor version numbers in this table.

src/preface.adoc
omasanori: I see. Thanks for explanation.

omasanori:
Revised to address comments.

aswaterman (approved):
Thank you!

PR #2861, opened 2026-04-09
unpriv: preface: Minor fixes

omasanori:
- Remove duplicated Zcb
- Add the fix of Zce examples to the changelog as it is a change to main text

### DIFF
--- a/src/preface.adoc
+++ b/src/preface.adoc
@@ -127,6 +127,7 @@ The changes in this version of the document include:
 
 * Addition of extensions that have already been ratified as part of the profile
   specifications.
+* Numerous non-normative descriptive improvements.
 
 [.big]*_Preface to Document Version 20260120_*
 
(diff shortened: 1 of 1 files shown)


omasanori:
@aswaterman @kasanovic Would you review them?

src/preface.adoc
```diff
@@ -128,6 +127,7 @@ The changes in this version of the document include:
 
 * Addition of extensions that have already been ratified as part of the profile
   specifications.
+* Fix of examples in Zce.
```
aswaterman: There's no reason to list the non-normative changes, of which this is one.  Folks interested in the details can look at the git log.  Let's do:

```
* Numerous non-normative descriptive improvements.
```

omasanori: I see, thanks!

omasanori:
Revised to address comment.

aswaterman (approved):

PR #2867, opened 2026-04-10
Clarify the word "canonical" in "canonical NaN"

dramforever:
We define a "canonical NaN" for each format, which is unrelated to the concept of canonical encodings in IEEE 754. The latter only makes sense for decimal encodings anyway. Add a non-normative note clarifying this.

IEEE 754-2019 adds a clarifying note that "In binary interchange formats, all number and NaN encodings are canonical." However, I think it's more confusing to explain it this way in the context of this note, so I didn't mention it.

cc @Icenowy

See also a similar clarification for WebAssembly: https://github.com/WebAssembly/spec/pull/1614

### DIFF
--- a/src/unpriv/f-st-ext.adoc
+++ b/src/unpriv/f-st-ext.adoc
@@ -195,6 +195,12 @@ quiet bit. For single-precision floating-point, this corresponds to the pattern 
 
 [NOTE]
 ====
+The canonical NaN as defined here is unrelated to the definition of
+canonical encodings in IEEE 754-2008 for decimal interchange formats.
+====
+'''
+[NOTE]
+====
 We considered propagating NaN payloads, as is recommended by IEEE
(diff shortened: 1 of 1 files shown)


aswaterman (approved):
Funny, even the Wikipedia article on NaN includes this caveat.

aswaterman:
@dramforever you need to sign your commit with your RSA key as described in https://github.com/riscv/riscv-isa-manual/blob/275f9c941ba4ee71f66cc338ad75e78e68f51a41/CONTRIBUTING.md#signing-github-commits--step-by-step then amend the commit and force-push over the branch.

dramforever:
Oops, missed that signing requirement. Should be signed now.

aswaterman:
@dramforever No problem; it is a recent requirement.

PR #2868, opened 2026-04-10
fixed #2849: Update zalrsc.adoc normative rule

UmerShahidengr:


mmhus (approved):

mmhus:
@UmerShahidengr could you update the reference and see if anything changed using `make update-ref`

src/unpriv/zalrsc.adoc
```diff
@@ -256,15 +256,15 @@ unconditionally fail any unconstrained LR/SC sequence.
...
+must guarantee that one of the following events eventually occurs#:
 
 * _H_ or some other hart executes a successful SC to the reservation set
 of the LR instruction in _H_'s constrained LR/SC loops.
 * Some other hart executes an unconditional store or AMO instruction to
 the reservation set of the LR instruction in _H_'s constrained LR/SC
 loop, or some other device in the system writes to that reservation set.
 * _H_ executes a branch or jump that exits the constrained LR/SC loop.
-* _H_ traps.#
+* _H_ traps.
```
jordancarlin: Should the bullets now be added as additional normative tags (potentially linked to the same normative rule)? Looks like this removes them from the normative rules and I think they are normative text.

mmhus (changes requested):
make update-ref

src/unpriv/zalrsc.adoc
UmerShahidengr: Yes, you are right, doing that already

UmerShahidengr:
@mmhus done

src/unpriv/zalrsc.adoc
UmerShahidengr: Done

ref/riscv-spec-norm-tags.json
```diff
@@ -144,30 +265,46 @@
     "norm:NTL-LR_SC_exception": "Since the NTL instructions are encoded as ADDs, they can be used within LR/SC loops without voiding the forward-progress guarantee. But, since using other loads and stores within an LR/SC loop does void the forward-progress guarantee, the only reason to use an NTL within such a loop is to modify the LR or the SC.",
     "norm:pause_op": "The PAUSE instruction is a HINT that indicates the current hart's rate\nof instruction retirement should be temporarily reduced or paused. The\nduration of its effect must be bounded and may be zero.",
     "norm:pause_enc_fence": "PAUSE is encoded as a FENCE instruction with pred=W, succ=0, fm=0,\nrd=x0, and rs1=x0.",
-    "norm:Zimop_mop-r_op": "The Zimop extension defines 32 MOP instructions named MOP.R.n, where\nn is an integer between 0 and 31, inclusive.\nUnless redefined by another extension, these instructions simply write 0 to\nx[rd].",
```
mmhus, line 147: @UmerShahidengr This change seems fishy, could you check if this norm was removed, why the cache block size rule was updated, perhaps someone else forgot to update refs earlier but this seems an unrelated change, but maybe needed

mmhus (changes requested):

ref/riscv-spec-norm-tags.json
UmerShahidengr: Thats strange, Zimop_mopr_op rule still exists [here](https://github.com/riscv/riscv-isa-manual/blob/eebfb186477a067c19e7d4fa0a436513917ffaaf/src/unpriv/zimop.adoc?plain=1#L30), but it is not showing up in riscv-spec-norm-tags.

aswaterman:
Ping me when it's ready to merge.

ref/riscv-spec-norm-tags.json
UmerShahidengr: This norm rule still exists in this file [here](https://github.com/UmerShahidengr/riscv-isa-manual/blob/5e19c3f1a619564e3a4e0412837bbdd2ae646fbd/ref/riscv-spec-norm-tags.json#L204), the diff is behaving differently on the changes, there is nothing missing here.

UmerShahidengr:
@mmhus

Justification for norm-rule changes:

* The norm rules `sret_dt`, `sret_v0`, and `sret_v1` were introduced in #2802 as replacements for `mret_dt`, `mret_v0`, and `mret_v1`.
* The norm rule `D_flen_64` was updated in #2832 with only a minor textual change, so there is no cause for concern.
* Other differences appear because the updated file has not been upstreamed since the migration to the latest version.

Merging this PR will synchronize `ref/riscv-spec-norm-tags.json` with the latest release, and subsequent changes will be tracked against the updated version.

@aswaterman this PR is good to go.

jordancarlin:
> Justification for norm-rule changes:
> 
> * The norm rules `sret_dt`, `sret_v0`, and `sret_v1` were introduced in [Fixed names of normative rules related to set in hypervisor #2802](https://github.com/riscv/riscv-isa-manual/pull/2802) as replacements for `mret_dt`, `mret_v0`, and `mret_v1`.
> * The norm rule `D_flen_64` was updated in [Define floating-point formats based on IEEE 754 #2832](https://github.com/riscv/riscv-isa-manual/pull/2832) with only a minor textual change, so there is no cause for concern.
> * Other differences appear because the updated file has not been upstreamed since the migration to the latest version.

Isn't the GitHub Actions workflow that @rpsene created supposed to automatically update that file when PRs are merged to avoid issues like this? Seems like it must not be working as expected.

mmhus (approved):

rpsene:
> > Justification for norm-rule changes:
> > 
> > * The norm rules `sret_dt`, `sret_v0`, and `sret_v1` were introduced in [Fixed names of normative rules related to set in hypervisor #2802](https://github.com/riscv/riscv-isa-manual/pull/2802) as replacements for `mret_dt`, `mret_v0`, and `mret_v1`.
> > * The norm rule `D_flen_64` was updated in [Define floating-point formats based on IEEE 754 #2832](https://github.com/riscv/riscv-isa-manual/pull/2832) with only a minor textual change, so there is no cause for concern.
> > * Other differences appear because the updated file has not been upstreamed since the migration to the latest version.
> 
> Isn't the GitHub Actions workflow that @rpsene created supposed to automatically update that file when PRs are merged to avoid issues like this? Seems like it must not be working as expected.

@jordancarlin I will check why it failed to update the files.

aswaterman (approved):

PR #2879, opened 2026-04-12
Assorted fixes to the naming conventions

### LINKED ISSUE #2876: Define Su* in ISA Extension Naming Conventions
As the Supm extension does not fit into the existing naming conventions, an update to the naming conventions specifying Su* for user-level extensions is desired.

### LINKED ISSUE #2878: On the second letter of Z* extensions
The specification defines the second letter of Z* extensions as:

> The first letter following the "Z" conventionally indicates the most closely related alphabetical extension category, IMAFDQLCBKJTPVH.

Zhinx is an exception of this rule, as it is related to Zfh, not H.

omasanori:
Fix https://github.com/riscv/riscv-isa-manual/issues/2876
Fix https://github.com/riscv/riscv-isa-manual/issues/2878

### DIFF
--- a/src/naming.adoc
+++ b/src/naming.adoc
@@ -70,6 +70,13 @@ All multi-letter extensions, including those with the "Z" prefix, must be
 separated from other multi-letter extensions by an underscore, e.g.,
 "RV32IMACZicsr_Zifencei".
 
+=== User-level Instruction-Set Extension Names
+
+Standard extensions that extend the user-level architecture are prefixed with
+the letters "Su".
+If multiple user-level extensions are listed, they should be ordered
+alphabetically.
+
 === Supervisor-level Instruction-Set Extension Names
 
 Standard extensions that extend the supervisor-level virtual-memory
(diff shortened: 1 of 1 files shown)


src/naming.adoc
```diff
@@ -58,19 +58,29 @@ The second letter from the end cannot be numeric if the
...
 "RV32IMACZicsr_Zifencei".
 
-=== Supervisor-level Instruction-Set Extension Names
+=== User-level Instruction-set Extension Names
+
+Standard extensions that extend the user-level architecture are prefixed with
+the letters "Su".
+If multiple user-level extensions are listed, they should be ordered
+alphabetically.
+
```
omasanori, line 82: It would be nice if a note explaining the differences between unprivileged extensions and user-level extensions were here, but I could not come up with a good explanation.

aswaterman (commented):
Capitalizing the the second word in a hyphenated phrase in a title is standard, e.g. from today's New York Times, "Mamdani Plans to Open City-Owned Grocery Store in East Harlem"

src/naming.adoc
aswaterman: The intro chapter explains that unprivileged extensions are those that are generally usable in any privilege mode.  And by contrast I think it's obvious obvious that user-mode extensions are those meant to be used specifically in user-mode.  So I think we are OK not re-explaining that here.

omasanori:
> Capitalizing the the second word in a hyphenated phrase in a title is standard, e.g. from today's New York Times, "Mamdani Plans to Open City-Owned Grocery Store in East Harlem"

OK, then I will use that style. Currently, some titles ex. *Supervisor-level Instruction-Set Extension Names* are inconsistent.

src/naming.adoc
omasanori: I see.

omasanori:
Unified the title case.

aswaterman (approved):

PR #2890, opened 2026-04-14
Assorted fixes to the naming conventions

### LINKED ISSUE #2889: Naming Conventions categorized H as an unprivileged extension but it is not
In the table in Naming Conventions chapter, the H extension is categorized as *Standard Unprivileged Extensions* but obviously it is not.

We should not move H to *Standard Hypervisor-Level Extensions* as doing so breaks the sorting order. We need to name the category of one-letter extensions. *Standard One-Letter Extensions,* *Standard Core Extensions,* whatever. Otherwise, is just *Standard Extensions* fine?

omasanori:
Fix https://github.com/riscv/riscv-isa-manual/issues/2889

### DIFF
--- a/src/naming.adoc
+++ b/src/naming.adoc
@@ -208,8 +208,6 @@ e.g., RV32IMACV is legal, whereas RV32IMAVC is not.
 
 |Vector Extension |V |D
 
-|Hypervisor Extension |H |
-
 3+^|*Additional Standard Unprivileged Extensions*
 
 |Additional Standard unprivileged extensions "abc" |Zabc |
@@ -219,6 +217,8 @@ e.g., RV32IMACV is legal, whereas RV32IMAVC is not.
 |Supervisor-level extension "def" |Sudef |
 
 3+^|*Standard Supervisor-Level Extensions*
+
+|Hypervisor Extension |H |
 
 |Supervisor-level extension "ghi" |Ssghi |
 
--- a/src/naming.adoc
+++ b/src/naming.adoc
@@ -77,8 +77,11 @@ separated from other multi-letter extensions by an underscore, e.g.,
 
 Standard extensions that extend the user-level architecture are prefixed with
 the letters "Su".
-If multiple user-level extensions are listed, they should be ordered
-alphabetically.
+
+Standard user-level extensions should be listed after standard unprivileged
+extensions, and like other multi-letter extensions, must be separated from other
+multi-letter extensions by an underscore. If multiple user-level extensions are
+listed, they should be ordered alphabetically.
 
 === Supervisor-Level Instruction-Set Extension Names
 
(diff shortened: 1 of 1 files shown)


src/naming.adoc
```diff
@@ -166,13 +171,25 @@ Hence, we define the default version of a standard extension to be the
...
+. One-letter extensions, IMAFDQCBPVH, in this order,
+. Z-prefixed extensions,
+. Su-prefixed extensions,
+. Ss-prefixed extensions,
+. Sv-prefixed extensions,
+. Sh-prefixed extensions,
+. Sm-prefixed extensions,
+. X-prefixed extensions.
+
+For instances, RV32IMACV is legal, whereas RV32IMAVC is not.
```
omasanori: Maybe *is canonical* is better than *is legal,* as it would be confusing if we define the term *legal* for specifying certain behavior in the future. Canonical is used in the section title, so it is unambiguous.

aswaterman (approved):

src/naming.adoc
aswaterman: I think canonical would be better.  Ping me when the PR is ready to merge.

omasanori:
@aswaterman Updated wording, it is ready now.

src/naming.adoc
```diff
@@ -166,13 +171,25 @@ Hence, we define the default version of a standard extension to be the
...
 
-=== Subset Naming Convention
+=== Canonical Order
 
-<<isanametable>> summarizes the standardized extension
-names. The table also defines the canonical
-order in which extension names must appear in the name string, with
-top-to-bottom in table indicating first-to-last in the name string,
-e.g., RV32IMACV is legal, whereas RV32IMAVC is not.
+Components in ISA name string must appear in the following order:
```
aswaterman: I realize the old text isn't exactly right here, but the current philosophy is that non-canonical orderings aren't inherently incorrect.  So, rather than saying "must" here, I'd say "In canonical ISA name strings, extensions appear in the following order"

src/naming.adoc
```diff
@@ -166,13 +171,25 @@ Hence, we define the default version of a standard extension to be the
...
+. One-letter extensions, IMAFDQCBPVH, in this order,
+. Z-prefixed extensions,
+. Su-prefixed extensions,
+. Ss-prefixed extensions,
+. Sv-prefixed extensions,
+. Sh-prefixed extensions,
+. Sm-prefixed extensions,
+. X-prefixed extensions.
+
+For instances, RV32IMACV is canonical, whereas RV32IMAVC is not.
```
aswaterman: "instance" should be singular here.

src/naming.adoc
omasanori: OK, I will update.

src/naming.adoc
omasanori: Thanks!

omasanori:
@aswaterman Addressed comments.

aswaterman (approved):
LGTM, thanks!

PR #2891, opened 2026-04-14
profiles: Update cross reference text

omasanori:
`<<vol:unpriv>>` expands to Volume I, Section I.


aswaterman (commented):
The reason we used "Volume I" here was to avoid an explicit section reference, back when it was impossible to do so programmatically.  Now that we can do so, I think we should refer to the more specific section, e.g. `<<fence, that specification>>`.

aswaterman (approved):

PR #2897, opened 2026-04-14
Reorganize Zc chapters

aswaterman:
In accordance with https://lists.riscv.org/g/sig-documentation/message/691


omasanori:
Just in case, Zimop contains Zcmop that would belong to the Zc chapter.

src/unpriv/q-st-ext.adoc
```diff
@@ -1,3 +1,4 @@
+[[quad-float]]
```
omasanori, line 1: ext:q? It is fine if you plan to rename it with F and D all at once.

aswaterman:
Yes, good point, Zcmop should be moved over.

src/unpriv/q-st-ext.adoc
aswaterman: It must be renamed eventually, but it does not need to be as part of this PR.

My algorithm has been that, when I need to use the new `ext:whatever[]` macro, I rename the `whatever` section.

omasanori: I see, please resolve this comment.

PR #2903, opened 2026-04-14
Align the extension order with the chapter order

omasanori:
While the Krste's plan puts P after V, the Z* extension order, IMAFDQLCBKJTPVH, puts P before V. Sooner is better if we align.

cc @kasanovic

### DIFF
--- a/src/naming.adoc
+++ b/src/naming.adoc
@@ -59,7 +59,7 @@ last letter is "p". For example, "Zifencei" names the instruction-fetch fence ex
 described in <<zifencei>>.
 
 The first letter following the "Z" conventionally indicates the most closely
-related alphabetical extension category, IMAFDQLCBKJTPVH. For the "Zfa"
+related alphabetical extension category, IMAFDQLCBJTVPKH. For the "Zfa"
 extension for additional floating-point instructions, for example, the letter
 "f" indicates the extension is related to the "F" standard extension.
 
@@ -179,7 +179,7 @@ For instance, "RV32IV" is equivalent to "RV32I2V1".
 In canonical ISA name strings, components appear in the following order:
 
 . Base ISA
-. One-letter extensions, MAFDQCBPVH, in this order
+. One-letter extensions, MAFDQCBVPH, in this order
 . Z-prefixed extensions
 . Su-prefixed extensions
 . Ss-prefixed extensions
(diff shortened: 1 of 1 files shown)


aswaterman (commented):
We should change the naming chapter instead of changing the chapter order.  The canonical ordering of extension names is unenforceable for extensions that don't exist yet, so moving P around doesn't break anything.

omasanori:
OK

topperc:
gcc and binutils both have P in their canonical order code because it is in the naming chapter. There's no problem changing the tools, we just need to make sure the right people know to do it.

omasanori:
I updated to align with Krste's plan.

LLVM has code depending on the extension order  including P at `llvm::RISCVISAUtils`. Other than toolchains, Linux kernel might have something for /proc/cpuinfo and such. There may be other tools depending on the order.

omasanori:
Note that the traditional order places K next to B while Krste's plan places K next to P that is next to V, same as the current chapter organization.

omasanori:
Not aligning the chapter order with the extension
order is inconsistent but cheap.

kito-cheng:
Just ack here as a GNU toolchain developer, P isn't ratified and upstream yet, so I am OK with changing that, and I will gonna to update the order once this PR got merged.

aswaterman (approved):

topperc:
So toolchain should output Zk extensions after Zv extensions now?

omasanori:
> So toolchain should output Zk extensions after Zv extensions now?

I think so.

omasanori:
I opened an issue on LLVM. Sorry if hasty.

aswaterman:
IMO we should not change the canonical order for any ratified extensions.

Not sure if we should change the chapter order or just accept that the chapter order disagrees.  It is only a very minor nit if they disagree.

topperc:
> IMO we should not change the canonical order for any ratified extensions.
> 
> Not sure if we should change the chapter order or just accept that the chapter order disagrees. It is only a very minor nit if they disagree.

So this PR should have only moved P?

aswaterman:
The PR description says that only P is being moved.  I didn't notice it moved K as well.  @omasanori can you revert the K change?

omasanori:
OK, I will.

omasanori:
@aswaterman https://github.com/riscv/riscv-isa-manual/pull/2910 Sorry for confusion.

aswaterman:
No prob!

PR #2909, opened 2026-04-15
unpriv: Add Normative Rules on IEEE 754 compliance

omasanori:
Fix https://github.com/riscv/riscv-isa-manual/issues/2835

### DIFF
--- a/src/unpriv/f-st-ext.adoc
+++ b/src/unpriv/f-st-ext.adoc
@@ -1,11 +1,11 @@
 [[single-float]]
-=== "F" Extension for Single-Precision Floating-Point, Version 2.2
+=== ext:f[] Extension for Single-Precision Floating-Point
 
-This chapter describes the F standard extension for _single-precision_
+This chapter describes the ext:f[] standard extension for _single-precision_
 floating-point, which adds [#norm:f_ieee_compliance]#computational instructions
 compliant with the <<ieee754-2008, IEEE 754-2008 arithmetic standard>>'s
-binary32 format and operations.# [#norm:f_depends_zicsr]#The F extension depends
-on the "Zicsr" extension for control and status register access.#
+binary32 format and operations.# [#norm:f_depends_zicsr]#The ext:f[] extension
+depends on the extlink:zicsr[] extension for control and status register access.#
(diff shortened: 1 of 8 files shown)


src/unpriv/f-st-ext.adoc
```diff
@@ -1,11 +1,11 @@
...
+=== ext:f[] Extension for Single-Precision Floating-Point
 
-This chapter describes the F standard extension for _single-precision_
+This chapter describes the ext:f[] standard extension for _single-precision_
 floating-point, which adds [#norm:f_ieee_compliance]#computational instructions
 compliant with the <<ieee754-2008, IEEE 754-2008 arithmetic standard>>'s
-binary32 format and operations.# [#norm:f_depends_zicsr]#The F extension depends
-on the "Zicsr" extension for control and status register access.#
+binary32 format and operations.# [#norm:f_depends_zicsr]#The ext:f[] extension
+depends on the extlink:zicsr[] extension for control and status register access.#
```
omasanori, line 8:
```suggestion
depends on the extlink:zicsr[] extension.#
```
Isn't *for control and status register access* obvious, by the way?

aswaterman (approved):

src/unpriv/f-st-ext.adoc
aswaterman: Not to people who don't know what Zicsr is :)

aswaterman:
Deferring to @mmhus / @jordancarlin

jordancarlin (commented):
I'm not convinced these normative rules are really needed since the F rule claiming compliance with IEEE-754 can be assumed to apply to the other precisions as well, but I don't see a significant issue with adding them either.

aswaterman:
I'd rather err on the side of not adding rules that aren't clearly beneficial.

jordancarlin:
> I'd rather err on the side of not adding rules that aren't clearly beneficial.

I generally agree. Unless someone else feels strongly that this rule is needed, I'd say drop it.

omasanori:
OK, let's leave them without tags. Thanks @aswaterman @jordancarlin !

PR #2911, opened 2026-04-16
zc: Move C before Zca

omasanori:
The C extension is placed after Zca, Zcf and Zcd, while the A and B extensions are placed before the sub-extensions. Defining C before Zca etc. does not harm readability IMHO so this PR aligns C with A and B, though the other direction (defining extensions after sub-extensions) is of course fine.


aswaterman (commented):
The current order follows Krste's suggestion, which makes sense to me because C depends on Zca/Zcf/Zcd.

omasanori:
Thanks for the info!

PR #2913, opened 2026-04-16
Use macros

omasanori:


aswaterman (commented):
New normative rules should be part of separate PRs.  Please tag @jordancarlin and @mmhus (or whoever else they delegate) to review those PRs.

src/profiles/rva23.adoc
```diff
@@ -115,11 +115,11 @@ future RVA profile.
 
 - <<zabha,**Zabha**>> Byte and halfword atomic memory operations.
 - <<zacas,**Zacas**>> Compare-and-Swap instructions.
-- <<ziccamoc,**Ziccamoc**>> Main memory regions with both the coherence and
-  cacheability PMAs must provide `AMOCASQ` level PMA support.
+- extlink:ziccamoc[] Main memory regions with both the coherence and
+  cacheability PMAs must provide AMOCASQ level PMA support.
```
aswaterman: AMOCASQ-level should be hyphenated.  (Same comment elsewhere in this PR.)

Can it be `insn:amocasq[]-level` to be robust to future formatting changes?

omasanori:
> New normative rules should be part of separate PRs. Please tag @jordancarlin and @mmhus (or whoever else they delegate) to review those PRs.

Thanks for your guidance. I put that part to <https://github.com/riscv/riscv-isa-manual/pull/2916>.

src/profiles/rva23.adoc
omasanori: OK. There were some without hyphens. I will update them too.

> Can it be insn:amocasq[]-level to be robust to future formatting changes?

In my opinion, while there is in fact `insn:amocas.q[]`, AMOCASQ-level should look like AMOArithmetic-level, so `insn:amocasq[]-level` seems an abuse of macros.

AMOSwap-level would be another example. It is directly related to `insn:amoswap[]` but not styled as `AMOSWAP`-level.

omasanori: See also: https://github.com/riscv/riscv-isa-manual/pull/2917

aswaterman (approved):

PR #2916, opened 2026-04-16
Add Normative Rule on Ziccamoa

### LINKED ISSUE #2908: On the definition of Ziccamoa and Normative Rules
Volume I defines Ziccamoa as follows:

> If the Ziccamoa extension is implemented, then main memory regions with both the cacheability and coherence PMAs must support all atomics in the Zaamo extension.

While in RVA20U64 and RVA22U64:

> Main memory regions with both the cacheability and coherence PMAs must support AMOArithmetic.

RVA23U64 and RVB23U64 have the same sentence as Volume I.

I prefer RVA20U64 version of wording as it is similar to Ziccamoc and it is as clear as the definition of Volume I. Anyways, having two definition seems unnecessary and unifying it would improve readability.

By the way, when I added Normative Rules for Ziccamoc and Zic64b, I did not find a similar rule for Ziccamoa in the AMO PMA section. Is it intentional?

omasanori:
In addition to the Ziccamoa rule, this also fixed the naming of rules on Ziccamoc and Zama16b.

Fix https://github.com/riscv/riscv-isa-manual/issues/2908

cc @jordancarlin @mmhus

### DIFF
--- a/normative_rule_defs/machine.yaml
+++ b/normative_rule_defs/machine.yaml
@@ -1316,8 +1316,10 @@ normative_rule_definitions:
     tag: "norm:pma_amo_zacas_req_arith"
   - name: pma_amo_zabha_req
     tag: "norm:pma_amo_zabha_req"
-  - name: pma_amo_ziccamoc_req
-    tag: "norm:pma_amo_ziccamoc_req"
+  - name: ziccamoa_pma_amo_req
+    tag: "norm:ziccamoa_pma_amo_req"
+  - name: ziccamoc_pma_amo_req
+    tag: "norm:ziccamoc_pma_amo_req"
   - name: pma_rsrv_levels
     tag: "norm:pma_rsrv_levels"
   - name: ziccrse_pma_rsrv_req
@@ -1346,8 +1348,8 @@ normative_rule_definitions:
     tag: "norm:pma_mag_op_vec"
   - name: pma_mag_op_implicit
     tag: "norm:pma_mag_op_implicit"
-  - name: pma_mag_zama16b_req
-    tag: "norm:pma_mag_zama16b_req"
+  - name: zama16b_pma_mag_req
+    tag: "norm:zama16b_pma_mag_req"
   - name: pma_mo_io_chan_0
     tags: [{name: "norm:pma_mo_io_chan_def", context: true}, "norm:pma_mo_io_chan_0"]
   - name: pma_mo_io_chan_1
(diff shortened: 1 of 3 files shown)


src/priv/machine.adoc
```diff
@@ -3034,6 +3034,10 @@ instructions require ability to perform an arithmetic comparison and a swap oper
 [#norm:pma_amo_zabha_req]#The AMOs specified by the Zabha extension require the same level of support as
 the corresponding instructions in the Zaamo standard extension or the Zacas extension.#
 
+[#norm:pma_amo_ziccamoa_req]#The ext:ziccamoa[] extension requires AMOArithmetic
```
jordancarlin: Per the normative rule naming convention (https://github.com/riscv/riscv-isa-manual/blob/main/tagging_normative_rules.adoc#naming-normative-rules), this should probably be named ziccamoa_pma_amo_req. Several other recently added normative rules in this section should be updated too.

omasanori: OK, I had used the rule above for reference, but it turned out that that rule also violates the convention...

jordancarlin: I think the rule above is different. The rule above describes a feature of how PMAs work while this rule defines an extension. I agree that the rule below this one is also wrong and needs to be updated.

omasanori: I see. I will update the Ziccamoa and Ziccamoc rules.

normative_rule_defs/machine.yaml
```diff
@@ -1316,8 +1316,10 @@ normative_rule_definitions:
     tag: "norm:pma_amo_zacas_req_arith"
   - name: pma_amo_zabha_req
     tag: "norm:pma_amo_zabha_req"
+  - name: ziccamoa_pma_amo_req
+    tag: "norm:ziccamoa_pma_amo_req"
   - name: pma_amo_ziccamoc_req
```
jordancarlin: This name should be changed too.

jordancarlin (changes requested):

normative_rule_defs/machine.yaml
omasanori:
```suggestion
   - name: ziccamoc_pma_amo_req
```

omasanori: Addressed. Thanks!

jordancarlin (approved):

mmhus:
Please run `make update-ref` to make sure everything is updated. The script may be buggy.

omasanori:
@mmhus Thanks! `make update-ref` done.

aswaterman (approved):

aswaterman:
This should auto-merge once the merge conflict is fixed.

omasanori:
@aswaterman Updated.

omasanori:
Ah, sorry, I broke the commit history and unrelated commit was in. Let me fix.

PR #2917, opened 2026-04-16
machine: Fix hyphenation in AMO PMAs

omasanori:
Note that this conflicts with <https://github.com/riscv/riscv-isa-manual/pull/2916> because they update the same section.

### DIFF
--- a/src/priv/machine.adoc
+++ b/src/priv/machine.adoc
@@ -3020,15 +3020,14 @@ possible.
 [#norm:pma_amo_zacas_levels1]#The Zacas extension defines three additional levels of support: AMOCASW,
 AMOCASD, and AMOCASQ.#
 
-[#norm:pma_amo_zacas_levels2]#AMOCASW indicates that in addition to instructions indicated by AMOArithmetic
-level support, the `AMOCAS.W` instruction is supported. AMOCASD indicates that
-in addition to instructions indicated by AMOCASW level support, the `AMOCAS.D`
+[#norm:pma_amo_zacas_levels2]#AMOCASW indicates that in addition to instructions indicated by AMOArithmetic-level support, the `AMOCAS.W` instruction is supported. AMOCASD indicates that
+in addition to instructions indicated by AMOCASW-level support, the `AMOCAS.D`
 instruction is supported. AMOCASQ indicates that in addition to instructions
-indicated by AMOCASD level support, the `AMOCAS.Q` instruction is supported.#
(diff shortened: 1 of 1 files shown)


src/priv/machine.adoc
```diff
@@ -2984,30 +2984,27 @@ all atomic operations# required by the attached processors.
...
-AMOArithmetic
-|_None_ +
-`amoswap` +
-above + `amoand`, `amoor`, `amoxor` +
-above + `amoadd`, `amomin`, `amomax`, `amominu`,
-`amomaxu`
+|AMONone |_None_
+AMOSwap |insn:amoswap[]
+AMOLogical |above + insn:amoand[], insn:amoor[], insn:amoxor[]
+AMOArithmetic |above + insn:amoadd[], insn:amomin[], insn:amomax[], insn:amominu[], insn:amomaxu[]
```
omasanori:
```suggestion
|AMOArithmetic |above + insn:amoadd[], insn:amomin[], insn:amomax[], insn:amominu[], insn:amomaxu[]
```

src/priv/machine.adoc
```diff
@@ -2984,30 +2984,27 @@ all atomic operations# required by the attached processors.
...
-AMOLogical +
-AMOArithmetic
-|_None_ +
-`amoswap` +
-above + `amoand`, `amoor`, `amoxor` +
-above + `amoadd`, `amomin`, `amomax`, `amominu`,
-`amomaxu`
+|AMONone |_None_
+AMOSwap |insn:amoswap[]
+AMOLogical |above + insn:amoand[], insn:amoor[], insn:amoxor[]
```
omasanori:
```suggestion
|AMOSwap |insn:amoswap[]
|AMOLogical |above + insn:amoand[], insn:amoor[], insn:amoxor[]
```

aswaterman (commented):
Can you avoid reformatting the text when making small changes?  It makes it difficult to tell where the changes are.  It is OK if lines sometimes cross 80-character boundaries as a result; that is better than making review difficult.

Adverbs ending in -ly should never be followed by a hyphen; e.g. it is always "naturally aligned".  Use hyphens for compound adjective phrases where the first word is not unambiguously an adverb; e.g. "processor-supported atomic operations" is correct.

omasanori:
Addressed the comment.

mmhus (commented):
I don't think this is a normative rule change. More like a textual improvement.

aswaterman (approved):
Thanks!

PR #2918, opened 2026-04-17
Use macros

omasanori:


omasanori:
> create_normative_rules.py: ERROR: Normative rule zicntr_rdcycle_op references non-existent tag norm:zicntr_rdcycle_op in file normative_rule_defs/counters.yaml

@jordancarlin @mmhus Could you check why this happens while the tag is not removed from the text? <https://github.com/riscv/riscv-isa-manual/commit/aaa99fb2f81c6d965a434b005f43845a1efedf47> is the suspected commit.

jordancarlin:
> > create_normative_rules.py: ERROR: Normative rule zicntr_rdcycle_op references non-existent tag norm:zicntr_rdcycle_op in file normative_rule_defs/counters.yaml
> 
> @jordancarlin @mmhus Could you check why this happens while the tag is not removed from the text? [aaa99fb](https://github.com/riscv/riscv-isa-manual/commit/aaa99fb2f81c6d965a434b005f43845a1efedf47) is the suspected commit.

I think @rpsene is the one who created that workflow. Any ideas?

omasanori:
Hmm, `make update-ref` does not help.

omasanori:
@jordancarlin @mmhus @rpsene @aswaterman By doing `make update-ref` in each commit, it becomes clear that <https://github.com/riscv/riscv-isa-manual/pull/2918/changes/f9b77d369a438caa104cdbfae5bac5348a7a319b> and <https://github.com/riscv/riscv-isa-manual/pull/2918/changes/32c5f10f93de87984a7c9498da8751d20b0999eb> drop tags. Using macros inside Normative Rules does not unconditionally drop tags but drops few tags for unknown reasons, or using macros itself is not the root of cause.

omasanori:
OK, I found the difference. If a macro is used at the end of line inside a Normative Rule, the rule is dropped.

NG:
```asciidoc
[#norm:zicntr_rdcycle_op]#The (...) of the csr:cycle[]
(...)
```

OK:
```asciidoc
[#norm:zicntr_rdcycle_op]#The (...)
(...) of the csr:cycle[] (...)
```

OK:
```asciidoc
[#norm:zicntr_rdcycle_op]#The (...) of the
csr:cycle[] (...)
```

While we can avoid the bug easily, it would be better to fix.

omasanori:
I will submit new PRs to avoid Norm tags conflicts.

PR #2924, opened 2026-04-18
Fix bug with BSD sed

### LINKED ISSUE #2923: Argument of `-i` is mandatory with BSD sed
The `-i` option of the `sed` command for in-place editing takes an argument for the extension of backup file, like `sed -i .bak -e (...)`. The argument is optional with GNU sed but mandatory with BSD sed. On macOS, as a result, `sed -i -e '$a\' ref/*.json` in `make update-ref` is recognized as **making backup files with -e suffix** and `'$a\'` is not recognized as a sed command.

```
$ git status
On branch use-macros
Your branch is up to date with 'origin/use-macros'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        ref/riscv-spec-norm-tags.json-e

nothing added to commit but untracked files present (use "git add" to track)
```

To fix it, do `sed -i .bak -e '$a\' ref/*.json` and then `rm ref/*.bak`.

omasanori:
Fix https://github.com/riscv/riscv-isa-manual/issues/2923

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -164,7 +164,8 @@ check-tags:
 # Required by GitHub pre-commit checks for this repo.
 update-ref: $(DOCS_NORM_TAGS)
 	cp -f $(DOCS_NORM_TAGS) ref
-	sed -i -e '$$a\' ref/*.json
+	sed -i .bak -e '$$a\' ref/*.json
+	rm ref/*.bak
 
 build-norm-rules-json: $(NORM_RULES_JSON)
 build-norm-rules-html: $(NORM_RULES_HTML)


Makefile
```diff
@@ -164,7 +164,8 @@ check-tags:
 # Required by GitHub pre-commit checks for this repo.
 update-ref: $(DOCS_NORM_TAGS)
 	cp -f $(DOCS_NORM_TAGS) ref
-	sed -i -e '$$a\' ref/*.json
+	sed -i .bak -e '$$a\' ref/*.json
+	rm ref/*.bak
```
aswaterman, line 168: Use sed -i '' (i.e. the empty string) and then you don’t need the rm.

omasanori: Thanks. I tried it on macOS but as a result it does not work as expected. (The last line still lacks LF)

omasanori: @aswaterman Is it critical? An alternative option is mandating GNU-compatible sed, but detecting the variant of sed just for avoiding backup files seems overengineering.

aswaterman: No, I'd like to know why it doesn't work but it isn't critical.

aswaterman (approved):

PR #2926, opened 2026-04-18
zifencei: Tweak wording

omasanori:


aswaterman (commented):
the existing text is better; MAJC has an instruction that stores an instruction word, a.k.a. a store instruction word instruction.

PR #2933, opened 2026-04-21
CMO: Use macros

omasanori:
By the way, the most of notes in this file are italic. Does it have a special meaning, or a style inconsistency?

Also, this file is the only user of `x{csrname}` that is expanded to `xenvcfg`. Is it necessary?


aswaterman (approved):
LGTM, thanks!

PR #2934, opened 2026-04-21
Tweak CMO

omasanori:
- Reduced the number of sections and its levels
- Unified the style of notes
- Removed '{csrname}' that is defined globally but used only in CMO

### DIFF
--- a/src/riscv-spec.adoc
+++ b/src/riscv-spec.adoc
@@ -38,7 +38,6 @@ endif::[]
 :data-uri:
 :hide-uri-scheme:
 :footnote:
-:csrname: envcfg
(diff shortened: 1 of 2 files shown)


src/riscv-spec.adoc
```diff
@@ -38,7 +38,6 @@ endif::[]
 :data-uri:
 :hide-uri-scheme:
 :footnote:
-:csrname: envcfg
```
aswaterman, line 41: Forgot about this one.  Looks very silly in retrospect :)

aswaterman (approved):
Thanks!

PR #2951, opened 2026-04-22
rvwmo: Tweak

omasanori:


src/unpriv/rvwmo.adoc
```diff
@@ -92,7 +92,7 @@ operation:
   (e.g., FLD/FSD in RV32) may each give rise to multiple memory operations.
 
 [#norm:rvwmo_v_p_multi_memops]#ISA extensions such as the ext:v[] and the upcoming
-ext:p[] may give rise to multiple memory operations.# However, the memory model for these extensions
+ext:p[] extensions may give rise to multiple memory operations.# However, the memory model for these extensions
```
aswaterman: I think this should be singular, as P is the name of a specific extension.  (The more defensive thing would be to say "P or Zp\*", but I think even that is more speculative, so just "P extension" is OK.)

aswaterman (approved):
LGTM, ping me when my comment is addressed.  Thank you!

src/unpriv/rvwmo.adoc
omasanori: So *the V extension and the upcoming P extension*?

aswaterman: Yeah.

aswaterman: Actually, I just remembered that the P extension doesn't add any new memory accesses.  (The Zilsd extension adds the ones that P can use on RV32.)  The Zilsd memory accesses _are_ formalized in the sense that they decompose into RVWMO primitives.  So maybe just delete the "and the upcoming P extension" and don't mention Zilsd.

omasanori:
Updated.

aswaterman (approved):

PR #2952, opened 2026-04-22
za: Remove unnumbered section

omasanori:
Unnumber section is used only here and bit weird in ToC.


aswaterman (commented):
I think unnumbered sections are a useful tool that we don't use effectively enough.  I agree this looks weird in the TOC, though.  If you add the `[discrete]` tag on the line before before the `===`, it will hide it from the TOC.

PR #2716, opened 2026-03-02
ci: harden Actions workflows and restore 'Normative Rule Changes Detected' PR comment

rpsene:
## Summary

This PR does two things:

### 1. Security hardening (original scope)
- Remove `pull_request_target` from normative tag checks.
- Split normative checks into two workflows:
  - untrusted PR workflow (`check-normative-tags-pr.yml`) with read-only token and no write operations.
  - trusted push/workflow_dispatch workflow (`check-normative-tags.yml`) for reference updates and issue creation.
- Add workflow-level least-privilege defaults (`permissions: contents: read`) and keep write scopes only where required.
- Pin all third-party GitHub Actions to immutable commit SHAs.
- Fix shellcheck/actionlint workflow script warnings.

### 2. Restore "Normative Rule Changes Detected" PR comment (follow-up)
Removing `pull_request_target` also removed the auto-posted PR comment. Restoring it via the GitHub-recommended `workflow_run` split so we can comment on PRs — **including fork PRs** — without ever running untrusted code with write permissions.

- **`check-normative-tags-pr.yml`** — also writes `pr_number.txt` next to `tag_changes.txt` and uploads both under the fixed artifact name `normative-tag-change-report`. The PR number must live in the artifact because `workflow_run.pull_requests` is empty for fork PRs.
- **`check-normative-tags-comment.yml`** (new) — runs on `workflow_run: completed`, executes only base-branch code, downloads the artifact, and posts/updates the PR comment through the existing `.github/scripts/comment-pr-changes.js` (which already de-dupes so re-runs update in place).

Addresses @jordancarlin's ask in #2958 (comment).

## Security model

| Workflow | Trigger | Permissions | Runs PR code? |
|---|---|---|---|
| `check-normative-tags-pr.yml` | `pull_request` | `contents: read` | Yes (untrusted, no write access) |
| `check-normative-tags-comment.yml` | `workflow_run` | `pull-requests: write`, `actions: read`, `contents: read` | No (base-branch copy only) |
| `check-normative-tags.yml` | `push` on main | `issues: write`, `contents: write` | No (trusted branch) |

Pattern reference: [GitHub Security Lab — Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/).

## Expected behaviour after merge
- Normative tag changes on a PR → bot comment appears within seconds of the PR check finishing (same for same-repo and fork PRs).
- No normative tag changes → no comment (grep gate).
- Failed PR check → no comment (conclusion guard).
[...]

PR #2958, opened 2026-04-23
Fix unwanted #

demin-han:


aswaterman (commented):
@jordancarlin @mmhus can one of you look at this?  There is indeed a stray # in the output, but deleting it would appear to break this normative rule.

mmhus:
I think the order of # and subsequent norm rule selection needs a bit of checking. The * in the subsequent norm rule in the list might break the norm rule extraction logic. @UmerShahidengr @jordancarlin What do you think?

jordancarlin:
It looks like these are extra `#`s because I see closing `#` markers for each of the tags immediately preceding these. I think this can be safely merged. It would be a lot easier to tell if the "normative rule changes detected" comment was working correctly. @rpsene any update on when that might be fixed?

rpsene:
> It looks like these are extra `#`s because I see closing `#` markers for each of the tags immediately preceding these. I think this can be safely merged. It would be a lot easier to tell if the "normative rule changes detected" comment was working correctly. @rpsene any update on when that might be fixed?

Thanks for the ping; let resume the fix locally.

rpsene:
@jordancarlin the fix is in #2716 — it re-introduces the 'Normative Rule Changes Detected' PR comment via a `workflow_run` follow-up workflow, which is the GitHub-recommended way to comment on PRs (including fork PRs) without reintroducing the `pull_request_target` risk we removed. Once #2716 lands on main, the bot comment will start appearing again on new PRs.

jordancarlin:
@rpsene I might not have been clear before, but the issue is not that the comment isn't appearing on PRs. The comment appears on PRs just fine (see the comment earlier in this PR thread). The issue is that the reference list of normative rules is not being automatically updated, so every time a PR is merged, it falls further and further out of date. The means the list of differences in the PR comment incorrectly includes changes from all of the other previously merged PRs and makes the comment difficult to use to actually assess changes.

aswaterman (approved):
Merging on the basis of https://github.com/riscv/riscv-isa-manual/pull/2958#issuecomment-4314538408

PR #2966, opened 2026-04-24
Tweak

### LINKED ISSUE #2959: On Zalasr and the “this constrains hardware reordering of memory accesses more than necessary” note in Zaamo
A note in Zaamo states that:

> The instructions in the "A" extension can be used to provide sequentially consistent loads and stores, but this constrains hardware reordering of memory accesses more than necessary. A C++ sequentially consistent load can be implemented as an LR with aq set. However, the LR/SC eventual success guarantee may slow down concurrent loads from the same effective address. A sequentially consistent store can be implemented as an AMOSWAP that writes the old value to x0 and has rl set. However the superfluous load may impose ordering constraints that are unnecessary for this use case. Specific compilation conventions may require both the aq and rl bits to be set in either or both the LR and AMOSWAP instructions.

My understanding is that Zalasr fills the gap, is it correct?

### LINKED ISSUE #2965: On “With the combination of Zaamo, Zabha, and Zalasr all C++ atomic operations can be supported with single instructions.”
The Zalasr section says:

> With the combination of Zaamo, Zabha, and Zalasr all C++ atomic operations can be supported with single instructions.

Is Zacas unnecessary here? As C++11 has `atomic_compare_exchange_weak` and `atomic_compare_exchange_strong`, I thought that we need AMOCAS to support them with single instructions.

omasanori:
Fix https://github.com/riscv/riscv-isa-manual/issues/2959
Fix https://github.com/riscv/riscv-isa-manual/issues/2965

### DIFF
--- a/src/unpriv/zaamo.adoc
+++ b/src/unpriv/zaamo.adoc
@@ -111,14 +111,20 @@ elision. cite:[Rajwar:2001:SLE]
 [NOTE]
 ====
 The instructions in the ext:a[] extension can be used to provide sequentially
-consistent loads and stores, but this constrains hardware
-reordering of memory accesses more than necessary.
-A C++ sequentially consistent load can be implemented as
-an insn:lr[] with _aq_ set. However, the LR/SC eventual
-success guarantee may slow down concurrent loads from the same effective
-address. A sequentially consistent store can be implemented as an insn:amoswap[]
-that writes the old value to `x0` and has _rl_ set. However the superfluous
-load may impose ordering constraints that are unnecessary for this use case.
-Specific compilation conventions may require both the _aq_ and _rl_
-bits to be set in either or both the insn:lr[] and insn:amoswap[] instructions.
+consistent loads and stores, but this constrains hardware reordering of memory
+accesses more than necessary.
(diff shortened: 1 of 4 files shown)


src/unpriv/zaamo.adoc
```diff
@@ -111,14 +111,20 @@ elision. cite:[Rajwar:2001:SLE]
...
+
+A sequentially consistent store can be implemented as an insn:amoswap[] that
+writes the old value to `x0` and has _rl_ set. However the superfluous load may
+impose ordering constraints that are unnecessary for this use case.
+
+Specific compilation conventions may require both the _aq_ and _rl_ bits to be
+set in either or both the insn:lr[] and insn:amoswap[] instructions.
+
+The ext:zalasr[] extension provides load-acquire and store-release instructions
+without aforementioned drawbacks.
```
aswaterman: "the" aforementioned drawbacks

aswaterman (approved):
Thanks!  Ping me once it's ready to merge.

omasanori:
@aswaterman Thanks for your review! Updated.

Issue #2975, opened 2026-04-24
[Clarification] - Is mstatus mandatory and do certain WARL fields have to support both 0 and 1?


davidharrishmc:
## Summary
Most CSRs state whether they are mandatory to exist.  Looking to make this explicit for the ones where it is not stated.

Is mstatus mandatory?  
Does MIE and MPIE have to support both 0 and 1?
If U is supported, does MPRV have to support both 0 and 1?
If S is supported, does MXR have to support both 0 and 1?
If S and satp vm modes are supported, does SUM have to support both 0 and 1?
If S is supported, does TVM have to support both 0 and 1?
If U is supported, does TW have to support both 0 and 1?
If S is supported, does TSR have to support both 0 and 1?

---

## Spec Reference
**Permalink:** https://github.com/riscv/riscv-isa-manual/blob/812318e97aadbba6334cc5a568b216a9b8e5d6fb/src/priv/machine.adoc

_Text Reference (Optional): e.g. Privileged 3.1.6

---

aswaterman:
> Is mstatus mandatory?

Yes.

> Does MIE and MPIE have to support both 0 and 1?

Yes.

> If U is supported, does MPRV have to support both 0 and 1?

Yes.

> If S is supported, does MXR have to support both 0 and 1?

If page-based virtual-memory is supported.

> If S and satp vm modes are supported, does SUM have to support both 0 and 1?

Same as above.

> If S is supported, does TVM have to support both 0 and 1?

Same as above.

> If U is supported, does TW have to support both 0 and 1?

Yes.

> If S is supported, does TSR have to support both 0 and 1?

Yes.

aswaterman:
If you're going to attack this, see if you can make the new text as concise as possible.  This is good stuff to clarify, but it risks adding clutter.

PR #2978, opened 2026-04-24
Update documentation for mstatus and mstatush registers

davidharrishmc:
Clarify requirements for `mstatus` and `mstatush` registers regarding implementation and writable bit fields per #2975.

Aimed for concise summary rather than per-field specification.

### DIFF
--- a/src/priv/machine.adoc
+++ b/src/priv/machine.adoc
@@ -352,6 +352,11 @@ include::images/wavedrom/mstatusreg.edn[]
 .Additional machine-mode status (`mstatush`) register  for RV32.
 include::images/wavedrom/mstatushreg.edn[]
 
+The `mstatus` (and RV32 `mstatush`) registers must be implemented.  
+The single bit fields must be writable with both 0 and 1 unless
+specified otherwise (e.g. many fields are read-only zero when
+an associated extension is not implemented.)
+
 [[privstack]]
 ===== Privilege and Global Interrupt-Enable Stack in `mstatus` register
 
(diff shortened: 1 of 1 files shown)


aswaterman:
Now that I'm seeing how this might play out, I wonder if I gave bad advice.  I think we can keep it concise without adding a level of indirection that's a little bit vague.  For example:

```
diff --git a/src/priv/machine.adoc b/src/priv/machine.adoc
index 43fa29d0..5a0498ac 100644
--- a/src/priv/machine.adoc
+++ b/src/priv/machine.adoc
@@ -626,7 +626,7 @@ privilege with which S-mode loads and stores access virtual memory.
 [#norm:mstatus_sum_op_no-vm]#SUM has no effect when page-based virtual memory is not in effect.#
 Note that, [#norm:mstatus_sum_op_mprv_mpp]#while SUM is ordinarily ignored when not
 executing in S-mode, it _is_ in effect when MPRV=1 and MPP=S.#
-[#norm:mstatus_sum_rdonly0]#SUM is read-only 0 if S-mode is not supported or if `satp`.MODE is read-only 0.#
+[#norm:mstatus_sum_rdonly0]#SUM is read-only 0 if S-mode is not supported or if `satp`.MODE is read-only 0; otherwise, it must be writable.#
 
 [[norm:mstatus_mxr_sum_op_acc_fault]]
 The MXR and SUM mechanisms only affect the interpretation of permissions
```

("Writable" suffices for 1-bit WARL fields.  For multi-bit WARL fields we probably need to be more specific.)  What do you think?

davidharrishmc:
Here's another stab at it with your suggestions.  Note that I added language that MXR and TVM must be 0 when satp.MODE is read-only zero.  Otherwise, we need to specify what behavior is allowed for these bits when S is supported but virtual memory is not supported.

aswaterman (approved):
LGTM (I made minor tweaks).

PR #2979, opened 2026-04-25
Add Zvfofp8min v0.9 spec

aswaterman:


omasanori:
Does this extension depend on Zvfbfmin, or is this extension independent of Zvfbfmin while requiring some sort of BF16 support?

aswaterman:
Zve32f is the only dependence.  Obviously this extension complements Zvfbfmin etc. but it is not required.

omasanori:
It sounds weird to me that it allows an implementation supporting conversion between BF16 and OFP8 but not supporting conversion between BF16 and FP32. Did I overlook anything?

aswaterman:
I agree many vendors will choose to bundle the extensions.  But the reason is that other extensions add BF16 matrix operations (e.g. forthcoming VME) that do not depend on Zvfbfmin, and there's no reason to require additional BF16 support if the only goal is to form BF16 operands to feed into matrix operations.

omasanori:
I see, thanks. A note in the section and/or a section in the rationale appendix would be nice.

aswaterman:
@omasanori agreed, done.

src/unpriv/zvfofp8min.adoc
```diff
@@ -0,0 +1,108 @@
+=== ext:zvfofp8min[] OFP8 Conversion Extension, Version 0.9
+
+The ext:zvfofp8min[] extension provides basic support for the 8-bit
+floating-point formats OFP8 E4M3 and OFP8 E5M2.
+
+The ext:zvfofp8min[] extension requires the extlink:zve32f[] extension.
```
nibrunieAtSi5, line 6: I am sorry if we already solve this, but we do not require any BF16 support (like Zvfbfmin or the future Zvfbfa) ?

nibrunieAtSi5: Already answered by @aswaterman:

>  Zve32f is the only dependence. Obviously this extension complements Zvfbfmin etc. but it is not required.

aswaterman: Yeah, I imagine many vendors will couple them but it is not strictly required to do so.

nibrunieAtSi5 (approved):
LGTM

Should we eventually embed the opcode encodings? (wavedrom snippets) and pseudo-code descriptions?

omasanori:
> Should we eventually embed the opcode encodings? (wavedrom snippets) and pseudo-code descriptions?

Did you mean per-instruction subsections like the current bitmanip and cryptography chapters? My understanding is that we will remove them after deploying per-instruction pages generated from UnifiedDB, and it will happen in not-so-far future.

PR #2981, opened 2026-04-25
Zf, Zvfbf: Use macros

omasanori:


src/unpriv/zfinx.adoc
```diff
@@ -1,47 +1,49 @@
-[[sec:zfinx]]
-=== "Zfinx", "Zdinx", "Zhinx", "Zhinxmin" Extensions for Floating-Point in Integer Registers, Version 1.0
+[[ext:zfinx]]
+=== ext:zfinx[], ext:zdinx[], ext:zhinx[], ext:zhinxmin[] Extensions for Floating-Point in Integer Registers
 
-This chapter defines the "Zfinx" extension (pronounced "z-f-in-x")
+This chapter defines the ext:zfinx[] extension (pronounced _z-f-in-x_)
```
aswaterman: I think quotation marks are better than italics for pronunciation guidance.

aswaterman (approved):

src/unpriv/zfinx.adoc
omasanori: OK. I wanted to avoid using straight quotes (") but revisit later when I found the proper way.

omasanori:
Updated to address comment.

aswaterman (approved):

PR #2996, opened 2026-04-27
Clarify that WRS interrupt interaction matches WFI

aswaterman:
The Zawrs spec already says that the WRS instructions follow WFI's rules for resumption upon interrupts, but the surrounding language suggests something different (that they need only be pending, not locally enabled).  Resolve this contradiction by changing the language to match that of WFI.

Note that implementations that resolve this contradiction in the opposite direction are technically still correct, since it is valid to break out of the WRS stall for any reason.


ved-rivos (approved):
LGTM. A nit about terminology. The privilege spec uses "locally enabled" and "individually enabled" interchangeably. Should this be unified to "individually enabled"?

aswaterman:
@ved-rivos I agree we should be consistent, but I'm going to make the opposite choice and standardize on "local" since it's the antonym of "global".

PR #2997, opened 2026-04-27
Add Zvt family of VME extensions

aswaterman:


src/unpriv/zvt.adoc
```diff
@@ -0,0 +1,835 @@
...
+raises an illegal-instruction exception if `mstatus`.MS=Off (or if V=1
+and `vsstatus`.MS=Off).  Otherwise, it changes `mstatus`.MS to Initial
+(and, if V=1, changes `vsstatus`.MS to Initial).
+This instruction is included in the Zvtbase extension.
+
+`vtdiscard` raises an illegal-instruction exception if `vtype`.vill=1, but
+does not raise an exception as a result of `mtype`.`mtwiden` being zero.
+
+NOTE: `vtdiscard` does not actually write the tile state.  For ABI
+purposes, the tile state becomes UNPREDICTABLE as a result of executing
```
omasanori: UNPREDICTABLE seems undefined throughout the specification. UNSPECIFIED?

src/unpriv/zvt.adoc
```diff
@@ -0,0 +1,843 @@
...
+
+The invalid and overflow scalar floating-point exception flags are
+updated by the results of floating-point matrix operations.  The
+inexact, underflow, and divide-by-zero flags are not updated,
+nor is the fixed-point saturation flag.
+
+For SEW {le} 16, floating-point products are bulk-normalized
+accumulated in fixed-point arithmetic, then converted to FP32, rounding to odd,
+according to the
+https://github.com/riscv/riscv-isa-manual/blob/03c4abb8b4b276938c3fd96e3eb53c5840e3c24b/src/zvdota.adoc#risc-v-bulk-normalization-algorithm[RVBNA Algoirhtm].
```
awiseAtCodasip: Spelling: "RVBNA Algoirhtm" should be "RVBNA Algorithm"

aswaterman: Oops, thanks.

src/unpriv/zvt.adoc
```diff
@@ -0,0 +1,859 @@
...
+existing vector instructions are used for pre- and post-processing of matrices.
+Hence, the Zvtbase extension requires the Zve32x extension.
+
+The Zvt extensions are compatible with the RV32E, RV32I, RV64E, and RV64I base
+ISAs, but it is expected that RV64I will be by far the most common base-ISA
+choice.
+
+The following table lists the computational extensions in the Zvt family.
+
+[cols="1,1,3,3", options="header"]
```
topperc: Based on how the PDF renders, these column sizes need to be adjusted.

aswaterman: @topperc thanks for pointing this out.  I've improved it.

src/unpriv/zvt.adoc
```diff
@@ -0,0 +1,859 @@
...
+
+The 3-bit field for pattern currently only has two encodings defined,
+row and column. The pattern field defines how the index field is
+interpreted.  For the currently defined row and column patterns, the
+index field is treated as an unsigned integer providing a
+one-dimensional index of the row or column.
+Row and column indices greater than or equal to ETE are reserved.
+
+NOTE: We expect most implementations will ignore reserved bits in TSS,
+taking `pattern` mod 2, and taking the index within the pattern mod ETE.
```
xypron, line 532: We should create an illegal instruction exception if reserved values are encountered.
Otherwise we will end up with software that relies on bits being ignored.

src/unpriv/zvt.adoc
```diff
@@ -0,0 +1,859 @@
...
+{ge} `min(vl, ETE)`.
+
+If `mtwiden`=0, these instructions are reserved.
+
+===== Matrix Arithmetic Instructions
+
+These instructions all have the form C += A^T * B, where C is held in
+a matrix tile and A and B are supplied by vector register groups.
+
+Unsupported `vtype` or `mtype` settings are reserved.
```
xypron, line 651: We should throw an illegal-instruction exception if a reserved value is hit.

src/unpriv/zvt.adoc
```diff
@@ -0,0 +1,859 @@
...
+
+If `vsstatus`.MS is Dirty, `vsstatus`.SD is 1;
+otherwise, `vsstatus`.SD is set in accordance with existing specifications.
+
+If `mstatus`.MS is Dirty, `mstatus`.SD is 1;
+otherwise, `mstatus`.SD is set in accordance with existing specifications.
+
+For the purposes of the `mstatus`.VS and `vsstatus`.VS fields, all Zvt
+instructions (including configuration instructions) are considered to modify
+vector state.
```
xypron, line 834: Why should an instruction that does not update vectors but only matrix tiles be considered as modifying vector status? This will lead to unnecessary saving of vector register values in code switching threads.

We should not require this.

xypron (changes requested):

PR #2999, opened 2026-04-28
Use {asterisk} in macros

omasanori:
The raw `*` may or may not be interpreted as `<strong>...</strong>`, `<li>...</li>` or the character itself, depending on context. For safety, instead of playing with asciidoctor and insert backslash where necessary, let's use `{asterisk}` where the character itself is needed. I did the conversion for macros since I did the markup so it was easy to me to spot where asterisks are used. There are other instances too, but no rush. Let's do it for the rest later.


aswaterman:
Personally, I would prefer to continuing to allow * where it is safe… though I acknowledge the context-dependency makes this a bit fraught.

omasanori:
OK, then I will close this PR after checking raw asterisks don't harm at these points.

omasanori:
I confirmed these ones are not misinterpreted. Thanks for feedback, @aswaterman!

aswaterman:
Cool, thanks for checking.

PR #3000, opened 2026-04-28
Clarify norm:fcvt_int_float_valid_input

omasanori:
The current sentence edited by me accidentally introduce the ambiguity that FCVT.D.S might be affected by norm:fcvt_int_float_valid_input. This PR clarifies that FCVT.W[U].S and FCVT.L[U].S are affected by the rule.

By the way, the `<<int_conv>>` table itself is not in any Normative Rules if I understand correctly. Is it OK for certification documents?


jordancarlin (approved):

jordancarlin:
> By the way, the `<<int_conv>>` table itself is not in any Normative Rules if I understand correctly. Is it OK for certification documents?

That seems like a tricky table to tag in a useful way and may not be worth it. If you can find a good way to do it, though, then go ahead.

aswaterman (approved):

PR #3003, opened 2026-04-29
altfmt_A/B and bs rearranged inside vtype CSR

efocht:
The order of the altfmt_A, altfmt_B, bs bits in the vtype CSR were rearranged such that
the altfmt bits are adjacent to unused reserved bits. This way a future extension of supported
data formats (e.g. IEEE P3109) could simply extend the altfmt bits in a contiguous way, and get
merged into a wider altfmt_AB field while staying compatible.


efocht:
Apologies, this PR landed in the wrong place, it should be on the riscv/integrated-matrix-extension repo. Closing it here.

PR #3004, opened 2026-04-30
Add norm:fcvt_int_quad_valid_input

### LINKED ISSUE #3001: [NormRules] - Q has no rules similar to norm:fcvt_int_double_valid_input
## Description
While the D extension has a Normative Rule that the range of valid inputs for FCVT.W[U].D and FCVT.L[U].D and their behevior for invalid inputs are the same as FCVT.W[U].S and FCVT.L[U].S, the Q extension has no such rule for FCVT.W[U].Q and FCVT.L[U].Q.

## Any changes required
Add fcvt_int_quad_valid_input similar to fcvt_int_double_valid_input in the Quad-Precision Convert and Move Instructions section.

(cc @riscv/csc)

omasanori:
Fix https://github.com/riscv/riscv-isa-manual/issues/3001


omasanori:
cc @jordancarlin @mmhus

mmhus:
Looks ok to me, though it looks like it also includes some extra formatting changes. @jordancarlin what do you think?

jordancarlin (approved):
The normative rule name changes look fine to me, but there are also some changes to the actual text that @aswaterman should review.

aswaterman (approved):

PR #3006, opened 2026-05-01
Update sstc.adoc with TM bit explanation

davidharrishmc:
Clarified the behavior of the TM bit in mcouteren when Sstc is enabled.


src/priv/sstc.adoc
```diff
@@ -22,3 +22,6 @@ extensions, and 'tc' for timecmp). [#norm:stimecmp_exist]#This extension adds th
 CSR (<<stimecmp>>)# [#norm:vstimecmp_exist]#and the VS-level `vstimecmp` CSR (<<vstimecmp>>).# [#norm:stce_bit_exist]#This extension adds the `STCE` bit to the `menvcfg`
 (<<sec:menvcfg>>) and `henvcfg`
 (<<sec:henvcfg>>) CSRs.#
+
+When Sstc is enabled, the TM bit in `mcouteren` cannot be read-only zero because that would prevent access
+to `stimecmp`.
```
aswaterman, line 27: I don't think adding this to the spec is necessary.  If stimecmp isn't readable at supervisor level, then the hart doesn't actually implement Sstc at supervisor level.  It doesn't matter why stimecmp wasn't readable. (An implementation that has writable TM but has TM set to 0 also doesn't implement Sstc at supervisor level, for example.)

PR #2991, opened 2026-04-27
Added normative rules of all unpriv profile extensions

UmerShahidengr:
This PR extracts normative rules for unprivileged extensions that were recently added to the ISA manual following the merger of RISC-V profiles into a unified document.

The following extensions now have properly tagged normative rules:
- **Memory Atomicity & Ordering**: Zama16b, Ziccamoa, Ziccamoc, Zicclsm, Ziccrse
- **Reservation Sets**: Za128rs, Za64rs
- **Cache Blocks**: Zic64b
- **Instruction Fetch**: Ziccif
- **Vector BF16**: Zvfbfmin, Zvfbfwma

Each extension now has:
1. Tagged normative text in the corresponding `.adoc` file
2. A YAML definition file in normative_rule_defs mapping normative rules to their tags

Also fixed #2947

PR #3008, opened 2026-05-02
Reorganize NormRules of PMA-related extensions from profiles

omasanori:
I had added some rules in machine.yaml but after reviewing <https://github.com/riscv/riscv-isa-manual/pull/2991> I thought they should have been placed on dedicated files. This PR did so.

Personally I think these NormRules are redundant but they stay as-is in this PR.

cc @jordancarlin @mmhus

### DIFF
--- a/normative_rule_defs/machine.yaml
+++ b/normative_rule_defs/machine.yaml
@@ -1316,14 +1316,8 @@ normative_rule_definitions:
     tag: "norm:pma_amo_zacas_req_arith"
   - name: pma_amo_zabha_req
     tag: "norm:pma_amo_zabha_req"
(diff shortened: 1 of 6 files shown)


jordancarlin:
> Personally I think these NormRules are redundant but they stay as-is in this PR.

If they are redundant with other normative rules, I am happy to review a PR that removes the redundancy. Just make sure to point out which other rule you think already covers the same thing to make review easy.

jordancarlin (approved):
LGTM. This is an improvement even if we refactor to drop redundancy later.

aswaterman (approved):

PR #3011, opened 2026-05-02
Fix Makefile portability: replace echo -e with printf

### LINKED ISSUE #2927: `echo -e` is not portable
Makefile uses `echo -e` but it is not supported on macOS (and \*BSD). A portable alternative to the `-e` option is using the `printf` command instead of `echo`.

mritunjayyy07:
Fixes #2927 
The Makefile uses `echo -e` for printing formatted output, but `-e` is not portable and is not supported consistently on macOS/BSD systems.

This change replaces `echo -e` with `printf`, which is POSIX-compliant and behaves consistently across platforms.

Also replaces `\e` with portable escape sequence `\033`.

Tested on macOS.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -193,19 +193,19 @@ $(BUILD_DIR)/%.pdf: $(SRC_DIR)/%.adoc $(ALL_SRCS)
 	$(WORKDIR_SETUP)
 	$(DOCKER_CMD) $(DOCKER_QUOTE) $(ASCIIDOCTOR_PDF) $(OPTIONS) $(REQUIRES) $< $(DOCKER_QUOTE)
 	$(WORKDIR_TEARDOWN)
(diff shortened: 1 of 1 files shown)


aswaterman:
@mritunjayyy07 thanks. You need to upload your SSH key to GitHub as a signing key for this PR to be acccepted; see https://github.com/riscv/riscv-isa-manual/blob/main/CONTRIBUTING.md#signing-github-commits--step-by-step

aswaterman:
And you probably need to redo the commit force-push over your branch afterwards.

mritunjayyy07:
I used GPG key because SSH was not working for some reason.
is it fine?

omasanori:
If you use a SSH key for signing, you need to register the public key to GitHub for signing even if you have already been register that key for authentication.

mritunjayyy07:
I had registered both signin and auth on github but still it did not work for me for some reason.

omasanori:
I see.

aswaterman (approved):

PR #3012, opened 2026-05-03
C, Zc: Use macros

omasanori:


src/unpriv/zca.adoc
```diff
@@ -49,9 +49,9 @@ conventions.
 However, for microcontrollers that only provide hardware single-precision
 floating-point units and whose programs use the ILP32F calling convention, the
 single-precision loads and stores are used at least as frequently as
-double-precision loads and stores are used in LP64D and ILP32D---hence the
+double-precision loads and stores are used in LP64D and ILP32D -- hence the
```
aswaterman: The original is correct: emdash with no spaces.

src/unpriv/zclsd.adoc
```diff
@@ -4,24 +4,24 @@
...
-Zclsd depends on Zilsd and Zca. It has overlapping encodings with Zcf and is thus incompatible with Zcf.
+ext:zclsd[] depends on ext:zilsd[] and ext:zca[]. It has overlapping encodings with ext:zcf[] and is thus incompatible with ext:zcf[].
 
 ==== Use of `x0` as operand
 
-[#norm:Zilsd_c-ldsp_x0]#For C.LDSP, usage of `x0` as the destination is reserved.#
+[#norm:Zilsd_c-ldsp_x0]#For insn:c.ldsp[], usage of `x0` as the destination is reserved.#
 
-If using `x0` as `src` of C.SDSP, the entire 64-bit operand is zero — i.e., register `x1` is not accessed.
+If using `x0` as `src` of insn:c.sdsp[], the entire 64-bit operand is zero -- i.e., register `x1` is not accessed.
```
aswaterman: If the goal were to use an emdash, this should also be `---` with no spaces, but I think this should just be a comma.

aswaterman (commented):
Looks great, thanks.  Ping me when it's ready to merge.

src/unpriv/zclsd.adoc
omasanori: On AsciiDoc, unlike LaTeX, two hyphens are replaced with a em dash and there are no replacement for en dash.

omasanori:
@aswaterman Updated. Note that `--`, not `---`, is replaced with an em dash on AsciiDoc (asciidoctor).

src/unpriv/zclsd.adoc
aswaterman: Ah, right.

aswaterman:
Thanks for the reminder!

aswaterman (approved):

omasanori:
I found that some `--` were not replaced with em dashes due to <https://github.com/asciidoctor/asciidoctor/issues/1578>.

PR #3014, opened 2026-05-03
Add symbols for em and en dashes

omasanori:
On AsciiDoc, in some context, `--` is replaced with an em dash (—) as specified in <https://docs.asciidoctor.org/asciidoc/latest/subs/replacements/>. However, currently asciidoctor has caveats as described in <https://github.com/asciidoctor/asciidoctor/issues/1578>. AsciiDoc has no similar replacement substitutions for en dash (–). This PR adds `{emdash}` and `{endash}` for convenience.

### DIFF
--- a/src/symbols.adoc
+++ b/src/symbols.adoc
@@ -16,3 +16,9 @@
 :pm: ±
 :rightarrow: →
 :asterisk: &#42;
+// https://docs.asciidoctor.org/asciidoc/latest/subs/replacements/
+// https://github.com/asciidoctor/asciidoctor/issues/1578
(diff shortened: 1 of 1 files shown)


omasanori:
The Documentation Developer Guide recommends to [avoid em dash](https://github.com/riscv/docs-dev-guide/blob/main/src/vale.adoc#ellipses), by the way.

omasanori:
Added zero-width joiners (U+200D) before and after en dash to prevent line breaks inside range (ex. `1{endash}20` or `x8{endash}x15`)

aswaterman (approved):

PR #3015, opened 2026-05-03
Removed Zcf instructions from pointer masking

davidharrishmc:
Pointer masking only operates in RV64, while Zcf instructions only can exist in RV32.  Therefore, pointer masking should not apply to any Zcf instructions.


aswaterman (commented):
Can't we just truncate this sentence after "All instructions mapping to any of the above"?  The contents after the "and" are just restatement of the contents before it.

Also, I'd replace "All instructions in RV32A and RV64A" with "All instructions in the A extension".

Between the two, we say the right things without bringing the base ISA distinction into the picture.  Better separation of concerns.

davidharrishmc:
The A extension is not complete because there is also Zabha and Zacas.  I've tried to capture that with amo*.

Shortened the compressed list.  Some readers might not remember that C.LWSP maps to LW, so I think it's worth mentioning.

If you don't like this, I'll change to exactly what you had written above.

aswaterman:
Yes, the fact that Zabha etc. are omitted is an indication that this text is problematic.  How about we just say something to the effect of "all explicit memory accesses, unless otherwise stated", and delete all the rest.

davidharrishmc:
Sorry, I might not be understanding you.  I updated atomics to "all explicit memory accesses..."  Did you prefer to delete the entire bulleted list instead?

aswaterman:
I didn't mean to add that under atomics--I meant just as a blanket statement for the whole ISA.  That covers e.g. LW, HLW, C.LWSP as well.

davidharrishmc:
I'm sure I'm not understanding you.  The opening sentence is already "Pointer masking applies to all explicit memory accesses."  Just cut it off after that? Doesn't flow very well to the next paragraph. If something else, could you just send the text as you want it?

One thing I like about the list is that CMOs and fences are on the list, with a comment about unused address fence address fields. That resolves some ambiguity about whether cbo.inval is a memory access, for instance.

aswaterman:
Sorry, I didn't realize the preceding sentence already says that PM is applied to all explicit memory accesses.  My point is I want to get rid of this list, because it doesn't serve a useful purpose and it creates a maintenance burden.

We can get rid of the next paragraph too by simply using an "unless otherwise specified" on the first sentence.

I am thinking something like this, but feel free to improve.


```
diff --git a/src/priv/zpm.adoc b/src/priv/zpm.adoc
index 9a6a901e..53c9fcd6 100644
--- a/src/priv/zpm.adoc
+++ b/src/priv/zpm.adoc
@@ -125,28 +125,11 @@ Linux places kernel addresses in the upper half of the address space and user ad
 
 ===== Memory Accesses Subject to Pointer Masking
 
-Pointer masking applies to all explicit memory accesses. Currently, in the Base and Privileged ISAs, these are:
-
-* **Base Instruction Set**: LB, LH, LW, LBU, LHU, LWU, LD, SB, SH, SW, SD.
-* **Atomics**: All instructions in RV32A and RV64A.
-* **Floating Point**: FLW, FLD, FLQ, FSW, FSD, FSQ.
-* **Compressed**: All instructions mapping to any of the above, and C.LWSP, C.LDSP, C.FLWSP, C.FLDSP, C.SWSP, C.SDSP, C.FSWSP, C.FSDSP.
-* **Hypervisor Extension**: HLV.\*, HSV.* (in some cases; see <<sec:hstatus>>).
-* **Cache Management Operations**: All instructions in Zicbom, Zicbop and Zicboz.
-* **Vector Extension**: All vector load and store instructions in the ratified RVV 1.0 spec.
-* **Zicfiss Extension**: SSPUSH, C.SSPUSH, SSPOPCHK, C.SSPOPCHK, SSAMOSWAP.W/D.
-* **Assorted**: FENCE, FENCE.I (if the currently unused address fields become enabled in the future).
+Unless otherwise specified, pointer masking applies to all explicit memory accesses.
 
 [NOTE]
 ====
-This list will grow over time as new extensions introduce new instructions that perform explicit memory accesses.
-====
-
-For other extensions, pointer masking applies to all explicit memory accesses by default. Future extensions may add specific language to indicate whether particular accesses are or are not included in pointer masking.
-
-[NOTE]
-====
-It is worth noting that pointer masking is not applied to `SFENCE.\*`, `HFENCE.*`, `SINVAL.\*`, or `HINVAL.*`. When such an operation is invoked, it is the responsibility of the software to provide the correct address.
+Pointer masking is not applied to `SFENCE.\*`, `HFENCE.*`, `SINVAL.\*`, or `HINVAL.*`. When such an operation is invoked, it is the responsibility of the software to provide the correct address.
 ====
 
 [#norm:pm_mprv_spvp]#MPRV and SPVP affect pointer masking as well, causing the pointer masking settings of the effective privilege mode to be applied.# [#norm:pm_mxr_exception]#When MXR is in effect at the effective privilege mode where explicit memory access is performed, pointer masking does not apply.#
```

davidharrishmc:
I can't access this through Github edit anymore due to lack of a DCO.

aswaterman:
@davidharrishmc just to be clear, are you going to fix the DCO, or should I ram through my proposed change?

davidharrishmc:
Go ahead and ram it through.  It would take me a little while to do this the proper way now.

aswaterman:
K.

PR #3016, opened 2026-05-03
Rearrange the description of floating-point register state

omasanori:
This moves the description of `f` registers, `fcsr` and FLEN to the introduction of the scalar floating-point chapter and leave the actual value of FLEN in F/D/Q sections, as I think it is more natural. Does it make sense?

### DIFF
--- a/src/unpriv/zf.adoc
+++ b/src/unpriv/zf.adoc
@@ -10,14 +10,14 @@ instructions for computation on single-precision floating-point values.
 The extlink:d[] and extlink:q[] extensions widen those registers to
 hold double- and quad-precision floating-point values, respectively, and add
 instructions for computation on those formats.
-Several additional extensions with the ext:zf[] and ext:zd[] prefixes provide
+Several additional extensions with the ext:zf[] prefix provide
 additional computational instructions.
 
-The extlink:zfinx[] and extlink:zdinx[] extensions add
-computational instructions analogous to those in the ext:f[] and ext:d[] extensions,
-but they instead operate on floating-point numbers in the `x` registers.
-These extensions, intended for lower-cost systems, are incompatible with the
-ext:f[] and ext:d[] extensions.
+The extlink:zfinx[] and extlink:zdinx[] extensions add computational
+instructions analogous to those in the ext:f[] and ext:d[] extensions, but they
+instead operate on floating-point numbers in the `x` registers. These extensions,
+intended for lower-cost systems, are incompatible with the ext:f[] and ext:d[]
+extensions.
 
 :sectnums!:
 [discrete]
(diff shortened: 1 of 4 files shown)


src/unpriv/zf.adoc
```diff
@@ -48,6 +41,87 @@ While the BF16 cite:[bfloat16], E4M3 and E5M2 formats are not defined in IEEE
...
+floating-point encoding techniques. Compiler support and calling conventions for
+split register file architectures are well understood, and using dirty bits on
+floating-point register file state can reduce context-switch overhead.
+
+The extlink:zfinx[] and extlink:zdinx[] extensions add computational
+instructions analogous to those in the ext:f[] and ext:d[] extensions, but they
+instead operate on floating-point numbers in the `x` registers. These extensions,
+intended for lower-cost systems, are incompatible with the ext:f[] and ext:d[]
+extensions.
+====
```
omasanori, line 118: This note might be suitable for the rationale appendix, but I'd leave it in a main chapter in this PR.

src/unpriv/zf.adoc
```diff
@@ -48,6 +41,87 @@ While the BF16 cite:[bfloat16], E4M3 and E5M2 formats are not defined in IEEE
 standard. E4M3 and E5M2 are specified by the Open Compute Project (OCP) as the
 OCP FP8 (OFP8) formats cite:[ocp-fp8].
 
+[discrete]
```
aswaterman, line 52: I would like to use the discipline of always writing

```
:sectnums!:
[discrete]
=== Title
:sectnums:
```

even when two unnumbered sections are next to each other.  This makes it more obvious what's going on when reading the code, and it makes it easier to move sections around.

src/unpriv/zf.adoc
aswaterman: Ack.  I'd like to focus on getting the manual in good shape before doing rationale-related refactoring.  And in most cases that kind of refactoring should be part of a separate PR, anyway.

src/unpriv/zf.adoc
```diff
@@ -10,19 +10,12 @@ instructions for computation on single-precision floating-point values.
...
-Several additional extensions with the ext:zf[] and ext:zd[] prefixes provide
+Several additional extensions with the ext:zf[] prefix provide
 additional computational instructions.
 
-The extlink:zfinx[] and extlink:zdinx[] extensions add
-computational instructions analogous to those in the ext:f[] and ext:d[] extensions,
-but they instead operate on floating-point numbers in the `x` registers.
-These extensions, intended for lower-cost systems, are incompatible with the
-ext:f[] and ext:d[] extensions.
-
```
aswaterman, line 21: I'd rather not move this text to commentary.  It's true that it adds no normative content, but it is part of the introduction to the Zf chapter, which acts as a guide for the chapter's overall contents.

src/unpriv/zf.adoc
aswaterman: Hmm, actually, if you agree with my request to restore the Zfinx paragraph in the previous section, then the remaining text here really does belong in the rationale appendix for the F extension.  If the split regfile rationale is left here, it looks like a contradiction.

So, I'd support moving the first paragraph to rationale sooner rather than later.  Feel free to do it atomically in this PR, since it's logically related to the changes you are making.

src/unpriv/zf.adoc
omasanori: I moved that paragraph back, while I added a new paragraph with slightly different wording here.

omasanori:
Updated to address comments.

src/unpriv/zf.adoc
omasanori: OK, I will do it in this PR

omasanori:
Moved a paragraph to the rationale appendix.

omasanori:
@aswaterman Ping

aswaterman (approved):
Looks good now, thanks!

aswaterman:
@omasanori I messed up the merge-conflict resolution via the web interface.  (Clearly I have no idea how to use the web interface.)  This restored some deleted text in `d-st-ext`.  Can you fix this up and ping me again when it's ready to merge?

omasanori:
@aswaterman Rebased.

aswaterman:
Thanks, sorry about that.

PR #3019, opened 2026-05-04
smstateen: clarify CTR state-enable bit behavior (#2672)

### LINKED ISSUE #2672: Smstateen.adoc does not define effect of msstateen0.CTR
Missing sentence.  Might be

The CTR bit in mstateen0 and hstateen0 controls access to the Sctr state described in Section 11.4.

amkr6207:
Fixes #2672.

Add the missing description of the `CTR` bit in `mstateen0` and `hstateen0` in `src/priv/smstateen.adoc`.

The detailed behavior is already defined in `src/priv/smctr.adoc`; this change just fills the missing explanation in `smstateen.adoc`.

### DIFF
--- a/src/priv/smstateen.adoc
+++ b/src/priv/smstateen.adoc
@@ -263,6 +263,10 @@ whether they really do.#
 [[norm:stateen0_jvt_op]]
 The JVT bit controls access to the `jvt` CSR provided by the Zcmt extension.
 
+[#norm:mstateen0_ctr_op]#The CTR bit in `mstateen0` controls access to the CTR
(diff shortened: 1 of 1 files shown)


aswaterman (commented):
We are going down the wrong path by describing CSR bits in multiple places.  If we think this is a better place for the description, we should delete the corresponding text in Smctr as part of the same PR.

amkr6207:
> We are going down the wrong path by describing CSR bits in multiple places. If we think this is a better place for the description, we should delete the corresponding text in Smctr as part of the same PR.

Thanks, you are correct. I updated the PR to leave only a cross-reference to Smctr.

aswaterman (approved):

Issue #2945, opened 2026-04-22
Delegate Source and Destination Register Listings in RVWMO to each instruction description


omasanori:
Adding every instruction to the RVWMO section by hand seems a waste of resources. The description of each instruction in UnifiedDB should have the relevant information.

When the integration of UnifiedDB is completed with dependency information, we should replace the hand-written listings from the RVWMO section by automatically generated ones, or remove the listings from the section.

PR #3020, opened 2026-05-04
rvwmo: split source/dest register listings into generated include (#2945)

amkr6207:
This PR is an initial step toward #2945.

It moves the large "Source and Destination Register Listings" block out of `src/unpriv/rvwmo.adoc` into a generated include file, and adds a small generator/check workflow for that content.

This should not change the document semantics.

A follow-up can switch the generator input to **UnifiedDB-backed metadata** once the preferred source and pinning approach are confirmed.


omasanori:
I don't think we should act on <https://github.com/riscv/riscv-isa-manual/issues/2945> right now, as some others have already been working on the UnifiedDB integration and we will see how it can help the dependency listing then.

PR #3021, opened 2026-05-04
Fix misformatted em dashes

omasanori:
I have checked the whole specification for em dash and fixed where the conversion did not work correctly. Some dashes are eliminated by tweaking the wording.

Reference: https://github.com/asciidoctor/asciidoctor/issues/1578

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ MONTHYEAR := $(shell date -j -f "%Y%m%d" "$(DATE)" +"%B %Y" 2>/dev/null || date 
 CITATION_DESCRIPTION := $(DATE)-$(RELEASE_TYPE)
 ifeq ($(RELEASE_TYPE), draft)
   WATERMARK_OPT := -a draft-watermark
-  RELEASE_DESCRIPTION := DRAFT---NOT AN OFFICIAL RELEASE
+  RELEASE_DESCRIPTION := DRAFT—NOT AN OFFICIAL RELEASE
 else ifeq ($(RELEASE_TYPE), intermediate)
   WATERMARK_OPT :=
   RELEASE_DESCRIPTION := Intermediate Release
--- a/src/priv/machine.adoc
+++ b/src/priv/machine.adoc
@@ -1205,15 +1205,9 @@ hand, we wish to allow flexibility for larger systems.
 [%autowidth,float="center",align="center",cols=">,^,<",options="header"]
 |===
 |Value |Name |Description
-|0 +
-1 +
-{ge}2
-|Direct +
-Vectored +
----
-|All traps set `pc` to BASE. +
-Asynchronous interrupts set `pc` to BASE+4{times}cause. +
-_Reserved_
+|0 |Direct |All traps set `pc` to BASE.
+|1 |Vectored |Asynchronous interrupts set `pc` to BASE+4{times}cause.
+|{ge}2 |-- |_Reserved_
 |===
 
 The encoding of the MODE field is shown in <<norm:mtvec_mode_enc>>.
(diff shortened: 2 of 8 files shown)


src/priv/machine.adoc
```diff
@@ -1210,7 +1210,7 @@ hand, we wish to allow flexibility for larger systems.
 {ge}2
 |Direct +
 Vectored +
----
+_n/a_
```
aswaterman: n/a is a little awkward; I think we should use `{emdash}` here.

src/unpriv/zihintntl.adoc
```diff
@@ -123,12 +123,12 @@ explicit cache management
...
-|Private L1; shared L2 |L1  |L1  |L2  |L2 |P1|ALL|---|---
-|Private L1; shared L2/L3 |L1 | L1 | L2 | L3 |P1  |S1   |ALL |---
-|Private L1/L2 |L1  |L2  |L2  |L2 | P1  |ALL  |--- |---
-|Private L1/L2; shared L3 |L1 | L2 | L3 | L3 | P1 | PALL| ALL |---
+| No caches 4+|_n/a_ 4+|none
+|Private L1 only |L1 |L1 |L1 |L1| ALL |_n/a_ |_n/a_ |_n/a_
+|Private L1; shared L2 |L1  |L1  |L2  |L2 |P1|ALL|_n/a_|_n/a_
+|Private L1; shared L2/L3 |L1 | L1 | L2 | L3 |P1  |S1   |ALL |_n/a_
+|Private L1/L2 |L1  |L2  |L2  |L2 | P1  |ALL  |_n/a_ |_n/a_
+|Private L1/L2; shared L3 |L1 | L2 | L3 | L3 | P1 | PALL| ALL |_n/a_
```
aswaterman: same

src/rv32.adoc
```diff
@@ -885,8 +885,8 @@ exploited by a hardware implementation, there is no advantage to making the
 instruction an option.
 ====
 
-[#norm:fence_unused_flds_rsv]#The unused fields in the FENCE
-instructions--_rs1_ and _rd_--are reserved
```
aswaterman, line 889: I prefer the old wording.

src/vector-crypto.adoc
```diff
@@ -1021,8 +1019,8 @@ instruction encoding such as the use of the zero register (`x0`), and, in the
 case of immediate forms of an instruction, the values in the immediate
 fields (i.e., imm, and uimm).
 
-In some cases --- which are explicitly specified in the lists below
---- operands that are used as control rather than data
+In some cases -- which are explicitly specified in the lists below
+-- operands that are used as control rather than data
```
aswaterman: delete spaces around --

Makefile
```diff
@@ -28,7 +28,7 @@ MONTHYEAR := $(shell date -j -f "%Y%m%d" "$(DATE)" +"%B %Y" 2>/dev/null || date
 CITATION_DESCRIPTION := $(DATE)-$(RELEASE_TYPE)
 ifeq ($(RELEASE_TYPE), draft)
   WATERMARK_OPT := -a draft-watermark
-  RELEASE_DESCRIPTION := DRAFT---NOT AN OFFICIAL RELEASE
+  RELEASE_DESCRIPTION := DRAFT—NOT AN OFFICIAL RELEASE
```
aswaterman, line 31: It's arguably better to avoid unicode in this file.  Does `--` work?

src/rv32.adoc
omasanori: How about using comma instead of dash?

Makefile
omasanori: No, it does not.

src/rv32.adoc
aswaterman: sure.

Makefile
aswaterman: OK, let's leave it like this, then.

omasanori:
Updated.

aswaterman (approved):

PR #3038, opened 2026-05-06
F, D, Q, Zf, Zfbfmin, Zfh: Rearrange canonical NaN description

omasanori:
cc @jordancarlin @mmhus

### DIFF
--- a/src/unpriv/f-st-ext.adoc
+++ b/src/unpriv/f-st-ext.adoc
@@ -6,8 +6,17 @@ floating-point, which adds [#norm:f_ieee_compliance]#computational instructions
 compliant with the IEEE 754-2008 arithmetic standard's binary32 format and
 operations.# [#norm:f_depends_zicsr]#The ext:f[] extension depends on the
 ext:zicsr[] extension.#
-[#norm:flen]#The ext:f[] extension adds 32 floating-point registers, `f0`-`f31`,
-each 32 bits wide (FLEN=32 in <<fprs>>)#.
+
+[[norm:flen]]
+The ext:f[] extension adds 32 floating-point registers, `f0`-`f31`, each 32 bits
+wide (FLEN=32).
+
+[[norm:F_canonical_NaN]]
+(((NaN, canonical, single-precision)))
+For single-precision floating-point, the canonical NaN has a positive sign and
+all significand bits clear except the MSB, the quiet bit. In other words, for
+single-precision floating-point, the canonical NaN corresponds to the pattern
+`0x7fc00000`.
 
 [[sec:fcsr]]
 ==== Floating-Point Control and Status Register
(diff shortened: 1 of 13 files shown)


src/unpriv/zf.adoc
```diff
@@ -119,6 +119,64 @@ instead.
...
+:sectnums!:
+[discrete]
+=== NaN Generation and Propagation
+:sectnums:
+
+[[norm:canonical_NaN]]
+(((NaN, canonical))) (((NaN, generation))) (((NaN, propagation)))
+Except when otherwise stated, if the result of a floating-point operation is
+NaN, it is _the canonical NaN_. The canonical NaN has a positive sign and all
+significand bits clear except the MSB (the quiet bit).
```
aswaterman: This statement about the canonical NaN encoding is not true for all FP extensions (e.g. Zvfofp8min contradicts this).  I do understand what you're trying to do here, but it might be cleaner to leave this in the F extension.

omasanori: OK

aswaterman: Same comment about the submnormals.  We have forthcoming matrix extensions that will disagree.  I think I would leave all of this alone (aside from the small localized improvements you made, which look good).

omasanori: I see. I will bring back the Subnormal Arithmetic section. OTOH, how about this new paragraph on the canonical NaN?

> Except when otherwise stated, if the result of a floating-point operation is NaN, it is _the canonical NaN_. Extensions introducing floating-point formats shall define the canonical NaN for each format.

We would need to add paragraphs to D, Q, Zfh and Zfbfmin but it is consistent with upcoming Zvfofp8min.

aswaterman: Let's see how it looks after making those modifications.  If we think it is an improvement after seeing the result, I'd be OK with it.

omasanori: OK, I will ping you after adding those paragraphs.

omasanori:
@aswaterman I added canonical NaN NormRules to floating-point extensions. They feel right to me. Your opinions?

src/unpriv/f-st-ext.adoc
```diff
@@ -6,8 +6,15 @@ floating-point, which adds [#norm:f_ieee_compliance]#computational instructions
...
-each 32 bits wide (FLEN=32 in <<fprs>>)#.
+
+[[norm:flen]]
+The ext:f[] extension adds 32 floating-point registers, `f0`-`f31`, each 32 bits
+wide (FLEN=32).
+
+[[norm:F_canonical_NaN]]
+(((NaN, canonical, single-precision)))
+For single-precision floating-point, the canonical NaN corresponds to the
+pattern `0x7fc00000`.
```
aswaterman: For the F extension only, let's revert to something like the old text that includes a fuller explanation:

```
For single-precision floating-point, the canonical NaN has a
positive sign and all significand bits clear except the MSB, a.k.a. the
quiet bit. For single-precision floating-point, this corresponds to the pattern `0x7fc00000`.
```

Your reduced text for D, Zfh, etc. is fine.  There's more explanatory stuff in the F chapter than the other chapters.

omasanori:
@aswaterman Rebased and updated to address the comment.

aswaterman (approved):

aswaterman:
Thanks!

PR #3041, opened 2026-05-06
Remove list of explicit memory accesses from pointer masking section

aswaterman:
The list is a maintenance burden.  Relying on the definition of "explicit memory access" instead of listing them all out avoids this pitfall.

See discussion on #3015.

### DIFF
--- a/.github/workflows/dco_check.yml
+++ b/.github/workflows/dco_check.yml
@@ -1,3 +1,4 @@
+---
 name: DCO Check
 
 on:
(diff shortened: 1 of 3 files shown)


aswaterman:
cc @davidharrishmc

aswaterman:
@wmat Looks like #3040 broke the build (didn't pass the pre-commit hooks for YAML formatting).  I fixed it here.

wmat:
Thanks for that, I didn't think precommit failures stopped the build. Noted for future.

PR #3042, opened 2026-05-07
Move remaining unpriv chapters

aswaterman:
Also fix up a few hundred extension-name links.


src/unpriv/base.adoc
```diff
@@ -0,0 +1,10 @@
+[[base]]
+== Base ISAs
+
+This chapter describes the RISC-V base ISAs.
+
+include::rv32.adoc[]
+include::rv32e.adoc[]
+include::rv64.adoc[]
```
omasanori, line 7: It is the same as the original, but RV64I after RV32E/RV64E seems odd to me. Will it stay as is after reorganization?

aswaterman: The long-term plan involves defining a broader schema for base ISAs, which will result in another reorganization.  But in the meantime, I agree with your criticism and will fix it.

PR #3052, opened 2026-05-08
Zv, vector-common: Use macros

### LINKED ISSUE #3035: On the vtype for small implementations note
In the vtype section, a note says:

> A small implementation supporting ELEN=32 requires only seven bits of state in vtype: two bits for ma and ta, two bits for vsew[1:0] and three bits for vlmul[2:0]. The illegal value represented by vill can be internally encoded using the illegal 64-bit combination in vsew[1:0] without requiring an additional storage bit to hold vill.

Are ma and ta here vma and vta?

### LINKED ISSUE #3036: Should vtype fields be marked up like separate CSRs?
The fields of the vtype CSR are in lower case, unlike other CSR fields. Should we markup them like `csr:vill[]` instead of `csr::[vill]`?

omasanori:
The vector chapter is so large that I decided to split the macro conversion PR into few parts.

Fix https://github.com/riscv/riscv-isa-manual/issues/3035
Fix https://github.com/riscv/riscv-isa-manual/issues/3036

### DIFF
--- a/src/unpriv/vector-common.adoc
+++ b/src/unpriv/vector-common.adoc
@@ -6,10 +6,10 @@ features, are defined in the subsequent sections.
 
 [NOTE]
 ====
-_The base vector extension is intended to provide general support for
+The base vector extension is intended to provide general support for
 data-parallel execution within the 32-bit instruction encoding space,
 with later vector extensions supporting richer functionality for certain
-domains._
+domains.
 ====
 
 ==== Implementation-defined Constant Parameters
@@ -21,9 +21,9 @@ must be a power of 2.#
 . [#norm:vlen]#The number of bits in a single vector register, _VLEN_ {ge} 8, which must be a power of 2, and must be no greater than 2^16^.#
 
 Standard vector extensions (<<sec-vector-extensions>>) and
-architecture profiles may set further constraints on _ELEN_ and _VLEN_.
+architecture profiles may set further constraints on ELEN and VLEN.
 
-NOTE: Following the ratification of the V extension, this specification
+NOTE: Following the ratification of the ext:v[] extension, this specification
 has been revised to admit the possibility of future extensions that
 allow ELEN > VLEN, wherein one element is held using bits from
 multiple vector registers.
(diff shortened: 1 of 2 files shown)


src/unpriv/vector-common.adoc
```diff
@@ -633,22 +632,22 @@ roundoff_signed(v, d) = (signed(v) >> d) + r
...
+===== Vector Fixed-Point Saturation Flag (csr:vxsat[])
 
 [[norm:vxsat_op_acc_sz]]
-The `vxsat` CSR has a single read-write least-significant bit
-(`vxsat[0]`) that indicates if a fixed-point instruction has had to
+The csr:vxsat[] CSR has a single read-write least-significant bit
+(csr:vxsat[][0]) that indicates if a fixed-point instruction has had to
 saturate an output value to fit into a destination format.
-Bits `vxsat[XLEN-1:1]` should be written as zeros.
+The remaining bits, csr:vxsat[][XLEN-1:1], should be written as zeros.
```
aswaterman, line 641: It's very difficult to do a good job reviewing these PRs when they mix text and formatting changes.  Let's try to keep the formatting changes completely independent of the text changes.

src/unpriv/vector-common.adoc
```diff
@@ -2384,8 +2383,8 @@ operands are always next to each other.
 
 [#norm:vwop_vd_eew_emul]#A few vector arithmetic instructions are defined to be __widening__
 operations where the destination vector register group has EEW=2*SEW
-and EMUL=2*LMUL.#  These are generally given a `vw*` prefix on the
-opcode, or `vfw*` for vector floating-point instructions.
+and EMUL=2*LMUL.#  These are generally given a insn:vw*[] prefix on the
+opcode, or insn:vfw*[] for vector floating-point instructions.
```
aswaterman, line 2387: Following this pattern, do you think it makes sense to refer to `insn:vsetvli[], insn:vsetivli[], and insn:vsetvl[]` as `insn:vset*[]`?  It is a pattern that shows up pretty often.

aswaterman (approved):
Thank you @omasanori, this was a lot of work!

src/unpriv/vector-common.adoc
omasanori: Yes, vset* may be nice. A concern to me is vset* is mentioned before vset* is specified in the specification so readers don't know what it is at that point.

src/unpriv/vector-common.adoc
omasanori: I see. Sorry for that.

src/unpriv/vector-common.adoc
omasanori: I sent a PR: https://github.com/riscv/riscv-isa-manual/pull/3056

PR #3054, opened 2026-05-08
Move NaN-boxing to Zf

omasanori:


aswaterman (commented):
I don't think this is a good idea.  It is better for narrative flow if the specs build upon each other.  NaN boxing is a complex topic for many readers; there is value in not revealing this complexity until it becomes necessary.

I do understand that there is a scoping argument for doing it your way: if both D and Zfh depend on it, then maybe it should be part of the common material.  It is a qualitative decision as to whether that argument wins the day.  In my judgment, we should lave this one as-is.

omasanori:
OK

PR #3064, opened 2026-05-10
Name extension sections bit more consistently

omasanori:


omasanori:
Another approach would be to name extensions *Zfoo (...) (Vector) Extension(s)* consistently, with exceptions for *Zfoo (...) Extension for Application Processors,* *Zfoo (...) Extension for Cryptography* and such.

aswaterman (approved):

aswaterman:
I think "Zfoo Extension for [...]" reads better than "Zfoo [...] Extension" in most cases (but it's arguable in some cases, too).

PR #3071, opened 2026-05-11
Use curved double quotes

### LINKED ISSUE #2992: Use curved quotation marks where appropriate
There are "straight quotation marks" and “curved quotation marks”. The former is the proper one in programming languages, but it is not suitable for publication typeset.

According to <https://docs.asciidoctor.org/asciidoc/latest/text/quotation-marks-and-apostrophes/>, backticks inside straight quotation marks can be used for curved quotation marks instead of using raw curved quotation marks `“...”` or character references `&#8220;...&#8221;`. How about using it where appropriate to improve typography?

omasanori:
I also fixed some markup errors found during conversion.

Fix https://github.com/riscv/riscv-isa-manual/issues/2992

### DIFF
--- a/src/priv/machine.adoc
+++ b/src/priv/machine.adoc
@@ -64,14 +64,14 @@ If zero, the hart is RV32, else it is RV64.
 
 [#norm:misa_extensions_enc_txt]#The Extensions field encodes the presence of the standard extensions,
 with a single bit per letter of the alphabet (bit 0 encodes presence of
-extension "A" , bit 1 encodes presence of extension "B", through to
-bit 25 which encodes "Z").#
-[#norm:misa_i_op]#The "I" bit will be set for the RV32I and RV64I base ISAs#,
-and [#norm:misa_e_op]#the "E" bit will be set for RV32E and RV64E.#
+extension ext:a[] , bit 1 encodes presence of extension ext:b[], through to
+bit 25 which encodes ext:z[]).#
+[#norm:misa_i_op]#The csr::[i] bit will be set for the RV32I and RV64I base ISAs#,
+and [#norm:misa_e_op]#the csr::[e] bit will be set for RV32E and RV64E.#
 [#norm:misa_extensions_warl_op]#The Extensions field is a *WARL* field that can contain writable bits where the
 implementation allows the supported ISA to be modified.#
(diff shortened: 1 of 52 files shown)


src/priv/machine.adoc
```diff
@@ -64,14 +64,14 @@ If zero, the hart is RV32, else it is RV64.
 
 [#norm:misa_extensions_enc_txt]#The Extensions field encodes the presence of the standard extensions,
 with a single bit per letter of the alphabet (bit 0 encodes presence of
-extension "A" , bit 1 encodes presence of extension "B", through to
-bit 25 which encodes "Z").#
-[#norm:misa_i_op]#The "I" bit will be set for the RV32I and RV64I base ISAs#,
-and [#norm:misa_e_op]#the "E" bit will be set for RV32E and RV64E.#
+extension ext:a[] , bit 1 encodes presence of extension ext:b[], through to
+bit 25 which encodes ext:z[]).#
+[#norm:misa_i_op]#The "`I`" bit will be set for the RV32I and RV64I base ISAs#,
```
aswaterman: I realize your goal was to preserve existing formatting for the most part, but these are CSR fields; don't you think they should be e.g. `csr::[i]` instead of "`I`"?

src/priv/smstateen.adoc
```diff
@@ -29,14 +29,14 @@ initialized to zero, either at machine level or by the OS itself initializing
 `sstatus`.
 
 Obviously, one way to prevent the use of new user-level CSRs as covert channels
-would be to add to `mstatus` or `sstatus` an "XS" field for each relevant
+would be to add to `mstatus` or `sstatus` an "`XS`" field for each relevant
```
aswaterman: Similarly, probably `csr::[xs]`

aswaterman (approved):
Looks good for the most part; ping me when you've addressed my comments.

src/priv/machine.adoc
omasanori: Yes.

src/priv/smstateen.adoc
omasanori: OK

omasanori:
@aswaterman Updated.

PR #3073, opened 2026-05-12
preface: Use macros

omasanori:


src/priv/preface.adoc
```diff
@@ -366,27 +366,27 @@ Changes from version 1.10 include:
 * Specified which interrupt sources are reserved for standard use.
 * Allocated some synchronous exception causes for custom use.
 * Specified the priority ordering of synchronous exceptions.
-* Added specification that xRET instructions may, but are not required
-to, clear LR reservations if A extension present.
+* Added specification that insn:xret[] instructions may, but are not required
```
aswaterman: I think `insn:xret[]` is not quite right here (and sorry if I failed to flag that elsewhere).  Capitalizing it as XRET makes it look like a proper instruction name.

Maybe until we come up with a better solution we should just use the literal string `xRET`?

omasanori: OK. How about insn:*ret[]?

aswaterman (approved):
LGTM modulo my one comment.  (Normally I am reluctant to change the preface, but these are just formatting changes.)

src/priv/preface.adoc
aswaterman: Maybe, but let's try it out first.  Can you try to reformat the following paragraph according to your scheme and show me what the rendered HTML looks like?  

https://github.com/riscv/riscv-isa-manual/blob/ef855dd44beae54f0cfbac239299c2f4308fbc01/src/priv/machine.adoc#L411-L416

omasanori:
```
An MRET or SRET instruction is used to return from a trap in M-mode or S-mode respectively.
[#norm:mstatus_xret_op]#When executing an insn:*ret[] instruction, supposing
csr::[*pp] holds the value _y_, csr::[*ie] is set to csr::[*pie]; the privilege mode is
changed to _y_; csr::[*pie] is set to 1; and csr::[*pp] is set to the
least-privileged supported mode (U if U-mode is implemented, else M). If
__y__{ne}M, insn:*ret[] also sets csr::[mprv]=0.#
```

Screenshot

It seems not good to me; *IE and *PIE is confusing.

aswaterman: Right.  `*` looks like "all of this category" whereas `__x__` looks like "the corresponding element in this category".  So, I think we need a way of saying `__x__RET`.  I propose we continue formatting it the old way until we find a better solution.  This pattern only shows up a few dozen times in the spec, so we can easily change it later if need be.

omasanori: I agree. I will ping you after reverting related ones; I also converted xtval, XIE and such in preface.

omasanori:
@aswaterman Updated. (I used `__x__` consistently so that we can grep them easily.)

aswaterman (approved):
Cool, thanks!

Issue #2985, opened 2026-04-26
Make the register state figures bit better


omasanori:
For instance, in the RV32I chapter, the text says:

> For RV32I, the 32 `x` registers are each 32 bits wide, i.e., XLEN=32. (...) There is one additional unprivileged register: the program counter `pc` holds the address of the current instruction.

followed by 33 boxes for each register. They increase the amount of scroll on smartphone without giving any idea beyond the quoted sentences. We have a similar figure in the F extension too.

We could do something better there. For example, Arm® Architecture Reference Manual
for A-profile architecture, B1.2 Registers in AArch64 Execution state illustrates the relationship between registers (64-bit GPR Xn, 32-bit GPR Wn, 128-bit SIMD&FP Register Qn, 64-bit SIMD&FP Register Dn, etc.).

If we have nothing to show there other than the fact that we count each bit in GPRs from XLEN-1 to 0, at least we can show just one box instead of 33 boxes.

aswaterman:
Not sure optimizing for smartphone viewing is really the goal of a manual that will ultimately be a few thousand pages long?

What about halving the length of the figure by making the table two columns?

omasanori:
Sure, optimizing the presentation on smartphone is not the goal. My point was the figures seem not useful whereas its large area. Making the figures multi-column in fact reduces the amount of scroll but it does not improve the importance of these figures much.

PR #3074, opened 2026-05-12
Replace fprs table with bytefield diagram

omasanori:
This is my proposal for <https://github.com/riscv/riscv-isa-manual/issues/2985>. I will draw another one for RV32I if this PR is accepted.

HTML version:

Screenshot

### DIFF
--- a/src/unpriv/images/bytefield/fprs.edn
+++ b/src/unpriv/images/bytefield/fprs.edn
@@ -1,0 +1,26 @@
+[bytefield]
+----
+(defattrs :plain [:plain {:font-family "M+ 1p Fallback" :font-size 16}])
+(def row-height 30)
+(def row-header-fn nil)
+(def left-margin 100)
+(def right-margin 100)
+(def boxes-per-row 16)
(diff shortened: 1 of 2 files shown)


aswaterman (approved):
Yep, nicely done.

aswaterman:
The figure suggests that `fcsr` is the same width as the other registers.  I don't have a great idea as to how to do it better, especially because the representation is actually correct for the F extension.  So, let's just run with it.

omasanori:
> The figure suggests that `fcsr` is the same width as the other registers.  I don't have a great idea as to how to do it better, especially because the representation is actually correct for the F extension.  So, let's just run with it.

Maybe we can omit the diagram for fcsr here, as the diagram for fcsr is in the dedicated section.

aswaterman:
I think it’s better to keep it since the purpose of the figure is to sketch all of the new architectural state associated with the extension. The graphical width is a minor nit by comparison.

omasanori:
I see.

PR #3079, opened 2026-05-12
Clarifying choices regarding underflow/inexact flags in RVBNA

nibrunieAtSi5:
Addressing some comments on RVBNA spec from https://lists.riscv.org/g/tech-vme/topic/119043904 in particular 

> §15.1.1.8 also drops inexact, underflow, and divide-by-zero from fflags for SEW≤16 inputs. There is no normative justification, and no acknowledgement that this is a deviation from the V-extension exception model.

* Adding a non normative note to explain some of the choices made regarding Floating-Point exception flag underflow and inexact in RVBNA (RISC-V Bulk Normalization Algorithm)
* The comment mentioned division-by-zero but I don't think anyone would actually expect a dot product (or any MAC-like operation) to raise this flag

An LLM was used to rewrite a first hand made version of this suggestion.

### DIFF
--- a/src/unpriv/rvbna.adoc
+++ b/src/unpriv/rvbna.adoc
@@ -306,6 +306,13 @@ If the result of a dot product operation is zero, then `+0` is returned, even if
 
 Under RVBNA, only the invalid operation and overflow exceptions can be raised.
 
+
+Note:: The underflow and inexact flags are not raised, for two different reasons.
++
+Detecting underflow accurately would add significant implementation complexity for little practical benefit. Furthermore, in some RVBNA configurations (e.g. product of OFP8 numbers into an FP32 accumulator), underflow cannot actually occur, because the product range is narrower than the range in which underflow could be triggered during accumulation.
++
+Inexact is suppressed because, given the compound nature of RVBNA, the flag would carry no meaningful numerical information: nearly every result would either always raise it or never raise it at all.
+
 ===== Invalid operation
 
 The invalid operation flag must be raised if at least one of the following conditions is met:


src/unpriv/rvbna.adoc
```diff
@@ -306,6 +306,15 @@ If the result of a dot product operation is zero, then `+0` is returned, even if
...
 Under RVBNA, only the invalid operation and overflow exceptions can be raised.
 
+
+Note:: The underflow and inexact flags are not raised, for two different reasons.
++
+Detecting underflow accurately would add significant implementation complexity for little practical benefit. In fact, in a lot of RVBNA configurations, underflow cannot actually occur, because the product range is narrower than the range in which underflow could be triggered during accumulation.
++
+Inexact is suppressed because, given the approximate nature of RVBNA, the flag would carry no meaningful numerical information -- nearly every result would either raise it or never raise it at all.
+// +
+// The divide-by-zero flag is never raised either, as RVBNA defines only MAC-style operations and performs no division-like operation.
```
nibrunieAtSi5: This should be removed, not commented out

aswaterman:
@wmat it seems like draft PRs fail to trigger the automated message that provides guidance with failing DCO checks.  Is that fixable?

src/unpriv/rvbna.adoc
```diff
@@ -306,6 +306,13 @@ If the result of a dot product operation is zero, then `+0` is returned, even if
 
 Under RVBNA, only the invalid operation and overflow exceptions can be raised.
 
+
+Note:: The underflow and inexact flags are not raised, for two different reasons.
++
+Detecting underflow accurately would add significant implementation complexity for little practical benefit. In fact, in a lot of RVBNA configurations (e.g. product of OFP8 numbers into an FP32 accumulator), underflow cannot actually occur, because the product range is narrower than the range in which underflow could be triggered during accumulation.
```
aswaterman: "In fact, in a lot of" => "Furthermore, in some"

src/unpriv/rvbna.adoc
```diff
@@ -306,6 +306,13 @@ If the result of a dot product operation is zero, then `+0` is returned, even if
...
++
+Inexact is suppressed because, given the compound nature of RVBNA, the flag would carry no meaningful numerical information -- nearly every result would either always raise it or never raise it at all.
```
aswaterman: no spaces around the emdash, but maybe a colon is better here?

aswaterman (approved):

wmat:
@aswaterman yes, I can update the dco and signing checks to trigger on  pull request events:
 - opened
 - reopened
 - synchronize
 - converted_to_draft

Would that work?

src/unpriv/rvbna.adoc
nibrunieAtSi5: changed for a `:`

src/unpriv/rvbna.adoc
nibrunieAtSi5: Changed

aswaterman:
@wmat sounds right to me, let's see how it works.

aswaterman:
Thanks @nibrunieAtSi5

PR #3087, opened 2026-05-14
base: Minor wording tweaks

omasanori:


src/unpriv/rv32.adoc
```diff
@@ -39,8 +39,8 @@ Most of the commentary for RV32I also applies to the RV64I base.
 ==== Programmers' Model for Base Integer ISA
 
 <<gprs>> shows the unprivileged state for the base integer ISA.
-[#norm:rv32i_xreg_sz]#For RV32I, the 32 `x` registers are each 32 bits wide,
-i.e., XLEN=32.# [#norm:x0eq0]#Register `x0` is hardwired with all bits equal to 0.#
+[#norm:rv32i_xreg_sz]#For RV32I, the 32 `x` registers are each 32 bits wide (XLEN=32).#
```
aswaterman: I think the `i.e.` is better (and maybe we should change the RV64 chapter to match).

omasanori:
@aswaterman Updated. I chose *in other words* instead of i.e. since the Docs Dev Guide discourages Latin phrases.

aswaterman (approved):
Ack.  I'm going to unilaterally overrule that one under the expectation we'll later change that rule.

PR #3090, opened 2026-05-14
Remove norm:imm_always_sex

### LINKED ISSUE #3089: [NormRules] - norm:imm_always_sex no longer strictly correct
## Description

norm:imm_always_sex states that *except for the 5-bit immediates used in CSR instructions, immediates are always sign-extended*. Strictly speaking, at this moment, there are instructions with zero-extended immediate other than in Zicsr (ex. in the C and V extensions).

## Any changes required

Change norm:imm_always_sex to *Unless otherwise noted, immediates are always sign-extended* or such.

(cc @riscv/csc)

omasanori:
Fix https://github.com/riscv/riscv-isa-manual/issues/3089

cc @jordancarlin @mmhus


jordancarlin (commented):
Should this become a new normative rule indicating all base ISA instruction immediates are sign-extended? I don't recall if the instruction specific normative rules already include that or not. If they do, then this looks fine as is.

aswaterman:
I agree that would fix it, but I guess my comment from #3089 still holds--it seems like a rule with little incremental value.  I'll merge this, but this can always be revisited if there's disagreement.

aswaterman (approved):

omasanori:
> Should this become a new normative rule indicating all base ISA instruction immediates are sign-extended? I don't recall if the instruction specific normative rules already include that or not. If they do, then this looks fine as is.

I have checked the Integer Register-Immediate Instructions section. Every instruction other than SLLI, SRLI, SRAI, LUI and AUIPC has the phrase *sign-extended* in the NormRule. The shifts, LUI and AUIPC do not sign-extension, so the current rules cover well.

PR #3096, opened 2026-05-17
F: Bring the canonical NaN definition back

omasanori:
After reading the scalar FP chapter several times, the canonical NaN definition seems not suitable for the introduction part, though having the canonical NaN normative rules in each extension feels correct to me. This brings back the canonical NaN part to the F extension while keeping the current text.


aswaterman (approved):
OK.  I was hesitant about this one beforehand, so if you also think it should be reversed, then I agree.

PR #3097, opened 2026-05-17
Markup fix

omasanori:


src/unpriv/cfi.adoc
```diff
@@ -16,7 +16,7 @@ RV32/RV64 provides two types of control transfer instructions - unconditional
 jumps and conditional branches. Conditional branches encode an offset in the
 immediate field of the instruction and are thus direct branches that are not
 susceptible to control-flow subversion. Unconditional direct jumps using insn:jal[]
-transfer control to a target that is in a +/- 1 MiB range from the current `pc`.
+transfer control to a target that is in a {pm}1{nbsp}MiB range from the current `pc`.
```
aswaterman, line 19: I think we need to introduce an idiom for the number-with-units pattern.  In the old LaTeX specs, we used `\wunits{1}{MiB}` to introduce the non-breaking space.

We could do something analogous in asciidoc: `wunits:1[MiB]`, I guess.  (The obvious alternative, `wunits::[1,MiB]`, is too cumbersome to type.)

src/unpriv/f-st-ext.adoc
```diff
@@ -54,7 +61,7 @@ the instruction, or a dynamic rounding mode held in `frm`. Rounding
 modes are encoded as shown in <<norm:dyn_round_enc>>. A value of 111 in the
 instruction's _rm_ field selects the dynamic rounding mode held in
 `frm`#. [#norm:roundingmode_rsv]#The behavior of floating-point instructions that depend on
-rounding mode when executed with a reserved rounding mode is _reserved_, including both static reserved rounding modes (101-110) and dynamic reserved rounding modes (101-111). Some instructions, including widening conversions, have the _rm_ field but are nevertheless mathematically unaffected by the rounding mode; software should set their _rm_ field to
+rounding mode when executed with a reserved rounding mode is _reserved_, including both static reserved rounding modes (101 and 110) and dynamic reserved rounding modes (101, 110, and 111). Some instructions, including widening conversions, have the _rm_ field but are nevertheless mathematically unaffected by the rounding mode; software should set their _rm_ field to
```
aswaterman: I think the old version is better in this case.

src/unpriv/cfi.adoc
aswaterman: (To be clear, this should be the topic of a future improvement.)

omasanori:
@aswaterman Updated.

aswaterman (approved):

aswaterman:
@omasanori I'll try to prototype the `wunits` thing.

omasanori:
Thanks!

PR #3098, opened 2026-05-17
Define units:[] macro; use in a couple places as an example

aswaterman:
cc @omasanori

### DIFF
--- a/src/lib/macros.rb
+++ b/src/lib/macros.rb
@@ -19,6 +19,8 @@ require 'asciidoctor/extensions'
 #
 # csrlink:foo[]           <<csr:foo, `foo`>>
 # csrlink:foo[bar]        <<csr:foo_bar, `foo.BAR`>>
+#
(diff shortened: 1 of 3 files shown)


omasanori:
LGTM. The `siunitx` and `physics` LaTeX packages have the `qty` (quantity) macro for this purpose, and quantity seems more suitable than unit (which means the KiB part of 1 KiB for instance), though.

aswaterman:
`qty` makes sense to me, too.

src/lib/macros.rb
```diff
@@ -107,6 +109,18 @@ def self.process_xref(processor, parent, name, ref)
       RVFormat.process_xref(self, parent, RVFormat.csr(name, field), ref)
     end
   end
+
+  inline_macro :qty do
+    process do |parent, value, args|
+      if args.empty?
+        parent.document.logger.fatal "macro qty:#{name}[] takes exactly one argument"
```
omasanori: s/name/value/ ?

PR #3099, opened 2026-05-17
Zb, Zk, Zv: Use macros

omasanori:


src/unpriv/zba.adoc
```diff
@@ -5,7 +5,7 @@ The ext:zba[] instructions can be used to accelerate the generation of addresses
 
 The shift and add instructions do a left shift of 1, 2, or 3 because these are commonly found in real-world code and because they can be implemented with a minimal amount of additional hardware beyond that of the simple adder. This avoids lengthening the critical path in implementations.
 
-[#norm:zba_slli_uw_function]#While the shift and add instructions are limited to a maximum left shift of 3, the slli instruction (from the base ISA) can be used to perform similar shifts for indexing into arrays of wider elements. The slli.uw -- added in this extension -- can be used when the index is to be interpreted as an unsigned word.#
+[#norm:zba_slli_uw_function]#While the shift and add instructions are limited to a maximum left shift of 3, the insn:slli[] instruction from the base ISA can be used to perform similar shifts for indexing into arrays of wider elements. The insn:slli.uw[] added in this extension can be used when the index is to be interpreted as an unsigned word.#
```
omasanori: By the way, is this really worth a normative rule? It sounds like a note for software developers to me.

aswaterman: Agreed.  I think it's OK for this to be part of the normative text for the purposes of narrative flow, but it doesn't make sense to me as a normative rule.

aswaterman: @omasanori ping when this one is ready to merge

omasanori:
@aswaterman ping

aswaterman (approved):

PR #3102, opened 2026-05-18
intro: Use qty

omasanori:


src/unpriv/intro.adoc
```diff
@@ -448,20 +448,20 @@ of 16-bit instruction _parcels_ in length and parcels are naturally
 aligned on 16-bit boundaries.
 The <<ext:zca,standard compressed ISA extension>> reduces code size by
 providing compressed 16-bit instructions and relaxes the alignment
-constraints to allow all instructions (16 bit and 32 bit) to be aligned
+constraints to allow all (16- and 32-bit) instructions to be aligned
```
aswaterman: "instructions" before "()" is better, so I suggest reverting that part, but I agree with your hyphenation.

src/unpriv/intro.adoc
```diff
@@ -448,20 +448,20 @@ of 16-bit instruction _parcels_ in length and parcels are naturally
...
 on any 16-bit boundary to improve code density.
 
 We use the term IALIGN (measured in bits) to refer to the
 instruction-address alignment constraint the implementation enforces.
-IALIGN is 32 bits in the base ISA, but some ISA extensions, including
-the compressed ISA extension, relax IALIGN to 16 bits. IALIGN may not
+IALIGN is 32 in the base ISA, but some ISA extensions, including
+the compressed ISA extension, relax IALIGN to 16. IALIGN may not
```
aswaterman, line 457: Agreed, IALIGN's definition should make it unitless.

omasanori:
@aswaterman I moved *(16- and 32-bit)* after *instructions* to address your comment.

aswaterman (approved):

PR #3104, opened 2026-05-18
Markup fix

omasanori:


omasanori:
@aswaterman By the way, rules for `__x__tval` does also apply to `mtval2`?

aswaterman (approved):
LGTM, thank you!

aswaterman:
> @aswaterman By the way, rules for `__x__tval` does also apply to `mtval2`?

@omasanori which rules specifically?

omasanori:
@aswaterman No particular rules in my mind, I wonder if `__x__tval` refers to not only `mtval` but also `mtval2`.

aswaterman:
Oh, I see.  No, it only means `__x__tval` with no suffix, as you'd expect.

The CSRs are very different, so we might never have any common rules, but if we did, we'd probably explicitly write `__x__tval or __x__tval2`.

omasanori:
I see. Thanks!

Issue #3112, opened 2026-05-19
Are vslideup.vx and vslidedown.vx included in Zvkt or not?


omasanori:
From Zvkt:

> In the vslide[up|down].vx instructions, the value in rs1 is used for control (i.e., slide amount) and therefore is exempt from DIEL.

> vslide[up|down].v[ix]

Which is correct?

omasanori:
Ah, OK. I understood the meaning.

Issue #3111, opened 2026-05-19
On the comment on shift instructions in Zvkt


omasanori:
> // The values in _vs1_, _rs1_, _imm_ are used for control (i.e., shift amount) and are exempt from DIEL.

Is it correct or outdated?

lfiolhais:
The original idea behind the DIEL extensions is that instructions with the same
code points, that are carry DIEL semantics, need to execute with DIEL. Shifts
are also part of the code points, hence DIEL is applicable to them.

From a security POV, the value of a shift should never be tied to any type of
secret data. The shift is encoded in the binary and an attacker doesn't have to
execute the binary to extract hints to the secret data. Definitionally, if the
shift amount for any DIEL instruction is tied to secret data, it is a threat
that is out-of-scope for DIEL.

Issue #3110, opened 2026-05-19
Are vrgather.vv and vrgatherei16.vv included in Zvkt or not?


omasanori:
> In the .vv and .xv forms of the vrgather[ei16] instructions, the values in vs1 and rs1 are used for control and therefore are exempt from DIEL.

> vrgather.v[ivx]
> vrgatherei16.vv

The former excludes vrgather.vv and vrgatherei16.vv but the latter includes them. Which is correct?

Also, .xv form seems not defined in the specification.

omasanori:
Forget it.

Issue #3109, opened 2026-05-19
Zvkt lacks vmv.x.s but vmv.v.s


omasanori:
Zvkt contains vmv.v.s that is not defined in the base vector architecture but it lacks vmv.x.s while it contains vmv.s.x. Should we replace vmv.v.s by vmv.x.s?

aswaterman:
Yes, that appears to be a typo.

PR #3113, opened 2026-05-19
Zvk: Use macros

omasanori:
With this change, we have converted the most of Volume I (except `qty`).

Note that this addresses the none of <https://github.com/riscv/riscv-isa-manual/issues/3107>, <https://github.com/riscv/riscv-isa-manual/issues/3108>, <https://github.com/riscv/riscv-isa-manual/issues/3109>, <https://github.com/riscv/riscv-isa-manual/issues/3110>, <https://github.com/riscv/riscv-isa-manual/issues/3111> and <https://github.com/riscv/riscv-isa-manual/issues/3112>.

### DIFF
--- a/src/unpriv/zvk.adoc
+++ b/src/unpriv/zvk.adoc
@@ -165,7 +165,7 @@ policies:
 
 Many vector crypto instructions operate on operands that are wider than elements (which are currently limited
 to 64 bits wide).
-Typically, these operands are 128- and 256-bits wide. In many cases, these operands are comprised of smaller
+Typically, these operands are 128 and 256 bits wide. In many cases, these operands are comprised of smaller
 operands that are combined (for example, each SHA-2 operand is comprised of 4 words). However, in other cases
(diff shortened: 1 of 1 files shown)


src/unpriv/zvk.adoc
```diff
@@ -165,7 +165,7 @@ policies:
 
 Many vector crypto instructions operate on operands that are wider than elements (which are currently limited
 to 64 bits wide).
-Typically, these operands are 128- and 256-bits wide. In many cases, these operands are comprised of smaller
+Typically, these operands are 128- and 256-bit wide. In many cases, these operands are comprised of smaller
```
aswaterman: I think this should be "128 and 256 bits wide" (no plural but no hyphen)

aswaterman (approved):
Nice, LGTM aside from my one comment.  Thanks!

src/unpriv/zvk.adoc
omasanori: Agreed.

omasanori:
@aswaterman Updated.

Issue #249, opened 2018-11-02
rdcycle and mcycle differ on multi-hart systems


timsifive:
The ISA spec says:
> The RDCYCLE pseudoinstruction reads the low XLEN bits of the cycle CSR which holds a count of the number of clock cycles executed by the *processor core on which the hart is running* from an arbitrary start time in the past.

The Privileged spec says:
> The mcycle CSR counts the number of cycles the hart has executed.
and
> The cycle, instret, and hpmcountern CSRs are read-only shadows of mcycle, minstret, and mhpmcountern, respectively.

The first statement says the counter applies to the entire core, the second says each hart has its own counter, and the third says they are the same thing. One of these must be false.

PR #3120, opened 2026-05-20
accessibility: add SVG metadata and alt text to wavedrom diagrams

### LINKED ISSUE #1129: Figures aren't accessible in HTML version
The HTML version has figures for things like instruction encodings, register layouts, etc., but they appear as just graphics and not in an accessible form (e.g. tables). I don't know if this is a wavedrom issue or not. I can look at the layouts by looking at the wavedrom source, but it's not the most desirable/preferable method.

Maanvi212006:
Fixes #1129

This PR adds accessibility improvements to WaveDrom and bytefield diagrams in the HTML output.

Changes:
- Add `docs-resources/tools/wavedrom_a11y.py` — a script that processes .edn files and injects <title>, <desc>, and aria-label into generated SVGs, and also updates the alt text in the [wavedrom, ,svg] block so screen readers can read it when embedded as base64
- Update b-immediate.edn, atomic-mem.edn, c-breakpoint-instr.edn as examples
b-immediate-accessible

b-immediate_alt-fix
- Add a11y-desc comment to fprs.edn for bytefield diagrams that can't be auto-parsed
byte-field

bytefield_alt

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -93,7 +93,15 @@ WORKDIR_SETUP = \
 WORKDIR_TEARDOWN = \
     mv $@.workdir/$@ $@ && \
     mkdir -p $(BUILD_DIR)/unpriv/images && \
-    find $@.workdir -name '*.svg' -exec cp {} $(BUILD_DIR)/unpriv/images/ \; && \
+    if [ "$(suffix $@)" = ".html" ]; then \
+        mkdir -p $(BUILD_DIR)/priv/images && \
+        grep -o 'src="priv/images/[^"]*\.svg"' $@ | sed 's/src="priv\/images\///;s/"//' | sort -u | \
+            while read f; do [ -f "$@.workdir/build/images-out/$$f" ] && cp "$@.workdir/build/images-out/$$f" $(BUILD_DIR)/priv/images/; done; \
+        grep -o 'src="unpriv/images/[^"]*\.svg"' $@ | sed 's/src="unpriv\/images\///;s/"//' | sort -u | \
+            while read f; do [ -f "$@.workdir/build/images-out/$$f" ] && cp "$@.workdir/build/images-out/$$f" $(BUILD_DIR)/unpriv/images/; done; \
+    else \
+        find $@.workdir -name '*.svg' -exec cp {} $(BUILD_DIR)/unpriv/images/ \; ; \
+    fi && \
     rm -rf $@.workdir
 endif
 
@@ -201,6 +209,7 @@ $(BUILD_DIR)/%.html: $(SRC_DIR)/%.adoc $(ALL_SRCS)
 	$(WORKDIR_SETUP)
 	$(DOCKER_CMD) $(DOCKER_QUOTE) $(ASCIIDOCTOR_HTML) $(OPTIONS) $(REQUIRES) $< $(DOCKER_QUOTE)
 	$(WORKDIR_TEARDOWN)
+	python3 post_process_html.py $@
 	@printf '\n  Built \033]8;;file://%s\033\\%s\033]8;;\033\\\n\n' "$(abspath $@)" "$@"
 
 $(BUILD_DIR)/%.epub: $(SRC_DIR)/%.adoc $(ALL_SRCS)
--- a/Makefile
+++ b/Makefile
@@ -92,6 +92,8 @@ WORKDIR_SETUP = \
 
 WORKDIR_TEARDOWN = \
     mv $@.workdir/$@ $@ && \
+    mkdir -p $(BUILD_DIR)/unpriv/images && \
+    find $@.workdir -name '*.svg' -exec cp {} $(BUILD_DIR)/unpriv/images/ \; && \
     rm -rf $@.workdir
 endif
 
--- a/post_process_html.py
+++ b/post_process_html.py
@@ -1,0 +1,35 @@
+#!/usr/bin/env python3
+import re, sys
+from pathlib import Path
+
+def process_html(html_path):
+    content = Path(html_path).read_text(encoding='utf-8')
+
+    pattern = re.compile(
+        r'<img([^>]*?)alt="([^"]*)"([^>]*?)>',
+        re.IGNORECASE | re.DOTALL
+    )
+
+    count = 0
+    def replace_match(m):
+        nonlocal count
+        before = m.group(1)
+        alt = m.group(2)
+        after = m.group(3)
+
+        if '.svg' not in (before + after):
+            return m.group(0)
+
+        if alt == "Diagram" or alt == "":
+            alt = "Instruction encoding diagram"
+
+        count += 1
+        # Wrap in figure with visible caption
+        return f'<figure><img{before}alt="{alt}"{after}><figcaption style="clip:rect(0 0 0 0);clip-path:inset(50%);height:1px;overflow:hidden;position:absolute;white-space:nowrap;width:1px">{alt}</figcaption></figure>'
+
+    new_content = pattern.sub(replace_match, content)
+    Path(html_path).write_text(new_content, encoding='utf-8')
+    print(f"Fixed {count} diagrams in {html_path}")
+
+if __name__ == '__main__':
+    process_html(sys.argv[1])
--- a/post_process_html.py.save
+++ b/post_process_html.py.save
@@ -1,0 +1,59 @@
+#!/usr/bin/env python3
+"""
+post_process_html.py — Replace base64-encoded SVG <img> tags with accessible inline <svg>
+Usage: python post_process_html.py build/riscv-spec.html
+"""
+import re, base64, sys
+from pathlib import Path
+
+def inject_a11y(svg_text, alt):
+    title = alt if alt and alt != "Diagram" else "Instruction encoding diagram"
+    
+    # Add role and aria-label to outer svg tag
+    svg_text = svg_text.replace('<svg ', f'<svg role="img" aria-label="{title}" ', 1)
+    
+    # Build hidden text + title + wrap inner content
+    hidden_text = f'<text x="0" y="0" style="position:absolute;width:1px;height:1px;overflow:hidden;opacity:0;font-size:1px">{title}</text>'
+    
+    def wrap_contents(m):
+        opening_tag = m.group(1)
+        rest = m.group(2)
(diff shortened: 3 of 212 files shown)


Maanvi212006:
I notice the CI is failing due to the script being placed in the docs-resources submodule. Should I move it to a tools/ folder in the main repo instead, or is there a preferred location for this kind of script?

omasanori:
I guess a PR would be sent to docs-resources and then another one to here after the first one is accepted, as this script would help other docs. @wmat @aswaterman Your opinions?

wmat:
Agreed, please add the script to docs-resources.  As soon as the PR is there, we can update to the latest version of docs-resources and test the accessibility examples.

Maanvi212006:
Alright! I'll open a PR in docs-resources first with the script, and then follow up with a PR here once it's accepted. Will close this one.

wmat:
Reopening so I can test locally.  DO NOT MERGE yet.

wmat:
So what's the exact test case for this?  I have checked out the PR for the accessibility changes to 3 figures, pointed my local source at the latest docs-resources and builts all of the docs with Make.  What should happen from an accessibility perspective?  I've tried with the Read Aloud chrome extension but don't get any results from the built HTML.

Maanvi212006:
It should actually pick up `alt` attribute.
It should now read something like:

`B-immediate instruction encoding`

But I guess Read Aloud is not able to pick it up. When I searched online, it says that screen readers like NVDA and VoiceOver can give more accurate results. 
But, we will have to make a fix that works for Read Aloud as well. 
So, a proper fix is to inline the SVGs directly into the HTML by disabling `data-uri:` for HTML builds. That way `title` and `desc` are directly readable by all screen readers including Read Aloud and `alt` attribute will then be irrelevant in this case.

Let me know if this works

wmat:
​Everything is at top of tree now, so you should be able to build and view.

Maanvi212006:
Hi @wmat! Good news — the alt text fix is working.

image

However I found that Read Aloud still skips the image even with the correct alt text when SVGs are base64 embedded. The only way to make it work is to disable :data-uri: for HTML builds.
I tried this and the SVGs are no longer base64. But there's a problem: when :data-uri: is disabled, the SVG files are generated in a temporary workdir and deleted after the build, so the HTML ends up with broken image links.

Two possible approaches:

1. Modify the Makefile WORKDIR_TEARDOWN to also copy the generated SVG files to the build folder
2. Keep :data-uri: but write a post-processor that replaces base64 SVGs with inline SVGs in the final HTML

Which approach would you prefer?

wmat:
I would suggest to try each solution and see which one works best.  Note that you have to consider the PDF as well, not just the HTML.

Maanvi212006:
Hi @wmat , just an update!
HTML accessibility is now working. Here's what ended up working:

- Reverted `:data-uri:` to PDF-only (via `ifdef::backend-pdf[]`) so SVGs are linked as files in HTML builds
- Updated `WORKDIR_TEARDOWN` in the Makefile to copy generated SVGs to `build/unpriv/images/` before the workdir is deleted
- Ran `wavedrom_a11y.py` across all `.edn` files in both `unpriv` and `priv` to populate meaningful `alt=` text
- Added `post_process_html.py` as a post-build script that wraps each SVG `` in a `<figure>` with a visually hidden `<figcaption>` — this is what Read Aloud actually picks up and reads aloud


PDF accessibility is the next step

Maanvi212006:
Hi @wmat, just an update!

Investigated the PDF pipeline. Root cause: `asciidoctor-pdf 2.3.24` does not write `/Alt` entries into the PDF tag structure at all — confirmed by inspecting the built output. The alt text from `.edn` files flows correctly through `asciidoctor-diagram` but is silently dropped at the `asciidoctor-pdf` rendering stage.

Planned fix: post-process the PDF after build using `pikepdf` to inject `/Alt` entries, sourcing alt text from the `.edn` files.

Maanvi212006:
Hi @wmat 

## PDF Accessibility — Investigation Summary

As a follow-up to the HTML fix, I investigated whether the same accessibility improvements could be applied to the PDF output. After tracing the full build pipeline, here is what I found.

### What was investigated

The WaveDrom diagram pipeline for PDF follows the same path as HTML up to a point: `.edn` files → `asciidoctor-diagram` gem → SVG → handed to `asciidoctor-pdf` for embedding. The `alt=` text we updated in the `.edn` files does survive through `asciidoctor-diagram` — it is correctly set on the image block passed to `asciidoctor-pdf`.

The problem is what happens next.

### Root cause

Inspecting the built `riscv-spec.pdf` directly reveals two things:

1. **The PDF contains no `StructTreeRoot` and no `MarkInfo`** — it is completely untagged. There is no accessibility tag tree at all, which means screen readers have no structured information to work with. They either skip content or read raw drawing instructions in paint order.

2. **Image XObjects carry no filenames or identifiers** — every diagram is stored as anonymous pixel data (`/I1`, `/I2`, etc.) with keys only for dimensions, encoding, and color space. There is no reference back to the source file or alt text.

A search through the entire `asciidoctor-pdf` 2.3.24 source found no code for writing tagged PDF structures, `StructTreeRoot`, or `/Alt` entries on images. The only alt-text-related code in the codebase is an `on_image_error` handler that renders alt text as visible fallback text when an image file fails to load — this is unrelated to screen reader accessibility.

### What this means

Injecting `/Alt` entries via post-processing would have no effect without a tag tree — screen readers only look for `/Alt` inside tagged structure elements. Building a full tag tree as a post-processing step would require re-describing the entire document structure (headings, paragraphs, figures, tables) from scratch, which is effectively reimplementing what `asciidoctor-pdf` should do natively.

### Recommendation

Tagged PDF output is not currently supported by `asciidoctor-pdf` 2.3.24. This is an upstream limitation outside the scope of what can be fixed in this repository. I would suggest:

- Opening a separate tracking issue for PDF accessibility, noting the upstream dependency
- Raising this with the `asciidoctor-pdf` project as a feature request for tagged PDF / PDF/UA support

The HTML fix in this PR remains fully working and addresses the accessibility need for the primary format most users interact with online.

wmat:
OK, sorry for taking a while on this one but I just checked out the PR and did make build-html and all of the images are broken in the HTML.  Does this need to touch the Makefile at all?  You can't flatten the images diretories, they need to stay where they were.

Maanvi212006:
Hey @wmat, thanks for testing.

Makefile isn't the cause — that image-flattening bug is recently pushed (9bb82cc4): HTML builds now reconstruct correct priv/unpriv image paths, PDF/EPUB untouched.

Separate issue, unrelated to your broken-images report: I also found and fixed a bug in docs-resources/tools/wavedrom_a11y.py (submodule) — it wasn't generating alt text correctly for the 92 bytefield diagrams (register layouts like misa, satp, mtvec). Since docs-resources is a submodule, that fix has to merge upstream first (opened riscv/docs-resources - [#249](https://github.com/riscv/docs-resources/pull/249)), then I'll bump the submodule pointer here. So this PR isn't fully mergeable until that lands too — will flag once it's in.

Quick summary of everything else in this PR so far:
- PDF accessibility: out of scope (asciidoctor-pdf has no tagged-PDF support) — commented separately.
- Bug 1 (Makefile flattening priv/unpriv images): fixed, pushed.
- Bug 2 (manual post-processing step): fixed, now auto-runs in the HTML build.
- Bug 3 (wrong/duplicate/missing alt text): root cause was the two wavedrom_a11y.py bugs above — fix pending the docs-resources merge.

Will ping you once the submodule PR lands and the pointer's updated so you can re-test.

wmat:
I have merged the fix to docs-resources.

Maanvi212006:
Hi @wmat — pushed 5 more commits, wanted to give you a full status update

**What's in this push:**

1. **docs-resources submodule pointer** — updated to current `main` (`50d4c15`), which includes the earlier bytefield alt-text fix (PR [#249](https://github.com/riscv/docs-resources/pull/249), already merged by you). It does *not* yet include a second tool-level fix I found and opened separately — [docs-resources PR #250](https://github.com/riscv/docs-resources/pull/250) — which is still open. More on that below.

2. **Process-gap fix** — `gprs.edn` and `fprs.edn` had never been run through the a11y tool at all in any prior batch, so they had no alt text. Added it.

3. **Misclassified multi-instruction tables** — `rvc-instr-quad0/1/2.edn` and `priv-instr-set.edn` are listing tables (multiple instructions per diagram), which the tool's single-register heuristic wasn't built to handle — it was generating garbled titles for these. I added hand-written `a11y-desc` comments and corrected alt text by hand, since the tool has no way for `a11y-desc` to override its auto-title logic.

4. **Pre-existing quote-escaping corruption** — 5 files (`csr-instr.edn`, `float-csr.edn`, `load-reserve-st-conditional.edn`, `m-st-ext-for-int-mult.edn`, `zifencei-ff.edn`) had section titles containing a literal `"`, which broke an older version of the tool's substitution regex and left garbage text trailing inside the macro brackets. The current tool escapes quotes properly at write time and wouldn't reproduce this on a fresh run, but it also can't self-repair lines that were already corrupted this way. Fixed by hand and confirmed in a real build — `zifencei-ff.edn` needed AsciiDoc's own `\"` escape specifically to avoid double-escaping in the rendered HTML.

5. **Live duplicate alt-text collapse** — this one wasn't a dormant risk, it had actually already happened: `double-ls.edn` (Load + Store panels) and `immediate-variants.edn` (all six R/I/S/B/U/J-Type panels) each had every panel showing the same alt text. Fixed by confirming each panel's real type against its own diagram config before correcting the text.

**Verification done this round:**
- Full clean rebuild succeeds, 487 diagrams processed, no errors.
- Fallback count ("Instruction encoding diagram" generic text) still 229 — unchanged, as expected.
- Spot-checked the trickier fixes (escaped quotes, the two collapse cases) directly against the built HTML output, not just the source `.edn` files.
- Ran repo-wide sweeps for the same corruption patterns (stray trailing text after `alt="..."`, duplicate alt text within a file, duplicate alt text *across* files) — nothing further turned up. I also checked the cross-file duplicates that did show up and they're legitimate (same section heading genuinely appears in two different chapters, e.g. RV32I vs RV64I integer computational instructions) rather than another collapse bug.

I've gone through several rounds of increasingly targeted greps and hand-verification, and each round has turned up fewer issues, but a script-based sweep can only catch patterns I've thought to look for. If anything else surfaces in review or later use, I'm glad to keep digging.

PR #3126, opened 2026-05-21
preface: Markup fix

omasanori:
As every extension in this manual is ratified, there is no longer any meaning to mark them in bold typeface. I updated only the latest tables.


aswaterman (approved):
Yeah, agreed.

PR #3127, opened 2026-05-21
Bring back fprs to F

omasanori:
I had moved the floating-point register state section to the introduction part, but now I thought that adding only the definition of FLEN to the introduction part would provide a better flow. In the end, the relevant main text was mostly reverted to the last ratified one with minor changes.


aswaterman (approved):
SGTM.

PR #3159, opened 2026-05-29
Update IEEE 754 specification reference

hdelassus:
In the unpriv preface, updated reference to the IEEE 754 specification to reflect the 2019 version.

### DIFF
--- a/src/unpriv/preface.adoc
+++ b/src/unpriv/preface.adoc
@@ -489,7 +489,7 @@ floating-point instructions in the 2-bit _fmt_ field.
 * Defined the signed-zero behavior of insn:fmin[] and insn:fmax[], and
 changed their behavior on signaling-NaN inputs to conform to the
 `minimumNumber` and `maximumNumber` operations in the IEEE 754-2019
-specification cite:[ieee754-2019].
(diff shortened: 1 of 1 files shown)


src/unpriv/preface.adoc
```diff
@@ -488,8 +488,8 @@ inaccuracies.
 floating-point instructions in the 2-bit _fmt_ field.
 * Defined the signed-zero behavior of insn:fmin[] and insn:fmax[], and
 changed their behavior on signaling-NaN inputs to conform to the
-`minimumNumber` and `maximumNumber` operations in the proposed IEEE 754-201x
-specification.
+`minimumNumber` and `maximumNumber` operations in the IEEE 754-2019
+specification cite:[ieee754-2019].
```
aswaterman: Thanks for pointing this one out.  I think the way we cited it in the floating-point chapter `<<ieee754-2008, IEEE 754-2008 arithmetic standard>>` renders a bit better; if you can fix it up, I'll merge

hdelassus: Oh I see, I was looking at another file (`f-st-ext.adoc`) instead of `zf.adoc`.

aswaterman (approved):

PR #3163, opened 2026-05-31
Sscofpmf cleanup

davidharrishmc:
Removed a clause about "if supervisor mode is implemented" because the extension name and preamble state that Sscofpmf is a supervisor-level extension.

Clean up CSR name font to be more consistent with the rest of the ISA manual.

### DIFF
--- a/src/priv/sscofpmf.adoc
+++ b/src/priv/sscofpmf.adoc
@@ -1,8 +1,8 @@
 [[Sscofpmf]]
 === ext:sscofpmf[] Extension for Count Overflow and Mode-Based Filtering
 
-The current Privileged specification defines mhpmevent CSRs to select and
-control event counting by the associated hpmcounter CSRs, but provides no
+The current Privileged specification defines `mhpmevent` CSRs to select and
+control event counting by the associated `hpmcounter` CSRs, but provides no
 standardization of any fields within these CSRs. For at least Linux-class
 rich-OS systems it is desirable to standardize certain basic features that are
 broadly desired (and have come up over the past year plus on RISC-V lists, as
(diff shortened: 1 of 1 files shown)


omasanori (commented):
`csr:mie[]` (see https://github.com/riscv/riscv-isa-manual/blob/main/src%2Flib%2Fmacros.rb for details) and such would be better.

However, do not use `csr:...[]` for CSRs with `__n__` because italic inside macros leads an error. For them, your edit is fine.

src/priv/sscofpmf.adoc
```diff
@@ -113,17 +113,17 @@ eventually overflow.
 [[norm:scountovf_op]]
 This extension adds the `scountovf` CSR,
 a 32-bit read-only register that contains shadow copies of
-the OF bits in the 29 mhpmevent CSRs (mhpmevent__3__ - mhpmevent__31__) - where
-scountovf bit _X_ corresponds to mhpmevent__X__.
+the OF bits in the 29 `mhpmevent` CSRs (`mhpmevent3` - `mhpmevent31`) - where
+`scountovf` bit _X_ corresponds to mhpmevent__X__.
```
omasanori:
```suggestion
`scountovf` bit _X_ corresponds to `mhpmevent`__X__.
```

davidharrishmc:
I'm not familiar with the csr:mie[] syntax and have at least made the chapter consistent with other chapters.  Suggest changing to csr:mie[] syntax globally later if that is an editorial policy, rather than doing chapter by chapter.

omasanori:
OK, I will do the conversion later then.

aswaterman (approved):

PR #3165, opened 2026-06-01
Clarify sip can contain machine interrupts even though diagram shows 0s

davidharrishmc:
Clarify the explanation regarding MEI, MTI, and MSI in the context of registers `sip` and `sie`.


aswaterman (commented):
This is true, but I don't think it's how it should be exposed in the supervisor spec.  They're effectively custom interrupts at this level, rather than well-defined standard ones.  I'll think about it and revisit this later.

PR #3169, opened 2026-06-01
Add v0.7.1 Frozen version of RVA23.1 and RVB23.1

edolnx:
Intended for TSC review and approval

This has no changes to the normative text contents of the old v0.7 release before the ISA Manual integration of the Profiles as archived at https://github.com/riscv/riscv-profiles/tree/rva23p1-rvb23p1


edolnx:
Abandonding, missed the requirement for signing and DCO

aswaterman:
Aye, thanks for jumping through those hoops.

PR #3170, opened 2026-06-01
RVA23.1 and RVB23.1 submission for TSC Approval

edolnx:
This PR is intended for TSC review and approval

Included is the v0.7.1 Frozen version of the RVA23.1 and RVB23.1 Profiles. The normative text is unchanged from the previous stand-alone documents archived at https://github.com/riscv/riscv-profiles/tree/rva23p1-rvb23p1

The intention is to move that into the new combined ISA manual, bringing only the minimum required changes.

### DIFF
--- a/src/profiles/rva23p1.adoc
+++ b/src/profiles/rva23p1.adoc
@@ -32,13 +32,13 @@ Expansion Options on top of the `rva23s64`.
 
 ===== Localized Options
 
-There are no privileged localized options in RVA23S64.
+There are no privileged localized options in `rva23p1s64`.
 
 ===== Development Options
 
-The following are new privileged development options in RVA23.1S64:
+The following are new privileged development options in `rva23p1s64`:
 
-- <<svrsw60t59b,**Svrsw60t59b**>> PTE reserved-for-software bits.
+- ext:svrsw60t59b[] PTE reserved-for-software bits.
 
 - **Ssdbltrap** Double Trap.
 
@@ -48,7 +48,7 @@ The following are new privileged development options in RVA23.1S64:
 
 The following are new privileged expansion options in RVA23.1S64:
 
-- **Ssctr** Control Transfer Records, CTR.
+- ext:ssctr[] Control Transfer Records, CTR.
 
 NOTE: *Ssctr* also implies inclusion of *Sscsrind* (indirect CSR access)
 
@@ -56,6 +56,6 @@ NOTE: *Ssctr* also implies inclusion of *Sscsrind* (indirect CSR access)
 
 ===== Transitory Options
 
-There are no privileged transitory options in RVA23S64.
+There are no privileged transitory options in `rva23p1s64`.
 
 //
--- a/src/profiles/rva23p1.adoc
+++ b/src/profiles/rva23p1.adoc
@@ -14,7 +14,14 @@ be replaced with the lower case letter 'p', as in `rva23p1s64`.
 Only supervisor-mode (`rva23p1s64`) profiles are specified in this
 family.
 
-==== `rva23p1s64` Optional Extensions
+=== The `rva23p1s64` Profile
+
+The `rva23p1s64` profile specifies the ISA features available to a
+supervisor-mode execution environment in 64-bit applications
+processors.  `rva23p1s64` is based on and requires an `rva23s64`
+implementation.
+
+==== Optional Extensions
 
 The `rva23p1s64` profile specifies the ISA features available to a
 supervisor-mode execution environment in 64-bit applications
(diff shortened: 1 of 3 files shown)


src/profiles/rva23p1.adoc
```diff
@@ -0,0 +1,61 @@
...
+
+===== Localized Options
+
+There are no privileged localized options in RVA23S64.
+
+===== Development Options
+
+The following are new privileged development options in RVA23.1S64:
+
+- <<svrsw60t59b,**Svrsw60t59b**>> PTE reserved-for-software bits.
```
aswaterman: `extlink:svrsw60t59b[]`

src/profiles/rva23p1.adoc
```diff
@@ -0,0 +1,61 @@
...
+The RVA23.1 profiles are intended to expand upon
+the RVA23 major release of the RVA Family of RISC-V Application
+Processor Profile.
+
+This is a minor release of the RVA23 Profile Family. As such, Minor
+Profiles do not contain any new mandatory extensions and are expected to be
+built on top of the extensions listed in the RVA23 profiles. For
+the sake of simplicity when referring to this profile in code or
+use in a command line argument, all instances of the period will
+be replaced with the lower case letter 'p', as in `rva23p1s64`.
```
aswaterman, line 12: `ext:rva23p1s64[]`

src/profiles/rvb23p1.adoc
```diff
@@ -0,0 +1,57 @@
...
+
+===== Expansion Options
+
+The following are new privileged expansion options in `rvb23p1s64`:
+
+- **Sscfg** Supervisor Counter Delegation.
+
+- **Ssctr** Control Transfer Records, CTR.
+
+NOTE: *Ssctr* also implies inclusion of *Sscsrind* (indirect CSR access)
```
aswaterman, line 53: `ext:ssctr[]`

aswaterman (commented):
Thanks, @edolnx!  Can you use the new formatting widgets (`ext` for extension/profile names, `extlink` for when you want a hyperlink)?  Alternatively, @omasanori or I can handle that later.

edolnx:
I will clean those up tomorrow - I was unclear on the syntax. I'll also try again to make commit signing work

omasanori:
The macros are implemented in [src/lib/macros.rb](https://github.com/riscv/riscv-isa-manual/blob/main/src%2Flib%2Fmacros.rb) and described at the beginning of the file.

src/profiles/rva23p1.adoc
omasanori: You will see `Rva23p1s64` by doing so, which is very different from the corresponding profiles in the past. Is it the new direction from the 23.1 profiles? Will it be applied retroactively?

edolnx:
@omasanori posted a comment that was lost in the shuffle:

> In [src/profiles/rva23p1.adoc](https://github.com/riscv/riscv-isa-manual/pull/3170#discussion_r3373810971):
> 
> > @@ -0,0 +1,61 @@
> +== RVA23.1 Profiles
> +
> +The RVA23.1 profiles are intended to expand upon
> +the RVA23 major release of the RVA Family of RISC-V Application
> +Processor Profile.
> +
> +This is a minor release of the RVA23 Profile Family. As such, Minor
> +Profiles do not contain any new mandatory extensions and are expected to be
> +built on top of the extensions listed in the RVA23 profiles. For
> +the sake of simplicity when referring to this profile in code or
> +use in a command line argument, all instances of the period will
> +be replaced with the lower case letter 'p', as in `rva23p1s64`.
> 
> You will see Rva23p1s64 by doing so, which is very different from the corresponding profiles in the past. Is it the new direction from the 23.1 profiles? Will it be applied retroactively?

This is the first minor profile and there are some quirks. To reduce confusion, we know using the period in the name will be a problem for compiler flags. As such, we moved to the `p` designation and I adopted that in this documentation as often as possible. To reduce confusion further, we found that the compilers do not use `RVA23` as the Profile but instead use `rva23` so the idea was to start changing the documents to match the compilers. The expectation is that the Profiles SIG will work with the TSC to make these changes retroactive as well.

omasanori:
I see. Then, leave them as-is without the `ext:` macro, as the macro will capitalize the first letter automatically. (Maybe adding `profile:` and `profilelink:` macros to the family would be nice.)

rpsene:
Thank you for putting this together, and apologies for the delay in getting it moving.

I am closing this PR in favour of #3298, which carries exactly the same content. The three files there are byte-identical to `7f75e156`, so none of the normative text has changed.

The reason for the resubmission is that this branch cannot pass the repository's contribution checks. `dco_check` needs a `Signed-off-by:` trailer on every commit and none of the seven commits here have one, and `signing_check` needs every commit cryptographically verified where four of the seven are unsigned. Fixing that requires the commit author to re-sign and force-push, which is no longer possible in this case, so RISC-V International staff are carrying the submission forward with the DCO sign-off made on our side.

The work and the original authoring are gratefully acknowledged in #3298.

PR #3173, opened 2026-06-02
Markup fix

omasanori:


src/priv/machine.adoc
```diff
@@ -915,9 +906,9 @@ hole, but this can be done without accessing memory. For example, the
...
 
-The FS and XS fields are read by the privileged code before saving the
-context. The FS field is set directly by privileged code when resuming a
-user context, while the XS field is set indirectly by writing to the
+The csr::[fs] and csr::[xs] fields are read by the privileged code before saving the
+context. The csr::[fs] field is set directly by privileged code when resuming a
+user context, while the csr::[xs] field is set indirectly by writing to the
 status register of the individual extensions.
 [#norm:mstatus_fs_vs_xs_update_indep_priv]#The status fields will
 also be updated during execution of instructions, regardless of privilege mode.#
```
omasanori, line 914: Why is VS not mentioned in this paragraph? Is it correct?

src/priv/machine.adoc
```diff
@@ -836,63 +835,55 @@ values being Off, Initial, Clean, and Dirty.
...
+The csr::[xs] field effectively reports the maximum status value across all
 user-extension status fields, though individual extensions can use a
-different encoding than XS.
+different encoding than csr::[xs].
 ====
 
-[#norm:mstatus_sd_acc]#The SD bit is a read-only bit# that
-[#norm:mstatus_sd_op1]#summarizes whether either the FS, VS,
-or XS fields signal the presence of some dirty state that will require
+[#norm:mstatus_sd_acc]#The csr::[sd] bit is a read-only bit# that
```
omasanori, line 882: Does SD stand for State Dirty? Adding full spell here?

aswaterman: It does, though I'd contend it's already implicitly spelled out ("SD [...] summarizes [...] presence of some dirty state"), which is consistent with how we handle FS etc.

src/priv/machine.adoc
aswaterman: VS looks like FS in this regard.  It's probably an editing error that it isn't mentioned.  I'll fix it.

aswaterman (approved):

src/priv/machine.adoc
aswaterman: https://github.com/riscv/riscv-isa-manual/pull/3174/

PR #3177, opened 2026-06-04
Removed tags from text that is not actually a normative rule

davidharrishmc:


aswaterman (approved):

aswaterman:
@davidharrishmc You're missing the `Signed-off-by` and hence failing the DCO check.  (Yes, this is separate from the SSH signing key.)  I'll bypass the rules and merge, but please get that sorted out for next time.  https://github.com/riscv/docs-dev-guide/blob/1eff69a7b7123f47e745e5f5ba3ddbcbe4691a20/CONTRIBUTING.md?plain=1#L50

PR #3178, opened 2026-06-04
Instruction indexes

wmat:
@aswaterman here is the instruction indexes for unpriv_rv32, unpriv_rv64, priv_rv32, and priv_rv64.  The 4 new adoc files were created from all_instructions.adoc in UDB.  I have a PR to add the script that creates them to riscv-unified-db.  Currently these are manually created, however, it could be automated and synced to riscv-isa-manual with a GH action.

Please review the indexes in the PDFs and HTML before merging.


aswaterman:
@wmat Thank you.  I am playing around with an alternate scheme to generate these from UDB as part of the spec build process, rather than committing the compiled result.  It might not pan out, in which case this route is OK.  Let's sit on this for now.

wmat:
Sounds good.  Note that I have a PR to UDB as well that splits the single adoc file.

In the meantime, I'll clean up the check failures.

wmat:
Also, the script I've pushed to UDB can be easily automated. I didn't immediately do it as I wanted to ensure the ISA docs were good.

aswaterman:
Yeah, whichever approach we take, we'll want to automate it.  Give me some time to keep pushing on my scheme and then we can decide.

wmat:
@aswaterman fixed up the DCO Check failure so once the Build ISA completes the artifacts will be available.  Note that the instructions in the body of the books will link to their single page description in their respective books.  This means that something like ADD will reside in both Unpriv and Priv indexes.  This is OK though, as those indexes are generated from UDB's single all_instructions.adoc file, not manually added twice.

Also, with the additions of theses indexes the ISA build step takes 1-2 an hours, so that will need considerable improvement.

aswaterman:
@wmat While I continue working on the generation aspect, can you research where all the runtime is going and start looking into solutions to improve it?

wmat:
Will do.

Issue #3156, opened 2026-05-29
[Clarification] - Zacas/PMA behaviors


wangpc-pp:
## Summary
We defined some new PMA levels for `Zacas`: AMOCASW, AMOCASD, and AMOCASQ.

The spec says, for example, `AMOCASQ indicates that in addition to instructions indicated by AMOCASD-level support, the AMOCAS.Q instruction is supported.`.

Then if the `AMOCASQ` level is not supported, we should raise an illegal instruction error for `AMOCAS.Q`?

---

## Spec Reference
https://github.com/riscv/riscv-isa-manual/blob/d1a5fe18e5cc85b1f288fc7c4c38d20605165867/src/priv/machine.adoc#amo-pma

> The Zacas extension defines three additional levels of support: AMOCASW, AMOCASD, and AMOCASQ.
> AMOCASW indicates that in addition to instructions indicated by AMOArithmetic-level support, the AMOCAS.W instruction is supported.
> AMOCASD indicates that in addition to instructions indicated by AMOCASW-level support, the AMOCAS.D instruction is supported.
> AMOCASQ indicates that in addition to instructions indicated by AMOCASD-level support, the AMOCAS.Q instruction is supported.
---

ved-rivos:
When an access is not supported by the PMA, it is recommended to cause an access-fault corresponding to the type of access - in this case a store/AMO access-fault - if the implementation provides precise trapping on PMA violations.

ved-rivos:
Hope that helped.

PR #3179, opened 2026-06-05
Split Zacas into Zacasw/Zacasd/Zacasq sub-extensions by operand width

wangpc-pp:
This change restructures the `Zacas` extension into three fine-grained sub-extensions:

- `Zacasw`: Word (32-bit) CAS via `amocas.w`
- `Zacasd`: Doubleword (64-bit) CAS via `amocas.d`
- `Zacasq`: Quadword (128-bit) CAS via `amocas.q` (RV64 only)

`Zacas` remains as a shorthand umbrella extension that implies all three sub-extensions, preserving backward compatibility for software that already depend on `Zacas`.

The sub-extensions form a progressive dependency chain, mirroring the PMA `AMOCASW/D/Q` progressive levels defined in the privileged spec:

```
  Zaamo -> Zacasw -> Zacasd -> Zacasq
```

This matches the PMA definition where `AMOCASD` requires `AMOCASW`, and `AMOCASQ` requires `AMOCASD`.

Motivation
============================================================

See also: https://github.com/riscv-non-isa/riscv-toolchain-conventions/issues/173.

1. The monolithic `Zacas` extension forces an all-or-nothing choice: either all three CAS widths or none.
2. Toolchains may generate `amocas.q` instructions for targets that only support `amocas.w/d`, causing access faults at runtime.
3. There is no way for software to discover which CAS widths are actually supported by the hardware.

Solution
=========

By splitting `Zacas` into per-width sub-extensions:

- Hardware can implement only the CAS widths it supports (e.g., a low-end RV64 core might implement `Zacasw` and `Zacasd` but not `Zacasq`).
- Toolchains can use the standard ISA extension discovery mechanism (misa, device tree, hwprobe) to determine which CAS widths are available, without needing to query PMA levels.
- Profiles can mandate specific sub-extensions independently (e.g., RVA30 could mandate Zacasw+Zacasd as mandatory and Zacasq as optional).
- The `Ziccamoc` extension (`AMOCASQ` PMA level) maps cleanly to `Zacasq`.

Comparison with X86 and AArch64 128-bit CAS
=============================================

1. X86 (`CMPXCHG16B`):

  - 128-bit CAS is provided by the `CMPXCHG16B` instruction, introduced in AMD64 Revision F (2006). It is a separate instruction from the 64-bit `CMPXCHG8B`.
  - `CMPXCHG16B` is an optional feature discovered via CPUID feature bit (CX16). Not all x86_64 processors support it (early AMD64 chips lacked it).
  - The instruction requires 16-byte aligned memory operands and uses RDX:RAX for the comparand and RCX:RBX for the exchange value.
  - This is analogous to `Zacasq` being a separately discoverable sub-extension rather than bundled with all CAS widths.

2. AArch64 — Three generations of 128-bit atomics:

  - AArch64 progressively added 128-bit atomic support across three separate, independently-discoverable extensions:

    - ARMv8.0 (baseline):

      - LDXP/STXP (Load/Store Exclusive Pair) provides 128-bit LL/SC-style atomics using register pairs. Requires a retry loop, needs write permission even for pure loads.

    - ARMv8.1-A FEAT_LSE (Large System Extensions):

      - CASP/CASPA/CASPAL/CASPL: Single-instruction 128-bit CAS operating on a pair of 64-bit registers (Xt, Xt2). This is the closest analog to Zacasq's amocas.q with register pairs.
      - Also adds atomic LDADD/CLR/EOR/SET/SMAX/SMIN/UMAX/UMIN and SWP for 8/16/32/64-bit widths. - Limitation: only solves 128-bit CAS. 128-bit atomic load still requires LDXP (needs write permission, inefficient). No 128-bit atomic fetch-and-op.

    - ARMv8.4-A FEAT_LSE2:

      - Makes aligned LDP/STP (load-pair/store-pair) single-copy atomic for 128-bit data. This means ordinary load-pair/store-pair can atomically read/write 128-bit values without needing exclusive instructions or write permission.
      - Does NOT add new instructions; only strengthens the atomicity guarantee of existing LDP/STP. - Solves the "128-bit atomic load/store without write permission" problem that LSE CASP left open.

    - ARMv9.4-A FEAT_LSE128:

      - Extends LSE with true 128-bit data-path atomic operations:

        - LDCLRP  — 128-bit atomic bit-clear (AND NOT)
        - SWPP    — 128-bit atomic swap (swap quadword)
        - LDIAPP  — 128-bit load-acquire with RCpc ordering
        - STILP   — 128-bit store-release with RCpc ordering

      - First time AArch64 supports non-CAS read-modify-write atomics
        on 128-bit data.
      - Independently discoverable optional feature.

  - Summary table:
    Feature       | ARM version | 128-bit CAS | 128-bit load/store | 128-bit fetch&op
    --------------|-------------|-------------|--------------------|------------------
    Baseline      | ARMv8.0     | LDXP/STXP   | LDXP/STXP (needs W) | none
    FEAT_LSE      | ARMv8.1-A   | CASP        | LDXP/STXP (needs W) | none
    FEAT_LSE2     | ARMv8.4-A   | CASP        | LDP/STP (atomic)    | none
    FEAT_LSE128   | ARMv9.4-A   | CASP        | LDP/STP + LDIAPP/STILP | SWPP, LDCLRP
  - Key takeaway:
    AArch64 took three generations of extensions to fully cover 128-bit
    atomics, each independently optional and discoverable. This strongly
    supports splitting RISC-V's Zacas into per-width sub-extensions, and
    leaves room for future extensions (e.g., 128-bit atomic fetch-and-op)
    without breaking the extension model.

Both X86 and AArch64 treat 128-bit CAS/atomics as separately discoverable features, not bundled with narrower atomic operations. This proposal aligns RISC-V with that established industry practice.

AI-Assisted: This commit was generated with AI assistance.


ved-rivos:
Conflating PMA - physical memory attributes - with instructions is not right. The premise that tool chains will query PMA to determine what instructions to emit is also not right since at the time of compilation one cannot determine how the memory operated on by the instruction will be malloc or mmap from.

wangpc-pp:
> Conflating PMA - physical memory attributes - with instructions is not right. 

I can understand that. The point is that we don't have a mechanism to know if an atomic instruction can be executed successfully. The PMA is designed as a fine-grained system but the atomic ISA is not.

> The premise that tool chains will query PMA to determine what instructions to emit is also not right since at the time of compilation one cannot determine how the memory operated on by the instruction will be malloc or mmap from. 

IIUC, you are referring to PMP instead of PMA here? PMA especially atomic levels can't be changed dynamicly? Correct me if I am wrong.

Besides, as noted, splitting Zacas makes it more fine-grained.

ved-rivos:
I am referring to PMA . For instance some memory may only support an access width PMA of 1 byte and may only support LB and SB accesses.

aswaterman:
New extensions should not be proposed in this forum; they must go through standard RVIA processes, even if they only consist of combinations of existing ratified features.

I tend to agree with Ved that this isn't the right solution to the problem.  As he points out, using ISA extensions to describe this would require going down to extremely fine granularity (e.g. some memory regions might only allow naturally aligned 4-byte loads and stores, and nothing else).

Compilers should generally assume that pointers correspond to cacheable main memory, and that all instructions in the ISA can be used on those pointers, unless informed otherwise by some kind of attribute (or by the programmer simply choosing not to use certain language-level atomics).  This is how other architectures work, and it doesn't seem wise to diverge from that approach in RISC-V.

wangpc-pp:
Sorry, I may not make it clear because of lack of some context. The actual motivation isn't to expose PMAs to ISA, but to facilitate a mechanism to check if `2-XLEN CAS`s like `amocas.d` on RV32 and `amocas.q` on RV64 are supported. Then we checked the whole unprivileged/privileged spec and found these sentences:
```
The Zacas extension defines three additional levels of support: AMOCASW, AMOCASD, and AMOCASQ.
AMOCASW indicates that in addition to instructions indicated by AMOArithmetic-level support, the AMOCAS.W instruction is supported.
AMOCASD indicates that in addition to instructions indicated by AMOCASW-level support, the AMOCAS.D instruction is supported.
AMOCASQ indicates that in addition to instructions indicated by AMOCASD-level support, the AMOCAS.Q instruction is supported.
```
Then we thought maybe we should expose some PMAs to user-space? Yeah, that was naive and not a common solution. I agree we should not go in this way. :-)

And then we think the once-for-all solution is splitting Zacas into smaller sub-extensions.

First, `Zacas` is very useful for large scale SoC. Based on our evaluation, LSE beats `LL/SC`, and so does RISC-V's `Zacas` and `Zalrsc`.

The key problem is that Zacas mandates `2-XLEN` CASs on register pairs. This requires the hardware to read `2-XLEN` registers. This has a big impact on the microarchitecture so `2-XLEN` CASs may not be supported.

But as described: 
1. The monolithic Zacas extension forces an all-or-nothing choice: either all three CAS widths or none. In the ISA level, all instructions should be implemented now.
2. Toolchains may generate `amocas.d`/`amocas.q` instructions for targets, causing access faults at runtime (https://github.com/riscv/riscv-isa-manual/issues/3156).
3. There is no way for software to discover which CAS widths are actually supported by the hardware.

We have investigated the usage of 128 bits CAS based on open-source softwares and our internal applications, and the result is that the scenarios are rare. Here are some typical usages:
1. Update freelist in slub:  https://github.com/torvalds/linux/blob/4549871118cf616eecdd2d939f78e3b9e1dddc48/mm/slab.h#L20
2. Lock-free stack in DPDK: https://github.com/DPDK/dpdk/blob/4757b8df04b67278e7d5b2bffa8e998e02e78d12/lib/stack/rte_stack_lf_generic.h#L63

And this is why I think we should split `Zacas`: makes `Zacas` more fine-grained and gains most of the benefit of CAS without introducing too much complexity.

WDYT?

ved-rivos:
> There is no way for software to discover which CAS widths are actually supported by the hardware.

A RV64 implementation that supports Zacas must implement all instructions provided by Zacas. There is no optionality. If Zacas and RV64 are supported then software should assume that all amocas instruction widths are supported.  

The PMA are attributes of memory regions. For main memory, a RV64 implementation that implements Zacas will support all widths of AMOCAS. For non-memory locations (device memory, etc.) some or none of the widths may be supported.

This is similar to how when when RV64 is supported, all widths of load/store instructions must be supported by hardware; however the PMA of some memory regions may not accept all widths. The other analogue are PMAs for AMO - AMOArithmetic, AMOLogical, AMOSwap, and AMONone. AMO is a property of the memory - a physical memory attribute. If some memory in the platform indicates AMONone indicating it does not support any AMO, it does not mean that the implementation does not support AMO on any memory.

So what you are proposing is conflating PMAs with extension support. Even with Zacas split out into 3 sub-extensions there will be memory regions in the platform that support all widths and some memory regions that may support none or some. Even if these sub extensions were created most RV64 implementations will advertize all of these since all of these would be supported on cacheable main memory regions but some/none of them may be supported on say IO memory regions. Creating these sub-extensions only creates more fragmentation and complexity for compliers but does not change the fact that not all memory regions - a PMA attribute - may support all widths. 

Compilers should generally assume that pointers correspond to cacheable main memory, and that all instructions in the ISA can be used on those pointers.

wangpc-pp:
Then why do we need `Ziccamoa`/`Ziccamoc`/... to mandate the atomicity in main memory region? Isn't that because atomic ISAs don't mandates PMAs and `Ziccamoa`/`Ziccamoc`/... are patches to fix the gap in the profile?

ved-rivos:
> Then why do we need Ziccamoa/Ziccamoc/... to mandate the atomicity in main memory region?

The base ISA provides maximal flexibility and optionality for implementations to customize the machine to the specific use cases it targets. The profiles take away some of that flexibility by placing mandates on what is optional. 

The profiles are intended to provide a uniform target for portable operating system, portable software, and portable tool chains to target. For instance, a machine built for a specific use case like a noise canceling ear plug may not want to comply with the profile because of its power, performance, or other optimization goals. However, there is also no expectation that a standard off-the-shelf package, OS, or application can be installed on that ear plug. On other hand, an implementer building a general purpose platform like a PC or a server will want to comply with the profiles and other platform specifications such as the Server SoC specification and Server Platform specification to enable portability of software between compliant platforms. 

Continuing with the ear plug example, the implementer developing the application understand what the properties of the memory regions are in that machine and will malloc memory from the right memory region for the right use case. On other hand application developers targeting general purpose platforms cannot be expected to know intricate details of each memory region. Such developers will expect uniformity in the properties that main memory provides and can assume that all main memory regions in a RVA23 compliant platform will provide AMOCASQ level atomicity support. The profile defined extensions such as Ziccamoa and Ziccamoc enable that portability.

> Isn't that because atomic ISAs don't mandates PMAs and Ziccamoa/Ziccamoc/... are patches to fix the gap in the profile?

Its not a gap in the ISA. The ISA intentionally provides maximal flexibility and optionality. The profiles tighten it for general purpose use to enable portability.

PR #3184, opened 2026-06-05
3182 local isa build time enhancements

### LINKED ISSUE #3182: Local ISA build time enhancements
Currently, the ISA build times locally are:

$time make
real    14m3.605s

wmat:
Adds some tweaks to the makefile for local build speed improvements and updates the README with additional instructions to take advantage of them.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -37,8 +37,10 @@ https://github.com/riscv/docs-dev-guide
 From the repo root:
 
 ```bash
-make build
+make -j2 build
 ```
+
+NOTE:  If you have a powerful machine with lots of RAM you can use -j4 or higher parallelism settings, however, -j2 is optimal for most machines.
 
 Common targets:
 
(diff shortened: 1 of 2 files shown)


aswaterman (commented):
The reason I set up private working directories for the build is precisely because sharing images between the documents results in race conditions that make the build unstable.  The simplest way to fix the race condition is to not do any sharing.

If we are doing parallel builds, then there isn't very much speed advantage in sharing the images build, anyway--we'll do some redundant work in parallel, but that's OK.

Once you get rid of the sharing, this basically degenerates to what we already did with `UNRELIABLE_BUT_FASTER_INCREMENTAL_BUILDS`.  So my recommendation is to update the readme to suggest that people who are iterating on spec changes use `make build-html UNRELIABLE_BUT_FASTER_INCREMENTAL_BUILDS=1`.

wmat:
Got it, thanks.

wmat:
This one should be good now, @aswaterman

aswaterman (approved):

PR #3188, opened 2026-06-07
Fix bit-manipulation reference pseudocode

abhinavagarwal07:
This fixes a few small mistakes in the bit-manipulation reference pseudocode:

- stop the `brev8` and `xperm8`/`xperm4` loops at the last in-range byte/nibble
- make `clzw` call the 32-bit helper it defines, instead of the XLEN-wide helper
- make the `ctz`/`ctzw` loops count upward, matching the prose and the `cpop`/`cpopw` pattern

These are text/pseudocode corrections only; they do not change the intended architectural behavior.

### DIFF
--- a/src/unpriv/zb.adoc
+++ b/src/unpriv/zb.adoc
@@ -872,7 +872,7 @@ function HighestSetBit32 x = {
 }
 
 let rs = X(rs);
-X[rd] = 31 - HighestSetBit(rs);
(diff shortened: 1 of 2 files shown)


aswaterman (approved):

abhinavagarwal07:
@aswaterman Gentle ping on this. This is approved but not merged.

aswaterman:
Thanks for the ping.

PR #3191, opened 2026-06-08
Use macros for Sv extensions

aswaterman:


omasanori (approved):
Thanks a lot, doing the conversion alone in my spare time is actually hard and will take a long time to complete.

aswaterman:
@omasanori Yeah, I didn't intend to leave all of the work to you, it's just that the amount of time I have to work on this varies a lot from week to week.

PR #3203, opened 2026-06-12
Markup fix

omasanori:


src/priv/machine.adoc
```diff
@@ -1515,59 +1515,59 @@ Otherwise, the corresponding bits in csr:sip[] and csr:sie[] are read-only zero.
...
 but event 0 is defined to mean "`no event.`"#
 [#norm:mhpmcounter_mandatory]#All counters should be implemented#, but
 [#norm:mhpmcounter_mhpmevent_rdonly0]#a legal implementation is to make both the counter and its corresponding event selector be read-only 0.#
 
-.Hardware performance monitor counters.
+.Hardware performance monitor counters
 include::images/bytefield/hpmevents.edn[]
 
-[#norm:mhpmcounter_warl]#The `mhpmcounters` are *WARL* registers# that
+[#norm:mhpmcounter_warl]#The `mhpmcounter__n__` are *WARL* registers# that
```
aswaterman: Maybe ``` `mhpmcounter`s ```? ("The mhpmcounter__n__ are" sounds like awkward English even if it's mathematically logical.)

omasanori: It sounds better to me. I will revise.

omasanori: @aswaterman Revised and pushed. I think alternative wordings would be `the mhpmcounter__n__ CSRs are WARL registers` or `csr:mhpmcounter*[] are WARL registers` or such. Your opinion?

aswaterman (approved):

src/priv/machine.adoc
aswaterman: Either seems fine to me.  I'll go ahead and stick with what's in this PR.

PR #3204, opened 2026-06-12
Don't use sed to satisfy pre-commit for generated files

Timmmm:
Instead of hackily using `sed` to remove a newline to make pre-commit happy about the tags JSON, just exclude that file from pre-commit checks because it's generated anyway.


aswaterman (approved):
Agreed, thanks.

aswaterman:
@Timmmm ping me when the DCO/signing checks are fixed, and I'll merge this.  Thanks again.

PR #3206, opened 2026-06-13
Clarify silent-store optimization for spinlocks

slate5:
This PR adds a non-normative note describing an implementation optimization that is not obvious from any single section of the ISA. The note is explanatory only and does not introduce a new algorithm. In contended spinlocks, silent-store detection can reduce write traffic for `amomax` or `amoor` in the retry loop.

```asm
        la      t0, lock
        li      t1, 1
again:
        amomax.w.aq     t1, t1, (t0)    # silent store when lock == 1
        bnez            t1, again
```
The effect is similar to test-and-test-and-set (TTAS) spinlock behavior, but without the initial optimistic load.

### DIFF
--- a/src/unpriv/mm-explanatory.adoc
+++ b/src/unpriv/mm-explanatory.adoc
@@ -1707,6 +1707,7 @@ are redundant with other PPO rules under RVTSO assumptions
 
 Other general notes:
 
+[[sec:silent-stores]]
 * Silent stores (i.e., stores that write the same value that already
 exists at a memory location) behave like any other store from a memory
 model point of view. Likewise, AMOs which do not actually change the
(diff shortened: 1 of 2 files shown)


aswaterman:
Linking these sections makes sense, but I'd rather not over-emphasize these particular AMOs or the spinlock use case.  Contended spinklocks should generally be implemented using test-and-test-and-set anyway, to avoid obtaining exclusive access to a lock that's currently held.  The silent-store optimization helps a little bit in the racy case where the lock appeared not to be held but is acquired between the load and the AMO, but ideally it's a second-order effect.

I'll lightly edit the text accordingly and merge.  Thanks!

aswaterman (approved):

slate5:
Thanks for the explanation, @aswaterman! Yes, silent-store optimization is a second-order effect. I also made a mistake by assuming that only logical or arithmetic AMOs (RMW) can benefit from the silent store, while amoswap cannot (hence I forced usage of those in spinlock).

PR #3210, opened 2026-06-16
Add validation for xref fallback text

slate5:
Find xrefs without explicit link text that fall back to the anchor name enclosed in brackets. This happens when the xref definition is not near the section title or reference text.

I made a mistake in a previous PR, so adding a test for this should help prevent regressions, at least my own :)


aswaterman (approved):
LGTM, thanks.

PR #3212, opened 2026-06-17
Add type attribute to colorize register diagrams

wmat:
Re-adding the type attribute back to the wavedrom files to colorize them.


aswaterman (commented):
What is the algorithm for how to color these?  I didn't go through the whole PR, but one thing that jumped out at me is that multi-bit mstatus fields are given a different type than single-bit mstatus fields.  What does that do?

wmat:
No algorithm, I just matched them to what was in v20240411 and added the same to the new ones.

PR #3214, opened 2026-06-17
Remove stale latexmath and unrelated dependencies

wmat:
- Drop mathematical and asciidoctor-mathematical from dependencies/Gemfile and the matching --require=asciidoctor-mathematical / -a mathematical-format=svg from the Makefile. No .adoc file under src/ uses stem:[]/latexmath/asciimath anymore (latexmath removal is complete on main), so these gems and their native build chain (mathematical, lasem/cairo/pango) are no longer exercised by any content.
- Drop ruby_dev from dependencies/Gemfile: an unmaintained, unrelated personal gem (not the ruby-dev apt package) that has no bearing on this build.


aswaterman (approved):
Always good to delete stuff.

PR #3222, opened 2026-06-20
Connect unprivileged to privileged references

slate5:
Since the two specifications were merged, each can cross-reference the other.


aswaterman (approved):
Thanks, I have been modifying these as I come across them but haven't reached these yet.

PR #3223, opened 2026-06-22
Fix Zvkt DIEL constant-time list (vmslt[u].vv, vrsub.vi)

abhinavagarwal07:
The Zvkt DIEL list names several vector forms that have no encoding and omits the real ones.

- `vmslt.vi` / `vmsltu.vi` have no immediate encoding (they assemble to `vmsle[u].vi`, already listed); the real `vmslt.vv` / `vmsltu.vv` were missing.
- `vrsub.vv` has no encoding; the real reverse-subtract immediate form is `vrsub.vi`.

This replaces the phantom forms with the real encoded instructions. Text only, no encoding change.


aswaterman (approved):

aswaterman:
Thanks!

PR #3224, opened 2026-06-22
Align Zicbom cache-block management faults with the CMO chapter

abhinavagarwal07:
The privileged spec currently lists `cbo.inval`/`cbo.clean`/`cbo.flush` with loads in the PMP and page-fault rules, implying read permission and load-class faults.

The ratified CMO chapter instead permits these instructions whenever load or store access is permitted, leaves the fetch-only case unspecified, and reports store-class faults when access is not permitted.

That permission rule is over the effective (translation and PMP) permissions, so it cannot be restated by the per-layer privileged fault rules without wrongly faulting read-only or fetch-only regions. This patch removes cache-block management from the load rules and defers its permission requirements to the CMO chapter, reporting the resulting faults as store-class: a store access-fault for PMP, and a store page-fault when address translation permits no access.

It also removes a duplicated "instruction" in the store page-fault sentence. `cbo.zero` is already store-type and is unchanged.

### DIFF
--- a/src/priv/machine.adoc
+++ b/src/priv/machine.adoc
@@ -3407,12 +3407,15 @@ include::images/bytefield/pmpcfg.edn[]
 
 [#norm:pmp_exec_fault]#Attempting to fetch an instruction from a PMP region that does not have
 execute permissions raises an instruction access-fault exception.#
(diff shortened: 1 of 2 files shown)


aswaterman:
deferring review of this one to @dkruckemyer-work

abhinavagarwal07:
gentle reminder @dkruckemyer-work

PR #3225, opened 2026-06-22
Revise the SPMP spec according to public review.

ybc-alkaid:
- Add normative rule definitions for `Sspmp`, `Sspmpen`, and `Smpmpdeleg`, and tag the corresponding normative text in `src/sspmp.adoc`.
- Clarify `Sspmp` dependency requirements, permission semantics, address matching behavior, no-match denial, SPMP/paging mutual exclusion, and ordering requirements after SPMP CSR writes.
- Clarify indirect CSR access through `siselect` and `miselect`, including out-of-bounds read/write behavior and M-mode access semantics.
- Revise `Sspmpen` and `Smpmpdeleg` wording around entry enable behavior, delegated PMP/SPMP ranges, locked entries, and CSR behavior.
- Update the SPMP encoding table image and CSR table descriptions for `spmpen` and `spmpenh`.


normative_rule_defs/sspmp.yaml
```diff
@@ -0,0 +1,247 @@
...
+      pmpnum cannot be lowered to or below the index of any locked PMP entry.
+    tags: ["norm:csrfld:mpmpdeleg:pmpnum:locked_pmp_constraint"]
+
+  ##############################################
+  # Smpmpdeleg Section: M-mode Access
+  ##############################################
+  - name: miselect_spmp_access
+    description: |
+      M-mode indirect access uses miselect for entry selection; mireg and mireg2 access spmpaddr/spmpcfg.
+    tags: ["norm:csr:miselect:spmp_access"]
```
jordancarlin: Please update the tags to follow the naming convention in https://github.com/riscv/riscv-isa-manual/blob/main/tagging_normative_rules.adoc. The only colon should be after norm. Look at the other normative rules for examples.

normative_rule_defs/sspmp.yaml
```diff
@@ -0,0 +1,247 @@
...
+  - name: mpmpdeleg_locked_pmp_constraint
+    description: |
+      pmpnum cannot be lowered to or below the index of any locked PMP entry.
+    tags: ["norm:csrfld:mpmpdeleg:pmpnum:locked_pmp_constraint"]
+
+  ##############################################
+  # Smpmpdeleg Section: M-mode Access
+  ##############################################
+  - name: miselect_spmp_access
+    description: |
```
jordancarlin: The description should be left blank for most normative rules unless the text in the ISA manual is not clear on its own. As this is a new spec, ideally all of the text is clear enough to not need the description.

jordancarlin (changes requested):

src/sspmp.adoc
```diff
@@ -437,14 +447,15 @@ M-mode employs different methods to access PMP and SPMP entries.
 PMP entries are accessed directly through their dedicated CSRs (i.e., `pmpcfg` and `pmpaddr`).
 Delegated SPMP entries, however, are accessed indirectly using the `xiselect` CSR (i.e., `siselect` and `miselect`).
 
-For these indirect accesses, `miselect` selects the target SPMP entry, `mireg` accesses its `spmpaddr` register, and `mireg2` accesses its `spmpcfg` register.
-The `mireg3` through `mireg6` are reserved.
+[#norm:csr:miselect:spmp_access]#For these indirect accesses, `miselect` selects the target SPMP entry, `mireg` accesses its `spmpaddr` register, and `mireg2` accesses its `spmpcfg` register.#
+The `mireg3` through `mireg6` are reserved (i.e., reads return zero, writes are ignored).
```
aswaterman: This is not what "reserved" means, so "i.e." is inappropriate.

I think this change should just be reverted.

src/sspmp.adoc
```diff
@@ -404,11 +411,11 @@ The `mpmpdeleg.pmpnum` field is a WARL field, which allows an implementation to
 *Addressing:*
 
 Both PMP and SPMP entries are indexed starting from zero.
-For example, in an implementation with 64 total entries where `pmpnum` is configured to 16:
+For example, in an implementation with 64 total entries where `pmpnum` is configured to N (asciimath:[N \le 64]):
```
aswaterman: We don't use the asciimath package in this repo; rewrite as `N {le} 64`

Please remove all other uses of asciimath, too.

src/sspmp.adoc
```diff
@@ -420,14 +427,17 @@ Its lower 8 bits are an alias for the 8-bit field in the corresponding M-mode PM
...
 . The `pmpnum` value can be set to override locked SPMP entries. For example, if `SPMP[0]` is locked, M-mode software can still increment `pmpnum`.
+. As `pmpnum` is reconfigured, SPMP entries with the largest indices change.
+For instance, raising `pmpnum` from N to N+2 truncates SPMP and spmpen ranges from `SPMP[0..63-N]`, `spmpen[0..63-N]` to `SPMP[0..61-N]`, `spmpen[0..61-N]`.
+Conversely, reducing `pmpnum` from N+2 to N expands SPMP and spmpen ranges from `SPMP[0..61-N]`, `spmpen[0..61-N]` to `SPMP[0..63-N]`, `spmpen[0..63-N]`.
 
 [NOTE]
 ====
 Changing `pmpnum` at runtime carries the side effect of inheriting configuration values from previous PMP or SPMP roles.
-Consequently, before writing a new value to `pmpnum`, the M-mode software should perform sanity checks, including storing and clearing the fields of the affected configuration registers.
+Consequently, before writing a new value to `pmpnum`, the M-mode software should perform confidence checks, including storing and clearing the fields of the affected configuration registers.
```
aswaterman: "Confidence" is not the right word.  "Checks" is also not right since we are recommending that M-mode software perform writes, not just reads.

The real problem is that this sentence is vague about what it is recommending.  "Including" implies that this recommendation is insufficient and other checks/updates are needed.  What are they?

If this recommendation really is sufficient, then drop the weasel words, e.g.:

Consequently, before writing a new value to `pmpnum`, M-mode software should initialize the affected configuration registers to a value suitable for their new role.

src/sspmp.adoc
```diff
@@ -371,23 +377,24 @@ The `spmpen` register offers significant benefits for context switch optimizatio
...
 
 
 
 [[pmp-entry-sharing]]
 ==== Resource Sharing between PMP and SPMP
 
-The `Smpmpdeleg` extension facilitates the delegation of PMP entries for use by S-level, thereby creating S-level PMP (SPMP) entries.
-This delegation is managed by an MXLEN-bit M-mode CSR named `mpmpdeleg`, whose layout is detailed in <<mpmpdeleg-format-rv64>>.
+The `Smpmpdeleg` extension facilitates the delegation of PMP entries for use by S-mode, thereby creating SPMP entries.
+This delegation is managed by an MXLEN-bit M-mode CSR named `mpmpdeleg` (CSR number 0x316), whose layout is detailed in <<mpmpdeleg-format-rv64>>.
```
aswaterman: We don't need to mention the CSR number here.  It should be in the `priv-csrs.adoc`.

src/priv-csrs.adoc
```diff
@@ -359,8 +359,8 @@ Supervisor indirect register alias 6.
 4+^|Supervisor Protection and Translation
 
 |`0x180` |SRW |`satp` |Supervisor address translation and protection.
-|`0x183` |SRW |`spmpen` |Supervisor physical memory protection mask.
-|`0x193` |SRW |`spmpenh` |Upper 32 bits of supervisor physical memory protection mask, RV32 only.
+|`0x183` |SRW |`spmpen` |Supervisor physical memory protection entry enable.
+|`0x193` |SRW |`spmpenh` |Upper 32 bits of supervisor physical memory protection entry enable, RV32 only.
```
aswaterman: Where is `mpmpdeleg`?

src/sspmp.adoc
```diff
@@ -325,21 +331,21 @@ Since, when V=1, SPMP is only active when hgatp.MODE=Bare, SPMP protection and G
...
 === "Sspmpen" Extension for Optimizing Context Switching of SPMP Entries
 
-In RV64, a context switch for the SPMP mechanism involves updating as many as 64 address registers and configuration registers.
+In RV64, a context switch for the SPMP mechanism involves updating up to 64 address registers and 64 configuration registers.
 The `Sspmpen` extension, introduced in this chapter, is an optional feature designed to enhance the performance of SPMP context switches.
 
-* For RV64 architectures, it introduces a 64-bit WARL CSR, `spmpen`.
-* For RV32 architectures, it introduces an additional 32-bit WARL CSR, `spmpenh`, which serves as an alias for the upper 32 bits of the `spmpen` register.
+* [#norm:ext:Sspmpen:spmpen_readwrite]#For RV64 architectures, it introduces a 64-bit WARL CSR, `spmpen` (CSR number 0x183).#
+* [#norm:ext:Sspmpen:spmpenh_alias]#For RV32 architectures, it introduces an additional 32-bit WARL CSR, `spmpenh` (CSR number 0x193), which serves as an alias for the upper 32 bits of the `spmpen` register.#
```
aswaterman: We don't need the CSR numbers here; putting them in `priv-csrs.adoc` suffices.

src/sspmp.adoc
```diff
@@ -293,7 +299,7 @@ While `SFENCE.VMA` and `HFENCE.GVMA` instructions normally order preceding store
 A failed SPMP check triggers an exception whose type corresponds to the memory operation (load, store/AMO, or instruction fetch).
 Each fault type is assigned a distinct exception code.
 
-The `Sspmp` extension co-opts the existing page fault exception codes for SPMP violations, as page fault handling is typically already delegated to S-mode.
+The `Sspmp` extension co-opts the existing page fault exception codes for SPMP violations, as page fault exceptions are commonly delegated to S-mode.
```
aswaterman: This isn't quite right: the fact that page-fault exceptions are commonly delegated isn't the main reason for this design.  Also, "co-opts" is odd for a technical spec.  I suggest:

The `Sspmp` extension reuses the existing page fault exception codes for SPMP violations.

src/sspmp.adoc
```diff
@@ -244,12 +249,13 @@ The following table dictates which isolation mechanism is active based on the sa
...
 Each value of `siselect` maps to a corresponding set of SPMP CSRs.
-`sireg` is used to access the `spmpaddr` register, while `sireg2` is used for the `spmpcfg` register.
-The registers `sireg3` through `sireg6` are reserved.
+[#norm:csr:siselect:spmpaddr_mapping]#`sireg` is used to access the `spmpaddr` register,# while [#norm:csr:siselect:spmpcfg_mapping]#`sireg2` is used for the `spmpcfg` register.#
+The registers `sireg3` through `sireg6` are reserved (i.e., reads return zero, writes are ignored).
 
 SPMP entries are indexed starting from zero.
-In a system with 48 SPMP entries, for example, S-mode can address `SPMP[0..47]` using `siselect#0..47`.
-An access to an out-of-bounds index via `siselect` will return zero on read and be ignored on write.
+For instance, in a system with N SPMP entries (asciimath:[N \le 64]),
```
aswaterman: If we are writing general text (using "N" instead of a constant), we should not use "For instance".  "For instance" suggests this is just an example, but in fact this covers all cases.

There are several other places where this should also be fixed.

src/sspmp.adoc
```diff
@@ -198,13 +203,13 @@ The semantics of `sstatus.SUM` within SPMP are consistent with its definition in
...
-. The lowest-numbered SPMP entry that matches any byte of an access (indicated by an address and the accessed length) determines whether that access is allowed or denied.
+. [#norm:ext:Sspmp:match_priority]#The lowest-numbered SPMP entry that matches any byte of an access (indicated by an address and the accessed length) determines whether that access is allowed or denied.#
 +
 . This matching SPMP entry must match *all* bytes of the access, or the access fails and an instruction, load, or store page-fault exception is generated (see <<exceptions>>).
 +
-. This matching is done irrespective of the SHARED, U, R, W, and X bits.
+. [#norm:ext:Sspmp:match_irrespective_perm_bits]#This matching is done irrespective of the SHARED, U, R, W, and X bits.#
 +
-. If the effective privilege mode of the access is S/U and no SPMP entry matches, but at least one SPMP entry is implemented, the access is `denied`.
+. [#norm:ext:Sspmp:no_match_deny]#If the effective privilege mode of the access is S-mode or U-mode and no SPMP entry matches, but at least one SPMP entry is implemented, the access is `denied`.#
```
aswaterman: You can just write "S or U" instead of "S-mdoe or U-mode" because you already used the word "mode" earlier in the sentence.  We use the same style elsewhere in the document.

src/sspmp.adoc
```diff
@@ -1,12 +1,17 @@
 [[sspmp]]
-== Supervisor-Level PMP Extensions, Version 1.0
-SPMP is designed to enforce memory isolation in environments where supervisor mode is implemented but virtual memory, managed by a Memory Management Unit (MMU), is not.
-Processors based on the RISC-V architecture are gaining significant traction in the burgeoning Internet of Things (IoT) and automotive sectors.
-In these domains, however, virtual memory is frequently avoided to comply with stringent constraints on system resources and execution latency.
-Consequently, in the absence of an MMU, effectively isolating S-mode operating systems (such as an RTOS) from U-mode applications presents a significant challenge.
-The SPMP extension adopts an architecture and programming model analogous to the existing Physical Memory Protection (PMP) standard.
-It equips an S-mode OS with the capability to guarantee secure processing and fault isolation for U-mode software.
-This protection is achieved by restricting the physical memory regions that U-mode software can access on a specific hart.
+== Supervisor-Level PMP Extensions, Version 0.9.1
```
aswaterman: Delete the version number.  When the PR is merged into master, it will be the final version.  As a stylistic choice, we no longer include the version number of ratified extensions by default.

normative_rule_defs/sspmp.yaml
ybc-alkaid: Hi @jordancarlin, thanks for your review, the PR is updated.

ybc-alkaid:
Hi @aswaterman, thanks for your comments, the PR is updated.

src/sspmp.adoc
jordancarlin: Seems like this should be a normative rule.

jordancarlin: Normative rule

jordancarlin: Normative rule with implementation defined tag.

src/sspmp.adoc
```diff
@@ -60,8 +65,8 @@ These registers collectively define a protected physical memory region and its a
 ====
 
 The SPMP address registers, named `spmpaddr0` through `spmpaddr63`, share the same layout as the M-mode PMP architecture.
-On RV32 systems, each `spmpaddr` register encodes a 34-bit physical address from bit 33 down to bit 2, as illustrated in <<spmpaddr-rv32>>.
-On RV64 systems, each `spmpaddr` register encodes a 56-bit physical address from bit 55 down to bit 2, as shown in <<spmpaddr-rv64>>.
+On RV32 systems, [#norm:spmpaddr_format_rv32]#each `spmpaddr` register encodes a 34-bit physical address from bit 33 down to bit 2, as illustrated in <<spmpaddr-rv32>>.#
+On RV64 systems, [#norm:spmpaddr_format_rv64]#each `spmpaddr` register encodes a 56-bit physical address from bit 55 down to bit 2, as shown in <<spmpaddr-rv64>>.#
```
jordancarlin: I suggest including the "On RVXX systems" part of the sentence in the normative rule so that the rule stands on its own more completely.

src/sspmp.adoc
```diff
@@ -60,8 +65,8 @@ These registers collectively define a protected physical memory region and its a
...
 An implementation may support fewer address bits on systems with a smaller physical address space.
 All writable SPMP entries should implement the same number of address bits.
```
jordancarlin: Is this "should" intended to mean implementations have to follow this rule? If so, can we change it to say "must" to avoid ambiguity? Assuming this is a "must", it should also be a normative rule.

src/sspmp.adoc
```diff
@@ -60,8 +65,8 @@ These registers collectively define a protected physical memory region and its a
...
 Since not all physical address bits must be implemented, the SPMP address registers are considered WARL, with exceptions defined by granularity rules.
```
jordancarlin: Normative rule indicating they are WARL.

src/sspmp.adoc
```diff
@@ -87,15 +92,15 @@ Every SPMP entry contains an SXLEN-bit configuration register, `spmpcfg[i]`.
 <<spmpcfg>> illustrates the layout of `spmpcfg[i]`.
 Permission rules and their encodings are detailed in <<encoding>>.
 
-. The R, W, and X bits govern permissions for read, write, and instruction execution, respectively.
-. The A field, which defines the address-matching mode, is detailed in <<address-matching>>.
+. [#norm:spmpcfg_rwx_bits]#The R, W, and X bits govern permissions for read, write, and instruction execution, respectively.#
+. [#norm:spmpcfg_a_field]#The A field, which defines the address-matching mode, is detailed in <<address-matching>>.#
 . Bits 5 and 6 are reserved for future standard use.
```
jordancarlin: Normative rule

src/sspmp.adoc
```diff
@@ -136,11 +141,11 @@ For comprehensive details, refer to <<pmp-address-matching>>.
 
 For a rule to be valid, its `spmpcfg[i].A` field must not be OFF.
 Furthermore, if `spmpcfg[i].A` is set to TOR, the condition `spmpaddr[i-1] < spmpaddr[i]` must hold.
-Particularly, if `spmpcfg[0].A` is set to TOR, zero is used for the lower bound.
+[#norm:addr_match_tor_entry0]#Particularly, if `spmpcfg[0].A` is set to TOR, zero is used for the lower bound.#
```
jordancarlin: If this is normative (which is implied by the normative rule) then it should not be in a note (which are generally considered to be non-normative).

src/sspmp.adoc
```diff
@@ -156,21 +161,21 @@ Because the `spmpcfg[i].A` field is WARL, an implementation is free to hardwire
...
-. The encoding `spmpcfg.SHARED == 1 and spmpcfg.U == 1` defines a *Shared-Region* rule. For shared regions, the state of the `sstatus.SUM` bit is irrelevant.
+. [#norm:shared_rule_sum_ignored]#The encoding `spmpcfg.SHARED == 1 and spmpcfg.U == 1` defines a *Shared-Region* rule. For shared regions, the state of the `sstatus.SUM` bit is irrelevant.#
 +
-. A *Shared-Region* rule is *enforced* for both Supervisor and User modes, but it imposes the constraint that read and write permissions for User mode are mutually exclusive.
+. [#norm:shared_rule_enforce]#A *Shared-Region* rule is *enforced* for both Supervisor and User modes, but it imposes the constraint that read and write permissions for User mode are mutually exclusive.#
 +
 . The R, W, and X bits collectively form a WARL field.
 +
-. The encoding `spmpcfg.RWX=010`, `spmpcfg.RWX=011`, and the combination `spmpcfg.SHARED == 1 and spmpcfg.U == 0` are all reserved for future standard use.
+. The encodings [#norm:rwx_010_reserved]#`spmpcfg.RWX=010`#, [#norm:rwx_011_reserved]#`spmpcfg.RWX=011`#, as well as the combination `spmpcfg.SHARED == 1 and spmpcfg.U == 0` are all reserved for future standard use.
```
jordancarlin: Make this a single normative rule.

src/sspmp.adoc
```diff
@@ -198,13 +203,13 @@ The semantics of `sstatus.SUM` within SPMP are consistent with its definition in
 ==== Matching Logic
 . SPMP entries are statically prioritized.
 +
-. The lowest-numbered SPMP entry that matches any byte of an access (indicated by an address and the accessed length) determines whether that access is allowed or denied.
+. [#norm:match_priority]#The lowest-numbered SPMP entry that matches any byte of an access (indicated by an address and the accessed length) determines whether that access is allowed or denied.#
 +
 . This matching SPMP entry must match *all* bytes of the access, or the access fails and an instruction, load, or store page-fault exception is generated (see <<exceptions>>).
```
jordancarlin: Normative rule

src/sspmp.adoc
```diff
@@ -244,12 +249,13 @@ The following table dictates which isolation mechanism is active based on the sa
...
-`sireg` is used to access the `spmpaddr` register, while `sireg2` is used for the `spmpcfg` register.
+[#norm:siselect_spmpaddr_mapping]#`sireg` is used to access the `spmpaddr` register,# while [#norm:siselect_spmpcfg_mapping]#`sireg2` is used for the `spmpcfg` register.#
 The registers `sireg3` through `sireg6` are reserved.
 
 SPMP entries are indexed starting from zero.
-In a system with 48 SPMP entries, for example, S-mode can address `SPMP[0..47]` using `siselect#0..47`.
-An access to an out-of-bounds index via `siselect` will return zero on read and be ignored on write.
+In a system with N SPMP entries (`N {le} 64`),
+`SPMP[0..N-1]` is accessed by writing `0x100+i` to the `siselect` register, where `i` represents the entry index (`0 {le} i {le} N-1`).
+[#norm:siselect_oob_read_zero]#An access to an out-of-bounds index via `siselect` will return zero on read# and [#norm:siselect_oob_write_ignored]#be ignored on write.#
```
jordancarlin: Either combine these into one normative rule or include both tags in the `siselect_oob_write_ignored` rule in the YAML file. Otherwise, the rule will just say "be ignored on write", which does not give enough context.

src/sspmp.adoc
```diff
@@ -273,10 +279,10 @@ For further details on indirect CSR access, refer to <<indirect-csr>>.
...
 Indirect accesses to SPMP CSRs are not ordered with respect to each other or
 with subsequent memory accesses.
-To enforce ordering for subsequent accesses whose effective privilege mode is
+[#norm:sfence_vma_ordering]#To enforce ordering for subsequent accesses whose effective privilege mode is
 S or U, software must execute an `SFENCE.VMA` instruction with `rs1=x0` and
 `rs2=x0`, which synchronizes subsequent memory accesses with all preceding
-SPMP CSR writes.
+SPMP CSR writes.#
 To enforce ordering for subsequent accesses whose effective privilege mode is
 VS or VU, an `HFENCE.GVMA` instruction with `rs1=x0` and `rs2=x0` is required.
```
jordancarlin: Normative rule

src/sspmp.adoc
```diff
@@ -307,9 +313,9 @@ Table of exception codes:
...
 |Interrupt|Exception Code|Description
-|0|12|Instruction page fault
-|0|13|Load page fault
-|0|15|Store/AMO page fault
+|0|12|[#norm:spmp_instr_page_fault]#Instruction page fault#
+|0|13|[#norm:spmp_load_page_fault]#Load page fault#
+|0|15|[#norm:spmp_store_page_fault]#Store/AMO page fault#
 |0|20|Instruction guest-page fault
 |0|21|Load guest-page fault
 |0|23|Store/AMO guest-page fault
```
jordancarlin: Why are only the non-hypervisor exceptions tagged as normative rules? Also, I think the more useful normative rule is in the preceding paragraphs. This table just lists the standard page fault exceptions which are already tagged elsewhere.

src/sspmp.adoc
```diff
@@ -325,31 +331,31 @@ Since, when V=1, SPMP is only active when hgatp.MODE=Bare, SPMP protection and G
 [[Sspmpen-extension]]
 === "Sspmpen" Extension for Optimizing Context Switching of SPMP Entries
 
-In RV64, a context switch for the SPMP mechanism involves updating as many as 64 address registers and configuration registers.
+In RV64, a context switch for the SPMP mechanism involves updating up to 64 address registers and 64 configuration registers.
 The `Sspmpen` extension, introduced in this chapter, is an optional feature designed to enhance the performance of SPMP context switches.
 
-* For RV64 architectures, it introduces a 64-bit WARL CSR, `spmpen`.
-* For RV32 architectures, it introduces an additional 32-bit WARL CSR, `spmpenh`, which serves as an alias for the upper 32 bits of the `spmpen` register.
+* [#norm:spmpen_readwrite]#For RV64 architectures, it introduces a 64-bit WARL CSR, `spmpen`.#
```
jordancarlin:
```suggestion
* [#norm:spmpen_warl]#For RV64 architectures, it introduces a 64-bit WARL CSR, `spmpen`.#
```

src/sspmp.adoc
```diff
@@ -325,31 +331,31 @@ Since, when V=1, SPMP is only active when hgatp.MODE=Bare, SPMP protection and G
...
+[#norm:spmpen_activation_condition]#An SPMP entry `i` is considered active only when both the `spmpen[i]` bit is set and the `spmpcfg[i].A` field is non-zero.#
+The `spmpen[i]` bit controls only whether SPMP entry `i` participates in address matching and enforcement; it does not control whether `spmpaddr[i]` or `spmpcfg[i]` is writable.
 
-When an entry `i` is locked, as indicated by `spmpcfg[i].L == 1`, its corresponding `spmpen[i]` bit becomes read-only.
+[#norm:spmpen_locked_readonly]#When an entry `i` is locked, as indicated by `spmpcfg[i].L == 1`, its corresponding `spmpen[i]` bit becomes read-only.#
 
-In an implementation featuring 64 PMP entries with 48 delegated to the supervisor level, the bits `spmpen[0..47]` control their respective SPMP entries `[0..47]`.
-Any attempt to write to the upper bits of the register, `spmpen[48..63]` are ignored.
+In an implementation featuring 64 PMP entries with N (`N {le} 64`) delegated to the supervisor level, the bits `spmpen[0..N-1]` control their respective SPMP entries `[0..N-1]`.
+Any attempt to write to the upper bits of the register are ignored.
```
jordancarlin: Normative rule

src/sspmp.adoc
```diff
@@ -371,23 +377,24 @@ The `spmpen` register offers significant benefits for context switch optimizatio
 The similar architecture of PMP and SPMP registers, including their shared address-matching logic, makes hardware reuse a practical approach for resource conservation.
 This chapter introduces the `Smpmpdeleg` extension, a mechanism that allows hardware resources to be dynamically allocated between PMP and SPMP.
 
-*This extension is mandatory for implementations that support Sspmp in conjunction with M-mode (i.e., `Sm1p13`).*
-To streamline the specification and reduce optional features, the `Smpmpdeleg` extension implements 64 PMP entries, but fewer can be writable; any non-writable entries behave as read-only zero.
+*This extension is mandatory for implementations that support both `Sspmp` and `Sm1p13`.*
+The `Smcsrind` extension for indirect CSR access must be implemented.
+For the purposes of this specification, `Smpmpdeleg` assumes an architectural space of 64 PMP entries, although fewer entries may be writable; any non-writable entries behave as read-only zero.
```
jordancarlin: Assumes or requires? If requires, then this should be a normative rule.

src/sspmp.adoc
```diff
@@ -371,23 +377,24 @@ The `spmpen` register offers significant benefits for context switch optimizatio
...
 
 
 
 [[pmp-entry-sharing]]
 ==== Resource Sharing between PMP and SPMP
 
-The `Smpmpdeleg` extension facilitates the delegation of PMP entries for use by S-level, thereby creating S-level PMP (SPMP) entries.
+The `Smpmpdeleg` extension facilitates the delegation of PMP entries for use by S-mode, thereby creating SPMP entries.
 This delegation is managed by an MXLEN-bit M-mode CSR named `mpmpdeleg`, whose layout is detailed in <<mpmpdeleg-format-rv64>>.
```
jordancarlin: See Table 1 of https://github.com/riscv/riscv-isa-manual/blob/main/tagging_normative_rules.adoc#naming-normative-rules. In addition to the fields below having normative rules, the CSR itself should have a normative rule. In this case, the primary purpose of that rule would be to indicate that this is an M-mode CSR.

src/sspmp.adoc
```diff
@@ -371,23 +377,24 @@ The `spmpen` register offers significant benefits for context switch optimizatio
...
 
-. A key component of this CSR is the `pmpnum` WARL field, which defines the starting index for delegation.
-All PMP entries with an index equal to or greater than `pmpnum` are delegated as SPMP entries.
+. [#norm:mpmpdeleg_pmpnum_field]#A key component of this CSR is the `pmpnum` WARL field, which defines the starting index for delegation.
+All PMP entries with an index equal to or greater than `pmpnum` are delegated as SPMP entries.#
 . If a write to `pmpnum` specifies a value exceeding the number of writable PMP entries, the field subsequently reads back the total count of writable entries. In such a case, no SPMP entries are delegated.
```
jordancarlin: This is a specific constraint on WARL behavior, so it should be a normative rule.

src/sspmp.adoc
```diff
@@ -404,11 +411,11 @@ The `mpmpdeleg.pmpnum` field is a WARL field, which allows an implementation to
...
+In an implementation with 64 total entries where `pmpnum` is configured to N (`N {le} 64`):
 
-. PMP entries 0 through 15 act as PMP (i.e., `PMP[0..15]`) and are accessible via standard PMP CSRs (i.e., `pmpcfg[0..3]` and `pmpaddr[0..15]` for RV32; `pmpcfg[0,2]` and `pmpaddr[0..15]` for RV64).
-. The remaining 48 entries are delegated as SPMP (i.e., `SPMP[0..47]`) and are indirectly accessed via `xiselect` (see <<access-method>> and <<m-mode-indirect-access>>).
-. Accesses to out-of-range indices, such as reading `PMP[16]` or writing to `SPMP[48]` in this scenario, are handled as follows: reads return zero, and writes are ignored.
+. PMP entries 0 through N-1 act as PMP (i.e., `PMP[0..N-1]`) and are accessible via standard PMP CSRs (e.g., `pmpcfg[0..3]` and `pmpaddr[0..15]` for RV32; `pmpcfg[0,2]` and `pmpaddr[0..15]` for RV64).
+. The remaining 64-N entries are delegated as SPMP (i.e., `SPMP[0..63-N]`) and are indirectly accessed via `xiselect` (see <<access-method>> and <<m-mode-indirect-access>>).
+. Accesses to out-of-range indices, such as reading `PMP[N]` or writing to `SPMP[64-N]` in this scenario, are handled as follows: reads return zero, and writes are ignored.
```
jordancarlin: Normative rule.

src/sspmp.adoc
```diff
@@ -420,14 +427,17 @@ Its lower 8 bits are an alias for the 8-bit field in the corresponding M-mode PM
 *Reconfiguration of delegated entries:*
 
 . M-mode software can dynamically adjust the allocation between PMP and SPMP by writing to the `mpmpdeleg` CSR.
-. The `pmpnum` value cannot be set to an index that is less than or equal to that of any locked PMP entry.
-For instance, if `PMP[7]` is locked, any attempt to write a value less than 8 to `pmpnum` is ignored, and the field retains its prior value.
+. [#norm:mpmpdeleg_locked_pmp_constraint]#The `pmpnum` value cannot be set to an index that is less than or equal to that of any locked PMP entry.
+For example, if `PMP[7]` is locked, any attempt to write a value less than 8 to `pmpnum` is ignored, and the field retains its prior value.#
 . The `pmpnum` value can be set to override locked SPMP entries. For example, if `SPMP[0]` is locked, M-mode software can still increment `pmpnum`.
+. As `pmpnum` is reconfigured, SPMP entries with the largest indices change.
```
jordancarlin: Normative rules.

src/sspmp.adoc
```diff
@@ -458,7 +469,7 @@ Any access attempt by either mode to an index outside this range (i.e., `i >= 48
 
 
 *Memory access in M-mode:*
-If the effective privilege mode of the memory access is M, the access is `allowed` regardless of the SPMP permissions.
+[#norm:mmode_mem_access_bypasses_spmp]#If the effective privilege mode of the memory access is M, the access is `allowed` regardless of the SPMP permissions.#
 
 If the Shbare extension is implemented in conjunction with `Sspmp`, then when V=1 and hgatp.MODE=Bare, the Hypervisor Virtual Machine Load and Store instructions (HLV, HLVX, HSV) executed from M-mode are subject to SPMP checks.
```
jordancarlin: Seems like a normative rule?

jordancarlin (changes requested):
Great to see normative rules included with new specifications! As this is one of the first new specs to include normative rules, I have a number of comments below.

ybc-alkaid:
Hi @jordancarlin , many thanks for your review. The PR is revised according to your feedback.

src/priv/sspmp.adoc
```diff
@@ -273,12 +280,12 @@ For further details on indirect CSR access, refer to <<indirect-csr>>.
 
 Indirect accesses to SPMP CSRs are not ordered with respect to each other or
 with subsequent memory accesses.
-To enforce ordering for subsequent accesses whose effective privilege mode is
+[#norm:sfence_vma_ordering]#To enforce ordering for subsequent accesses whose effective privilege mode is
 S or U, software must execute an `SFENCE.VMA` instruction with `rs1=x0` and
```
aswaterman, line 285: The Sspmp spec should say that the Sspmp extension includes the SFENCE.VMA instruction.

Furthermore, the M-mode chapter has a sentence about when the TVM bit is writable.  It should be modified to say that the TVM bit is "read-only 0 when S-mode is not supported, or if satp.MODE is read-only 0 and the Sspmp extension is not implemented; otherwise, it must be writable".

However, please rebase on the HEAD of main before editing the M-mode chapter.  Otherwise, there will be major merge conflicts.

ybc-alkaid:
Thanks @aswaterman , the PR is revised accordingly. 

PS: #3225 was opened against #2573’s head branch because I cannot push directly to riscv:ybc-alkaid-Sspmp. After rebasing onto main, #3225 now contains the complete SPMP proposal. You can update #2573 by force-updating its head branch from ybc-alkaid:pr-2573.

rpsene:
Hi @ybc-alkaid,

The `signing_check` job on this PR is failing because riscv-isa-manual requires every commit in a pull request to carry a verified cryptographic signature, and all 19 commits on `pr-2573` are currently reported as `unsigned`. Note that this is not the DCO `Signed-off-by` trailer, it is an actual GPG or SSH signature that GitHub can verify.

Since the commits are already pushed, signing them means rewriting the branch and force-pushing. Here is the full sequence.

### 1. Create a signing key and register it on GitHub (one time)

SSH signing is the simplest route if you already use an SSH key with GitHub:

```bash
# reuse an existing key, or create one:
ssh-keygen -t ed25519 -C "your-github-email@example.com"
```

Then go to https://github.com/settings/keys, click **New SSH key**, paste the contents of `~/.ssh/id_ed25519.pub`, and set **Key type: Signing Key**. This is a separate entry from your authentication key, even when it is the same key material.

### 2. Configure git to sign (one time)

```bash
git config --global gpg.format ssh
git config --global user.signingkey ~/.ssh/id_ed25519.pub
git config --global commit.gpgsign true
```

### 3. Sign the 19 commits already on the branch

```bash
git checkout pr-2573
git rebase --exec 'git commit --amend --no-edit -n -S' HEAD~19
```

Using `HEAD~19` rewrites only your own commits and keeps the same base, so there is nothing to resolve. If you would rather also rebase onto the current `main`, replace `HEAD~19` with `upstream/main`.

### 4. Force-push

```bash
git push --force-with-lease origin pr-2573
```

The checks re-run automatically after the push and `signing_check` should turn green.

### Two things worth knowing

1. Six of the commits are authored by @aswaterman. The rebase preserves the Author field, so his authorship is untouched. You are already the committer on all 19 commits, so you are only adding your own signature to commits you already committed. Nothing gets misattributed.
2. The email address in your commits must be a verified address on your GitHub account, otherwise GitHub still shows the commits as "Unverified" even with a valid signature. You can review your addresses at https://github.com/settings/emails.

GitHub's reference: https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

This applies to this pull request only. Your open PRs in riscv-unified-db and riscv-arch-test do not run this check.

Let me know if you run into trouble and I will help.

ybc-alkaid:
Hi @rpsene , thanks for the detailed guide, I've updated the PR.

PR #3232, opened 2026-07-06
Add missing AIA CSRs to priv/csrs.adoc

topperc:
Add the supervisor, hypervisor, and VS CSRs from https://github.com/riscv/riscv-aia/blob/main/src/CSRs.adoc

I also moved mtopi and mtopei from Machine Trap Setup to Machine Trap Handling. That made more sense to me after reading the descriptions. I can change it back and move stopi and stopei if the old location was correct.

### DIFF
--- a/src/priv/csrs.adoc
+++ b/src/priv/csrs.adoc
@@ -204,6 +204,7 @@ Note that not all registers are required on all implementations.
 |`0x104` |SRW |csr:sie[]        |Supervisor interrupt-enable register
 |`0x105` |SRW |csr:stvec[]      |Supervisor trap handler base address
 |`0x106` |SRW |csr:scounteren[] |Supervisor counter enable
+|`0x114` |SRW |csr:sieh[]       |Upper qty:32[bits] of csr:sie[], RV32 only
 
 4+^|Supervisor Configuration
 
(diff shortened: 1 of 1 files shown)


src/priv/csrs.adoc
```diff
@@ -325,12 +338,16 @@ Note that not all registers are required on all implementations.
...
 |`0x204` |HRW |csr:vsie[]      |Virtual supervisor interrupt-enable register
 |`0x205` |HRW |csr:vstvec[]    |Virtual supervisor trap handler base address
+|`0x214` |HRW |csr:vsieh[]     |Upper qty:32[bits] of csr:vsie[], RV32 only
 |`0x240` |HRW |csr:vsscratch[] |Virtual supervisor scratch register
 |`0x241` |HRW |csr:vsepc[]     |Virtual supervisor exception program counter
 |`0x242` |HRW |csr:vscause[]   |Virtual supervisor trap cause
 |`0x243` |HRW |csr:vstval[]    |Virtual supervisor trap value
 |`0x244` |HRW |csr:vsip[]      |Virtual supervisor interrupt pending
+|`0x254` |HRW |csr:vsiph[]     |Upper qty:32[bits] of csr:vsip[], RV32 only
+|`0x25C` |HRW |csr:vstopei[]   |Virtual supervisor top external interrupt (only with an IMSIC)
```
aswaterman: Can you delete the parenthetical (both here and above) since it's stylistically inconsistent?

Thanks for the PR.

topperc: Done

aswaterman (approved):
Thanks again!

PR #3233, opened 2026-07-06
Add Zvzip extension and integrate with normative-rule pipeline

ptomsich:
This PR adds the draft Zvzip extension chapter to the unprivileged manual: five instructions for interleaving, deinterleaving, and pairwise transposition of vector elements (vzip.vv, vunzipe.v, vunzipo.v, vpaire.vv, vpairo.vv), targeting workloads such as complex-number packing, pixel-component (de)interleaving, and small-matrix transposition. It supersedes the earlier draft, which defined the count-changing instructions over an effective vector length of `2*vl`.

### DIFF
--- a/src/unpriv/zvzip.adoc
+++ b/src/unpriv/zvzip.adoc
@@ -1,0 +1,381 @@
+=== "Zvzip" Extension for Reordering Structured Data, Version 0.1
+
+This chapter describes the Zvzip standard extension for reordering structured
+data in vector registers. These instructions address usages such as packing and
+unpacking data structures such as color components of a pixel, real and
+imaginary components of complex numbers, transposing small matrices, among
+others.
+
+[%autowidth]
+[%header,cols="2,4"]
+|===
+|Mnemonic      |Instruction
+| vzip.vv      | <<insns-vzip>>
+| vunzipe.v    | <<insns-vunzipe>>
+| vunzipo.v    | <<insns-vunzipo>>
+| vpaire.vv    | <<insns-vpaire>>
+| vpairo.vv    | <<insns-vpairo>>
+|===
+
+The Zvzip Extension depends on the Zve32x Extension.
+
+<<<
+
+[[insns-vzip, Vector Zip]]
+==== Vector Zip Instruction
+
+Synopsis::
+
+Interleave elements from source vector register groups into destination vector
+register groups.
+
+Mnemonic::
+
+vzip.vv vd, vs2, vs1, vm
+
+Encoding::
+
+[wavedrom, , svg]
+....
+{reg:[
+  {bits: 7, name: 'OP-V'},
+  {bits: 5, name: 'vd'},
+  {bits: 3, name: 'OPMVV'},
+  {bits: 5, name: 'vs1'},
+  {bits: 5, name: 'vs2'},
+  {bits: 1, name: 'vm'},
+  {bits: 6, name: '111110'},
+], config:{lanes: 1, hspace:1024}}
+....
+
+Description::
+
+Vector Zip (VZIP) instruction interleaves elements from two source vector
+register groups (`vs2` and `vs1`) into one destination vector register group
+(`vd`) by alternating elements from the two sources.                           +
+                                                                               +
+For destination element index `i`, if `i` is even then `vd[i] = vs2[i/2]`, and
+if `i` is odd then `vd[i] = vs1[i/2]`.                                         +
+                                                                               +
+Equivalently, the result order is:
+`vd = [vs2[0], vs1[0], vs2[1], vs1[1], ... ]`                                  +
+                                                                               +
+This instruction operates with an effective vector length (EVL) of 2*VL. The
+destination EMUL is 2xLMUL. The instruction is reserved when LMUL is 8.
+Prestart, inactive, and tail element handling follows the standard vector
+rules, applied over the EVL.                                                   +
+                                                                               +
+The destination vector register group may overlap the source vector register
+group if the overlap is in the highest-numbered part of the destination
+register group and the source EMUL is at least 1. If the overlap violates these
+constraints, the instruction encoding is reserved.
+
+Operation::
+
+[source,sail]
+--
+function clause execute (VZIP(vs2, vs1, vd, vm)) = {
+  EVL = 2 * VL;
+  foreach (i from vstart to EVL-1) {
+    let j   = i / 2;
+    let op1 = get_velem(vs1, SEW, j);
+    let op2 = get_velem(vs2, SEW, j);
+    let res = if (i % 2 == 0) then op2 else op1;
+    if (vm == 0b1) | (v0[i] == 0b1) then
+      set_velem(vd, EEW=SEW, i, res);
+    // inactive element handling follows VMA
+  }
+  // tail element handling follows VTA
+  RETIRE_SUCCESS
+}
+--
+
+<<<
+
+[[insns-vunzipe, Vector Unzip Even]]
+==== Vector Unzip Even Instruction
+
+Synopsis::
+
+Extract even-indexed elements from source vector register group into the
+destination vector register group.
+
+Mnemonic::
+
+vunzipe.v vd, vs2, vm
+
+Encoding::
+
+[wavedrom, , svg]
+....
+{reg:[
+  {bits: 7, name: 'OP-V'},
+  {bits: 5, name: 'vd'},
+  {bits: 3, name: 'OPMVV'},
+  {bits: 5, name: '01011'},
+  {bits: 5, name: 'vs2'},
+  {bits: 1, name: 'vm'},
+  {bits: 6, name: '010010'},
+], config:{lanes: 1, hspace:1024}}
+....
+
+Description::
+
+The vector unzip-even instruction (VUNZIPE) extracts VL even-indexed elements
+from the source vector register group into the destination vector register
+group.                                                                         +
+                                                                               +
+This instruction accesses 2*VL elements in the source vector register group and
+the source EMUL is 2xLMUL. The instruction is reserved when LMUL is 8.         +
+                                                                               +
+Prestart, inactive, and tail element handling follow the standard vector
+rules and are defined over the destination element indices (`0` to `VL-1`).    +
+                                                                               +
+The destination vector register group may overlap the source vector register
+group only if the overlap is in the lowest-numbered part of the source register
+group. If the overlap violates these constraints, the instruction encoding is
+reserved.
+
+Operation::
+
+[source,sail]
+--
+function clause execute (VUNZIPE(vs2, vd, vm)) = {
+  foreach (i from vstart to VL-1) {
+    let j = i * 2;
+    if (vm == 0b1) | (v0[i] == 0b1) then
+      set_velem(vd, EEW=SEW, i, get_velem(vs2, SEW, j));
+    // inactive element handling follows VMA
+  }
+  // tail element handling follows VTA
+  RETIRE_SUCCESS
+}
+--
+
+<<<
+
+[[insns-vunzipo, Vector Unzip Odd]]
+==== Vector Unzip Odd Instruction
+
+Synopsis::
+
+Extract odd-indexed elements from source vector register group into the
+destination vector register group.
+
+Mnemonic::
+
+vunzipo.v vd, vs2, vm
+
+Encoding::
+
+[wavedrom, , svg]
+....
+{reg:[
+  {bits: 7, name: 'OP-V'},
+  {bits: 5, name: 'vd'},
+  {bits: 3, name: 'OPMVV'},
+  {bits: 5, name: '01111'},
+  {bits: 5, name: 'vs2'},
+  {bits: 1, name: 'vm'},
+  {bits: 6, name: '010010'},
+], config:{lanes: 1, hspace:1024}}
+....
+
+Description::
+
+The vector unzip-odd instruction (VUNZIPO) extracts VL odd-indexed elements
+from the source vector register group into the destination vector register
+group.                                                                         +
+                                                                               +
+This instruction accesses 2*VL elements in the source vector register group and
+the source EMUL is 2xLMUL. The instruction is reserved when LMUL is 8.         +
+                                                                               +
+Prestart, inactive, and tail element handling follow the standard vector
+rules and are defined over the destination element indices (`0` to `VL-1`).    +
+                                                                               +
+The destination vector register group may overlap the source vector register
+group only if the overlap is in the lowest-numbered part of the source register
+group. If the overlap violates these constraints, the instruction encoding is
+reserved.
+
+Operation::
+
+[source,sail]
+--
+function clause execute (VUNZIPO(vs2, vd, vm)) = {
+  foreach (i from vstart to VL-1) {
+    let j = (i * 2) + 1;
+    if (vm == 0b1) | (v0[i] == 0b1) then
+      set_velem(vd, EEW=SEW, i, get_velem(vs2, SEW, j));
+    // inactive element handling follows VMA
+  }
+  // tail element handling follows VTA
+  RETIRE_SUCCESS
+}
+--
+
+<<<
+
+[[insns-vpaire, Vector Pair Even]]
+==== Vector Pair Even Instruction
+
+Synopsis::
+
+Interleave the even-indexed elements of the source vector register groups into
+the destination vector register group.
+
+Mnemonic::
+
+vpaire.vv vd, vs2, vs1, vm
+
+
+Encoding::
+
+[wavedrom, , svg]
+....
+{reg:[
+  {bits: 7, name: 'OP-V'},
+  {bits: 5, name: 'vd'},
+  {bits: 3, name: 'OPIVV'},
+  {bits: 5, name: 'vs1'},
+  {bits: 5, name: 'vs2'},
+  {bits: 1, name: 'vm'},
+  {bits: 6, name: '001111'},
+], config:{lanes: 1, hspace:1024}}
+....
+
+Description::
+
+The vector pair-even instruction (VPAIRE) interleaves the even-indexed
+elements of the source vector register groups into the destination vector
+register group.                                                                +
+                                                                               +
+For destination element index `i`, if `i` is even then `vd[i] = vs2[i]`, and if
+`i` is odd then `vd[i] = vs1[i - 1]`.                                          +
+                                                                               +
+Equivalently, the result order is:
+`vd = [vs2[0], vs1[0], vs2[2], vs1[2], ... ]`                                  +
+                                                                               +
+The destination vector register group cannot overlap the source vector register
+groups and, if masked, cannot overlap the mask register. If the overlap
+violates these constraints, the instruction encoding is reserved.              +
+                                                                               +
+Prestart, inactive, and tail element handling follow the standard vector rules.
+
+Operation::
+
+[source,sail]
+--
+function clause execute (VPAIRE(vs2, vs1, vd, vm)) = {
+  foreach (i from vstart to VL-1) {
+    let j = if (i % 2) == 0 then i else (i - 1);
+    let res = if (i % 2) == 0
+              then get_velem(vs2, SEW, j)
+              else get_velem(vs1, SEW, j);
+    if (vm == 0b1) | (v0[i] == 0b1) then
+      set_velem(vd, EEW=SEW, i, res);
+    // inactive element handling follows VMA
+  }
+  // tail element handling follows VTA
+  RETIRE_SUCCESS
+}
+--
+
+<<<
+
+[[insns-vpairo, Vector Pair Odd]]
+==== Vector Pair Odd Instruction
+
+Synopsis::
+
+Interleave the odd-indexed elements of the source vector register groups into
+the destination vector register group.
+
+Mnemonic::
+
+vpairo.vv vd, vs2, vs1, vm
+
+
+Encoding::
+
+[wavedrom, , svg]
+....
+{reg:[
+  {bits: 7, name: 'OP-V'},
+  {bits: 5, name: 'vd'},
+  {bits: 3, name: 'OPMVV'},
+  {bits: 5, name: 'vs1'},
+  {bits: 5, name: 'vs2'},
+  {bits: 1, name: 'vm'},
+  {bits: 6, name: '001111'},
+], config:{lanes: 1, hspace:1024}}
+....
+
+Description::
+
+The vector pair-odd instruction (VPAIRO) interleaves the odd-indexed
+elements of the source vector register groups into the destination vector
+register group.                                                                +
+                                                                               +
+For destination element index `i`, if `i` is even then `vd[i] = vs2[i + 1]`, and
+if `i` is odd then `vd[i] = vs1[i]`.                                           +
+                                                                               +
+Equivalently, the result order is:
+`vd = [vs2[1], vs1[1], vs2[3], vs1[3], ... ]`                                  +
+                                                                               +
+The destination vector register group cannot overlap the source vector register
+groups and, if masked, cannot overlap the mask register. If the overlap
+violates these constraints, the instruction encoding is reserved.              +
+                                                                               +
+Prestart, inactive, and tail element handling follow the standard vector rules.
+                                                                               +
+VPAIRO may read one element past `VL` in `vs2` when `VL` is odd. If the element
+index is greater than or equal to VLMAX in the source vector register group,
+the value 0 is returned for that element.
+
+Operation::
+
+[source,sail]
+--
+function clause execute (VPAIRO(vs2, vs1, vd, vm)) = {
+  foreach (i from vstart to VL-1) {
+    let j = if (i % 2) == 0 then (i + 1) else i;
+    let res =
+      if (j >= vlmax) then zeros()
+      else if (i % 2) == 0 then get_velem(vs2, SEW, j)
+      else                      get_velem(vs1, SEW, j);
+    if (vm == 0b1) | (v0[i] == 0b1) then
+      set_velem(vd, EEW=SEW, i, res);
+    // inactive element handling follows VMA
+  }
+  // tail element handling follows VTA
+  RETIRE_SUCCESS
+}
+--
+
+<<<
+
+[NOTE]
+====
+
+The following example illustrates use of the vector pair-even and pair-odd to
+transpose VL/4 independent 4x4 matrices packed across vector registers.
+
+The first stage operates on 32-bit elements. The second stage packs adjacent
+pairs into 64-bit elements to complete the transpose.
+
+----
+vsetivli t0, zero, e32, m1, ta, ma
+vpaire.vv v5, v1, v2 #|a|b|c|d|A|B|C|D|..    |a|e|c|g|A|E|C|G|..
+vpairo.vv v6, v1, v2 #|e|f|g|h|E|F|G|H|.. -> |b|f|d|h|B|F|D|H|..
+vpaire.vv v7, v3, v4 #|i|j|k|l|I|J|K|L|..    |i|m|k|o|I|M|K|O|..
+vpairo.vv v8, v3, v4 #|m|n|o|p|M|N|O|P|..    |j|n|l|p|J|N|L|P|..
+
+vsetivli t0, zero, e64, m1, ta, ma
+vpaire.vv v1, v5, v7 #|a e|c g|A E|C G|..    |a e|i m|A E|I M|..
+vpaire.vv v2, v6, v8 #|b f|d h|B F|D H|.. -> |b f|j n|B F|J N|..
+vpairo.vv v3, v5, v7 #|i m|k o|I M|K O|..    |c g|k o|C G|K O|..
+vpairo.vv v4, v6, v8 #|j n|l p|J N|L P|..    |d h|l p|D H|L P|..
+----
+
+====
--- a/src/unpriv/zvzip.adoc
+++ b/src/unpriv/zvzip.adoc
@@ -1,23 +1,114 @@
-=== "Zvzip" Extension for Reordering Structured Data, Version 0.1
+[[ext:zvzip]]
+=== ext:zvzip[] Vector Extension for Reordering Structured Data, Version 0.3
 
-This chapter describes the Zvzip standard extension for reordering structured
-data in vector registers. These instructions address usages such as packing and
-unpacking data structures such as color components of a pixel, real and
-imaginary components of complex numbers, transposing small matrices, among
-others.
+The ext:zvzip[] extension provides vector instructions for reordering
+structured data in vector registers: interleaving and deinterleaving
+element sequences, and pairwise transposition of elements from two
+vector register groups. These instructions address usages such as
+packing and unpacking data structures (e.g., the color components of a
+pixel, or the real and imaginary components of complex numbers) and
+transposing small matrices.
+
+NOTE: This extension is in *Draft* status and has not been ratified.
+The instruction encodings shown in this chapter are provisional,
+pending codepoint allocation, and are subject to change.
 
 [%autowidth]
 [%header,cols="2,4"]
 |===
-|Mnemonic      |Instruction
-| vzip.vv      | <<insns-vzip>>
-| vunzipe.v    | <<insns-vunzipe>>
-| vunzipo.v    | <<insns-vunzipo>>
-| vpaire.vv    | <<insns-vpaire>>
-| vpairo.vv    | <<insns-vpairo>>
+|Mnemonic                     |Instruction
+| vzip.vv vd, vs2, vs1, vm    | <<insns-vzip>>
+| vunzipe.v vd, vs2           | <<insns-vunzipe>>
+| vunzipo.v vd, vs2           | <<insns-vunzipo>>
+| vpaire.vv vd, vs2, vs1, vm  | <<insns-vpaire>>
+| vpairo.vv vd, vs2, vs1, vm  | <<insns-vpairo>>
 |===
 
-The Zvzip Extension depends on the Zve32x Extension.
+[#norm:Zvzip_dependent_Zve32x]#The ext:zvzip[] extension depends upon
+the ext:zve32x[] extension.#
+
+[#norm:Zvzip_eew]#All operands of all ext:zvzip[] instructions have
+EEW=SEW. insn:vpaire.vv[], insn:vpairo.vv[], insn:vunzipe.v[], and
+insn:vunzipo.v[] support all SEW settings supported by the
+implementation; for insn:vzip.vv[], SEW support is additionally
+subject to the SEW/LMUL constraint stated with that instruction.#
+
+[#norm:Zvzip_diel]#If the ext:zvkt[] extension is implemented, all
+ext:zvzip[] instructions shall be executed with data-independent
+execution latency.#
+
+NOTE: Data-reordering instructions are natural building blocks for
+constant-time cryptographic kernels, where they replace
+insn:vrgather.vv[] sequences. Since insn:vrgather.vv[] and the slide
+instructions are on the ext:zvkt[] instruction list, code migrating to
+ext:zvzip[] instructions must not lose the data-independent-latency
+guarantee.
+
+==== Mixed-EMUL operation
+
+insn:vzip.vv[], insn:vunzipe.v[], and insn:vunzipo.v[] convert between
+an _interleaved_ element sequence and a pair of _deinterleaved_
+element sequences of half the length.
+[#norm:Zvzip_operand_emul]#For all ext:zvzip[] instructions,
+csr:vtype[] and csr:vl[] describe the destination: `vl` gives the
+number of destination elements, and the destination is a vector
+register group with EMUL=LMUL. The source operands of insn:vzip.vv[]
+are vector register groups with EMUL=LMUL/2; the source operand of
+insn:vunzipe.v[] and insn:vunzipo.v[] is a vector register group with
+EMUL=2*LMUL.#
+
+[#norm:Zvzip_operand_align]#The standard vector register alignment
+rules apply to each operand according to its EMUL.#
+
+[NOTE]
+====
+For every ext:zvzip[] instruction, `vl` gives the number of elements
+written to the destination, matching the base specification's
+definition of csr:vl[], and masking, prestart, body, and tail element
+handling follow the standard vector rules, applied to the destination
+element indices. The unzip instructions read source elements at
+indices up to `2*vl-1`, following the pattern of insn:vslidedown[]
+and insn:vrgather[], whose source reads are likewise not bounded by
+`vl`.
+
+Two alternative formulations were considered. Defining the family as
+widening and narrowing operations on element pairs, with one pair as
+a single element of EEW=2*SEW, would require EEW=128 for SEW=64, the
+one case for which no base-vector emulation sequence exists; the
+vector cryptography extension faced the same question and adopted
+element groups rather than element widths beyond ELEN. Splitting each
+instruction into a pair of same-length low/high (or even/odd) halves
+would double the instruction count for full-width reordering.
+====
+
+insn:vpaire.vv[] and insn:vpairo.vv[] do not change the number of
+elements: all their operands have EMUL=LMUL, and the standard vector
+rules apply without modification.
+
+The base specification's universal rules are unaffected; in
+particular, the destination of a masked insn:vzip.vv[],
+insn:vpaire.vv[], or insn:vpairo.vv[] cannot overlap the mask
+register `v0`, and such encodings remain reserved as specified by the
+base vector specification.
+
+The illegal-instruction exceptions specified in this chapter for
+unsupported SEW/LMUL settings, register-group overlap violations, and
+`vm=0` on the unzip instructions are a deliberate strengthening
+relative to the base vector specification, which treats such
+encodings as reserved; the vector cryptography extension provides
+precedent for mandating an illegal-instruction exception, but only
+for its unsupported-EMUL condition, not for register-group overlap.
+
+In the operation blocks in this chapter, `handle_illegal()` raises an
+illegal-instruction exception for an illegal encoding, following the
+convention of the vector cryptography chapter, and
+`vreg_groups_overlap(x, ex, y, ey)` is true when the vector register
+group starting at register `x` with EMUL=`ex` and the vector register
+group starting at register `y` with EMUL=`ey` share at least one
+vector register. `LMUL` denotes the exact (possibly fractional) value
+of the csr:vtype[] LMUL setting, and arithmetic on it is exact
+rational arithmetic; a vector register group whose EMUL is less than
+or equal to 1 occupies a single vector register.
 
 <<<
 
@@ -26,14 +117,17 @@ The Zvzip Extension depends on the Zve32x Extension.
 
 Synopsis::
 
-Interleave elements from source vector register groups into destination vector
-register groups.
+Interleave the elements of two source vector register groups into one
+destination vector register group of twice the EMUL.
 
 Mnemonic::
 
-vzip.vv vd, vs2, vs1, vm
+insn:vzip.vv[vd,vs2,vs1,vm]
 
 Encoding::
+
+[#norm:vzip-vv_enc]#insn:vzip.vv[] is encoded in the OP-V major opcode
+with funct3=OPMVV and funct6=0b111110.#
 
 [wavedrom, , svg]
 ....
@@ -50,43 +144,69 @@ Encoding::
 
 Description::
 
-Vector Zip (VZIP) instruction interleaves elements from two source vector
-register groups (`vs2` and `vs1`) into one destination vector register group
-(`vd`) by alternating elements from the two sources.                           +
-                                                                               +
-For destination element index `i`, if `i` is even then `vd[i] = vs2[i/2]`, and
-if `i` is odd then `vd[i] = vs1[i/2]`.                                         +
-                                                                               +
-Equivalently, the result order is:
-`vd = [vs2[0], vs1[0], vs2[1], vs1[1], ... ]`                                  +
-                                                                               +
-This instruction operates with an effective vector length (EVL) of 2*VL. The
-destination EMUL is 2xLMUL. The instruction is reserved when LMUL is 8.
-Prestart, inactive, and tail element handling follows the standard vector
-rules, applied over the EVL.                                                   +
-                                                                               +
-The destination vector register group may overlap the source vector register
-group if the overlap is in the highest-numbered part of the destination
-register group and the source EMUL is at least 1. If the overlap violates these
-constraints, the instruction encoding is reserved.
+The vector zip instruction interleaves the elements of two source
+vector register groups (`vs2` and `vs1`, each with EMUL=LMUL/2) into
+the destination vector register group (`vd`, with EMUL=LMUL) by
+alternating elements from the two sources.
+
+[#norm:vzip-vv_op]#For destination element index `i`, if `i` is even
+then `vd[i] = vs2[i/2]`, and if `i` is odd then `vd[i] = vs1[i/2]`.#
+Equivalently, the result order is
+`vd = [vs2[0], vs1[0], vs2[1], vs1[1], ...]`.
+
+The destination is the interleaved operand: it has `vl` body elements,
+and the source element indices read are less than `ceil(vl/2)` in
+`vs2` and less than `floor(vl/2)` in `vs1`. Masking, prestart,
+inactive, and tail element handling follow the standard vector rules,
+applied to the destination element indices.
+
+NOTE: When `vl` is odd, `vs2` supplies one element more than `vs1`.
+Software must ensure that `vs2` holds at least `ceil(vl/2)` valid
+elements (e.g., by producing it with an AVL of `ceil(vl/2)`);
+otherwise `vd[vl-1]` reflects a tail element of the instruction that
+produced `vs2`.
+
+[#norm:vzip-vv_emul_constr]#insn:vzip.vv[] raises an
+illegal-instruction exception when `2*SEW > LMUL * min(ELEN, VLEN)`,
+i.e., when a source operand with EMUL=LMUL/2 cannot support EEW=SEW.
+This condition includes LMUL=1/8 for every SEW.#
+
+[#norm:vzip-vv_vreg_constr]#A source vector register group of
+insn:vzip.vv[] may overlap the destination vector register group only
+if the source EMUL is at least 1 and the highest-numbered vector
+register in the source vector register group is also the
+highest-numbered vector register in the destination vector register
+group. Otherwise, an overlapping encoding raises an
+illegal-instruction exception.#
+
+NOTE: This is the same overlap rule that applies to the destination of
+widening vector arithmetic instructions, restated for a widened
+destination of the same EEW.
 
 Operation::
 
 [source,sail]
 --
 function clause execute (VZIP(vs2, vs1, vd, vm)) = {
-  EVL = 2 * VL;
-  foreach (i from vstart to EVL-1) {
-    let j   = i / 2;
-    let op1 = get_velem(vs1, SEW, j);
-    let op2 = get_velem(vs2, SEW, j);
-    let res = if (i % 2 == 0) then op2 else op1;
-    if (vm == 0b1) | (v0[i] == 0b1) then
-      set_velem(vd, EEW=SEW, i, res);
-    // inactive element handling follows VMA
+  let src_EMUL = LMUL / 2;
+  let vs2_overlap_ok = not(vreg_groups_overlap(vs2, src_EMUL, vd, LMUL))
+                       | ((src_EMUL >= 1) & (vs2 + src_EMUL == vd + LMUL));
+  let vs1_overlap_ok = not(vreg_groups_overlap(vs1, src_EMUL, vd, LMUL))
+                       | ((src_EMUL >= 1) & (vs1 + src_EMUL == vd + LMUL));
+  if (2 * SEW > LMUL * min(ELEN, VLEN))
+     | not(vs2_overlap_ok) | not(vs1_overlap_ok)
+  then { handle_illegal(); RETIRE_FAIL }  // illegal encoding
+  else {
+    foreach (i from vstart to vl - 1) {
+      let src = if (i % 2 == 0) then vs2 else vs1;
+      let res = get_velem(src, SEW, i / 2);  // source EMUL is LMUL/2
+      if (vm == 0b1) | (v0.mask[i] == 0b1) then
+        set_velem(vd, SEW, i, res);
+      // inactive elements follow the vma policy
+    };
+    // tail elements (indices >= vl) follow the vta policy
+    RETIRE_SUCCESS
   }
-  // tail element handling follows VTA
-  RETIRE_SUCCESS
 }
 --
 
@@ -97,14 +217,18 @@ function clause execute (VZIP(vs2, vs1, vd, vm)) = {
 
 Synopsis::
 
-Extract even-indexed elements from source vector register group into the
-destination vector register group.
+Extract the even-indexed elements of a source vector register group
+of twice the EMUL into the destination vector register group.
 
 Mnemonic::
 
-vunzipe.v vd, vs2, vm
+insn:vunzipe.v[vd,vs2]
 
 Encoding::
+
+[#norm:vunzipe-v_enc]#insn:vunzipe.v[] is encoded in the OP-V major
+opcode with funct3=OPMVV, funct6=0b010010, the value 0b01011 in the
+vs1 field, and vm=1.#
 
 [wavedrom, , svg]
 ....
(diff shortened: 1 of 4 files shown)


aswaterman:
Thanks, @ptomsich.  I'll mark the PR as draft until these are ratified.

src/unpriv/zvzip.adoc
```diff
@@ -0,0 +1,713 @@
...
+If `vstart >= evl`, no destination elements are updated.#
+
+NOTE: Unlike insn:vlm.v[], whose effective vector length likewise
+differs from `vl` but whose destination is always written with a
+tail-agnostic policy, the unzip instructions handle destination tail
+elements according to the vta setting.
+
+[#norm:vunzipe-v_vm_constr]#insn:vunzipe.v[] is encoded as an
+unmasked instruction (`vm=1`). The masked encoding (`vm=0`) raises an
+illegal-instruction exception.#
```
nibrunieAtSi5, line 285: In that case, I think it would make more sense to defined this encoding as being reserved rather than forcing it to raise an illegal instruction exception (this latter behavior can still be obtained with the `Ssstrict` extension).

nibrunieAtSi5 (commented):
I think some of the "illegal instruction exception" conditions listed in this draft should be better of defined as "reserved" (in particular the masked encoding and the overlap).

wangpc-pp:
I thought `Zvzip` has frozen for a while but this PR makes some huge semantics changes. :-(

ptomsich:
> I thought `Zvzip` has frozen for a while but this PR makes some huge semantics changes. :-(

Zvzip never went to Internal Review (which is before Freeze).
It's being pushed here only to generate the artifacts (i.e., PDF) for the Internal Review.

ptomsich:
> I thought `Zvzip` has frozen for a while but this PR makes some huge semantics changes. :-(

Summary of the rationale for turning the evl-calculation upside down:
- the earlier version violated norm:vl_op without a carve-out (only the whole-register loads/stores had such a carve-out and did so by bypassing vl altogether); this brings all operations in alignment with norm:vl_op
- the earlier version generated an anomaly around vstart (using it in SAIL to indicate that these instructions were restartable) potentiall producing vstart > VL (through evl), which conflicts with norm:vstart_val_rsv
- the precedent of vlm.v was adjusting avl downward (so the elements written/read can be less than vl, but not more)

I hope that explains the direction of changes.

wangpc-pp:
> > I thought `Zvzip` has frozen for a while but this PR makes some huge semantics changes. :-(
> 
> Summary of the rationale for turning the evl-calculation upside down:
> 
> * the earlier version violated norm:vl_op without a carve-out (only the whole-register loads/stores had such a carve-out and did so by bypassing vl altogether); this brings all operations in alignment with norm:vl_op
> * the earlier version generated an anomaly around vstart (using it in SAIL to indicate that these instructions were restartable) potentiall producing vstart > VL (through evl), which conflicts with norm:vstart_val_rsv
> * the precedent of vlm.v was adjusting avl downward (so the elements written/read can be less than vl, but not more)
> 
> I hope that explains the direction of changes.

Make sense to me. I am working on the revised LLVM support now. Is there any risk that Zvzip will have anther big change after this? I hope we can freeze it ASAP.

src/unpriv/zvzip.adoc
ptomsich: We put this explicitly in to drive the discussion whether to instead repurpose vm=0/vm=1 to differentiate between even and odd.  That would put the encoding at VXUNARY0, same vs1 for both (i.e. vunzip*), and vm=0/vm=1 being the even/odd (vunzipe/vunzipo) selector.  There is a NOTE in the document to point toward such an 'alternative' encoding.

This is the sort of question that ARC will have to settle either during Internal Review or during ARC review.
Keep in mind: encodings are not final.

ptomsich:
> > > I thought `Zvzip` has frozen for a while but this PR makes some huge semantics changes. :-(
> > 
> > 
> > Summary of the rationale for turning the evl-calculation upside down:
> > 
> > * the earlier version violated norm:vl_op without a carve-out (only the whole-register loads/stores had such a carve-out and did so by bypassing vl altogether); this brings all operations in alignment with norm:vl_op
> > * the earlier version generated an anomaly around vstart (using it in SAIL to indicate that these instructions were restartable) potentiall producing vstart > VL (through evl), which conflicts with norm:vstart_val_rsv
> > * the precedent of vlm.v was adjusting avl downward (so the elements written/read can be less than vl, but not more)
> > 
> > I hope that explains the direction of changes.
> 
> Make sense to me. I am working on the revised LLVM support now. Is there any risk that Zvzip will have anther big change after this? I hope we can freeze it ASAP.

The goal is to send this for Internal Review this week (~2 weeks).
Any feedback from Internal Review may upset the direction again, before this going to ARC Review.
I'd expect at least encoding changes to come out of ARC (unless they want to relitigate the `evl`/`vstart`-discussion).
Only then, can we go to freeze.

At the very least, you should expect the encodings and "what makes an illegal instruction" to change.

wangpc-pp:
One possible problem found when implementing the new spec: we set `vl/type` on interleaved operand now, which causes more `vset(i)vli` instructions for `unzip`. For example, if we want to `unzip` a `<2 x i64>` (`v10`) from `<4 x i64>` (`v8`) and store:
* Before (v0.1):
```asm
vsetivli zero, 2, e64, m1, ta, ma
vunzipe.v v10, v8
vse64.v v10, (a1)
```
* After (v0.2):
```asm
vsetivli zero, 4, e64, m2, ta, ma
vunzipe.v v10, v8
vsetivli zero, 2, e64, m1, ta, ma
vse64.v v10, (a1)
```
However, for `zip`, the new spec reduces `vset(i)vli` instructions. For example, if we want to `zip` a `<4 x i64>` (`v12`) from two `<2 x i64>` (`v8` and `v10`) and store:
* Before (v0.1):
```asm
vsetivli zero, 2, e64, m1, ta, ma
vzip.vv v12, v8, v10
vsetvli zero, 4, e64, m2, ta, ma
vse64.v v12, (a1)
```
* After (v0.2):
```asm
vsetvli zero, 4, e64, m2, ta, ma
vzip.vv v12, v8, v10
vse64.v v12, (a1)
```

----

So I think instead of setting `vl/vtype` on interleaved operand, we may consider to set `vl/vtype` on result (`vd`)?

rpsene:
@ptomsich, the failing check here is the **Commit Signing Check**, which requires commits to be **cryptographically signed** (GPG or SSH, so they show as "Verified" on GitHub). It is separate from the DCO `Signed-off-by` line, which both of your commits already have, so this is not a sign-off problem.

Both commits currently show as unverified (`unknown_key`):
- `c9d1b9b` "Add Zvzip extension..." (Ved Shanbhogue, author)
- `39fb18a` "Zvzip: redefine count-changing..." (you)

Since you own the branch, the simplest fix is to re-sign both commits with your own registered key and force-push. Ved stays the author (and keeps his Signed-off-by) on the first commit, so credit is preserved:

1. Register a signing key at github.com/settings/keys (GPG or SSH).
2. `git config commit.gpgsign true` (for SSH, also set `gpg.format ssh` and `user.signingkey`).
3. `git rebase --exec "git commit --amend --no-edit -S" HEAD~2 && git push --force-with-lease`

That should turn both commits "Verified" and clear the check. Let me know if you'd like a hand.

camel-cdr:
I'd like to repeat [my earlier request](https://github.com/riscv/integrated-matrix-extension/issues/58) to add notes on how to implement the new instructions with base RVV instructions.

Finding the best patterns is non-trivial, as can be seen by the [IME draft spec using sub optimal ones](https://github.com/riscv/integrated-matrix-extension/issues/58).

Here is the overview again:
```
vzip.vv vd, vs1, vs2 -> vwaddu.vv vd, vs1, vs2; vwmaccu.vx vd, -1, vs2

vunzipe.v vd, vs1 -> vnsrl.vx vd, vs1, 0
vunzipo.v vd, vs1 -> vnsrl.vx vd, vs1, \SEW/8

vpaire.v vd, vs1, vd -> vslide1up.vx   vd, vs1, 0, vm=odd-bits
vpairo.v vd, vs1, vd -> vslide1down.vx vd, vs1, 0, vm=even-bits
```

ptomsich:
> One possible problem found when implementing the new spec: we set `vl/type` on interleaved operand now, which causes more `vset(i)vli` instructions for `unzip`. For example, if we want to `unzip` a `<2 x i64>` (`v10`) from `<4 x i64>` (`v8`) and store:
> 
> * Before (v0.1):
> 
> ```assembly
> vsetivli zero, 2, e64, m1, ta, ma
> vunzipe.v v10, v8
> vse64.v v10, (a1)
> ```
> 
> * After (v0.2):
> 
> ```assembly
> vsetivli zero, 4, e64, m2, ta, ma
> vunzipe.v v10, v8
> vsetivli zero, 2, e64, m1, ta, ma
> vse64.v v10, (a1)
> ```
> 
> However, for `zip`, the new spec reduces `vset(i)vli` instructions. For example, if we want to `zip` a `<4 x i64>` (`v12`) from two `<2 x i64>` (`v8` and `v10`) and store:
> 
> * Before (v0.1):
> 
> ```assembly
> vsetivli zero, 2, e64, m1, ta, ma
> vzip.vv v12, v8, v10
> vsetvli zero, 4, e64, m2, ta, ma
> vse64.v v12, (a1)
> ```
> 
> * After (v0.2):
> 
> ```assembly
> vsetvli zero, 4, e64, m2, ta, ma
> vzip.vv v12, v8, v10
> vse64.v v12, (a1)
> ```
> 
> So I think instead of setting `vl/vtype` on interleaved operand, we may consider to set `vl/vtype` on result (`vd`)?

`vzip` is the same either way: its `vd` is the interleaved operand, so "`vl` on `vd`" only changes the two `unzip`s.

On the `vset(i)vli` count: the extra toggle only shows up in register-to-register code. If the interleaved data comes from memory, the load already needs the 2n/m2 setting, so both variants use two `vset(i)vli` per loop iteration. So the cost is one extra instruction, in reg-to-reg unzip sequences like your example. On cores that rename `vtype`, that toggle should be cheap.

To be fair, you proposal has a certain charme:
- norm:vl_op ("vl specifies the number of elements to be updated with results") would hold as written for all five instructions.
- vstart would be in the same element space as vl and the destination. The two vstart rules and the whole evl mechanism in the current draft, including the ceil(vl/2)/floor(vl/2) asymmetry, would just go away.
- The even/odd bookkeeping moves into the AVL computation, which is arguably where it belongs.

Both variants have precedent in the ratified spec:
- The current unzips work like vcompress: vl counts the consumed operand, the destination gets a smaller, derived body, and "the remaining elements of vd are treated as tail elements according to the current tail policy". Ours is the same shape with a static count instead of a data-dependent one, and the same unmasked-only encoding.
- "vl on vd" works like vrgather/vslidedown: vl bounds the destination, and the source may be read "at any index < VLMAX regardless of csr:vl[]".

The difference between the two is the reason we changed the orientation in the first place. `vrgather` and `vslidedown` *read past vl only when software asks for it* (through indices or an offset) and each has a normative sentence saying so.
`vcompress` *never reads past vl* at all.
A vd-oriented unzip would read source elements up to
  index 2*vl-1 every time, as a built-in side effect. No ratified instruction does that.

With the current spec, forgetting the second `vsetvli` produces an oversized store: visible, and often a trap on group misalignment. With "vl on vd", a loop that loads n elements and then unzips reads n more elements of stale register content, silently. And the unzip source becomes a 2×LMUL group, wider than vtype implies, with the legality corners split across the family (LMUL=8 excluded for the unzips, small LMUL excluded for vzip).

A count-changing instruction always has a factor of 2 somewhere; the only choice is where to put it, and whether  forgetting it fails (potentially) "with a bang" (i.e., traps) or silently. We put it on the consumer side of the unzips because that side is less likely to fail silently.
If the group cares more about compiled code size (a compiler never forgets the vsetvli; the silent failure mostly bites hand-written code), then "vl on vd" is workable. It would need a vrgather-style sentence allowing the past-vl source reads, and it brings back potential stale-data issues (flagged as "further inspection is needed" by the security folks on our team).

ptomsich:
> Philipp Tomsich (@ptomsich), the failing check here is the **Commit Signing Check**, which requires commits to be **cryptographically signed** (GPG or SSH, so they show as "Verified" on GitHub). It is separate from the DCO `Signed-off-by` line, which both of your commits already have, so this is not a sign-off problem.
> 
> Both commits currently show as unverified (`unknown_key`):
> 
> * `c9d1b9b` "Add Zvzip extension..." (Ved Shanbhogue, author)
> * `39fb18a` "Zvzip: redefine count-changing..." (you)
> 
> Since you own the branch, the simplest fix is to re-sign both commits with your own registered key and force-push. Ved stays the author (and keeps his Signed-off-by) on the first commit, so credit is preserved:
> 
> 1. Register a signing key at github.com/settings/keys (GPG or SSH).
> 2. `git config commit.gpgsign true` (for SSH, also set `gpg.format ssh` and `user.signingkey`).
> 3. `git rebase --exec "git commit --amend --no-edit -S" HEAD~2 && git push --force-with-lease`
> 
> That should turn both commits "Verified" and clear the check. Let me know if you'd like a hand.

It was the 'register a signing key' step that was missing — the commits had been signed, but Github didn't know the the key was trusted.

src/unpriv/zvzip.adoc
nibrunieAtSi5: One of the issue I see with this, is that if we reused the encoding but with `vm=1` for another extension, and that Zvzip is also supported, as written the specification would make this impossible (since it would mandate raising the illegal instruction exception).

Strengthening "reserved" to illegal instruction exception may make sense in some cases, but I do not think this is one of those cases.

QiuJing-Damo:
> I'd like to repeat [my earlier request](https://github.com/riscv/integrated-matrix-extension/issues/58) to add notes on how to implement the new instructions with base RVV instructions.
> 
> Finding the best patterns is non-trivial, as can be seen by the [IME draft spec using sub optimal ones](https://github.com/riscv/integrated-matrix-extension/issues/58).
> 
> Here is the overview again:
> 
> ```
> vzip.vv vd, vs1, vs2 -> vwaddu.vv vd, vs1, vs2; vwmaccu.vx vd, -1, vs2
> 
> vunzipe.v vd, vs1 -> vnsrl.vx vd, vs1, 0
> vunzipo.v vd, vs1 -> vnsrl.vx vd, vs1, \SEW/8
> 
> vpaire.v vd, vs1, vd -> vslide1up.vx   vd, vs1, 0, vm=odd-bits
> vpairo.v vd, vs1, vd -> vslide1down.vx vd, vs1, 0, vm=even-bits
> ```

I also agree that vunzipe and vunzipo may not be strictly necessary, since they can be expressed reasonably well using existing base RVV instructions.

However, I believe vpaire and vpairo are very important and should be kept. These two instructions are critical parts of the codec encode/decode flow, and in this context it is very important to complete the operation with as few instructions as possible. Reducing the instruction count here has a direct impact on efficiency.I don't think slide can replace the pair operations.
image
image

The importance of vpaire/vpairo can also be seen from the data I presented at the US Summit last year. Please refer to the figure for the relevant results.

QiuJing-Damo:
> > I'd like to repeat [my earlier request](https://github.com/riscv/integrated-matrix-extension/issues/58) to add notes on how to implement the new instructions with base RVV instructions.

To make the discussion more concrete, let's focus on x264's SATD 4×4 DCT as a representative example. This is a workload accelerated by the zip extension alone — no other custom extensions required. The column interleave in the Hadamard transform is where vpaire/vpairo earns its keep.

<head></head><h2 id="two-implementations" data-pm-slice="1 1 []" style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-style: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">Two Implementations</h2><h3 id="scheme-a-—-vpaire-/-vpairo-(58-instructions-total,-8-for-interleave)" style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-style: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">Scheme A —<span class="Apple-converted-space"> </span><code>vpaire</code><span class="Apple-converted-space"> </span>/<span class="Apple-converted-space"> </span><code>vpairo</code><span class="Apple-converted-space"> </span>(58 instructions total, 8 for interleave)</h3><pre data-language="c" style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;"><code>void x264_8_sub4x4_dct_rvv(int16_t dct[16], uint8_t *pix1, uint8_t *pix2)
{
    size_t gvl;
    asm volatile(
        /* ---- Load 4x4 pixels (e8, mf8) ---- */
        "vsetvli    %[gvl], zero, e8, mf8, ta, ma           \n\t"
        "vle8.v     v0,  (%[pix1])                           \n\t"
        "vle8.v     v1,  (%[pix2])                           \n\t"
        "addi       %[pix1], %[pix1], 16                     \n\t"
        "addi       %[pix2], %[pix2], 32                     \n\t"
        "vle8.v     v2,  (%[pix1])                           \n\t"
        "vle8.v     v3,  (%[pix2])                           \n\t"
        "addi       %[pix1], %[pix1], 16                     \n\t"
        "addi       %[pix2], %[pix2], 32                     \n\t"
        "vle8.v     v4,  (%[pix1])                           \n\t"
        "vle8.v     v5,  (%[pix2])                           \n\t"
        "addi       %[pix1], %[pix1], 16                     \n\t"
        "addi       %[pix2], %[pix2], 32                     \n\t"
        "vle8.v     v6,  (%[pix1])                           \n\t"
        "vle8.v     v7,  (%[pix2])                           \n\t"
        /* ---- Widen subtract (e8 → e16) ---- */
        "vwsubu.vv  v8,  v0, v1                              \n\t"
        "vwsubu.vv  v9,  v2, v3                              \n\t"
        "vwsubu.vv  v10, v4, v5                              \n\t"
        "vwsubu.vv  v11, v6, v7                              \n\t"
        /* ---- Row butterfly (e16, mf4) ---- */
        "vsetvli    %[gvl], zero, e16, mf4, ta, ma           \n\t"
        "vadd.vv    v12, v8,  v11        /* s03 */            \n\t"
        "vadd.vv    v13, v9,  v10        /* s12 */            \n\t"
        "vsub.vv    v14, v8,  v11        /* d03 */            \n\t"
        "vsub.vv    v15, v9,  v10        /* d12 */            \n\t"
        "vadd.vv    v16, v12, v13                             \n\t"
        "vadd.vv    v17, v14, v15                             \n\t"
        "vsub.vv    v18, v12, v13                             \n\t"
        "vsub.vv    v19, v14, v15                             \n\t"
        "vadd.vv    v17, v17, v14                             \n\t"
        "vsub.vv    v19, v19, v15                             \n\t"
        /* ---- Column interleave Stage 1 (e32, mf2, VL=4) ---- */
        "vsetvli    %[gvl], zero, e32, mf2, ta, ma           \n\t"
        "vpaire.vv  v28, v18, v16                             \n\t"
        "vpairo.vv  v30, v18, v16                             \n\t"
        "vpaire.vv  v29, v19, v17                             \n\t"
        "vpairo.vv  v31, v19, v17                             \n\t"
        /* ---- Column interleave Stage 2 (e16, mf2, VL=8) ---- */
        "vsetvli    %[gvl], zero, e16, mf2, ta, ma           \n\t"
        "vpaire.vv  v8,  v29, v28                             \n\t"
        "vpairo.vv  v9,  v29, v28                             \n\t"
        "vpaire.vv  v10, v31, v30                             \n\t"
        "vpairo.vv  v11, v31, v30                             \n\t"
        /* ---- Column butterfly (e16, mf4) ---- */
        "vsetvli    %[gvl], zero, e16, mf4, ta, ma           \n\t"
        "vadd.vv    v12, v8,  v11        /* s03 */            \n\t"
        "vadd.vv    v13, v9,  v10        /* s12 */            \n\t"
        "vsub.vv    v14, v8,  v11        /* d03 */            \n\t"
        "vsub.vv    v15, v9,  v10        /* d12 */            \n\t"
        "vadd.vv    v16, v12, v13                             \n\t"
        "vadd.vv    v17, v14, v15                             \n\t"
        "vsub.vv    v18, v12, v13                             \n\t"
        "vsub.vv    v19, v14, v15                             \n\t"
        "vadd.vv    v17, v17, v14                             \n\t"
        "vsub.vv    v19, v19, v15                             \n\t"
        /* ---- Store ---- */
        "vse16.v    v16, (%[dct])                             \n\t"
        "addi       %[dct], %[dct], 8                         \n\t"
        "vse16.v    v17, (%[dct])                             \n\t"
        "addi       %[dct], %[dct], 8                         \n\t"
        "vse16.v    v18, (%[dct])                             \n\t"
        "addi       %[dct], %[dct], 8                         \n\t"
        "vse16.v    v19, (%[dct])                             \n\t"
        : [pix1]"+r"(pix1), [pix2]"+r"(pix2), [gvl]"=&amp;r"(gvl), [dct]"+r"(dct)
        : : "memory");
}</code></pre><hr style="font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;"><h3 id="scheme-b-—-vmv1r-+-vslide1up/vslide1down-(76-instructions-total,-25-for-interleave)" style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-style: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">Scheme B —<span class="Apple-converted-space"> </span><code>vmv1r</code><span class="Apple-converted-space"> </span>+<span class="Apple-converted-space"> </span><code>vslide1up</code>/<code>vslide1down</code><span class="Apple-converted-space"> </span>(76 instructions total, 25 for interleave)</h3><p style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">Uses the masked-slide pattern from your overview:</p><pre style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;"><code>vpaire.vv vd, vs2, vs1  →  vmv1r.v vd, vs1
                           vslide1up.vx vd, vs2, x1   (vm=odd-bits, mu)

vpairo.vv vd, vs2, vs1  →  vslide1down.vx vd, vs1, x1 (vm=even-bits, mu)
                           (result written back to vs1 — source destroyed)</code></pre><pre data-language="c" style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;"><code>void x264_8_sub4x4_dct_rvv_slide(int16_t dct[16], uint8_t *pix1, uint8_t *pix2)
{
    size_t gvl;
    size_t tmp;
    asm volatile(
        /* ---- Mask pattern setup (once, VL-independent) ---- */
        "addi       %[tmp], zero, 5                           \n\t"
        "vmv.v.x    v24, %[tmp]          /* v24 = 0x5 (even) */ \n\t"
        "addi       %[tmp], zero, 10                          \n\t"
        "vmv.v.x    v25, %[tmp]          /* v25 = 0xa (odd)  */ \n\t"
        "addi       %[tmp], zero, 1      /* shift = 1        */ \n\t"
        /* ---- Load 4x4 pixels (e8, mf8) ---- */
        "vsetvli    %[gvl], zero, e8, mf8, ta, ma           \n\t"
        "vle8.v     v0,  (%[pix1])                           \n\t"
        "vle8.v     v1,  (%[pix2])                           \n\t"
        "addi       %[pix1], %[pix1], 16                     \n\t"
        "addi       %[pix2], %[pix2], 32                     \n\t"
        "vle8.v     v2,  (%[pix1])                           \n\t"
        "vle8.v     v3,  (%[pix2])                           \n\t"
        "addi       %[pix1], %[pix1], 16                     \n\t"
        "addi       %[pix2], %[pix2], 32                     \n\t"
        "vle8.v     v4,  (%[pix1])                           \n\t"
        "vle8.v     v5,  (%[pix2])                           \n\t"
        "addi       %[pix1], %[pix1], 16                     \n\t"
        "addi       %[pix2], %[pix2], 32                     \n\t"
        "vle8.v     v6,  (%[pix1])                           \n\t"
        "vle8.v     v7,  (%[pix2])                           \n\t"
        /* ---- Widen subtract (e8 → e16) ---- */
        "vwsubu.vv  v8,  v0, v1                              \n\t"
        "vwsubu.vv  v9,  v2, v3                              \n\t"
        "vwsubu.vv  v10, v4, v5                              \n\t"
        "vwsubu.vv  v11, v6, v7                              \n\t"
        /* ---- Row butterfly (e16, mf4) ---- */
        "vsetvli    %[gvl], zero, e16, mf4, ta, ma           \n\t"
        "vadd.vv    v12, v8,  v11        /* s03 */            \n\t"
        "vadd.vv    v13, v9,  v10        /* s12 */            \n\t"
        "vsub.vv    v14, v8,  v11        /* d03 */            \n\t"
        "vsub.vv    v15, v9,  v10        /* d12 */            \n\t"
        "vadd.vv    v16, v12, v13                             \n\t"
        "vadd.vv    v17, v14, v15                             \n\t"
        "vsub.vv    v18, v12, v13                             \n\t"
        "vsub.vv    v19, v14, v15                             \n\t"
        "vadd.vv    v17, v17, v14                             \n\t"
        "vsub.vv    v19, v19, v15                             \n\t"
        /* ---- Column interleave Stage 1 (e32, mf2, VL=4) ---- */
        "vsetvli    %[gvl], zero, e32, mf2, ta, mu           \n\t"
        "vmv1r.v    v0,  v25              /* load odd mask   */ \n\t"
        /* vpaire(v28,v18,v16) */
        "vmv1r.v    v28, v18              /* ① copy src→dst  */ \n\t"
        "vslide1up.vx v28, v16, %[tmp]   /* ② slide odd pos */ \n\t"
        /* vpaire(v29,v19,v17) */
        "vmv1r.v    v29, v19                                  \n\t"
        "vslide1up.vx v29, v17, %[tmp]                       \n\t"
        "vmv1r.v    v0,  v24              /* load even mask  */ \n\t"
        /* vpairo → v16 */
        "vslide1down.vx v16, v18, %[tmp] /* ① write to vs1! */ \n\t"
        /* vpairo → v17 */
        "vslide1down.vx v17, v19, %[tmp] /* v16,v17 destroyed */ \n\t"
        /* ---- Column interleave Stage 2 (e16, mf2, VL=8) ---- */
        "vsetvli    %[gvl], zero, e16, mf2, ta, mu           \n\t"
        "vmv1r.v    v0,  v25                                  \n\t"
        "vmv1r.v    v8,  v29                                  \n\t"
        "vslide1up.vx v8, v28, %[tmp]                        \n\t"
        "vmv1r.v    v10, v17                                  \n\t"
        "vslide1up.vx v10, v16, %[tmp]                       \n\t"
        "vmv1r.v    v0,  v24                                  \n\t"
        "vslide1down.vx v28, v29, %[tmp]                     \n\t"
        "vslide1down.vx v16, v17, %[tmp]                     \n\t"
        /* ---- Column butterfly (e16, mf4) ---- */
        "vsetvli    %[gvl], zero, e16, mf4, ta, ma           \n\t"
        "vadd.vv    v12, v8,  v28        /* s03 */            \n\t"
        "vadd.vv    v13, v10, v16        /* s12 */            \n\t"
        "vsub.vv    v14, v8,  v28        /* d03 */            \n\t"
        "vsub.vv    v15, v10, v16        /* d12 */            \n\t"
        "vadd.vv    v16, v12, v13                             \n\t"
        "vadd.vv    v17, v14, v15                             \n\t"
        "vsub.vv    v18, v12, v13                             \n\t"
        "vsub.vv    v19, v14, v15                             \n\t"
        "vadd.vv    v17, v17, v14                             \n\t"
        "vsub.vv    v19, v19, v15                             \n\t"
        /* ---- Store ---- */
        "vse16.v    v16, (%[dct])                             \n\t"
        "addi       %[dct], %[dct], 8                         \n\t"
        "vse16.v    v17, (%[dct])                             \n\t"
        "addi       %[dct], %[dct], 8                         \n\t"
        "vse16.v    v18, (%[dct])                             \n\t"
        "addi       %[dct], %[dct], 8                         \n\t"
        "vse16.v    v19, (%[dct])                             \n\t"
        : [pix1]"+r"(pix1), [pix2]"+r"(pix2), [gvl]"=&amp;r"(gvl),
          [dct]"+r"(dct), [tmp]"=&amp;r"(tmp)
        : : "memory", "v0", "v24", "v25");
}</code></pre><hr style="font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;"><h2 id="focused-diff-—-column-interleave-only" style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-style: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">Focused Diff — Column Interleave Only</h2><blockquote style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;"><p>Red (<code>-</code>) = Scheme B (slide), Green (<code>+</code>) = Scheme A (paire/pairo)</p></blockquote><h3 id="stage-1-(e32,-mf2,-vl=4)" style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-style: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">Stage 1 (e32, mf2, VL=4)</h3><pre data-language="diff" style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;"><code>- vsetvli %[gvl], zero, e32, mf2, ta, mu   # must switch to mu
+ vsetvli %[gvl], zero, e32, mf2, ta, ma
- vmv1r.v v0,  v25               # load odd mask
- vmv1r.v v28, v18               # ① copy src→dst (vd ≠ vs1)
- vslide1up.vx v28, v16, %[tmp]  # ② slide odd positions
- vmv1r.v v29, v19
- vslide1up.vx v29, v17, %[tmp]
- vmv1r.v v0,  v24               # load even mask
- vslide1down.vx v16, v18, %[tmp] # ① write back to vs1 (v16 destroyed!)
- vslide1down.vx v17, v19, %[tmp]
- # → 10 instructions
+ vpaire.vv v28, v18, v16
+ vpairo.vv v30, v18, v16
+ vpaire.vv v29, v19, v17
+ vpairo.vv v31, v19, v17
+ # → 4 instructions</code></pre><h3 id="stage-2-(e16,-mf2,-vl=8)" style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-style: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">Stage 2 (e16, mf2, VL=8)</h3><pre data-language="diff" style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;"><code>- vsetvli %[gvl], zero, e16, mf2, ta, mu
+ vsetvli %[gvl], zero, e16, mf2, ta, ma
- vmv1r.v v0,  v25
- vmv1r.v v8,  v29
- vslide1up.vx v8, v28, %[tmp]
- vmv1r.v v10, v17
- vslide1up.vx v10, v16, %[tmp]
- vmv1r.v v0,  v24
- vslide1down.vx v28, v29, %[tmp]
- vslide1down.vx v16, v17, %[tmp]
- # → 10 instructions
+ vpaire.vv v8,  v29, v28
+ vpairo.vv v9,  v29, v28
+ vpaire.vv v10, v31, v30
+ vpairo.vv v11, v31, v30
+ # → 4 instructions</code></pre><hr style="font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;"><h2 id="what-the-one-line-mapping-misses" style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-style: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">What the One-Line Mapping Misses</h2><p style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">Your overview gives:</p><pre style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;"><code>vpaire.v vd, vs1, vd -&gt; vslidelup.vx vd, vs1, 0, vm=odd-bits
vpairo.v vd, vs1, vd -&gt; vslidedown.vx vd, vs1, 0, vm=even-bits</code></pre><p style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">This is correct when<span class="Apple-converted-space"> </span><code>vd = vs1</code><span class="Apple-converted-space"> </span>(or<span class="Apple-converted-space"> </span><code>vd = vs2</code>), but in practice the SATD code needs<span class="Apple-converted-space"> </span><strong>different destination registers</strong>:</p>
  | vpaire.vv vd, vs2, vs1 (general) | Your mapping (vd = vs1)
-- | -- | --
When vd ≠ vs1 | Need extra vmv1r.v vd, vs2 to pre-load | N/A
When vd = vs1 | Direct 1-to-1 | Direct 1-to-1

<p style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">paire/pairo performs the interleave directly in hardware — no mask involved at all.</p><h3 id="2.-source-register-destruction" style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-style: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">2. Source Register Destruction</h3><p style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">Slide-based<span class="Apple-converted-space"> </span><code>vpairo</code><span class="Apple-converted-space"> </span>writes results back to<span class="Apple-converted-space"> </span><code>vs1</code><span class="Apple-converted-space"> </span>(exploiting<span class="Apple-converted-space"> </span><code>mu</code><span class="Apple-converted-space"> </span>to preserve the original odd positions). This<span class="Apple-converted-space"> </span><strong>destroys the source register</strong><span class="Apple-converted-space"> </span>— in the SATD code,<span class="Apple-converted-space"> </span><code>v16</code>/<code>v17</code><span class="Apple-converted-space"> </span>are overwritten in Stage 1, and Stage 2 can only use the surviving<span class="Apple-converted-space"> </span><code>v28</code>/<code>v29</code>. Register allocation becomes tightly constrained.</p><p style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">paire/pairo allows<span class="Apple-converted-space"> </span><code>vd</code><span class="Apple-converted-space"> </span>to be any independent register — no such constraint.</p>

The one-line mapping is a useful reference, but as the SATD example shows, the real-world cost includes mask setup, mu policy switches, and register allocation constraints that don't appear in an isolated instruction equivalence. This is exactly why finding the best patterns is non-trivial — and why I believe vpaire/vpairo should stay in the extension.

camel-cdr:
> I also agree that vunzipe and vunzipo may not be strictly necessary, since they can be expressed reasonably well using existing base RVV instructions.
> However, I believe vpaire and vpairo are very important and should be kept

I think you misunderstand the point of my post, I wasn't suggesting that the Zvzip instructions aren't needed.
Even vunzip is important, to cover the SEW=64 case.

It's just that these operations are so common, we should document how to do them in base "V" (and RVA23).
People will wonder how to do those transposes and zips, search the spec, find Zvzip and then struggle to write the fallback case when Zvzip isn't available.

Instruction count is a bad metric to talk about this.
Consider the 4x4 transpose code from the current spec:

```
vsetvli t0, x0, e32, m1, ta, ma
vpaire.vv v5, v1, v2
vpairo.vv v6, v1, v2
vpaire.vv v7, v3, v4
vpairo.vv v8, v3, v4

vsetvli t0, x0, e64, m1, ta, ma
vpaire.vv v1, v5, v7
vpaire.vv v2, v6, v8
vpairo.vv v3, v5, v7
vpairo.vv v4, v6, v8
```

In base "V" you can do this with:

```
vsetvli t0, x0, e64, m1, ta, mu
vmv1r.v v8, v3
vmv1r.v v9, v4

vslide1down.vx v3, v1, x0, v0.t
vslide1down.vx v4, v2, x0, v0.t
vmv1r.v v0, v21
vslide1up.vx v1, v8, x0, v0.t
vslide1up.vx v2, v9, x0, v0.t

vsetvli t0, x0, e32, m1, ta, mu
vmv1r.v v8, v1
vmv1r.v v9, v3

vslide1up.vx v1, v2, x0, v0.t
vslide1up.vx v3, v4, x0, v0.t
vmv1r.v v0, v22
vslide1down.vx v2, v8, x0, v0.t
vslide1down.vx v4, v9, x0, v0.t
```

This introduces a good amount of extra moves, but on architectures with renaming, the actual executed instruction count is the same.
The other reason we want dedicated vpaire* instructions, is that they are extremely cheap to implement, compared to mask vslide1up.
So you would expect them to operate at a higher throughput and lower latency and use fewer register file ports.

QiuJing-Damo:
> 
> This introduces a good amount of extra moves, but on architectures with renaming, the actual executed instruction count is the same. The other reason we want dedicated vpaire* instructions, is that they are extremely cheap to implement, compared to mask vslide1up. So you would expect them to operate at a higher throughput and lower latency and use fewer register file ports.

Thanks for the detailed analysis and the base V fallback examples — really appreciate you putting that together.

Just to make sure we're on the same page about the scope here: I think what we're discussing at this point is whether Zvzip should go through the fast track process, rather than whether it needs to be mandated as part of RVA23. My understanding is that Zvzip is a small, focused extension to the base vector ISA, quite similar in spirit to Vector Crypto — useful and well-defined, but not a required piece of the RVA23 profile. There's no expectation that it has to go into RVA23, same as Vector Crypto.

aswaterman:
> I think what we're discussing at this point is whether Zvzip should go through the fast track process, rather than whether it needs to be mandated as part of RVA23

No, that is not on the table. RVA23U64 is long since ratified.

I believe @camel-cdr has the right of this one.

ptomsich:
> One possible problem found when implementing the new spec: we set `vl/type` on interleaved operand now, which causes more `vset(i)vli` instructions for `unzip`. For example, if we want to `unzip` a `<2 x i64>` (`v10`) from `<4 x i64>` (`v8`) and store:
> 
> * Before (v0.1):
> 
> ```assembly
> vsetivli zero, 2, e64, m1, ta, ma
> vunzipe.v v10, v8
> vse64.v v10, (a1)
> ```
> 
> * After (v0.2):
> 
> ```assembly
> vsetivli zero, 4, e64, m2, ta, ma
> vunzipe.v v10, v8
> vsetivli zero, 2, e64, m1, ta, ma
> vse64.v v10, (a1)
> ```
> 
> However, for `zip`, the new spec reduces `vset(i)vli` instructions. For example, if we want to `zip` a `<4 x i64>` (`v12`) from two `<2 x i64>` (`v8` and `v10`) and store:
> 
> * Before (v0.1):
> 
> ```assembly
> vsetivli zero, 2, e64, m1, ta, ma
> vzip.vv v12, v8, v10
> vsetvli zero, 4, e64, m2, ta, ma
> vse64.v v12, (a1)
> ```
> 
> * After (v0.2):
> 
> ```assembly
> vsetvli zero, 4, e64, m2, ta, ma
> vzip.vv v12, v8, v10
> vse64.v v12, (a1)
> ```
> 
> So I think instead of setting `vl/vtype` on interleaved operand, we may consider to set `vl/vtype` on result (`vd`)?

I though long and hard about this: first I was tempted to reorient (as described by you), but the increased/changed error surface makes me remain careful

The proposal only changes the unzips: vzip's vd is the interleaved operand, so it is identical under both schemes.

The vsetvli saving is real but narrow. It only appears in reg-to-reg sequences: when the interleaved data comes from memory, the load needs the 2n/m2 configuration anyway, so both schemes execute two vset(i)vli per loop iteration. And a vtype-only toggle is cheap on cores that rename vtype.

Against that saving, "vl on vd" makes the unzips read source elements up to index 2*vl-1 unconditionally. No ratified instruction does that: vrgather and vslidedown read past vl only when software asks for it through indices or an offset, and each carries an explicit normative sentence for it; vcompress — the instruction the current unzips are modeled on (vl counts the consumed operand, destination gets a derived body, remainder is tail per vta) — never reads past vl at all. It also flips the common bug from loud to silent: today, a missing reconfiguration produces an oversized store, visible and usually trapping on group misalignment; under "vl on vd", a loop that loads n elements silently unzips n more elements of stale register content.

A count-changing instruction always has a factor of 2 somewhere. The only real choice is where it sits and how forgetting it fails. We placed it on the side where the failure is loud, and paid one vsetvli in reg-to-reg code for it. A compiler never forgets the vsetvli, so the density argument protects code that needs no protection, while the silent-read hazard bites exactly the hand-written kernels this extension targets.

camel-cdr:
If it's only unzip, the we have the best of both worlds, as you can change to vnsrl (unless SEW=64 as in the example):

```
vsetivli zero, 2, e8, m1, ta, ma
vnsrl.wi v10, v8, 8
vse8.v v10, (a1)
```

@ptomsich  I do wonder if defining Zvzip in terms of widening/narrowing is a better solution. This woud fit more directly into how existing things behave, for SEW=64 we would need to define an EEW=128 though.

ptomsich:
> If it's only unzip, the we have the best of both worlds, as you can change to vnsrl (unless SEW=64 as in the example):
> 
> ```
> vsetivli zero, 2, e8, m1, ta, ma
> vnsrl.wi v10, v8, 8
> vse8.v v10, (a1)
> ```
> 
> Philipp Tomsich (@ptomsich) I do wonder if defining Zvzip in terms of widening/narrowing is a better solution. This woud fit more directly into how existing things behave, for SEW=64 we would need to define an EEW=128 though.

Yes, it is a closer to narrowing/widening.  I just don't see how we could overcome the EEW=128 wall, though.

ptomsich:
> > If it's only unzip, the we have the best of both worlds, as you can change to vnsrl (unless SEW=64 as in the example):
> > ```
> > vsetivli zero, 2, e8, m1, ta, ma
> > vnsrl.wi v10, v8, 8
> > vse8.v v10, (a1)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Philipp Tomsich (Philipp Tomsich (@ptomsich)) I do wonder if defining Zvzip in terms of widening/narrowing is a better solution. This woud fit more directly into how existing things behave, for SEW=64 we would need to define an EEW=128 though.
> 
> Yes, it is a closer to narrowing/widening. I just don't see how we could overcome the EEW=128 wall, though.

While playing around with EEW=128, I realized that the current definition is intended as exactly your framing, but transposed: the pair-doubling lives on EMUL (element count) (that fits SEW=ELEN), instead of on EEW.
Below ELEN, vnsrl remains a fine deinterleave.

aswaterman:
> I'd expect at least encoding changes to come out of ARC (unless they want to relitigate the evl/vstart-discussion).

I haven't thought about this enough to have formed an opinion yet, but I can say for sure that the ARC will at least discuss this topic.

I'm not persuaded by the concern about software forgetting the extra vsetvli.  The better software metric is how many vsetvli instructions are needed in the most common cases.  Presumably one of the two schemes is better than the other in this regard.

@ptomsich when this goes to ARC, can you write us a note explaining the changes and evaluating the pros and cons of the two approaches?  Kind of like you've already done in this thread.

lukel97:
> One possible problem found when implementing the new spec: we set `vl/type` on interleaved operand now, which causes more `vset(i)vli` instructions for `unzip`. For example, if we want to `unzip` a `<2 x i64>` (`v10`) from `<4 x i64>` (`v8`) and store:
> 
> * Before (v0.1):
> 
> ```assembly
> vsetivli zero, 2, e64, m1, ta, ma
> vunzipe.v v10, v8
> vse64.v v10, (a1)
> ```
> 
>  * After (v0.2):
> 
> ```assembly
> vsetivli zero, 4, e64, m2, ta, ma
> vunzipe.v v10, v8
> vsetivli zero, 2, e64, m1, ta, ma
> vse64.v v10, (a1)
> ```
> 
> However, for `zip`, the new spec reduces `vset(i)vli` instructions. For example, if we want to `zip` a `<4 x i64>` (`v12`) from two `<2 x i64>` (`v8` and `v10`) and store:
> 
> * Before (v0.1):
> 
> ```assembly
> vsetivli zero, 2, e64, m1, ta, ma
> vzip.vv v12, v8, v10
> vsetvli zero, 4, e64, m2, ta, ma
> vse64.v v12, (a1)
> ```
> 
> * After (v0.2):
> 
> ```assembly
> vsetvli zero, 4, e64, m2, ta, ma
> vzip.vv v12, v8, v10
> vse64.v v12, (a1)
> ```
> 
> So I think instead of setting `vl/vtype` on interleaved operand, we may consider to set `vl/vtype` on result (`vd`)?

FWIW, if we're going to prioritise removing a vsetvli in either one of the zip or unzip store sequences, a `vzip.vv` + `vse64.v` is already expressible as a `vsseg2e64.v` provided that `vl` is a multiple of 2. So it's probably more important to remove the extra `vsetvli` from the unzip case.

`vl/vtype` on `vd` also makes it somewhat similar to how a `vlseg2eEEW.v` reads `vl * 2` elements from memory.

wangpc-pp:
Reverse ping :-)

garthlei:
I was wondering why we chose length-changing zip/unzip semantics in Zvzip. Could anyone tell me the major use cases? I guess they can be beneficial when we actually load something, zip/unzip that, and then store into the memory. However, for general vector permutations, I guess a design with the same element counts in source and destination registers can be more helpful?

Note that AArch64 has zip/uzp instructions that have same source/destination element counts, and two-vector permutations like (say, element count=4) [1, 3, 5, 7] and [2, 6, 3, 7] can be implemented using Arm's instructions `uzp2` and `zip2`. These are common vector permutation patterns that show up in vector (especially auto-vectorized) implementations of encoders.

With the current instructions in the Zvzip, we will have to use two instructions (vslideup+vunzipo for uzp2, vzip+vslidedown for zip2) for each permutation when LMUL is fractional, and we cannot avoid frequent vtype toggling, which can further affect the performance.

What about having zip/unzip instructions that do not change the vector length? For zip, we can have something like vziplow and vziphigh, and for unzip, vunzip[eo] instructions with two source operands. Current semantics would then have to be implemented using two instructions, however, and I would like to know which are the most common scenarios that we all care about.

camel-cdr:
@garthlei 

> was wondering why we chose length-changing zip/unzip semantics in Zvzip. Could anyone tell me the major use cases? I guess they can be beneficial when we actually load something, zip/unzip that, and then store into the memory. However, for general vector permutations, I guess a design with the same element counts in source and destination registers can be more helpful?

For the same reason RVV has vwaddu, and not UADDW and UADDW2.

The operation both ISAs try to model with vzip* is interleaving the elements of two vector registers.
Arm needs two instructions to model this, RVV one.
When you can't utilize the entire vector register on Arm, you may use ZIP1 in isolation, in RVV you can decrease vl and/or LMUL.
I don't know of any uses of ZIP2 without also a ZIP1 on the same operands.


> Note that AArch64 has zip/uzp instructions that have same source/destination element counts, and two-vector permutations like (say, element count=4) [1, 3, 5, 7] and [2, 6, 3, 7] can be implemented using Arm's instructions uzp2 and zip2. These are common vector permutation patterns that show up in vector (especially auto-vectorized) implementations of encoders.

Where do only the ZIP2 instructions come up, without a corresponding ZIP1?

The only real use case I can think of is naive emulation/translation of NEON to RVV. This type of application should try to combine the corresponding zip1 and zip2 NEON instructions instead.

> With the current instructions in the Zvzip, we will have to use two instructions (vslideup+vunzipo for uzp2, vzip+vslidedown for zip2) for each permutation when LMUL is fractional, and we cannot avoid frequent vtype toggling, which can further affect the performance.

What behavior do you expect from ZIP1/ZIP2 in RVV? How does it interact with LMUL and vl?
You seem to suggest that LMUL defines where the upper and lower parts of the vector register are.
But then you can't even use it to emulate the NEON instructions.
Having it based on vl is even more unprecedented behavior, has to deal with odd vl values and results in much more expensive hardware.

> What about having zip/unzip instructions that do not change the vector length? For zip, we can have something like vziplow and vziphigh, and for unzip, vunzip[eo] instructions with two source operands. Current semantics would then have to be implemented using two instructions, however, and I would like to know which are the most common scenarios that we all care about.

There is some utility on having a vunzip* from two vector register groups, as this may save you a move, if the vector registers you want to unzip aren't adjacent.
But how relevant is that really? Is it worth the opcode space?
It would make the usual case where you work on a wider LMUL than then for the unzip less nice.

garthlei:
Hi @camel-cdr ! Thank you for your clarification.

> Bohan (Garth) Lei (@garthlei)
> 
> > was wondering why we chose length-changing zip/unzip semantics in Zvzip. Could anyone tell me the major use cases? I guess they can be beneficial when we actually load something, zip/unzip that, and then store into the memory. However, for general vector permutations, I guess a design with the same element counts in source and destination registers can be more helpful?
> 
> For the same reason RVV has vwaddu, and not UADDW and UADDW2.

This makes sense, as NEON is a SIMD architecture. zip/unzip are a bit different than vwaddu though, as zip/unzip change the element count, while vwaddu changes the width. 

> 
> The operation both ISAs try to model with vzip* is interleaving the elements of two vector registers. Arm needs two instructions to model this, RVV one. When you can't utilize the entire vector register on Arm, you may use ZIP1 in isolation, in RVV you can decrease vl and/or LMUL. I don't know of any uses of ZIP2 without also a ZIP1 on the same operands.

> 
> Where do only the ZIP2 instructions come up, without a corresponding ZIP1?

I'm sorry for not expressing it clearly. ZIP2 rarely shows up without a ZIP1, and using something alone was not the typical scenario I was thinking about. I was thinking about scenarios like this:

```
zip1 v27.4h, v22.4h, v25.4h
zip2 v0.4h, v22.4h, v25.4h
zip1 v29.4h, v23.4h, v24.4h
zip2 v30.4h, v23.4h, v24.4h
```

In RISC-V (suppose VLEN=128b), with Zvzip, we would have something like:

```
vsetvli zero,zero,e16,m1,ta,ma
vzip.vv v27,v22,v25
vzip.vv v29,v23,v24
vslidedown.vi	v0,v27,4
vslidedown.vi	v30,v29,4
```

Also 4 instructions (apart from vsetvli), but with vslidedown depending on vzip results.

Similar for uzp scenarios:

```
uzp1 v27.4h, v22.4h, v25.4h
uzp2 v0.4h, v22.4h, v25.4h
uzp1 v29.4h, v23.4h, v24.4h
uzp2 v30.4h, v23.4h, v24.4h
```

which would be

```
vsetvli zero,zero,e16,m1,ta,ma
vslideup.vi v22,v25,4
vslideup.vi v23,v24,4
vunzipe.v v27,v22
vunzipo.v v0,v22
vunzipe.v v29,v23
vunzipo.v v30,v23
```

We have to use vslideup instructions before doing the vunzip.
Things will be a lot easier if the element count is 8 in my cases, though, as with M2, we can naturally have upper and lower parts in different registers.

> What behavior do you expect from ZIP1/ZIP2 in RVV? How does it interact with LMUL and vl? You seem to suggest that LMUL defines where the upper and lower parts of the vector register are. But then you can't even use it to emulate the NEON instructions. Having it based on vl is even more unprecedented behavior, has to deal with odd vl values and results in much more expensive hardware.

This indeed is a problem. Personally relying on vl sounds better to me, but I am not quite sure about this part. 

> 
> > What about having zip/unzip instructions that do not change the vector length? For zip, we can have something like vziplow and vziphigh, and for unzip, vunzip[eo] instructions with two source operands. Current semantics would then have to be implemented using two instructions, however, and I would like to know which are the most common scenarios that we all care about.
> 
> There is some utility on having a vunzip* from two vector register groups, as this may save you a move, if the vector registers you want to unzip aren't adjacent. But how relevant is that really? Is it worth the opcode space? It would make the usual case where you work on a wider LMUL than then for the unzip less nice.

As I showed above, not only move. When the vectors related are shorter than a vector register, we would need slides. I am not really sure which is the more common use case for us, which is part of the reason why I replied to the PR.

camel-cdr:
@garthlei 
> zip/unzip are a bit different than vwaddu though, as zip/unzip change the element count, while vwaddu changes the width.

Depends on how you define it. vzip could also be defined as a widening instruction, which keeps vl constant, but changes SEW.

> I'm sorry for not expressing it clearly. ZIP2 rarely shows up without a ZIP1, and using something alone was not the typical scenario I was thinking about. I was thinking about scenarios like this:

Ok, this makes more sense. Do these cases require the results to be in two separate registers?

I've looked through some ffmpeg kernels and found some cases where they use zip1/zip2 in separate registers and do different things for both registers.
However, the ones I've looked at have good/better RVV solutions, without ZIP1/ZIP2.

Can you share a full function, that does something relevant, where the current zvzip is problematic?

garthlei:
@camel-cdr 

> I've looked through some ffmpeg kernels and found some cases where they use zip1/zip2 in separate registers and do different things for both registers. However, the ones I've looked at have good/better RVV solutions, without ZIP1/ZIP2.
> 
> Can you share a full function, that does something relevant, where the current zvzip is problematic?

In fact we can avoid using ZIP1/ZIP2/UZP1/UZP2 when we want to when writing assembly versions, but that can be a bit hard for compilers to generate such code.

GCC (aarch64) currently vectorizes `sub4x4_dct` like this:

```
sub4x4_dct:
        ldr     d31, [x1]
        ldr     d29, [x2]
        ldr     d2, [x1, 32]
        ldr     d25, [x2, 64]
        ldr     s30, [x1, 16]
        ldr     s28, [x2, 32]
        ldr     s0, [x1, 48]
        ldr     s23, [x2, 96]
        zip1    v30.2s, v31.2s, v30.2s
        zip1    v28.2s, v29.2s, v28.2s
        zip1    v0.2s, v2.2s, v0.2s
        zip1    v23.2s, v25.2s, v23.2s
        usubl   v28.8h, v30.8b, v28.8b
        usubl   v23.8h, v0.8b, v23.8b
        dup     d19, v28.d[1]
        dup     d4, v23.d[1]
        uzp2    v22.4h, v28.4h, v19.4h
        uzp1    v19.4h, v28.4h, v19.4h
        uzp2    v20.4h, v23.4h, v4.4h
        uzp1    v4.4h, v23.4h, v4.4h
        uzp1    v21.4h, v22.4h, v20.4h
        uzp2    v17.4h, v19.4h, v4.4h
        uzp2    v20.4h, v22.4h, v20.4h
        uzp1    v4.4h, v19.4h, v4.4h
        sub     v18.4h, v21.4h, v17.4h
        add     v17.4h, v21.4h, v17.4h
        sub     v7.4h, v4.4h, v20.4h
        add     v4.4h, v4.4h, v20.4h
        add     v16.4h, v18.4h, v18.4h
        sub     v27.4h, v4.4h, v17.4h
        add     v4.4h, v4.4h, v17.4h
        sub     v16.4h, v7.4h, v16.4h
        add     v7.4h, v7.4h, v7.4h
        add     v7.4h, v7.4h, v18.4h
        uzp2    v24.4h, v27.4h, v16.4h
        uzp1    v27.4h, v27.4h, v16.4h
        uzp2    v6.4h, v4.4h, v7.4h
        uzp1    v4.4h, v4.4h, v7.4h
        uzp1    v5.4h, v6.4h, v24.4h
        uzp2    v26.4h, v4.4h, v27.4h
        uzp2    v24.4h, v6.4h, v24.4h
        uzp1    v27.4h, v4.4h, v27.4h
        sub     v3.4h, v5.4h, v26.4h
        add     v26.4h, v26.4h, v5.4h
        sub     v1.4h, v27.4h, v24.4h
        add     v27.4h, v24.4h, v27.4h
        add     v2.4h, v3.4h, v3.4h
        sub     v30.4h, v27.4h, v26.4h
        add     v27.4h, v26.4h, v27.4h
        sub     v2.4h, v1.4h, v2.4h
        add     v1.4h, v1.4h, v1.4h
        add     v1.4h, v1.4h, v3.4h
        zip1    v29.4h, v30.4h, v2.4h
        zip2    v30.4h, v30.4h, v2.4h
        zip1    v0.4h, v27.4h, v1.4h
        zip2    v27.4h, v27.4h, v1.4h
        zip1    v31.2s, v0.2s, v29.2s
        zip1    v25.2s, v27.2s, v30.2s
        zip2    v0.2s, v0.2s, v29.2s
        zip2    v27.2s, v27.2s, v30.2s
        stp     d31, d0, [x0]
        stp     d25, d27, [x0, 16]
        ret
```

We do have better implementations with TRN/vpair instructions, though, but not so straightforward for compiler vectorization.

ZIP1/ZIP2 instructions are used in this version. The first four ZIP1 instructions in this example can be safely replaced by vpaire/vslideup instructions. The last eight ZIP1/ZIP2 instructions can also actually be implemented with fewer vzip instructions, but this requires special compiler optimizations. AFAIK currently neither GCC nor LLVM seems to merge permutations like [0,4,1,5] and [2,6,3,7] into a long [0,4,1,5,2,6,3,7] (vzip semantics) though. That would require walking the def-use chain to see if we could use the long vzip, which should be done as a part of a special optimization for this. Same-length permutations are easier to implement on compilers.

The UZP1/UZP2 sequence in this version can be replaced by an equivalent TRN1/TRN2/vpaire/vpairo sequence, but this is also not so straightforward for the compiler to generate, and requires a special optimization.

SATD kernel functions contains a similar UZP1/UZP2 sequence, which is harder to optimize into other instructions (we can still use vpaire/vpairo in this function, though, but it's hard for the compiler to perform such transformations):

```
        saddw   v23.4s, v23.4s, v29.4h
        saddw2  v16.4s, v16.4s, v29.8h
        saddw   v22.4s, v22.4s, v30.4h
        saddw2  v27.4s, v27.4s, v30.8h
        uzp2    v21.4s, v23.4s, v16.4s
        uzp1    v16.4s, v23.4s, v16.4s
        uzp2    v20.4s, v22.4s, v27.4s
        uzp1    v27.4s, v22.4s, v27.4s
        sub     v19.4s, v16.4s, v21.4s
        add     v16.4s, v16.4s, v21.4s
        sub     v18.4s, v27.4s, v20.4s
        add     v27.4s, v27.4s, v20.4s
```

PR #3046, opened 2026-05-07
Zabhlrsc extension

tariqkurd-repo:
Replacement for https://github.com/riscv/riscv-isa-manual/pull/1922 brought up to date to the new format

PR #3234, opened 2026-07-08
Clean attempt at Zabhlrsc

tariqkurd-repo:
replacement for #3046 this time passing the integration checks


src/unpriv/zabhlrsc.adoc
```diff
@@ -0,0 +1,56 @@
...
+====
+The absence of LR/SC operations for subword data types
+is problematic for CHERI software (RVY base architectures).
+Non-CHERI RISC-V software can use LR/SC on larger data types than are strictly
+required for the memory access to register the reservation set.
+RVY base architectures check memory bounds and so it is not possible to round a subword
+access up to a word or larger to gain the reservation set.
+It is not always possible to emulate narrow accesses using wider accesses on a CHERI system, due to the memory bounds.
+Such emulation can be problematic for other reasons as explained in the rationale for ext:zabha[].
+====
```
tariqkurd-repo, line 19:
```suggestion
Given that this emulation scheme is practical and performant on non-CHERI systems, the RVA and RVB profiles are unlikely to mandate this extension or make it an option.
====
```

src/unpriv/zabhlrsc.adoc
```diff
@@ -0,0 +1,57 @@
...
+The absence of LR/SC operations for subword data types
+is problematic for CHERI software (RVY base architectures).
+Non-CHERI RISC-V software can use LR/SC on larger data types than are strictly
+required for the memory access to register the reservation set.
+RVY base architectures check memory bounds and so it is not possible to round a subword
+access up to a word or larger to gain the reservation set.
+It is not always possible to emulate narrow accesses using wider accesses on a CHERI system, due to the memory bounds.
+Such emulation can be problematic for other reasons as explained in the rationale for ext:zabha[].
+Given that this emulation scheme is practical and performant on non-CHERI systems, the RVA and RVB profiles are unlikely to mandate this extension or make it an option.
+====
```
tariqkurd-repo, line 19:
```suggestion
[NOTE]
====
The absence of LR/SC for subword data types is problematic for CHERI software
(RVY base architectures). Non-CHERI RISC-V software can register the
reservation set using an access wider than the operand actually requires, but
RVY base architectures check memory bounds, so widening a subword access in
this way is not always possible. Such emulation can be problematic for other
reasons too, as explained in the rationale for ext:zabha[]. Given that the
emulation scheme is practical and performant on non-CHERI systems, the RVA and
RVB profiles are unlikely to mandate this extension or make it an option.
====
```

tariqkurd-repo: The objective is to simplify the note as it wa quite wordy before.

arichardson:
Might want to update the PR description to something like the intro paragraph, so that merging creates a sensible commit message:

```
The Zalrsc extension offers LR/SC (load reserved/store conditional) instructions
for _words_ and _doublewords_. Zabhlrsc extends this by adding _byte_ and
_halfword_ versions. The Zabhlrsc extension depends upon the Zalrsc
standard extension.
```

src/unpriv/zabhlrsc.adoc
```diff
@@ -0,0 +1,57 @@
...
+The absence of LR/SC operations for subword data types
+is problematic for CHERI software (RVY base architectures).
+Non-CHERI RISC-V software can use LR/SC on larger data types than are strictly
+required for the memory access to register the reservation set.
+RVY base architectures check memory bounds and so it is not possible to round a subword
+access up to a word or larger to gain the reservation set.
+It is not always possible to emulate narrow accesses using wider accesses on a CHERI system, due to the memory bounds.
+Such emulation can be problematic for other reasons as explained in the rationale for ext:zabha[].
+Given that this emulation scheme is practical and performant on non-CHERI systems, the RVA and RVB profiles are unlikely to mandate this extension or make it an option.
```
jrtc27: I've always taken the view that, if your cache already supports LB/LH/SB/SH and LR.W/SC.W then surely supporting LR.B/LR.H/SC.B/SC.H is basically free? You already have the hardware to extract or insert a byte/halfword, and you already have the hardware to reserve cache lines, so you just need to decode the instruction and do both? Having this extension makes the compiler's life easier when not using an AMO and generates shorter code.

aswaterman: That analysis sounds right to me.  But this is one of those extensions where adding it as an option, rather than a mandate, isn't terribly useful for standard software; and the door to adding mandatory extensions to the RVA/RVB series has closed for the next 4+ years.  So the ARC asked for this note to be added just to head off questions about it.

jrtc27: Hm, but would this note not imply that even in 4+ years it is unlikely to be mandatory?

jrtc27: (I agree it's not so useful as an optional extension, but would that not make it all the more important for it to, eventually, end up as mandatory? Unless the intent is that no implementation will ever support it unless also supporting Y, where it and Zabha are implicitly part of A out of necessity?)

aswaterman: I don’t have a crystal ball but I’d expect that CAS is going to become mandatory in the next major RVA profile, which takes pressure off of LR/SC. But your point is taken. I’d be fine changing the text to “unlikely to make this extension a profile option”.

tariqkurd-repo: Should it not be a development option like `Zabha`? It's essentially the same spec, but if RVI is moving away from LR/SC then "unlikely" would be better I guess.

aswaterman: It should only be a development option if it's on track to be mandated in the future, which at this time it is not.

PR #3243, opened 2026-07-12
rvwmo: add source/destination register listings for Zacas, Zabha, Zalasr

### LINKED ISSUE #3241: [NormRules] - RVWMO dependency metadata missing for Zacas, Zabha, and Zalasr
The RVWMO [source/destination register listings](https://github.com/riscv/riscv-isa-manual/blob/79b241cb6d2f07167366747ae49df94b88ad4d3b/src/unpriv/rvwmo.adoc#L371-L388) say they define syntactic dependencies, but the [atomic tables cover only RV32A and RV64A](https://github.com/riscv/riscv-isa-manual/blob/79b241cb6d2f07167366747ae49df94b88ad4d3b/src/unpriv/rvwmo.adoc#L584-L641). The ratified Zacas, Zabha, and Zalasr instructions have no entries.

For the non-CAS Zabha operations and Zalasr, the missing metadata looks straightforward — they're width analogues of the existing AMO / load / store rows.

AMOCAS is the part that isn't. It [reads `rd` as the compare value and then overwrites it](https://github.com/riscv/riscv-isa-manual/blob/79b241cb6d2f07167366747ae49df94b88ad4d3b/src/unpriv/zacas.adoc#L26-L37), whereas existing AMOs write `rd` but never read it — and Zabha adds `amocas.b`/`amocas.h` when Zacas is also implemented, so those inherit the same question. [#2940](https://github.com/riscv/riscv-isa-manual/issues/2940) established that AMOCAS is treated like other AMOs for the PPO rules (by the listings saying nothing, per the [discussion](https://github.com/riscv/riscv-isa-manual/issues/2940#issuecomment-4306140881)). But I don't think that settles how the `rd`/`rd+1` compare *input* is classified: is it deliberately not a syntactic source, or should it be an address / data / plain source?

This feels like a concrete case for the per-instruction metadata work already tracked in [#2945](https://github.com/riscv/riscv-isa-manual/issues/2945). If that's the intended home for it, happy to frame it that way.

abhinavagarwal07:
The RVWMO atomic register listings cover RV32A/RV64A but omit the ratified Zacas, Zabha, and Zalasr instructions. This adds them: Zabha at the same profile as the word/doubleword AMOs, Zalasr with the same register-dependency profile as ordinary loads and stores, and amocas with the old rd as a data source (the comparand) and also the destination — footnoted for the rd+1/rs2+1 pair roles on amocas.d/.q. Daggers match the existing AMOs and loads.

Fixes #3241.

### DIFF
--- a/src/unpriv/rvwmo.adoc
+++ b/src/unpriv/rvwmo.adoc
@@ -640,6 +640,70 @@ register(s) to destination register(s) as specified
 
 |===
 
+.Zabha Standard Extension
+[%autowidth.stretch,float="center",align="center",cols="<,<,<,<,<",options="header"]
+|===
+||Source Registers |Destination Registers |Accumulating CSRs|
+
+|insn:amoswap.b/h[]† |_rs1_ ^A^, _rs2_ ^D^ |_rd_ | |
+
+|insn:amoadd.b/h[]† |_rs1_ ^A^, _rs2_ ^D^ |_rd_ | |
+
+|insn:amoxor.b/h[]† |_rs1_ ^A^, _rs2_ ^D^ |_rd_ | |
+
+|insn:amoand.b/h[]† |_rs1_ ^A^, _rs2_ ^D^ |_rd_ | |
+
+|insn:amoor.b/h[]† |_rs1_ ^A^, _rs2_ ^D^ |_rd_ | |
+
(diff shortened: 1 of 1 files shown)


aswaterman (approved):

aswaterman:
Thanks!

PR #3246, opened 2026-07-13
Clarify that M-mode interrupts are optional

davidharrishmc:
Added a paragraph to Machine Interrupt section clarifying that mip and mie are WARL and any field can be ROZ.  Language parallels corresponding paragraph in Supervisor Interrupt section.

Updated normative rules to match.  Deleted LCOFI_INTR_IMPL rule because it is implied by Sscofpmf.

If this PR is not accurate, there should be the opposite language that every bit MEI/MTI must be writable.  However, this would be inconsistent with the supervisor interrupts, and also strange that a very simple core would necessarily implement interrupts.

Also a question about whether the following statement is accurate.  Can delete if not accurate.
"If a bit of csr:mie[] is read-only zero, the corresponding bit of csr:mideleg[] is also read-only zero."

### DIFF
--- a/src/priv/machine.adoc
+++ b/src/priv/machine.adoc
@@ -1313,7 +1313,7 @@ include::images/bytefield/mideleg.edn[]
 
 [#norm:mideleg_enc_txt]#csr:mideleg[] holds trap delegation bits for individual interrupts, with the
 layout of bits matching those in the csr:mip[] register# (i.e., STIP
-interrupt delegation control is located in bit 5).
+interrupt delegation control is located in bit 5).  If a bit of csr:mie[] is read-only zero, the corresponding bit of csr:mideleg[] is also read-only zero.
 
 [#norm:medeleg_when_rd0]#For exceptions that cannot occur in less privileged modes, the
 corresponding csr:medeleg[] bits should be read-only zero. In particular,
@@ -1420,6 +1420,13 @@ reflected in csr:mip[msip] eventually, but not necessarily immediately.#
 if a platform standard supports the delivery of machine-level
 interprocessor interrupts through external interrupts (MEI) instead,
 then csr:mip[msip] and csr:mie[msie] may both be read-only zeros.#
+
+[#norm:std_mach_intr_impl]#Each standard machine interrupt type (MEI, MTI, MSI) may not be implemented,
+in which case the corresponding interrupt-pending and interrupt-enable
+bits are read-only zeros.#
+[#norm:mip_mie_warl]#All bits in `mip` and `mie` are *WARL* fields. The
+implemented interrupts may be found by writing one to every bit location
+in `mie`, then reading back to see which bit positions hold a one.#
 
 [#norm:mip_sxip_mie_sxie_rdonly0]#If supervisor mode is not implemented, bits csr::[seip], csr::[stip], and csr::[ssip] of
 csr:mip[] and csr::[seie], csr::[stie], and csr::[ssie] of csr:mie[] are read-only zeros.#
--- a/src/priv/machine.adoc
+++ b/src/priv/machine.adoc
@@ -1424,9 +1424,9 @@ then csr:mip[msip] and csr:mie[msie] may both be read-only zeros.#
 [#norm:std_mach_intr_impl]#Each standard machine interrupt type (MEI, MTI, MSI) may not be implemented,
 in which case the corresponding interrupt-pending and interrupt-enable
 bits are read-only zeros.#
-[#norm:mip_mie_warl]#All bits in `mip` and `mie` are *WARL* fields. The
+All bits in `mip` and `mie` are *WARL* fields. The
 implemented interrupts may be found by writing one to every bit location
-in `mie`, then reading back to see which bit positions hold a one.#
+in `mie`, then reading back to see which bit positions hold a one.
 
 [#norm:mip_sxip_mie_sxie_rdonly0]#If supervisor mode is not implemented, bits csr::[seip], csr::[stip], and csr::[ssip] of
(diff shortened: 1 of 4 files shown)


src/priv/machine.adoc
```diff
@@ -1421,6 +1421,13 @@ if a platform standard supports the delivery of machine-level
 interprocessor interrupts through external interrupts (MEI) instead,
 then csr:mip[msip] and csr:mie[msie] may both be read-only zeros.#
 
+[#norm:std_mach_intr_impl]#Each standard machine interrupt type (MEI, MTI, MSI) may not be implemented,
```
aswaterman: We should avoid "may not" because it has two meanings in conventional English: "might not" or "must not".  Here we obviously mean "might not", so I'd recommend that.

src/priv/machine.adoc
```diff
@@ -1421,6 +1421,13 @@ if a platform standard supports the delivery of machine-level
...
+in which case the corresponding interrupt-pending and interrupt-enable
+bits are read-only zeros.#
+All bits in `mip` and `mie` are *WARL* fields. The
+implemented interrupts may be found by writing one to every bit location
+in `mie`, then reading back to see which bit positions hold a one.
```
aswaterman: The final sentence doesn't need to be part of the normative text, as it's just software guidance.  I'd probably just delete it, but if you think it's helpful, you could make it a one-sentence non-normative note following this paragraph.

src/priv/machine.adoc
davidharrishmc: Done.  The language is a copy of what was already in supervisor.adoc, so I changed "may not" to "might not" in that file as well.

src/priv/machine.adoc
davidharrishmc: Deleted, and adjusted the corresponding normative rule in supervisor.adoc to not include the non-normative note.

davidharrishmc:
Changes made based on review.  Please confirm whether the statement about mideleg being read-only zero is correct.  I've become uncertain about the intent of the spec, and am concerned I might be changing the meaning of the spec.

src/priv/machine.adoc
```diff
@@ -1313,7 +1313,7 @@ include::images/bytefield/mideleg.edn[]
 
 [#norm:mideleg_enc_txt]#csr:mideleg[] holds trap delegation bits for individual interrupts, with the
 layout of bits matching those in the csr:mip[] register# (i.e., STIP
-interrupt delegation control is located in bit 5).
+interrupt delegation control is located in bit 5).  If a bit of csr:mie[] is read-only zero, the corresponding bit of csr:mideleg[] is also read-only zero.
```
aswaterman: Since you mentioned it, I do believe this is an unintended normative change.  (It probably would've been the right thing to specify, but the spec doesn't currently mandate it, AFAICS.) So we should back it out.

src/priv/machine.adoc
```diff
@@ -1421,6 +1421,11 @@ if a platform standard supports the delivery of machine-level
 interprocessor interrupts through external interrupts (MEI) instead,
 then csr:mip[msip] and csr:mie[msie] may both be read-only zeros.#
 
+[#norm:std_mach_intr_impl]#Each standard machine interrupt type (MEI, MTI, MSI) might not be implemented,
+in which case the corresponding interrupt-pending and interrupt-enable
+bits are read-only zeros.#
```
aswaterman, line 1426: And now I'm questioning the wisdom behind adding this part.  It's true for non-machine interrupts, too...

src/priv/machine.adoc
```diff
@@ -1421,6 +1421,11 @@ if a platform standard supports the delivery of machine-level
...
+All bits in `mip` and `mie` are *WARL* fields.
```
aswaterman, line 1427: ...and so maybe this is all we're left with?

davidharrishmc: You see the corresponding language about SEI, STI, SSI is in the supervisor section.  It has been confusing to have a lack of parallel structure between the two sections.  ACT has spent too long trying to find evidence of whether MEI, MTI, MSI were mandatory, given that they the supervisor interrupts are specified as optional and the machine were not specified as optional.  

I find this clearest as written now.  But if you don't want the "might not be implemented" language, here's an alternative:

machine.adoc:
All bits in `mip` and `mie` are *WARL* fields and are read-only zero if an interrupt type is not supported.  If a bit in one register is read-only zero, the corresponding bit in the other must be read-only zero as well.

supervisor.adoc:
All bits in `sip` and `sie` are *WARL* fields and are read-only zero if an interrupt type is not supported or cannot be delegated to S-mode.  If a bit in one register is read-only zero, the corresponding bit in the other must be read-only zero as well.

allenjbaum:
I was going to push back on this one, because I thought it was
requiring some specific HW behavior... but realized that, since we are
talking about bit fields, that sentence is guidance and does add any new
information.
If the WARL fields in question were more than one bit in width, that would
be a different story.

allenjbaum:
Ugh, this chain is getting long. I was referring to the
writing&reading back comment only.

davidharrishmc:
@aswaterman could you look at this once more?  The current PR language exactly matches the supervisor section and resolves the ambiguity that the difference in language between machine and supervisor suggested different behavior.  If you don't like the fields being called out explicitly, here is alternate language to make machine and supervisor parallel, repeated from above.

machine.adoc:
All bits in mip and mie are WARL fields and are read-only zero if an interrupt type is not supported. If a bit in one register is read-only zero, the corresponding bit in the other must be read-only zero as well.

supervisor.adoc:
All bits in sip and sie are WARL fields and are read-only zero if an interrupt type is not supported or cannot be delegated to S-mode. If a bit in one register is read-only zero, the corresponding bit in the other must be read-only zero as well.

aswaterman (approved):

PR #3229, opened 2026-07-03
Fix RLB wording and tag hgatp ordering note

abhinavagarwal07:
Fixes a broken RLB sentence and removes a redundant phrase.

Also adds the missing norm anchor to the hgatp ordering note, matching the existing satp note.

Issue #3186, opened 2026-06-05
Tags JSON / normative rules are not checked in CI properly


Timmmm:
I'm not sure exactly what's going on with the generation of `ref/riscv-spec-norm-tags.json` but the way I was expecting it to work is that CI runs

```
make update-ref   # Rebuild the json file
git diff --exit-code  # Verify it's up-to-date in git
```

But what it actually does is something far more complex involving `docs-resources/tools/detect_tag_changes.py`, which does not verify that the file is 100% up-to-date (e.g. it ignores formatting changes).

The result is that the JSON is *not* currently up-to-date.

PR #3249, opened 2026-07-15
regenerate normative tag reference file

wmat:
ref/riscv-spec-norm-tags.json has drifted out of sync with the sources. It was
maintained by `detect_tag_changes.py --update-reference`, which only ever merges
*additions*: modifications and deletions are reported and then discarded. The
"sections" subtree was never updated at all, because that tool only reads and
writes data["tags"]. See #3186.

This commit regenerates the file from the current sources. It contains no
specification change of its own -- every difference below is text already merged
in main that simply never reached the reference file.

Tags: 2922 -> 2921

1 added
    norm:hgatp_op_hfence_gvma
        Anchor added by "Fix RLB wording and tag hgatp ordering note" (#3229).

2 deleted
    norm:pm_family_extensions
    norm:pm_tag_check_impl
        Both anchors were removed from src/ by "Removed tags from text that is
        not actually a normative rule" (#3177). #3177 predates the most recent
        ref/ commit (44dab723), so these deletions survived an "update"
        untouched -- the additions-only bug in action.

220 modified

  203 are markup or notation only, and change no requirement:
        the csr:/ext:/insn: macro migration
            "mip"             -> "csr:mip[]"
            "the F extension" -> "the ext:f[] extension"
            "FS"              -> "csr::[fs]"
        binary literal renotation, "11b" -> "0b11"
        en-dash entities, "31--1" -> "31{endash}1"

   17 are substantive wording, each traceable to a merged PR:

        Zalasr added to the misaligned atomicity granule PMA scope (#3231)
            norm:misaligned_atomicity_granule_size
            norm:pma_mag_insts
            norm:rvwmo_misaligned_in_atomic_pma

        MML S/U-only effective privilege clarified (#3230)
            norm:mseccfg_mml_M_rule_op
            norm:mseccfg_mml_SorU_rule_op

        RLB wording fixed (#3229)
            norm:mseccfg_rlb_op_warl

        Typo and grammar fixes
            norm:Zawrs_wrs-sto_stall_duration
                "implementation-define" -> "implementation-defined"
            norm:rvwmo_ppo_sync2
                duplicated "annotation"

        Other already-merged edits
            norm:pm_uxl_clear         MXL dropped from UXL/SXL/MXL
            norm:pmp_no_entry_match   no-match behaviour reworded
            norm:mstatus_fs_vs_xs_update_indep_priv
                                      tense reworded
            norm:mstatus_sd_rdonly0   field list reordered to FS/VS/XS
            norm:cycle_instret_hpmcounter_op_rdonly
                                      "mhpmcounter n" -> "mhpmcountern"
            norm:ssmp_ss_page_illegeal_access
                                      "pte.xwr=001" -> "pte.xwr=0b001"
            norm:xlen_reduction_hint_op1
                                      insn: macro with operands
            norm:misa_mxl_enc         table cell extraction
            norm:mstatus_fs_vs_xs_enc table cell extraction

The markup/substantive split above comes from a heuristic classifier and is
offered as a review aid rather than a guarantee. The 17 items listed are the
ones worth reading closely; the remaining 203 are mechanical.

Reproduce with `make update-ref`. Note that on main that target is currently
BSD-sed-only and fails on GNU sed, so this file was generated using the
portability fix from the follow-up change that adds the CI freshness gate
(#3186).

Signed-off-by: Bill Traynor <wmat@riscv.org>
Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>


jordancarlin:
I believe the workflow that @rpsene wrote was supposed to automatically keep this up to date, but it seems like something about the update process isn't working right.

jordancarlin (approved):

aswaterman (approved):
I'm deferring to @jordancarlin.

PR #3250, opened 2026-07-15
Check the normative tags JSON properly in CI

### LINKED ISSUE #3186: Tags JSON / normative rules are not checked in CI properly

wmat:
Fixes #3186.

## Problem

`ref/riscv-spec-norm-tags.json` was never actually verified by CI. The workflow used `detect_tag_changes.py --update-reference` as its freshness mechanism, which cannot do that job:

- it only ever merges **additions** — modifications and deletions are reported and then discarded;
- it compares text with whitespace and AsciiDoc formatting **normalized away**, so formatting-only drift is invisible to it;
- it never touches the `"sections"` subtree at all, because it only reads and writes `data["tags"]`.

The result was silent, unbounded drift. By the time #3249 regenerated the file it was 220 tags stale, and two tags removed from `src/` by #3177 had survived a *later* "update" of `ref/` completely untouched.

Separately, `make update-ref` — the documented way to regenerate the file — used `sed -i .bak`, which is BSD syntax. GNU sed parses `.bak` as an input filename and fails with `can't read .bak`. So the only supported regeneration path worked on macOS only, and CI never invoked it.

## Approach

The root problem is that two different jobs were conflated into one tool:

1. **Freshness** — "is `ref/` byte-identical to a fresh build?" Mechanical, no judgement involved.
2. **Review** — "which normative rules meaningfully changed?" A human signal, where normalizing away formatting is *correct*.

`detect_tag_changes.py` is good at (2) and structurally incapable of (1). This PR gives (1) its own mechanism and leaves (2) to the tool that does it well.

## Changes

**`make check-ref`** (new) — the freshness gate. Regenerates `ref/` and fails if `git diff` reports anything. It is built *on top of* `update-ref` rather than as a parallel comparison, so the check can never drift from the fixer. On failure the corrected files are already in the working tree, ready to review and commit.

**`make update-ref`** — `sed -i .bak` replaced with a portable trailing-newline append. Now works under both GNU and BSD sed.

**Workflow**

- `pull_request_target` + `contents: write` → `pull_request` + `contents: read`. This job builds pull-request-authored code and should not hold a write-scoped token while doing so.
- Auto-update and auto-commit of `ref/` removed. These *were* the drift mechanism; contributors now update `ref/` in their own PR, which is also what makes a read-only token sufficient.
- `make check-ref` added as a hard, non-bypassable gate.
- The change report now compares `ref/` **as of the PR's base commit** against the fresh build. It previously compared a PR's `ref/` against that same PR's own build, which becomes a guaranteed no-op once the gate enforces freshness.
- The report is published to the job summary. Fork PRs receive a read-only token on `pull_request`, so bot comments aren't available; the pass/fail status plus the summary carry the signal.
- Label text moved out of inline `${{ }}` interpolation into `env:`.

**Retired** — `scripts/check-tag-changes.sh` and the `check-tags` target. `build-norm-rules` no longer depends on it, so `make build-norm-rules` no longer silently rewrites `ref/` as a side effect. The norm-rules JSON is generated from `build/`, never from `ref/`, so there was never a real dependency there.

**`ref/README.md`** — rewritten. It documented two files that no longer exist, omitted `"sections"` from the schema, and claimed `make check-tags` updates the reference "with any changes", which was never true.

## Preserved

- The `normative-change-approved` label still gates modifications and deletions. It now applies **only** to the review check — freshness is mechanical and cannot be approved away. An approved normative change still ships its regenerated `ref/`.
- Issue creation on merge to `main` is unchanged.

## Deliberately not in this PR

Three `docs-resources` items, better handled in that repo:

- `detect_tag_changes.py`'s normalizer doesn't know the `csr:`/`ext:`/`insn:` macro forms, so the macro migration surfaced as 203 "Modified" tags in #3249 — semantically identical text reported as breaking. The next macro sweep will do the same. This is the biggest remaining win for report signal-to-noise.
- `--update-reference` cannot produce a correct reference file and should be deprecated now that `make update-ref` owns regeneration.
- `tags.rb` could emit the trailing newline itself, letting `update-ref` drop the append entirely.

None of these block this PR; the freshness fix is self-contained here.

## Verification

- `make check-ref` passes on this branch, and fails correctly against a stale `ref/` — both directions tested on Linux, where the old target could not run at all.
- Confirmed `check-ref` ignores `ref/README.md`: an earlier revision scoped the diff to the whole `ref/` directory and spuriously failed when only the hand-written docs changed.
- `actionlint` clean, including shellcheck of the `run:` blocks; all pre-commit hooks pass.
- Report logic simulated against a real base commit, and against the all-zeros base of a new-branch push.

Depends on #3249 (merged), which brought `ref/` up to date. Without it, this gate would correctly fail every PR.


.github/workflows/check-normative-tags.yml
```diff
@@ -1,187 +1,239 @@
...
+#   2. Governance (human judgement). Additions/modifications/deletions of
+#      normative text are reported for review, and modifications or deletions
+#      fail unless the PR carries the 'normative-change-approved' label.
+#
+# SECURITY: this job builds pull request code, so it must stay on `pull_request`
+# (fork PRs get a read-only token) and never `pull_request_target`. It formerly
+# ran on pull_request_target with `contents: write` while checking out the PR
+# head and invoking a PR-controlled Makefile, which allowed any PR author to push
+# to the repository. Because the token is read-only for forks, the report is
+# published to the job summary rather than posted as a PR comment.
```
jordancarlin, line 22: If we decide that the comment is important down the road, I believe this can be safely accomplished using a GitHub bot instead of pull_request_target. Fine with this as is for now though.

.github/workflows/check-normative-tags.yml
```diff
@@ -1,187 +1,239 @@
...
 on:
   push:
     paths:
       - 'src/**'
       - 'normative_rule_defs/**'
       - 'ref/**'
-      - 'scripts/check-tag-changes.sh'
-  pull_request_target:
+      - 'Makefile'
+      - '.github/workflows/check-normative-tags.yml'
```
jordancarlin: At this point it looks like just about every file is included. Should we change this to just run on all pushes so this list doesn't need to be maintained?

.github/workflows/check-normative-tags.yml
```diff
@@ -1,187 +1,239 @@
...
+  pull_request:
     paths:
       - 'src/**'
       - 'normative_rule_defs/**'
       - 'ref/**'
-      - 'scripts/check-tag-changes.sh'
+      - 'Makefile'
+      - '.github/workflows/check-normative-tags.yml'
```
jordancarlin: Same comment as above.

aswaterman:
@wmat I'll approve this so that you are able to merge it after @jordancarlin weighs in.

aswaterman (approved):

Makefile
```diff
@@ -172,19 +174,44 @@ build-tags: $(DOCS_NORM_TAGS)
...
-check-tags:
-	@bash ./scripts/check-tag-changes.sh
 
-# Copy built normative tags JSON files to ref directory and use sed to ensure they end with a newline.
-# Required by GitHub pre-commit checks for this repo.
+# Regenerate the checked-in normative tags JSON in $(REF_DIR) from the sources.
+#
+# The tags backend (docs-resources/converters/tags.rb) emits JSON with no
+# trailing newline, but pre-commit's end-of-file-fixer requires one, so append
+# it when it is missing. Inside a command substitution `tail -c 1` yields the
```
jordancarlin, line 182: We should push a change to the docs-resources script eventually so the output is natively correct.

Makefile
```diff
@@ -172,19 +174,44 @@ build-tags: $(DOCS_NORM_TAGS)
...
+# documentation living alongside them must not affect this check.
+check-ref: update-ref
+	@if git diff --quiet -- $(REF_NORM_TAGS); then \
+	  echo "Normative tag reference files are up to date."; \
+	else \
+	  echo "ERROR: normative tag reference files are out of date:"; \
+	  echo; \
+	  git diff --stat -- $(REF_NORM_TAGS); \
+	  echo; \
+	  echo "Run 'make update-ref', review the diff, and commit the result."; \
```
jordancarlin: Isn't the ref already updated? I think the instructions can be simplified to not mention running `make update-ref`.

.github/workflows/check-normative-tags.yml
```diff
@@ -1,187 +1,239 @@
...
   push:
     paths:
       - 'src/**'
       - 'normative_rule_defs/**'
       - 'ref/**'
-      - 'scripts/check-tag-changes.sh'
-  pull_request_target:
+      - 'Makefile'
+      - '.github/workflows/check-normative-tags.yml'
+  pull_request:
```
jordancarlin, line 26: Discovered this the hard way on the sail-riscv repo. You need to add the following.
```
    # Explicit listing of types is necessary to run the workflow when a label is added
    types:
      - opened
      - synchronize
      - reopened
      - labeled
      - unlabeled
```

ref/README.md
```diff
@@ -1,87 +1,100 @@
...
+    "norm:tag-name": "The normative text content...",
+  },
+  "sections": {
+    "title": "...",
+    "id": "...",
+    "children": [],
+    "tags": []
+  }
+}
+```
```
jordancarlin, line 37: I think technically this isn't valid JSON without this change.
```suggestion
```json
{
  "tags": {
    "norm:tag-name": "The normative text content..."
  },
  "sections": {
    "title": "...",
    "id": "...",
    "children": [],
    "tags": []
  }
}
```
```

jordancarlin (changes requested):
A few minor things here and there, but generally this looks much better. One thing we might want to consider is adding [autofix.ci](https://autofix.ci/) so that the PR can be automatically updated without using `pull_request_target`. We already use it in riscv-unified-db and riscv-arch-test, and it works great.

.github/workflows/check-normative-tags.yml
wmat: Agreed.  I was expecting that this might be an issue.  Deferred for now.

.github/workflows/check-normative-tags.yml
wmat: Done.

Makefile
wmat: Should be fixed by last push.

Makefile
wmat: Noted. I am opening an issue there to track what needs to be fixed.

.github/workflows/check-normative-tags.yml
wmat: Should be fixed with last push.

.github/workflows/check-normative-tags.yml
wmat: Should be fixed with last push.

ref/README.md
wmat: Pretty sure I included this in my last push to fix.

jordancarlin (approved):
LGTM. I still think we should add autofix.ci (see my previous comment), but that can be part of a separate PR.

wmat:
Yep, opened an issue for the addition of autofix.ci

PR #3252, opened 2026-07-16
[BLOCKED on docs-resources#246] Makefile: drop the normative-tags trailing-newline workaround

wmat:
> [!WARNING]
> **DRAFT — BLOCKED. Do not merge yet.**
> Merging this before the tags backend emits the newline would make `make check-ref` fail (a regenerated `ref/` would lose its trailing newline and pre-commit's `end-of-file-fixer` would reject it).
>
> **Merge only after both:**
> 1. riscv/docs-resources#246 is merged, **and**
> 2. this repo's `docs-resources` submodule is bumped to a commit that includes it.
>
> Ideally fold the submodule bump into this PR (or land it first), then rebase.

The tags backend now emits a trailing newline directly (riscv/docs-resources#246), so `update-ref` no longer needs to append one.

The former fix-up used `tail -c 1` in a command substitution; its original BSD `sed -i .bak` form failed on GNU sed, so the only documented way to regenerate the reference file worked on macOS only — one of the factors behind `ref/` drifting out of date (riscv/docs-resources#245, item 3).

### Validation
Previewed by pointing this repo's `docs-resources` submodule at a branch containing docs-resources#246: `make update-ref` produces a `ref/riscv-spec-norm-tags.json` **byte-for-byte identical** to the committed one, and `make check-ref` passes — i.e. with #246 in place the removed `tail -c 1` step is provably a no-op.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -177,15 +177,10 @@ check-xref-fallbacks: $(DOCS_HTML)
 
 # Regenerate the checked-in normative tags JSON in $(REF_DIR) from the sources.
 #
-# The tags backend (docs-resources/converters/tags.rb) emits JSON with no
-# trailing newline, but pre-commit's end-of-file-fixer requires one, so append
-# it when it is missing. Inside a command substitution `tail -c 1` yields the
-# empty string exactly when the file already ends in a newline.
+# The tags backend (docs-resources/converters/tags.rb) emits a trailing newline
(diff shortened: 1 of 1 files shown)


jordancarlin (approved):
I reviewed the docs-resources changes, and they look fine. Once they are merged and the submodule is updated, this LGTM.

rpsene (approved):

Issue #2613, opened 2026-01-27
How to match tlb entry under  vsatp.MODE == BARE and PMM on?


AlexGJL:
1.when vsatp.mode==BARE, the effective virtual address is 50bits([49:0]) for sv48, should bits[49:48] re-fill in tlb, and join match-checking for tlb access? should check bits[49:48] when sfence.vma/hfence.vvma/hfence.gvma flsuh tlb entry?
2.both vsatp.MODE == BARE and PMM on, bits[49:48] are both in pmm feild when PMLEN=XLEN-48, and virtual address feild, should bits[49:48] re-fill in tlb, and join match-checking for tlb access? should check bits[49:48] when sfence.vma/hfence.vvma/hfence.gvma flsuh tlb entry?

aswaterman:
The answer depends to some extent on your microarchitecture.  But in a conventional design, yes, you will need bits 49:48 in the TLB tag.  You'll also probably need a field in the TLB entry indicating that this is an entry corresponding to vsatp.MODE=Bare, so that you know not to match on this entry when vsatp.MODE != Bare or V=0.

SFENCE.VMA when V=1 doesn't have any effect on TLB entries established for V=1 and vsatp.MODE=Bare (and therefore HFENCE.VVMA doesn't, either).  HFENCE.GVMA does.

AlexGJL:
thanks for your reply. 
1. bits[49:48] are common field for PMM and  virtual address when PMLEN=XLEN-48 and pmm==on，how to use the common field？ should bits[49:48] be masked when tlb be refilled or do entry match？
2. Based on your reply, my understanding is as follows：a. bare entry is total different from non-bare entry，bare entry only be refilled and cam when MODE = Bare and V=1，and use HFENCE.GVMA to flush this entries, these tlb entries vaild va width is 50bits; b. non-bare entries‘ va width is 48bits ，refiling，caming and SFENCE.VMA/HFENCE.VVMA/HFENCE.GVMA match 48 bits is enough. is these right?

aswaterman:
1. If you implement pointer masking, you need to correctly implement the pointer-masking spec.  There are a variety of techniques, including explicitly masking the address bits before TLB access, or disabling matching on those two TLB tag bits.  I'll leave the uarch details as an exercise to the reader.
[...]

PR #3255, opened 2026-07-18
Clarify PMM fence handling

abhinavagarwal07:
Clarifies pointer masking behavior, following #2613:

- Extends the GPA masking and `HFENCE.GVMA` guidance (for `vsatp.MODE=Bare`) to `hstatus.HUPMM`.
- Specifies in the supervisor chapter that `senvcfg.PMM` changes take effect immediately without `SFENCE.VMA`, avoiding TLB flushes on context switches.

### DIFF
--- a/src/priv/hypervisor.adoc
+++ b/src/priv/hypervisor.adoc
@@ -2135,6 +2135,9 @@ supervisor physical addresses. When running with virtualization in VS/VU mode
 with csr:vsatp[mode]=Bare, this means that those qty:2[bits] may be subject to
 pointer masking, depending on csr:hgatp[mode] and csr:senvcfg[pmm] or csr:henvcfg[pmm] (for
 VU/VS mode). If csr:vsatp[mode]{ne}Bare, this issue does *not* apply.
+Those qty:2[bits] may likewise be subject to pointer masking according to
+csr:hstatus[hupmm] for explicit memory accesses made by `HLV.\*` or `HSV.*`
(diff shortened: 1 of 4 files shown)


src/priv/hypervisor.adoc
```diff
@@ -2147,10 +2150,16 @@ to be flushed when PMLEN changes in a way that may affect these duplicate
 entries.
 ====
 
+Since VS-mode cannot execute `HFENCE.GVMA`,
+[#norm:senvcfg_pmm_gstage_cache]#implementations must ensure that changes to
+`senvcfg.PMM` take effect for subsequent VU-mode explicit memory accesses
+without requiring an `HFENCE.GVMA` instruction.#
+
```
aswaterman: "VS-mode cannot execute HFENCE.GVMA" is not the reason.  The reason is that we didn't want to impose the overhead of a TLB flush on context switches.

This also isn't the right place to indicate these semantics, since it's an S-mode property, not a hypervisor property.

I suggest adding the following paragraph after the S-mode paragraph that describes the PMM field:

Changing senvcfg.PMM takes effect immediately, without the need to execute an SFENCE.VMA instruction.

abhinavagarwal07: Sure. Will do.

aswaterman: Thanks!

aswaterman (approved):

Issue #402, opened 2019-06-28
Proposal to  enhancing HPM


NonerKao:
The current facility provided by RISC-V hardware performance monitors (HPM) requires more capabilities to support software tools like **perf** in Linux or **hwpmc** in BSDs.  In this proposal, we will discuss the deficiencies in the privileged spec v1.0, taking Linux perf as an example.  Further, we will share our solution as a reference.

### Perf Sampling in RISC-V

**Scenario:** A user want to sample a program every 1000000 cycles in U-mode and S-mode

**Input:** `period = 1000000; events = [cycles]; modes = [u, s]`

**Output:** A trace of sampled pcs of the program

1. Set a counter to its maximum value minus `period`
2. Set the corresponding event mask `events`
3. Set the corresponding mode selector `modes`
4. Start the target program
5. Increase the counter by one each time the events marked by `events` happens in privileged modes `modes`
6. Trigger an interrupt when the counter overflows
7. Sample current pc
8. Reset counter as step 2
9. Resume the program, go to step 4

### Problems

* (Functional) No counter-triggered interrupt
* (Functional) No mode selection
* (Performance) No writable HPM counters/controls in S-mode
Currently, HPM CSRs are only writable in M-mode, so perf must make SBI calls to set those CSRs, which obviously causes large overheads. 

### Implementation in Our Prototype

We enhance RISC-V HPM by adding critical CSRs.  Note that all following CSRs have the same length as `*counteren` registers:

|CSR name|Description|Related to Problem|
|---|---|---|
|`mcounterwen`|Each counter has one bit to allow/deny write accesses to HPM CSRs from S-mode.  The firmware/M-mode initialization can thus turn this register on during boot time so that the OS can directly manipulate any required CSRs.|write access from S-mode|
|`[m\|s]countermask_[m\|s\|u]`|Each counter has three mode-selection mask bits (u-/s-/m-mode) that profiling software can manipulate.|mode selection|
|`[m\|s]counterinen`(\*)|Each counter has one bit indicating to or not to trigger an interrupt when an overflow happens.|counter-triggered interrupt|
|`[m\|s]counterovf`(\*)|In ISR, scanning this CSR to know which counters have overflowed.|counter-triggered interrupt|
|`scounterinhibit`|While we have `mcounterinhibit` in current spec, we implement this as an alias that can be directly controlled by S-mode software once the corresponding bits in `mcounterwen` is set .|others|
[...]

PR #3269, opened 2026-07-23
Zijf (Far jumps and calls)

aswaterman:
This proposed extension adds two new instructions, JF (Jump Far) and JFAL (Jump Far and Link), which improve code density and performance for large programs by replacing a 64-bit AUIPC+JALR sequence with a single 32-bit instruction.


wangpc-pp:
Some questions here:
1. We had some discussions in https://lists.riscv.org/g/tech-unprivileged/topic/119028439 and the conclusion is:
> Note that fusion of an AUIPC/JALR (or LUI/JALR) instruction pair provides a much larger jump range.  Even without fusion, this code idiom is the standard way to encode a larger jump range.  But without recognizing this idiom as a "far" JAL that doesn't need to use an indirect branch predictor, then one gets the extra pressure on your indirect branch predictor.

> Redefining the existing JAL is not on the table for compatibility reasons.  There are proposals to add 48-bit instructions that have greater displacement than JAL (32-ish bits), and proposals to add 32-bit instructions that have slightly greater displacement than JAL (24-ish bits).  But even if these proposals eventually go through, they are not of use for implementations being developed now, so I'd follow Greg's advice and optimize AUIPC/JALR.

2. According to our profiling of ByteDance's applications, we need at least a `28 bits` displacement to cover 90% scenarios. The JAL/JALR have already covered about 62%, and `23 bits`/`24 bits` only covers about 70%/72%.
3. Do we have a prototype to show the gains of code size and performance?

aswaterman:
Thanks for chiming in @wangpc-pp.

An additional 10% coverage is material, even if not earth-shattering.

To do any better we will need longer instructions. Doing as much as we can within the 32-bit encoding space is useful even if we eventually have 48-bit jump instructions, since these instructions are denser.

Also note that even implementations that optimize the AUIPC+JALR fusion pair will benefit from this extension, because of reduced code size and because a physical register allocation is skipped in the JF case.

src/unpriv/zijf.adoc
```diff
@@ -0,0 +1,37 @@
...
+
+insn:jf[] sets the `pc` to `pc` + `offset`.
+insn:jf[] can target a range of qty:±8[MiB].
+
+include::images/wavedrom/jf.edn[]
+
+insn:jfal[] writes the address of the instruction following the
+jump (`pc` + 4) into register `x1`.
+Then, insn:jfal[] sets the `pc` to `pc` + `offset`.
+insn:jfal[] can target a range of qty:±16[MiB].
```
Wren6991, line 16: The range +- 1 MiB is redundant  with the existing `jal`. This range is easily decoded: bits `31`, `19`, `14`, `13`, `11` are either all-zeroes or all-ones.

Should these encodings be marked as reserved or are they allocated for `jfal`?

Similar point applies to `jf` with bits `31`, `19`, `13`, `11`.

Wren6991: I can see it being simpler for the tools to not have to patch the opcode during relaxation.

To put some numbers to it: there are 2^21 redundant opcodes in total, which costs the sames as 64 new R-type instructions, or 1/2 of a new I-type.

aswaterman: @Wren6991 It's a valid point.  We had been approaching this from a different angle--we were looking at the possibility of expanding the ranges of the two instructions to ±9 MiB and ±17 MiB--but found that choice to be awkward.

I agree we should at least consider reserving the redundant cases, but if we do so, we might want to change the immediate encoding so that the high-order immediate bits are in a less inconvenient place.  (Right now they overlap the rs1 and rd register specifiers.)

Anyway, we'll think about it.

Wren6991: > Right now they overlap the rs1 and rd register specifiers

~~It varies a bit. When `31`, `14` (`jfal` only) and `13` are not all alike, you have completely free choice over `rs1`, `rs2` and `rd`. That's half of the `jf` and three quarters of the `jfal` encoding space, so 40 clean R-types.~~ Sorry, I had this backwards. The redundant opcodes have those bits all alike.

You're right though, I missed that, and maybe it could be avoided altogether with some shuffling.

brucehoult:
Why not make the target of these instructions address be a multiple of 16 or 64 bytes? That would increase the range from 16MB to 128MB or 512MB.

If someone is worried about the performance overhead of (LUI/AUIPC)+JALR then they are probably also prepared to align those important target addresses at least a little.

You'd need to discard the low bits of the PC before/after adding the offset.

bgrayson-mips:
I am on the fence on whether the increase in coverage is worth allocating more opcode space. It seems a heavy price to pay considering it doesn't provide **that** much more reach, especially for programs that may have instructions scattered across much of the 64b address space.

But a few thoughts:
- do we really need both jf and jfal? I would expect most jumps that go beyond 1MB are likely in another function, and eliminating jf would provide one more address bit (at the expense of eliminating tail-calls, but if you are jumping megabytes away, you probably have bigger performance issues than saving an occasional `ret`).
- this seems to overlap somewhat with `cm.jt`, which already saves even more opcode space and doesn't require a temporary register. It has some issues though as a replacement for jf/jfal:

1. It requires the Zcmt extension, which reuses opcodes that may collide with other C extension flavors that one may want to implement (Zcd).
2. Even though cm.jt doesn't require a temporary register and saves even more opcode space, it can only handle 256 different destinations.
3. cm.jt also requires extra memory accesses to provide its reach. But this is solvable microarchitecturally -- see below.

So my counter-proposal strawman (based on 5 minutes' worth of thought, so take it with a grain of salt!) is to create a different extension ("Zijt"?) that uses the same approach as Zcmt, except that it provides a larger table index, and is a 32b opcode. This maintains the space savings of Zijf, avoids the temporary register of `auipc/jalr`, expands the reach from 25b to 64b(!!!!) (or even 128b for RV128), is similar to something that is already ratified and "thought through" (reducing toolchain effort and overall risk), and does all of this while utilizing **less** opcode space than Zijf.

The performance impact of the extra memory access can be avoided microarchitecturally by having a Zijt predictor that can be sized according to the implementation's choice.

@wangpc-pp do you know how many distinct destinations were observed in each program in your analysis? We could use several `jvt` modes to specify several different table sizes if something like 1024 would not provide enough coverage, or possibly just specify a single fairly large table size (16K entries?) based on what analysis shows.

XIVN1987:
Rather than adding a new instruction, it is better to add a switch that extends the immediate field of the existing jal instruction to 24 bits. As for compatibility concerns, this switch could be mandated to only be supported in domains where compatibility requirements are low, such as the microcontroller space. Domains with stringent compatibility requirements, like RVA, must not be allowed to support this switch.

wangpc-pp:
> @wangpc-pp do you know how many distinct destinations were observed in each program in your analysis?

I didn't do this kind of profiling, but based on the size of binary artifacts (they are at GB level), I guess the number won't be small. :-)

src/unpriv/zijf.adoc
```diff
@@ -0,0 +1,37 @@
...
+
+include::images/wavedrom/jfal.edn[]
+
+[NOTE]
+====
+To reduce hardware cost, offset bits 11{endash}1 and 19{endash}15 are in the
```
tariqkurd-repo, line 22: The {endash} renders very strangely, why not write this as [11:1] and [19:15}?

tariqkurd-repo: Image

aswaterman: We use `endash` elsewhere in the spec, so we're better off fixing how it renders rather than avoiding it.  I'll fix that on the main branch.

mehnadnerd:
I don't believe JFAL should be added because it makes decoding the destination register more complicated. I believe this creates the first case where a 32-bit instruction has rd in a field other than [11:7]. Doing so directly lengthens the timing paths for decoding register destinations and CAMing for the previous writer to a register for rename.

I don't think JF has any of these issues.

tariqkurd-repo:
> I don't believe JFAL should be added because it makes decoding the destination register more complicated. I believe this creates the first case where a 32-bit instruction has rd in a field other than [11:7]. Doing so directly lengthens the timing paths for decoding register destinations and CAMing for the previous writer to a register for rename.
> 
> I don't think JF has any of these issues.

This is certainly true if C isn't implemented, as soon as C _is_ implemented the decode is more complex anyway.

Wren6991:
> I believe this creates the first case where a 32-bit instruction has rd in a field other than [11:7].

A minor example is RV32 `ld` from Zilsd, which writes back to both `{instr[11:8], 1'b0}` and `{instr[11:8], 1'b1}`. I can't think of any non-Zca, non-register-pair examples though.

src/unpriv/zijf.adoc
aswaterman: OK, fixed.

chandlerc:
Just wanted to share a few observations about this extension, both based on my experience optimizing Google's software for different ISAs, and working extensively on LLVM over the years:

First and foremost, I do think it is highly valuable to expand the range reachable. While there will be cases larger than what is reachable here, more coverage remains good.

We've also definitely seen single function bodies larger than 16MiB. And when doing profile-based hot/cold partitioning, conditionally branching to a different section is pretty common. That said, if the conditional variant is especially more problematic, its not the end of the world to emulate with the unconditional.

I'd still like to see a 64-bit encoding with a _very_ large offset (ideally >40-bits, maybe more) to give us a solid backstop. But that doesn't obviate the utility here.

Some specific improvements I would suggest here:

- First and foremost, I would suggest making the destination the 4-byte aligned address computed by roughly `(pc & ~0b11) + (offset << 2)`. Getting the range to +/- 64MiB (if I did my math right) seems _very_ valuable. TBH, I wonder if it'd be worth making the alignment larger just for the increased range. aligning functions to 4- or 8- bytes isn't the worst thing, and would give you single instruction coverage of many more binaries.
- Second, if you're not going to reserve the +/- 1MiB range, you should compute the offset as +/- (offset + 1MiB) to get an extra 1MiB range in both directions. FWIW, I wouldn't expect it to be hard for linkers to manage this -- I'd expect the compiler to emit a relocatable pseudo-instruction, and the linker to be able to update any bits needed, for example switching between the two encodings based on whether > or <= 1MiB displacement.

jnk0le:
> (pc & ~0b11) + (offset << 2). Getting the range to +/- 64MiB (if I did my math right) 

32MiB, `jal`/`jfal` offsets are already shifted by 1. Only `jalr` contains imm0.

lenary:
I was asked about this on the Long Instructions TG Call, especially with regards to whether the PC-relative offset should have more than a 2-byte alignment (in order to get more range).

I said on the call that I thought it was difficult. Thinking about it over the last 24 hours, I'm convinced it's not possible.

I'm not going to rehash that presentation again in depth. For the medium code models, the key points are:
- All calls (including tail calls) are emitted as a `AUIPC`+`JALR`, which has PC±2GiB range by the compiler, to match the full range of the medium code models. These code models do not use range extension thunks.
- All (unconditional) jumps within functions are currently emitted by the compiler as a `JAL`, which has PC±1MiB range. In general an `AUIPC`+`JALR` pair cannot be used for a longer jump within a function, because the relocation on that pair may go via the PLT, which will clobber the `t*` registers. This PC±1MiB range is the limiting factor on large functions - functions larger than 1MiB cannot jump from any block to any other block, so you begin to hit linking issues.

With this far jumps extension:
- We would emit jumps within functions using `JF`, which I think is proposed to have a PC±8MiB range, which increases the function size before you start running into problems to 8MiB.
- Calls would still be emitted as `AUIPC`+`JALR` (because they need to match the code model), and we would design a linker relaxation to turn this into `JFAL` when things are in range.

If the `JFAL` offset has a higher alignment requirement than to IALIGN, relaxation will never converge for that instruction. This is because any other relaxation (they delete multiples of 16 bits) can bring the PC-relative offset in and out of alignment. Over-aligning functions doesn't help these offsets converge at all, because there's no guarantee that the `JFAL` instruction is aligned any more than to the nearest 16 bits, the the offset is PC-relative. If `JFAL` has complex requirements on the PC-relative offset, linkers will just not implement the linker relaxation, and there will be no improvement in code density in linked binaries.

It is not a problem for relaxation convergence if the offset range limits are not nice powers of two, because during relaxation most PC-relative offsets will stay the same or get smaller.

How the large code model works is different, but it is not compatible with PIC/PIE/dynamic linking, so is not particularly relevant to conventional Linux systems. That code model uses literal pools, which is quite different to the operation of Zijf.

tariqkurd-repo:
As this proposal uses up 6 of the remaining values in [14:12] of the JALR encoding:

000: JALR
010: reserved
1x0: JF
xx1: JFAL

is there any plan for the final reserved encoding?

aswaterman:
> is there any plan for the final reserved encoding?

Just keeping it reserved for now, out of prudence.

tariqkurd-repo:
> > is there any plan for the final reserved encoding?
> 
> Just keeping it reserved for now, out of prudence.

In the CHERI TG we're discussing a new type of JALR combined with a speculation barrier similar to TEB (https://github.com/riscv/riscv-spec-barrier/blob/main/src/zisb-insn.adoc) but maybe not quite the same semantics - we're still working that one through. Essentially it would be a "look before you leap" JALR to be careful about speculating the target because it's jumping into a different secure compartment. In the CHERI case we don't need an offset (as it's a sentry) but for general RISC-V then you would.

If TEB+JALR was an interesting concept for general RISC-V maybe it could go in that slot, if not then we can keep the proposal CHERI specific.

aswaterman:
That variant of JALR probably does not deserve an immediate offset, and in any event it's not clear that it's better than just using a TEB + JALR.  Anyway, that can be a topic for another day.

hlb194802:
We have been looking at the same problem from the OpenJDK side. The proposed
±16 MiB JFAL range would already be helpful, but it may still be short for
large JIT code caches.

One option we would appreciate feedback on is splitting the current `xx1`
JFAL encoding space:

```text
funct3[2:0]

000    JALR
010    reserved
1x0    JF
x01    JFAL32    32-byte displacement granularity
x11    JFAL       2-byte displacement granularity
```

Both variants use a 23-bit signed immediate and write the return address to
`x1`:

```text
JFAL:
    x1 = pc + 4
    pc = pc + (sext(imm23) << 1)
    range: [-8 MiB, +8 MiB - 2]

JFAL32:
    x1 = pc + 4
    pc = pc + (sext(imm23) << 5)
    range: [-128 MiB, +128 MiB - 32]
```

This would require no new major opcode and would leave `funct3=010` reserved.
The tradeoff is reducing ordinary JFAL from ±16 MiB to ±8 MiB in exchange for
a ±128 MiB variant.

## OpenJDK motivation

OpenJDK places JIT-compiled methods in code caches whose default maximum size
is approximately 240 MiB; some large production applications use substantially
more. Method entry points may also change because of tiered compilation and
deoptimization, so OpenJDK uses patchable call sites and trampolines. When a
target is outside direct-call range, execution takes a slower path with an
additional jump, memory load, and indirect branch.

The relevant ranges are:

| Instruction | Range |
|---|---:|
| RISC-V JAL | ±1 MiB |
| Current Zijf JFAL proposal | ±16 MiB |
| Proposed JFAL32 | ±128 MiB |
| AArch64 BL | ±128 MiB |

JFAL32 would not cover every call, particularly in very large code caches,
but it could make many more calls direct than a ±16 MiB instruction. Remaining
out-of-range calls could continue to use trampolines.

## Why ±128 MiB?

AArch64 `BL` uses a signed 26-bit immediate scaled by four, giving a 32-bit
instruction approximately ±128 MiB of PC-relative reach.

An earlier OpenJDK experiment on AArch64 restricted direct calls from the
normal ±128 MiB range to ±1 MiB, forcing more calls through trampolines. The
reported SPECjbb2015 results were:

| System | ±128 MiB | Restricted to ±1 MiB | Regression |
|---|---:|---:|---:|
| Raspberry Pi 4B | 1186 | 1153 | −2.74% |
| Kunpeng 920 | 66287 | 64134 | −3.25% |
| Arm Neoverse N1 | 15487 | 15288 | −1.29% |

These results do not directly predict JFAL32 performance on RISC-V, but they
show that direct-call reach can measurably affect Java workloads. The original
data and OpenJDK trampoline analysis are in
[RISC-V SIG-J message #402](https://lists.riscv.org/g/sig-j/message/402).

## Scope and alignment

JFAL32 requires:

```text
(target - pc) % 32 == 0
```

We understand the linker-relaxation convergence concern raised in this thread.
JFAL32 would not be intended as a general linker-relaxation target. Its main
use case would instead be JIT-generated code, where the runtime knows the final
addresses and controls method alignment and call-site placement. A JIT could
select `JAL`, `JFAL`, `JFAL32`, or a trampoline according to the final range
and alignment, while calls that do not satisfy the constraint would continue
to use the existing fallback path. Any padding introduced to place JFAL32
calls would remain part of the code-size tradeoff.


## References

- [Zijf PR #3269](https://github.com/riscv/riscv-isa-manual/pull/3269)
- [Original SPECjbb2015 data and OpenJDK trampoline analysis](https://lists.riscv.org/g/sig-j/message/402)
- [Complete SIG-J discussion: “Considering RISC-V's JAL”](https://lists.riscv.org/g/sig-j/topic/considering_risc_v_s_jal/95108467)
- [Arm architectural definition of BL](https://developer.arm.com/documentation/ddi0602/2021-12/Base-Instructions/BL--Branch-with-Link-)

Wren6991:
> This is because any other relaxation (they delete multiples of 16 bits) can bring the PC-relative offset in and out of alignment. Over-aligning functions doesn't help these offsets converge at all, because there's no guarantee that the JFAL instruction is aligned any more than to the nearest 16 bits, the the offset is PC-relative

@lenary: I think this is resolved by aligning the sum as well as the addend? So rather than `pc + imm`, where `imm % 4 == 0`, instead calculate as `(pc + imm) & -4` with the same restriction on `imm`. That would make any 4-byte-aligned address reachable from any 2-byte-aligned PC, and has ~zero incremental hardware cost.

The alignment restriction is then just "targets are 4-byte aligned" and my understanding is the linker already handles similar cases by eliding NOPs during relaxation.

lenary:
> > This is because any other relaxation (they delete multiples of 16 bits) can bring the PC-relative offset in and out of alignment. Over-aligning functions doesn't help these offsets converge at all, because there's no guarantee that the JFAL instruction is aligned any more than to the nearest 16 bits, the the offset is PC-relative
> 
> Sam Elliott (@lenary): I think this is resolved by aligning the sum as well as the addend? So rather than `pc + imm`, where `imm % 4 == 0`, instead calculate as `(pc + imm) & -4` with the same restriction on `imm`. That would make any 4-byte-aligned address reachable from any 2-byte-aligned PC, and has ~zero incremental hardware cost.
> 
> The alignment restriction is then just "targets are 4-byte aligned" and my understanding is the linker already handles similar cases by eliding NOPs during relaxation.

Sure. This would line up with Chandler's proposal, but is not what I was asked about on the Long Instruction TG Call.

jnk0le:
> An earlier OpenJDK experiment on AArch64 restricted direct calls from the normal ±128 MiB range to ±1 MiB, forcing more calls through trampolines. The reported SPECjbb2015 results were:
> System 	±128 MiB 	Restricted to ±1 MiB 	Regression
> Raspberry Pi 4B 	1186 	1153 	−2.74%
> Kunpeng 920 	66287 	64134 	−3.25%
> Arm Neoverse N1 	15487 	15288 	−1.29%
> 
> These results do not directly predict JFAL32 performance on RISC-V, but they show that direct-call reach can measurably affect Java workloads. The original data and OpenJDK trampoline analysis are in [RISC-V SIG-J message #402](https://lists.riscv.org/g/sig-j/message/402).

What is the regression against +-16MiB or 32MiB range?

Forcing 4 byte alignment is not that invasive as 2+ MiB binaries are already aligning all function  entries (and even loops) to 4 bytes. Any more might be problematic.

hlb194802:
> > An earlier OpenJDK experiment on AArch64 restricted direct calls from the normal ±128 MiB range to ±1 MiB, forcing more calls through trampolines. The reported SPECjbb2015 results were:
> > System 	±128 MiB 	Restricted to ±1 MiB 	Regression
> > Raspberry Pi 4B 	1186 	1153 	−2.74%
> > Kunpeng 920 	66287 	64134 	−3.25%
> > Arm Neoverse N1 	15487 	15288 	−1.29%
> > These results do not directly predict JFAL32 performance on RISC-V, but they show that direct-call reach can measurably affect Java workloads. The original data and OpenJDK trampoline analysis are in [RISC-V SIG-J message #402](https://lists.riscv.org/g/sig-j/message/402).
> 
> What is the regression against +-16MiB or 32MiB range?
> 
> Forcing 4 byte alignment is not that invasive as 2+ MiB binaries are already aligning all function entries (and even loops) to 4 bytes. Any more might be problematic.

It may be helpful to put the SPECjbb2015 result in context. The [OpenJDK AArch64 code-cache study](https://mail.openjdk.org/pipermail/hotspot-dev/2022-February/057692.html) reports average maximum usage of 12,491 KiB in the non-profiled heap, 13,968 KiB in the profiled heap, and 4,274 KiB in the non-nmethod heap—about 30 MiB in total.
With a code footprint of this size, a ±32 MiB reach should be sufficient under a compact layout, while ±16 MiB could also reasonably be expected to cover most call sites. It would therefore not be surprising if SPECjbb2015 showed little or no performance regression with either range.
However, SPECjbb2015 is a benchmark and may not fully represent large, long-running enterprise applications. For example, Atlassian has reported that Jira deployments with many plugins can exhaust the default 240 MiB code cache, resulting in cache flushing and significant performance degradation. Jira subsequently adopted a 512 MiB reservation ([JRASERVER-66796](https://jira.atlassian.com/browse/JRASERVER-66796)).
Therefore, a lack of regression on SPECjbb2015 would be encouraging, but it may not tell us whether ±16 MiB or ±32 MiB is sufficient for substantially larger enterprise workloads. These larger deployments are the main reason for considering a ±128 MiB form.

aswaterman:
Speaking for only for myself, not on behalf of the Unpriv IC: I don't think we should add a 32-byte-aligned jump in the 32-bit encoding space.  Calls of that distance are better served by 48-bit instructions that don't need to impose such a strict alignment requirement.  I do think the 4-byte alignment for JFAL (but not JF) is a possibility, though not a slam dunk.

PR #3276, opened 2026-07-27
Normative-rule keyword/tag fixes requiring Andrew's review

UmerShahidengr:
As part of the normative-keyword review led by @allenjbaum and @mmhus (tracking sheet "[keyword_matches](https://docs.google.com/spreadsheets/d/1mqHjJRukAbOGoicweirfzAxXE1ydsMAxhEBQlTSdHUQ/edit?usp=sharing)"),
these are the changes marked **"WG: Needs fixing" = TRUE** and **"Needs Review by AW" = TRUE**. The Workgroup requested these changes; they are drafted here so @aswaterman can review the concrete diff.

Scope: `can/should` → `will/are/is` keyword tightening, a WG-supplied rewrite, sentence reordering, and tag split/add operations (with matching YAML defs).

## Changes applied
| Chapter | File | Tag | Fix (WG request) |
|---|---|---|---|
| F extension | `f-st-ext.adoc` | `fflags_frm_op_sz_acc` | Reworded the opening sentence to acknowledge `frm`/`fflags` exist as standalone CSRs. |
| Hypervisor | `hypervisor.adoc` | `hstatus_hu_op` | "can be executed" → **"will be executed"**. |
| Hypervisor | `hypervisor.adoc` | `hvip_sz_op` | WG-supplied text: "read/write register that a hypervisor can write" → **"WARL register that a hypervisor writes"**. |
| Hypervisor | `hypervisor.adoc` | `H_guest_page_fault` | "may be delegated" → **"are delegated"**; tag narrowed to the 1st (normative) sentence — sentences 2 (parameter) and 3 (documentation) pulled out of the rule. |
| Hypervisor | `hypervisor.adoc` | `sfence_vma_v1` | "the effective ASID can be considered to be" → **"the effective ASID is"**. |
| Hypervisor | `hypervisor.adoc` | `H_cause_ecall` | "so they can be delegated separately" → **"so they are delegated separately"**. |
| Hypervisor | `hypervisor.adoc` | `H_tinst_addr_offset` *(new)* | Added a normative tag over the Addr.-Offset definition (excluding the mismatched last sentence). |
| Counter Delegation | `smcdeleg.adoc` | `smcdeleg_delegated_access` *(new)* | "can be read or written" → **"is read or written"**; "while `siselect` holds" → **"when `siselect` is"**; tagged the rule (the state-mapping table is not normatively tagged). |
| Smcntrpmf | `smcntrpmf.adoc` | `rv32_high_access` | Both "can be accessed" → **"are accessed"**. |
| Smctr | `smctr.adoc` | `ctr_trap_disabled_tgt` (context) | "how they can be recorded" → **"how they are recorded"** (the sentence was already outside the tag). |
| Smctr | `smctr.adoc` | `exttrap_def` | Reordered the 3rd sentence to 2nd and retagged so the rule covers sentences 1–2; "opt-in" → "enable(d)"; "By default" → **"When the `xctrctl.xTE` bits are clear"**. |
| Smctr | `smctr.adoc` | `sctrstatus_frozen_inhibit` *(new)* | Tagged the FROZEN sentence and removed the duplicated "This bit can be set by hardware …" reference. |

New tags (`H_tinst_addr_offset`, `smcdeleg_delegated_access`, `sctrstatus_frozen_inhibit`)
each have a matching definition in their chapter's `normative_rule_defs/*.yaml`.


src/priv/hypervisor.adoc
```diff
@@ -264,7 +264,7 @@ further in <<hinterruptregs>>.
 [[norm:hstatus_hu_op]]
 Field HU (Hypervisor in U-mode) controls whether the virtual-machine
 load/store instructions, HLV, HLVX, and HSV, can be used also in U-mode.
-When HU=1, these instructions can be executed in U-mode the same as in
+When HU=1, these instructions will be executed in U-mode the same as in
```
aswaterman: This change is incorrect.  HU=1 means the instructions are allowed to be executed in U-mode.  It doesn't force the instructions to be executed in U-mode.  "Can" is the correct verb.

src/priv/hypervisor.adoc
```diff
@@ -477,8 +477,8 @@ Store/AMO guest-page fault
 ==== Hypervisor Interrupt (`hvip`, `hip`, and `hie`) Registers
 
 [[norm:hvip_sz_op]]
-Register `hvip` is an HSXLEN-bit read/write register that a hypervisor
-can write to indicate virtual interrupts intended for VS-mode. Bits of
+Register `hvip` is an HSXLEN-bit *WARL* register that a hypervisor
```
aswaterman: Undo this change.  The fields have their own WARL etc. designation; the register as a whole is correctly described as "read/write".

src/priv/hypervisor.adoc
```diff
@@ -477,8 +477,8 @@ Store/AMO guest-page fault
...
+writes to indicate virtual interrupts intended for VS-mode. Bits of
```
aswaterman: Undo this change. "Can write" is a clearer description as it describes what software is allowed to do, not what software must do.

src/priv/hypervisor.adoc
```diff
@@ -2041,10 +2041,9 @@ whether this usage will ever become common, we chose not to preclude it.
 
 ==== Guest-Page Faults
 
-[[norm:H_guest_page_fault]]
-Guest-page-fault traps may be delegated from M-mode to HS-mode under the
+[#norm:H_guest_page_fault]#Guest-page-fault traps are delegated from M-mode to HS-mode under the
```
aswaterman, line 2044: Undo this change.  The old wording is clearer as this sentence describes what software is allowed to do, not what must happen.

src/priv/hypervisor.adoc
```diff
@@ -2082,7 +2081,7 @@ When V=1, the virtual-address argument to SFENCE.VMA is a guest virtual
 address within the current virtual machine, and the ASID argument is a
 VS-level ASID within the current virtual machine. The current virtual
 machine is identified by the VMID field of CSR `hgatp`, and the
-effective ASID can be considered to be the combination of this VMID with
+effective ASID is the combination of this VMID with
```
aswaterman: The old wording is better because this is describing a notional concept, not a formal ISA property.

src/priv/hypervisor.adoc
```diff
@@ -2321,7 +2320,7 @@ _Reserved_
 |===
 
 [[norm:H_cause_ecall]]
-HS-mode and VS-mode ECALLs use different cause values so they can be
+HS-mode and VS-mode ECALLs use different cause values so they are
```
aswaterman: The old wording is better because it is describing a capability that software is allowed to take advantage of.

src/priv/smcdeleg.adoc
```diff
@@ -36,11 +36,11 @@ The Smcdeleg and Ssccfg extensions both depend on the Sscsrind extension.
...
 privilege mode with read access to select counters.[#norm:smcdeleg_cde_en]#When the Smcdeleg/Ssccfg extensions are enabled (`menvcfg`.CDE=1), it further allows M-mode to delegate select counters to S-mode.#
 
-The `siselect` (and `vsiselect`) index range 0x40-0x5F is reserved for
+[#norm:smcdeleg_delegated_access]#The `siselect` (and `vsiselect`) index range 0x40-0x5F is reserved for
 delegated counter access. When a counter _i_ is delegated
 (`mcounteren`[_i_]=1 and `menvcfg`.CDE=1), the register state associated
-with counter _i_ can be read or written via `sireg*`, while `siselect` holds
-0x40+__i__. The counter state accessible via alias CSRs is shown in
+with counter _i_ is read or written via `sireg*`, when `siselect` is
+0x40+__i__.# The counter state accessible via alias CSRs is shown in
```
aswaterman, line 43: Undo these wording changes for reasons similar to my other comments.

src/priv/smcntrpmf.adoc
```diff
@@ -37,7 +37,7 @@ Smcntrpmf remedies these issues by introducing privilege mode filtering for the
 
 [#norm:unimplemented_mode_bits]#For each bit in 61:58, if the associated privilege mode is not implemented, the bit is read-only zero.#
 
-[#norm:rv32_high_access]#For RV32, bits 63:32 of mcyclecfg can be accessed via the mcyclecfgh CSR, and bits 63:32 of minstretcfg can be accessed via the minstretcfgh CSR.#
+[#norm:rv32_high_access]#For RV32, bits 63:32 of mcyclecfg are accessed via the mcyclecfgh CSR, and bits 63:32 of minstretcfg are accessed via the minstretcfgh CSR.#
```
aswaterman, line 40: Undo these wording changes for reasons similar to my other comments.

src/priv/smctr.adoc
```diff
@@ -460,7 +460,7 @@ recorded, if anything, depends on whether the source privilege mode
 and/or target privilege mode are enabled for recording, and on the transfer type (trap
 or trap return).
 
-[#norm:ctr_trap_enabled]#Traps between enabled privilege modes are recorded as normal.#  [#norm:ctr_trap_disabled_src]#Traps from a disabled privilege mode to an enabled privilege mode are partially recorded, such that the `ctrsource`.PC is 0.# [#norm:ctr_trap_disabled_tgt]#Traps from an enabled mode to a disabled mode, known as external traps, are not recorded by default.#  See <<External Traps>> for how they can be recorded.
+[#norm:ctr_trap_enabled]#Traps between enabled privilege modes are recorded as normal.#  [#norm:ctr_trap_disabled_src]#Traps from a disabled privilege mode to an enabled privilege mode are partially recorded, such that the `ctrsource`.PC is 0.# [#norm:ctr_trap_disabled_tgt]#Traps from an enabled mode to a disabled mode, known as external traps, are not recorded by default.#  See <<External Traps>> for how they are recorded.
```
aswaterman, line 463: Undo this change because "can be" is more appropriate, as that section describes a facility that software can optionally use.

src/priv/smctr.adoc
```diff
@@ -504,7 +504,7 @@ _Consider an exception that traps from VU-mode to HS-mode, with `vsctrctl`.U=1 a
 
 ====== External Traps
 
-[#norm:exttrap_def]#External traps are traps from a privilege mode enabled for CTR recording to a privilege mode that is not enabled for CTR recording.  By default external traps are not recorded, but privileged software running in the target mode of the trap can opt-in to allowing CTR to record external traps into that mode. The `__x__ctrctl`.__x__TE bits allow M-mode, S-mode, and VS-mode to opt-in separately.#
+[#norm:exttrap_def]#External traps are traps from a privilege mode enabled for CTR recording to a privilege mode that is not enabled for CTR recording.  The `__x__ctrctl`.__x__TE bits allow M-mode, S-mode, and VS-mode to be enabled separately.#  When the `__x__ctrctl`.__x__TE bits are clear, external traps are not recorded, but privileged software running in the target mode of the trap can enable CTR to record external traps into that mode.
```
aswaterman, line 507: "allow M-mode, S-mode, and VS-mode to be enabled" makes it sound like this feature controls whether those privilege modes are available.  The correct wording would be "allow M-mode, S-mode, and VS-mode to each control whether external traps are enabled for CTR recording".

Or just undo this change, as the old wording seems fine to me.

src/priv/smctr.adoc
```diff
@@ -725,7 +725,7 @@ an implementation may opt to record CCV=0 for all calls, or those whose parent c
 
 ===== Freeze
 
-When `sctrstatus`.FROZEN=1, transfer recording is inhibited.  This bit can be set by hardware, as described below, or by software.
+[#norm:sctrstatus_frozen_inhibit]#When `sctrstatus`.FROZEN=1, transfer recording is inhibited.#
```
aswaterman, line 728: Why did you delete the second sentence?  It makes the spec more readable.

src/unpriv/f-st-ext.adoc
```diff
@@ -46,8 +46,9 @@ the original value into integer register _rd_, and then writing a new
 value obtained from integer register _rs1_ into csr:fcsr[].
 
 [[norm:fflags_frm_op_sz_acc]]
-The fields within the csr:fcsr[] can also be accessed individually through
-different CSR addresses, and separate assembler pseudoinstructions are defined
+The Rounding Mode and Accrued Exception Flags fields of csr:fcsr[] are also
+accessible individually as the csr:frm[] and csr:fflags[] CSRs at different CSR
+addresses, and separate assembler pseudoinstructions are defined
```
aswaterman, line 51: This one looks OK to me.

allenjbaum:
In general, it sounds like there  are a lot of sentences that describe what
software is permitted to do, rather than a requirement.
Is that a sign that say those sentences should be removed from a normative
rule tag?

src/priv/hypervisor.adoc
mmhus: WG: Agreed with Andrew

src/priv/hypervisor.adoc
mmhus: **Work Group (WG):** 
We should revert this to read/write but we need a normative statement to show that some of the bits are WARL. The image below:

Image

Fig 76. implies that it is all WARL but Fig 77. implies that it isn't. So there seems to be a discrepancy here.

src/priv/hypervisor.adoc
mmhus: WG: Agreed. Should be reverted. 

@UmerShahidengr 
`Bits of hvip that are not writable are read-only zeros.` should be a separate normative rule

src/priv/hypervisor.adoc
```diff
@@ -477,8 +477,8 @@ Store/AMO guest-page fault
 ==== Hypervisor Interrupt (`hvip`, `hip`, and `hie`) Registers
 
 [[norm:hvip_sz_op]]
-Register `hvip` is an HSXLEN-bit read/write register that a hypervisor
-can write to indicate virtual interrupts intended for VS-mode. Bits of
+Register `hvip` is an HSXLEN-bit *WARL* register that a hypervisor
+writes to indicate virtual interrupts intended for VS-mode. Bits of
 `hvip` that are not writable are read-only zeros.
```
mmhus, line 482: **WG feedback:** 
Another point here is that the spec isn't clear about the upper [HSXLEN-1:16] bits, whether they are reserved or allowed for custom platform use and whether they are writable. Based on that they could be writable not read-only. The spec in other places like mstatus, clearly mentions if bits are reserved or not.

src/priv/hypervisor.adoc
mmhus: **WG comments:**
`under the control of CSR` implies that this is not optional for the implementation. Software may choose to delegate, but the CSR always controls it.

src/priv/hypervisor.adoc
mmhus: **WG:** Agreed. Revert. But we want to move it to the end of the paragraph so that we can take it out of the normative rule.

src/priv/hypervisor.adoc
mmhus: **WG:**
Revert it, but also remove the normative rule. This should be tagged in the mcause table.

src/priv/smcdeleg.adoc
mmhus: **WG:**
The only way to read and write is via sireg when it is delegated. Software can choose to read it but the normative statement shows CSR control rather than software.

Image

mmhus (commented):
WG:
Feels it would be better to discuss these in a meeting directly with Andrew. Will send out an email.

aswaterman:
@mmhus it isn't practical to manage this as a single PR, given that we will be discarding some of it but not all of it.  For the parts that we are going to proceed with (and the parts that we are still debating), please create individual PRs and copy over the discussion so far, then close this one.

PR #3280, opened 2026-07-28
rvwmo: add Zicfiss source/destination register listings

abhinavagarwal07:
RVWMO's source/destination tables are meant to be complete but Zicfiss isn't in them. This adds the rows (sspush, sspopchk, ssrdp, ssamoswap, compressed forms) plus an ssp entry in the CSR granularity table, like the Zabha/Zacas/Zalasr change in riscv/riscv-isa-manual#3243.

Push/pop are marked to carry ssp->ssp only — the new ssp is just old ssp +/- XLEN/8, not the pushed value, so the Accumulating CSRs column (which would pull in the GPR) doesn't fit.

On sspopchk the update is gated on rs1 matching, so rs1->ssp is arguable; listed ssp->ssp only since the new ssp derives from the old one. See also riscv/riscv-isa-manual#2782.

### DIFF
--- a/src/unpriv/rvwmo.adoc
+++ b/src/unpriv/rvwmo.adoc
@@ -363,6 +363,7 @@ infinite sequence of other memory operations.
 |csr:fflags[] |Bits 4, 3, 2, 1, 0 |csr:fcsr[]
 |csr:frm[] |entire CSR |csr:fcsr[]
 |csr:fcsr[] |Bits 7-5, 4, 3, 2, 1, 0 |csr:fflags[], csr:frm[]
+|csr:ssp[] |entire CSR |
 |===
 
 Note: read-only CSRs are not listed, as they do not participate in the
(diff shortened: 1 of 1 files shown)


abhinavagarwal07:
gentle reminder @aswaterman

aswaterman:
Can you post a screenshot of the result?  At the moment, I don't have access to a machine on which I can build the document.

src/unpriv/rvwmo.adoc
```diff
@@ -363,6 +363,7 @@ infinite sequence of other memory operations.
 |csr:fflags[] |Bits 4, 3, 2, 1, 0 |csr:fcsr[]
 |csr:frm[] |entire CSR |csr:fcsr[]
 |csr:fcsr[] |Bits 7-5, 4, 3, 2, 1, 0 |csr:fflags[], csr:frm[]
+|csr:ssp[] |entire CSR |
```
abhinavagarwal07, line 366: Image

src/unpriv/rvwmo.adoc
```diff
@@ -704,6 +705,34 @@ register(s) to destination register(s) as specified
 5+| ^*^ insn:ld.{aq,aqrl}[] and insn:sd.{rl,aqrl}[] are RV64-only
 |===
 
+.Zicfiss Standard Extension
```
abhinavagarwal07, line 708: Image

abhinavagarwal07 (commented):
PFA rendered doc images

src/unpriv/rvwmo.adoc
```diff
@@ -704,6 +705,34 @@ register(s) to destination register(s) as specified
...
+
+|insn:ssamoswap.w[]† |_rs1_ ^A^, _rs2_ ^D^ |_rd_ | |
+
+|insn:ssamoswap.d[]†^*^ |_rs1_ ^A^, _rs2_ ^D^ |_rd_ | |
+
+5+| ^*^ insn:ssamoswap.d[] is RV64-only.
+
+5+| insn:c.sspush[] and insn:c.sspopchk[] require ext:zcmop[].
+
+5+| The insn:sspush[], insn:sspopchk[], and insn:ssrdp[] rows apply when xSSE=1; when Zicfiss is inactive these encodings follow their ext:zimop[]/ext:zcmop[] behavior.
```
aswaterman: I'd like to delete these notes.  We don't need to list the extensions that instructions belong to, and it creates a new maintenance headache: sometimes new extensions include existing instructions, and now there's a new list that needs to be updated.

Relatedly, we don't need to mention the xSSE constraint.  When it's 0, these instructions simply don't exist.  (And the Zimop/Zcmop spec already says that MOPs carry no syntactic dependencies unless overridden.)

LGTM with these deletions.  Thanks, @abhinavagarwal07!

abhinavagarwal07: Thanks for the feedback. Ready

aswaterman (approved):

PR #3286, opened 2026-07-29
Make the workflow callable from other repos

### LINKED ISSUE #3285: Make the dco_check.yml workflow callable by other repos.
In the spirit of not duplicating effort, this workflow should be callable by other repos.

wmat:
Add workflow_call attribute.

Signed-off-by: Bill Traynor wmat@riscv.org

### DIFF
--- a/.github/workflows/dco_check.yml
+++ b/.github/workflows/dco_check.yml
@@ -8,6 +8,7 @@ on:
       - reopened
       - synchronize
       - converted_to_draft
+  workflow_call:
 
 jobs:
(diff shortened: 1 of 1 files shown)


kersten1 (approved):

jordancarlin:
If we are making it callable by other repos, should we move the workflow to a common repository instead (maybe docs-resources) and call if from there?

wmat:
Probably.  I won't do that yet though until I see how rollout goes on the more active repos.

PR #3289, opened 2026-08-01
Fix Encoding of satp MODE field table formatting.

wmat:
Fixes the layout of the Encoding of `satp` MODE field table in the supervisor chapter.

Signed-off-by: Bill Traynor wmat@riscv.org


aswaterman (approved):
Thanks!

PR #3296, opened 2026-08-10
Zvbext: element-wise bit compression and expansion

wangpc-pp:
Zvbext provides VBCOMPRESS.VV/VX and VBEXPAND.VV/VX for rearranging bits independently within each vector element. VBCOMPRESS gathers the bits selected by a selection mask into contiguous least-significant positions, while VBEXPAND scatters contiguous least-significant bits into the positions selected by the mask. Unlike VCOMPRESS.VM, neither operation moves values between vector elements.

The vector-vector forms support a different selection mask for each element. The vector-scalar forms broadcast a selection mask from an integer register, using the standard RVV sign-extension and truncation rules. The operations cover the semantics of parallel bit extract and deposit and benefit bit packing, Base64, variable-length integer, and Morton-code workloads.

Zvbext depends on Zve32x and supports EEW=8, 16, and 32, with EEW=64 available when Zve64x is supported.

Supporting materials:
1. Applications: https://github.com/wangpc-pp/zvbext-applications
2. LLVM: https://github.com/wangpc-pp/llvm-project/tree/main-riscv-zvbext
3. QEMU: https://github.com/wangpc-pp/qemu/tree/master-zvbext

### DIFF
--- a/normative_rule_defs/zvbext.yaml
+++ b/normative_rule_defs/zvbext.yaml
@@ -1,0 +1,62 @@
+---
+# yaml-language-server: $schema=../docs-resources/schemas/defs-schema.json
+
+$schema: "../docs-resources/schemas/defs-schema.json#"
+
+chapter_name: Vector Bit Compression and Expansion Extension
+
+normative_rule_definitions:
+
+  - name: Zvbext_dependent_Zve32x
+    tags: ["norm:Zvbext_dependent_Zve32x"]
+
+  - name: Zvbext_eew
+    tags: ["norm:Zvbext_eew", "norm:Zvbext_eew_Zve64x"]
+
+  - name: Zvbext_selection_mask
+    tags: ["norm:Zvbext_selection_mask"]
+
+  - name: Zvbext_eew_emul
+    tags: ["norm:Zvbext_eew_emul"]
+
+  - name: Zvbext_bitwise
+    tags: ["norm:Zvbext_bitwise"]
+
+  - name: Zvbext_execution
+    tags: ["norm:Zvbext_execution"]
+
+  - name: Zvbext_overlap
+    tags: ["norm:Zvbext_overlap"]
+
+  - name: Zvbext_fixed_point_state
+    tags: ["norm:Zvbext_fixed_point_state"]
+
+  - name: vbcompress_mask
+    tags: ["norm:vbcompress_mask"]
+    kind: instruction
+    instances: [vbcompress.vv, vbcompress.vx]
+
+  - name: vbcompress_sew_rsv
+    tags: ["norm:vbcompress_sew_rsv", "norm:vbcompress_sew64_rsv"]
+    kind: instruction
+    instances: [vbcompress.vv, vbcompress.vx]
+
+  - name: vbcompress_op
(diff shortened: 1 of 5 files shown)


camel-cdr:
Thanks for pushing this proposal further.

A few comments that [I shared previously](https://lists.riscv.org/g/sig-vector/message/557) on the mailing list:

1. I think there should be a subextension for only the .vx variants, because the decoding of the mask to control bits for the butter-fly networks is the most expensive part and with .vx you'd only need one copy of that circuity.

2. We should also define vcompressflip and vexpandflip, which move the unmasked bits to the place right in inverted order, as this is the default behavior when implementing bit-compress/expand with butterfly networks.
These instruction variants can be used to do any 2^n bit permutation with n of these instructions. So any 64-bit bit permutation in six or fewer vcompressflip operations, this isn't possible with the non flip variants and you'd need to do `(vcompress(m) | vexpand(m) << 32)` six times instead.

vcompressflip:

image

vexpandflip:

image

* vcompress(x,mask) == vcompressflip(x,mask)&mask
* vexpand(x,mask) == vexpandflip(x,mask)&mask

2.5. I think it would be valuable to have the sail code use the butterfly networks or have a non-normative note that this can be implemented by butterfly networks, because that may not be obvious to implementers and it otherwise may look very expensive to implement. See also https://github.com/clairexen/edu-sag

3. I'd love to also get vbmatflip (8x8 bit-matrix transpose), which is very useful for things like [histogramming](https://gitlab.com/-/snippets/3745720) ([huge speedup](https://xcancel.com/CamelCdr/status/1833566218222977329)), [inverse vrgather](https://87flowers.com/inverse-pshufb/). I mention this here, because vbmatflip can be implemented by chaining the bfly and ibfly networks that are used to implement vcompress/vexpand.

wangpc-pp:
@camel-cdr Thanks, these are useful points!

I agree that the mask decoder is a major implementation cost and that the `.vx` forms offer an important opportunity to decode once and broadcast the steering controls. I would like to collect implementation/PPA feedback before deciding whether this warrants a separately named `.vx`-only subset, since a low-area implementation can also serialize mask decoding for `.vv` while still implementing the complete extension.

The gather-flip/scatter-flip operations are interesting and the arbitrary-permutation property is compelling. I think they need a precise semantic definition and workload/compiler analysis before being added to this PR (currently I have no idea how these two instructions can be used separately). BTW, under the conventional compress-flip definition, the compress operation is recovered as `compress_flip(x & m, m)`, while `expand(x, m) = expand_flip(x, m) & m` (the first identity in the comment appears to need correction).

I agree that the specification should include a non-normative note about butterfly/inverse-butterfly implementations and mask-decoder sharing. I will do that (and may also the SAIL code).

`vbmatflip` also looks useful, but I suggest treating it as a separate follow-up proposal because its semantics and software mapping are distinct from generic `pext/pdep`.

----

This PR is just for public tracking. I will advertise it in vector-sig in a few days and restart the discussion in this [thread](https://lists.riscv.org/g/sig-vector/topic/113206944). :-)

aswaterman:
Please do not propose ISA extensions as PRs to the ISA manual. They need to go through RVIA processes first.

PR #3297, opened 2026-08-10
Zvmatch: vector element-set matching

wangpc-pp:
Zvmatch defines VMMATCH.VX and VMMATCH.VV to classify 8-bit and 16-bit vector elements against a nonempty key set and produce a standard vector mask.

VMMATCH.VX interprets an integer register as XLEN/SEW packed keys. It provides a bounded path for small byte and halfword sets. VMMATCH.VV compares each active query element against the complete VLMAX key domain in a vector register group, allowing larger sets without introducing separate key-length state.

The instructions support tokenizer boundary detection, delimiter and symbol classification, and similar workloads that otherwise require repeated equality comparisons and mask OR operations. They retain the base vector model for masking, tail handling, vstart, exceptions, and restart.

The provisional OP-V encodings use funct6=110100 in the OPIVX and OPIVV formats.

Supporting materials:
1. LLVM: https://github.com/wangpc-pp/llvm-project/tree/main-riscv-zvmatch
2. QEMU: https://github.com/wangpc-pp/qemu/tree/master-zvmatch

### DIFF
--- a/normative_rule_defs/zvmatch.yaml
+++ b/normative_rule_defs/zvmatch.yaml
@@ -1,0 +1,100 @@
+---
+# yaml-language-server: $schema=../docs-resources/schemas/defs-schema.json
+
+$schema: "../docs-resources/schemas/defs-schema.json#"
+
+chapter_name: Vector Element-Set Matching Extension
+
+normative_rule_definitions:
+
+  - name: Zvmatch_dependent_V
+    tags: ["norm:Zvmatch_dependent_V"]
+
+  - name: Zvmatch_sew
+    tags: ["norm:Zvmatch_sew"]
+
+  - name: Zvmatch_lmul
+    tags: ["norm:Zvmatch_lmul"]
+
+  - name: Zvmatch_bit_pattern
+    tags: ["norm:Zvmatch_bit_pattern"]
+
+  - name: vmmatch-vx_sew_rsv
+    tags: ["norm:vmmatch-vx_sew_rsv"]
+    kind: instruction
+    instance: vmmatch.vx
+
+  - name: vmmatch-vx_key_extraction
+    tags: ["norm:vmmatch-vx_key_extraction"]
+    kind: instruction
+    instance: vmmatch.vx
+
+  - name: vmmatch-vx_op
+    tags: ["norm:vmmatch-vx_op"]
+    kind: instruction
+    instance: vmmatch.vx
+
+  - name: vmmatch-vx_all_slots
+    tags: ["norm:vmmatch-vx_all_slots"]
+    kind: instruction
+    instance: vmmatch.vx
+
+  - name: vmmatch-vx_x0
+    tags: ["norm:vmmatch-vx_x0"]
+    kind: instruction
+    instance: vmmatch.vx
+
+  - name: vmmatch-vv_sew_rsv
+    tags: ["norm:vmmatch-vv_sew_rsv"]
+    kind: instruction
+    instance: vmmatch.vv
+
+  - name: vmmatch-vv_key_domain
+    tags: ["norm:vmmatch-vv_key_domain"]
+    kind: instruction
+    instance: vmmatch.vv
+
+  - name: vmmatch-vv_op
+    tags: ["norm:vmmatch-vv_op"]
+    kind: instruction
+    instance: vmmatch.vv
+
+  - name: vmmatch-vv_all_keys
+    tags: ["norm:vmmatch-vv_all_keys"]
+    kind: instruction
+    instance: vmmatch.vv
+
+  - name: vmmatch-vv_no_empty_set
+    tags: ["norm:vmmatch-vv_no_empty_set"]
+    kind: instruction
+    instance: vmmatch.vv
+
+  - name: vmmatch_execution
+    tags: ["norm:vmmatch_execution"]
+    kind: instruction
+    instances: [vmmatch.vx, vmmatch.vv]
+
+  - name: vmmatch_no_empty_set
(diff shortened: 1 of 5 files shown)


camel-cdr:
## On vmatch.vx

I really like the idea of `vmatch.vx`, it will be very useful string processing applications.

It also plays well with LMUL>1, because the destination in a mask register and the input is just one vector register, implementations like [Tenstorrent Ascalon X](https://camel-cdr.github.io/rvv-bench-results/tt_asc_x/index.html) would be able to have 2x throughput on this instruction (see ["What's going on with the comparison instructions?"](https://camel-cdr.github.io/rvv-bench-results/tt_asc_x/index.html)).

## On vmatch.vv

I'm not sure `vmatch.vv` can be efficiently implemented for large VLEN or/and LMUL>1. (you could be doing something on the order of 2^30 comparisons in one instruction, if you max out VLEN and LMUL)
This really should be in separate extensions and vs1 should be limited to LMUL=1, as it's way more complex and costly to implement.
Also, why is it only defined for SEW=8 and SEW=16?

The concept that the key-set needs to be filled with VLMAX elements directly goes against the concept of an Ovlt implementation.

There was some discussion about something similar for duplicate detection on the mailing list 6 years ago: https://lists.riscv.org/g/tech-vector-ext/topic/74915359
The idea hear was to use a probabilistic approach. The proposed `vmhash.vv` behaved like `vmatch.vv`, but was allowed to have false positives, which was ok for that usecase. (You would need to define it a bit more rigorously, but the "false positives" allow an efficient hardware implementation for long vectors) 

--- 

I think a better idea is to have `vmatch<N>.vs` instructions, which matches against the first N elements in vs2 (e.g. vmatch16.vs matches against first 16 elements in vs2). This has a much simpler VLEN and LMUL scaling.

Then for the duplicate detection usecase have something like the probabilistic `vmhash.vv`. (I think the `vmhash.vv` idea needs a bit of refining, but could be good)

Which important usages are covered by `vmatch.vv`, but not by `vmatch.vs`, `vmhash.vv` and a few variants of `vmatch<N>.vs` (say 4/8/16)?

wangpc-pp:
@camel-cdr

Thanks, this is very helpful!

* I agree that the current full-VLMAX definition of `VMMATCH.VV` is too unconstrained. Its work scales as `O(vl * VLMAX)`, and reducing vl does not reduce the key-domain work. At the architectural maximum of VLEN=2^16 and LMUL=8, VLMAX is 65536 for SEW=8, so a full operation could require up to 2^32 element comparisons. For SEW=16, the corresponding bound is 2^30. Restricting the key source to LMUL=1 would help, but the cost would still scale with VLEN.

* The SEW=8 and SEW=16 restriction came from the intended exact classification workloads: bytes, UTF-16 code units, and pre-decoded 16-bit symbols. It is also particularly relevant to `VMMATCH.VX`: `SEW=32` provides only two packed keys on RV64 and one on RV32, while `SEW=64` provides one on RV64 and no complete slot on RV32. I should make this rationale explicit. A separately defined vector-key form could have its own SEW rules (probably?).

* For tokenizer and delimiter classification, false positives are not acceptable without a subsequent verification step, so VMHASH.VV would be useful for duplicate detection but would not replace the exact matching operation.

That said, I do not currently have an important use case that requires an exact full-VLMAX key set in one instruction and cannot instead use multiple fixed-N matches followed by `VMOR.MM`. The main advantage of `VMMATCH.VV` is reduced instruction count for larger exact sets, rather than unique functionality.

I therefore plan to revise the proposal as follows:

  * Keep `VMMATCH.VX` as the bounded base operation.
  * Investigate `.VV` forms such as `VMMATCH<1|2|4|8>R.VV`, in which vs1 is a LMUL-1/2/4/8 register (we may define 1/2 only).

camel-cdr:
If you goal is delimiter classification, why not make vmatch8.vx and vmatch16.vx, and have them work with every SEW and only match if the value is equal (the upper bits zero and the lower matching)?
For UTF-16 and UTF-32, the ascii values would still be in the range 0-128, so that case would be covered.

Something like this:
```
vmatch8.vx:
for i from vstart to vl - 1:
    if vm == 1 or v0.mask[i] == 1:
        vd.mask[i] = OR over j in [0, 8):
                         vs2[i] == zero_extend(x[rs1][8*j +: 8])

vmatch16.vx:
for i from vstart to vl - 1:
    if vm == 1 or v0.mask[i] == 1:
        vd.mask[i] = OR over j in [0, 8):
                         vs2[i] == zero_extend(x[rs1][16*j +: 16])
```

---

Idk why you would need to match more than 16 characters (16 is already excessive).
I would drop vmatch.vv all thogether (vmhash may still be interesting for the other use case).

If you want to do something like match the base64 char set, or alphanumeric, there are way more efficient ways than comparing against every member of the set.

For base64, we can simply construct two 16-byte LUTs, one for the lower and one for the upper four bits:

	static const uint8_t err_lo[16] = { 21, 17, 17, 17, 17,  17,  17,  17,  17, 17, 19, 26, 27, 27, 27, 26 };
	static const uint8_t err_hi[16] = { 16, 16, 1,  2,  4,   8,   4,   8,   16, 16, 16, 16, 16, 16, 16, 16 };

this can detect elements not in the set as follows:

	err = vmsne(vand(vrgather(vErrLo, vLo), vrgather(vErrHi, vHi)), 0);

Whats really needed to make those use cases nice is a constraint/in-lane vrgather, in this case in 128-bit lane, like pshufb in x86.
That would accelerate a lot of other problems as well.

aswaterman:
Please do not propose ISA extensions as PRs to the ISA manual.  They need to go through RVIA processes first.

PR #3298, opened 2026-08-10
RVA23.1 and RVB23.1 submission for Public Review

rpsene:
This PR supersedes #3170.

It is intended for Public Review.

Included is the v0.7.1 Frozen version of the RVA23.1 and RVB23.1 Profiles. The normative text is unchanged from the previous stand-alone documents archived at https://github.com/riscv/riscv-profiles/tree/rva23p1-rvb23p1

The intention is to move that into the new combined ISA manual, bringing only the minimum required changes.

### Why this replaces #3170

The content of this PR is byte-identical to the head of #3170 (`7f75e156`) for all three files. It is resubmitted here because the original branch could not satisfy the repository's contribution checks:

- `dco_check` requires a `Signed-off-by:` trailer on every non-bot commit. None of the 7 commits on #3170 carried one.
- `signing_check` requires every commit to be cryptographically verified. 4 of the 7 were unsigned.

The original contributor is no longer able to contribute to RISC-V International, so the sign-off and signature could not be added to those commits. RISC-V International staff are therefore carrying the submission forward, with the DCO sign-off made by the submitter of this PR. The 7 commits (including 2 merge commits) are collapsed into a single signed, signed-off commit.

### Public review comments

Addresses #3314.

Issue #3314 raised five editorial points on these documents. All five are addressed here. The changes are editorial only: no normative requirement and no extension list is altered. As a consequence the chapters are no longer byte-identical to the archived stand-alone documents at `riscv/riscv-profiles`. The differences are limited to:

- singular "Profile" in both chapter headings and the surrounding prose
- the grammar correction in the RVA23.1 optional extensions paragraph, using the wording proposed in #3314
- removal of a sentence duplicated within the RVA23.1 chapter
- rewriting the RVB23.1 opening, which incorrectly claimed to specify the entire RVB23 profile family

### DIFF
--- a/src/profiles/profiles.adoc
+++ b/src/profiles/profiles.adoc
@@ -9,7 +9,9 @@ include::rvi20.adoc[]
 include::rva20.adoc[]
 include::rva22.adoc[]
 include::rva23.adoc[]
+include::rva23p1.adoc[]
 include::rvb23.adoc[]
+include::rvb23p1.adoc[]
 
 // Appendices
 :appendix-number: @
--- a/src/profiles/rva23p1.adoc
+++ b/src/profiles/rva23p1.adoc
@@ -1,0 +1,61 @@
+== RVA23.1 Profiles
+
+The RVA23.1 profiles are intended to expand upon
+the RVA23 major release of the RVA Family of RISC-V Application
+Processor Profile.
+
+This is a minor release of the RVA23 Profile Family. As such, Minor
+Profiles do not contain any new mandatory extensions and are expected to be
+built on top of the extensions listed in the RVA23 profiles. For
+the sake of simplicity when referring to this profile in code or
+use in a command line argument, all instances of the period will
+be replaced with the lower case letter 'p', as in `rva23p1s64`.
+
+Only supervisor-mode (`rva23p1s64`) profiles are specified in this
+family.
+
+=== The `rva23p1s64` Profile
+
+The `rva23p1s64` profile specifies the ISA features available to a
+supervisor-mode execution environment in 64-bit applications
+processors.  `rva23p1s64` is based on and requires an `rva23s64`
+implementation.
+
+==== Optional Extensions
+
+The `rva23p1s64` profile specifies the ISA features available to a
+supervisor-mode execution environment in 64-bit applications
+processors.  `rva23p1s64` is designed to be added on to and requires
+and implementation of `rva23s64` for Mandatory Base and Mandatory
+Extensions. `rva23p1s64` only provides new Development Options and
+Expansion Options on top of the `rva23s64`.
+
+===== Localized Options
+
+There are no privileged localized options in `rva23p1s64`.
+
+===== Development Options
+
+The following are new privileged development options in `rva23p1s64`:
+
+- ext:svrsw60t59b[] PTE reserved-for-software bits.
+
+- **Ssdbltrp** Double Trap.
+
+- **Ssccfg** Supervisor Counter Delegation.
+
+===== Expansion Options
+
+The following are new privileged expansion options in `rva23p1s64`:
+
+- ext:ssctr[] Control Transfer Records, CTR.
+
+NOTE: *Ssctr* also implies inclusion of *Sscsrind* (indirect CSR access)
+
+- <<ssqosid,**Ssqosid**>> QoS identifiers.
+
+===== Transitory Options
+
+There are no privileged transitory options in `rva23p1s64`.
+
+//
--- a/src/profiles/rva23p1.adoc
+++ b/src/profiles/rva23p1.adoc
@@ -1,6 +1,6 @@
-== RVA23.1 Profiles
+== RVA23.1 Profile
 
-The RVA23.1 profiles are intended to expand upon
+The RVA23.1 profile is intended to expand upon
 the RVA23 major release of the RVA Family of RISC-V Application
 Processor Profile.
 
@@ -11,7 +11,7 @@ the sake of simplicity when referring to this profile in code or
 use in a command line argument, all instances of the period will
 be replaced with the lower case letter 'p', as in `rva23p1s64`.
 
-Only supervisor-mode (`rva23p1s64`) profiles are specified in this
+Only a supervisor-mode (`rva23p1s64`) profile is specified in this
 family.
 
 === The `rva23p1s64` Profile
@@ -23,12 +23,10 @@ implementation.
 
 ==== Optional Extensions
 
-The `rva23p1s64` profile specifies the ISA features available to a
-supervisor-mode execution environment in 64-bit applications
-processors.  `rva23p1s64` is designed to be added on to and requires
-and implementation of `rva23s64` for Mandatory Base and Mandatory
-Extensions. `rva23p1s64` only provides new Development Options and
-Expansion Options on top of the `rva23s64`.
+`rva23p1s64` is designed as an add-on and requires an implementation
+of `rva23s64` for its Mandatory Base and Mandatory Extensions.
+`rva23p1s64` only provides new Development Options and Expansion
+Options on top of the `rva23s64`.
 
 ===== Localized Options
 
--- a/src/profiles/rva23p1.adoc
+++ b/src/profiles/rva23p1.adoc
@@ -1,15 +1,12 @@
-== RVA23.1 Profile
+== RVA23.1 Profile Family
 
-The RVA23.1 profile is intended to expand upon
-the RVA23 major release of the RVA Family of RISC-V Application
-Processor Profile.
+The RVA23.1 profile family is intended to expand upon
+the RVA23 family of RISC-V Application Processor Profiles.
 
-This is a minor release of the RVA23 Profile Family. As such, Minor
-Profiles do not contain any new mandatory extensions and are expected to be
-built on top of the extensions listed in the RVA23 profiles. For
-the sake of simplicity when referring to this profile in code or
-use in a command line argument, all instances of the period will
-be replaced with the lower case letter 'p', as in `rva23p1s64`.
+This is a minor release of the RVA23 profile family. As such, it
+contains only optional extensions. The minor release number is part
+of the profile name, with the period replaced by the lowercase
+letter 'p', as in `rva23p1s64`.
 
 Only a supervisor-mode (`rva23p1s64`) profile is specified in this
 family.
--- a/src/profiles/rvb23p1.adoc
+++ b/src/profiles/rvb23p1.adoc
@@ -1,8 +1,8 @@
-== RVB23.1 Profiles
+== RVB23.1 Profile
 
-This chapter specifies the RVB23 Profile Family. RVB23.1 is the
-first minor release intended to expand upon the RVB23 major release
-of the RVB23 Family of RISC-V Application Processor Profile.
+The RVB23.1 profile is intended to expand upon
+the RVB23 major release of the RVB Family of RISC-V Application
+Processor Profile.
 
 This is a minor release of the RVB23 Profile Family. As such, Minor Profiles
 do not contain any new mandatory extensions and are expected to be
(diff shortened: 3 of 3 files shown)


rpsene:
[riscv-spec.pdf](https://github.com/user-attachments/files/30919137/riscv-spec.pdf)
[rvb23p1.pdf](https://github.com/user-attachments/files/30919577/rvb23p1.pdf)
[rva23p1.pdf](https://github.com/user-attachments/files/30919578/rva23p1.pdf)

src/profiles/rva23p1.adoc
```diff
@@ -0,0 +1,61 @@
...
+
+There are no privileged localized options in `rva23p1s64`.
+
+===== Development Options
+
+The following are new privileged development options in `rva23p1s64`:
+
+- ext:svrsw60t59b[] PTE reserved-for-software bits.
+
+- **Ssdbltrap** Double Trap.
```
ChunyuLiao, line 43: Is this Ssdbltrp?

src/profiles/rva23p1.adoc
```diff
@@ -0,0 +1,61 @@
...
+
+- **Sscfg** Supervisor Counter Delegation.
```
ChunyuLiao, line 45: Ditto Ssccfg？

src/profiles/rva23p1.adoc
rpsene: Yes — good catch, thank you. It was misspelled as `Ssdbltrap`; the ratified name is `Ssdbltrp`. Fixed.

The same typo was also present in `rvb23p1.adoc`, so I corrected it there as well.

src/profiles/rva23p1.adoc
rpsene: Correct as well — it read `Sscfg`, and the ratified name is `Ssccfg`. Fixed here and in the matching line in `rvb23p1.adoc`.

Both names verified against riscv-unified-db. Thanks for catching these — they were the only substantive feedback the profiles received during public review, and they were worth having.

rpsene:
@aswaterman we are getting ready for TSC vote, a review here would be welcome. It should not be merged yet.

rpsene:
Built specification PDF for review, from the current head of this PR.

**Run:** [RISC-V ISA Build #34419381227](https://github.com/riscv/riscv-isa-manual/actions/runs/34419381227) — success, on `4dc6677`
**Artifact:** `riscv-spec-7ec1a7f.pdf` (908 pages, 5.3 MB)

The artifact is built from this PR's merge commit, so it reflects the manual as it would read once this lands. It contains both new profiles, with the two extension-name corrections from this review applied — `Ssdbltrp` and `Ssccfg`.

The same PDF is attached to the TSC approval tickets for voters:

- RVB23.1 — https://riscv.atlassian.net/browse/RVS-5031
- RVA23.1 — https://riscv.atlassian.net/browse/RVS-5034

Separately, and not part of this PR: the build still contains one `Ssdbltrap`, in `src/priv/hypervisor.adoc:1855` — "The Ssdbltrap extension (See `<<ssdbltrp>>`)". That is already on `main`, so it is in the published manual today. The cross-reference anchor is correct; only the prose is misspelled. Raised separately as #3370.

rpsene:
Raised the `Ssdbltrap` typo separately as #3370 — one word in `src/priv/hypervisor.adoc:1855`, unrelated to this PR.

src/profiles/rva23p1.adoc
```diff
@@ -0,0 +1,59 @@
+== RVA23.1 Profile
```
aswaterman, line 1: Profile Family

src/profiles/rva23p1.adoc
```diff
@@ -0,0 +1,59 @@
...
+
+The RVA23.1 profile is intended to expand upon
```
aswaterman, line 3: profile family

src/profiles/rva23p1.adoc
```diff
@@ -0,0 +1,59 @@
...
+the RVA23 major release of the RVA Family of RISC-V Application
```
aswaterman, line 4: RVA23 family of RISC-V Application Processor Profiles

src/profiles/rva23p1.adoc
```diff
@@ -0,0 +1,59 @@
...
+Processor Profile.
+
+This is a minor release of the RVA23 Profile Family. As such, Minor
+Profiles do not contain any new mandatory extensions and are expected to be
+built on top of the extensions listed in the RVA23 profiles. For
+the sake of simplicity when referring to this profile in code or
+use in a command line argument, all instances of the period will
+be replaced with the lower case letter 'p', as in `rva23p1s64`.
```
aswaterman, line 12: This is a little misleading. rva23p1s64 _is_ the name of the profile; it’s not just how it’s referred to in code.

src/profiles/rva23p1.adoc
```diff
@@ -0,0 +1,59 @@
+== RVA23.1 Profile
+
+The RVA23.1 profile is intended to expand upon
+the RVA23 major release of the RVA Family of RISC-V Application
+Processor Profile.
+
+This is a minor release of the RVA23 Profile Family. As such, Minor
+Profiles do not contain any new mandatory extensions and are expected to be
```
aswaterman, line 8: This sentence reads like general guidance for defining minor profiles, rather than a statement about this profile. You might instead simply say, “As such, it contains only optional extensions.”

rpsene:
Thanks @aswaterman — all five comments are addressed in d48d970, and I applied the same corrections to `rvb23p1.adoc`, which carried word-for-word identical wording.

- Both sections are now titled as profile families, and the opening sentence refers to the RVA23.1 / RVB23.1 profile family.
- "the RVA23 major release of the RVA Family of RISC-V Application Processor Profile" is now "the RVA23 family of RISC-V Application Processor Profiles".
- The general guidance about what minor profiles may contain is replaced with your wording: "As such, it contains only optional extensions."
- The naming sentence now states that the minor release number is part of the profile name, instead of implying `rva23p1s64` is only how the profile is written in code or on a command line.

The built PDF from that commit is attached below (908 pages); the RVA23.1 and RVB23.1 chapters are 6 and 8.

[riscv-spec-rva23p1-rvb23p1-d48d970.pdf](https://github.com/user-attachments/files/32126605/riscv-spec-rva23p1-rvb23p1-d48d970.pdf)

topperc:
Do we expect toolchains to accept rva23p1u64 in their `-march` option? With no new mandatory extensions it wouldn't do anything different than rva23u64.

aswaterman:
@topperc do you currently treat rva23u64 differently from rva23s64?

In general, application programs should be ignorant of the supervisor features, so rva23p1s64 should do the same thing as either of the aforementioned two for user code.

But what about OS code using new supervisor instructions/CSRs?

topperc:
> Craig Topper (@topperc) do you currently treat rva23u64 differently from rva23s64?
> 
> In general, application programs should be ignorant of the supervisor features, so rva23p1s64 should do the same thing as either of the aforementioned two for user code.
> 
> But what about OS code using new supervisor instructions/CSRs?

Sorry I should have written rva23p1s64 and rva23s64 in my comment. -march=rva23s64 enables all the mandatory extensions user and supervisor extensions as if they had been written explicitly in the march string. So my question was whether we should accept -march=rva23p1s64 since it would be identical to -march=rva23s64.

aswaterman:
I see your point.  Yes, we should accept it.  Generally, the minor releases will just be aliases since they add no mandatory extensions, but it will be easier to explain it to most people if we accept all valid strings.

src/profiles/rva23p1.adoc
```diff
@@ -0,0 +1,56 @@
...
+supervisor-mode execution environment in 64-bit applications
+processors.  `rva23p1s64` is based on and requires an `rva23s64`
+implementation.
+
+==== Optional Extensions
+
+`rva23p1s64` is designed as an add-on and requires an implementation
+of `rva23s64` for its Mandatory Base and Mandatory Extensions.
+`rva23p1s64` only provides new Development Options and Expansion
+Options on top of the `rva23s64`.
```
jordancarlin, line 26: The text immediately following the "Optional Extensions" header is dramatically different between RVA23.1 and RVB23.1 despite both profiles otherwise following the same structure. They should match each other here as well. I prefer the wording currently used for RVB23.1 since the RVA23.1 version seems repetitive with the text immediately preceding this section, but either way is fine as long as the two profiles are aligned.

src/profiles/rva23p1.adoc
```diff
@@ -0,0 +1,56 @@
...
+
+===== Localized Options
+
+There are no privileged localized options in `rva23p1s64`.
+
+===== Development Options
+
+The following are new privileged development options in `rva23p1s64`:
+
+- ext:svrsw60t59b[] PTE reserved-for-software bits.
```
jordancarlin, line 36: The other profiles use the `extlink:XX[]` macro when listing extensions. This should follow that convention instead of using a mix of `ext:XX[]` and regular asciidoc bold formatting with asterisks.

jordancarlin (changes requested):

src/profiles/rva23p1.adoc
rpsene: Thanks @jordancarlin, agreed that both chapters should read the same here. The difference is not only wording, though: the RVB23.1 sentence defines the unprivileged options of `rvb23p1s64` by reference to `rvb23u64`, and RVA23.1 has no equivalent statement. Aligning them in either direction would change normative content, and these profiles are Frozen with the TSC vote in progress, so I am not changing this PR.

I opened #3390 to align the two after ratification, using the RVB23.1 structure you prefer.

src/profiles/rva23p1.adoc
rpsene: Agreed, this should use `extlink:` like the other profile chapters. It is an editorial change, but it needs new `[[ext:…]]` anchors first: of the extensions listed here, only Svrsw60t59b has one today (Ssdbltrp, Ssccfg, Ssctr, Sscsrind and Ssqosid do not), so switching now would produce broken links. Since the TSC is voting on these chapters now, I am keeping this PR unchanged and tracking the anchors plus the `extlink:` conversion in #3390, to be done after ratification.

rpsene:
@aswaterman  @topperc  should we capture this as a toolchain note in Volume III, in the profiles introduction, so it applies to every profile and not only RVA23.1/RVB23.1? Something along the lines of: toolchains should accept all valid profile names, including minor releases, and a minor release that adds no mandatory extensions selects the same extensions as its major release (so `-march=rva23p1s64` behaves like `-march=rva23s64`). That way GCC, LLVM and other tools behave consistently, and implementers don't have to infer it.

src/profiles/profiles.adoc
lenary, line 6: This contains a list of profiles, does it need to be updated?

rpsene: Good catch, thanks @lenary. Yes, it does: the preface lists the profiles in Volume III and says all of them are ratified, and this PR adds two without touching that list.

Because of that "ratified" wording, I will add `rva23p1s64` and `rvb23p1s64` to the preface once RVA23.1 and RVB23.1 are ratified, rather than changing it while the TSC vote is open. Tracking it in #3400.

camel-cdr:
> I see your point. Yes, we should accept it. Generally, the minor releases will just be aliases since they add no mandatory extensions, but it will be easier to explain it to most people if we accept all valid strings

Nobody should ever use the profile point-release march instead of the profile march it self though.
The only reason to use them is, if you don't understand how the point releases work.

I suppose it's fine to accept them, to make the toolchain more resilient, but there should imo definitely be a warning that it doesn't add anything new.
I think most people coming from Arm would expect them to add new mandatory features.

aswaterman:
A warning isn't appropriate since semantically it's entirely valid.  I still think accepting them as aliases is fine, but making a blanket policy to never add point releases to `-march` is better than accepting them and warning about something that isn't unsafe or incorrect.

PR #3300, opened 2026-08-12
Normative rule change recommended by CSC WG

UmerShahidengr:
Hypervisor: guest-page-fault traps 'are delegated'; scope tag to first sentence only

### DIFF
--- a/src/priv/hypervisor.adoc
+++ b/src/priv/hypervisor.adoc
@@ -2044,10 +2044,9 @@ whether this usage will ever become common, we chose not to preclude it.
 
 ==== Guest-Page Faults
 
-[[norm:H_guest_page_fault]]
-Guest-page-fault traps may be delegated from M-mode to HS-mode under the
+[#norm:H_guest_page_fault]#Guest-page-fault traps are delegated from M-mode to HS-mode under the
 control of CSR csr:medeleg[], but cannot be delegated to other privilege
-modes. On a guest-page fault, CSR csr:mtval[] or csr:stval[] is written with the
+modes.# On a guest-page fault, CSR csr:mtval[] or csr:stval[] is written with the
 faulting guest virtual address as usual, and csr:mtval2[] or csr:htval[] is
 written either with zero or with the faulting guest physical address,
 shifted right by qty:2[bits]. CSR csr:mtinst[] or csr:htinst[] may also be written


aswaterman (commented):
@kasanovic this is part of a batch of wording improvements proposed by the CSC.  I'm curious what you think about this one, since it's a case where the new wording is semantically accurate, but seems to destroy some of the prose's intent--"may be" is shorthand for "software can decide what to do here".

src/priv/hypervisor.adoc
```diff
@@ -2044,10 +2044,9 @@ whether this usage will ever become common, we chose not to preclude it.
 
 ==== Guest-Page Faults
 
-[[norm:H_guest_page_fault]]
-Guest-page-fault traps may be delegated from M-mode to HS-mode under the
+[#norm:H_guest_page_fault]#Guest-page-fault traps are delegated from M-mode to HS-mode under the
```
aswaterman, line 2047: "are delegated" is not necessarily true: they might not be delegated.

"are delegable" would be correct--it means the same thing as "may be delegated"--but the former is more awkwardly phrased.

kasanovic: Agree with Andrew.  In original, "may be" has usual English meaning, indicating "it is possible to" - the remaining part of that clause says how this is done through the registers.  To be more parallel with the "cannot be" in the second clause, I would propose changing "may be" to "can be" in first clause.

jordancarlin: Looking back at the notes from the WG, I think the concern was the potential association between "may be" and `medeleg` controlling the delegation. It is already mentioned elsewhere in the spec that exceptions are delegated from M to S/HS mode using `medeleg`, so what about the following:

```
Guest-page-fault traps can be delegated from M-mode to HS-mode,
but cannot be delegated to other privilege modes.
```

allenjbaum:
Yes, otherwise (my interpretation is) the implication would be that
regardless of medeleg bits, it might not be delegated.

aswaterman:
> Yes, otherwise (my interpretation is) the implication would be that
regardless of medeleg bits, it might not be delegated.

This is not a valid interpretation, because it contradicts the definition of medeleg, which says that the bits being set means the exception is delegated.

src/priv/hypervisor.adoc
aswaterman: This seems like a regression to me.  It says the same normative thing as the old sentence but is less informative to the reader.

allenjbaum:
It's that pesky word "may". If something may happen, it also may not
happen, and the phrase "under the control of medeleg[]" is less precise
than it should be
Maybe more simply
"Guest page faults can only be delegated to HS mode"
and  don't bother with the "under control of", etc, since, as you say,
delegation is defined elsewhere.

aswaterman:
@allenjbaum As I explained above, I do not agree that the current wording is ambiguous.  I can see that you don't like the current wording, but by itself that isn't a good enough reason to change it.

Ping me if the PR is later updated with alternate wording for our review.

allenjbaum:
OK, will do, but after Hotchips and travel, so let's let this one stew a
while.

PR #3301, opened 2026-08-12
Keyword changes in Hypervisor chapter

UmerShahidengr:


aswaterman (changes requested):
The old wording is better, as it makes it clear that it is introducing the new, notional concept of "effective ASID".  The new wording implies there is a well-defined concept of "effective ASID" that readers should know about.

src/priv/hypervisor.adoc
```diff
@@ -2085,7 +2085,7 @@ When V=1, the virtual-address argument to SFENCE.VMA is a guest virtual
 address within the current virtual machine, and the ASID argument is a
 VS-level ASID within the current virtual machine. The current virtual
 machine is identified by the VMID field of CSR `hgatp`, and the
-effective ASID can be considered to be the combination of this VMID with
+effective ASID is the combination of this VMID with
```
jordancarlin, line 2088: @aswaterman how about the following:

```adoc
machine is identified by the VMID field of CSR `hgatp`, and the
ASID is effectively the combination of this VMID with
```

aswaterman: It's not really the machine that's being identified; it's the task.

I'm waffling, but now I actually think the "is" is OK.

aswaterman (approved):

PR #3302, opened 2026-08-12

UmerShahidengr:


aswaterman (changes requested):
This change doesn't seem right, as it is describing what software can do, rather than what must be.  It isn't the case that these exceptions must be separately delegated; it's that they _can_ be.

jordancarlin:
> This change doesn't seem right, as it is describing what software can do, rather than what must be. It isn't the case that these exceptions must be separately delegated; it's that they _can_ be.

Software can choose to delegate one or both, but they are always delegated separately. There is nothing linking the delegation of the two together.

allenjbaum:
Yes; the keyword "can" implies that an implementation could  disallow them
to be delegated separately

aswaterman:
The problem is really with the phrase "are delegated", because the exceptions might not be delegated.  What you really mean to say is that they "are separately delegable", which refers to the ability to delegate as opposed to the current state of delegation.

I'd be fine with that alternative wording... but it means the exact same thing as "can be delegated separately", so I don't see what the change accomplishes.

PR #3303, opened 2026-08-12

UmerShahidengr:


aswaterman (changes requested):
This change is incorrect, as definitionally the instruction normally raises only one exception (i.e. in the case that it "may raise" multiple exceptions, it does not raise multiple of them).

allenjbaum:
I'm don't I agree with that. The full sentence is:
If an instruction may raise multiple synchronous exceptions, the decreasing
priority order of <<HSyncExcPrio>> indicates which exception is taken and
reported in `mcause` or `scause`.

That seems to indicate that it is raising multiple exception, but only one
is reported. The sentence distinguishes between raising and reporting.
Perhaps a slight rewording:
If an instruction "is capable of raising"multiple synchronous exceptions
rather than "may raise"

aswaterman:
@allenjbaum Yeah, I prefer the definitional choice that multiple exceptions may be raised but that they are prioritized, resulting in one trap.

aswaterman (approved):

PR #3304, opened 2026-08-12

UmerShahidengr:
keyword_matches row 1305. WG: change 'can be accessed' to 'are accessed' in both places.

### DIFF
--- a/src/priv/smcntrpmf.adoc
+++ b/src/priv/smcntrpmf.adoc
@@ -37,7 +37,7 @@ Smcntrpmf remedies these issues by introducing privilege mode filtering for the 
 
 [#norm:unimplemented_mode_bits]#For each bit in 61:58, if the associated privilege mode is not implemented, the bit is read-only zero.#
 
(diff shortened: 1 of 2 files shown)


src/priv/smcntrpmf.adoc
```diff
@@ -37,7 +37,7 @@ Smcntrpmf remedies these issues by introducing privilege mode filtering for the
 
 [#norm:unimplemented_mode_bits]#For each bit in 61:58, if the associated privilege mode is not implemented, the bit is read-only zero.#
 
-[#norm:rv32_high_access]#For RV32, bits 63:32 of mcyclecfg can be accessed via the mcyclecfgh CSR, and bits 63:32 of minstretcfg can be accessed via the minstretcfgh CSR.#
+[#norm:rv32_high_access]#For RV32, bits 63:32 of mcyclecfg are accessed via the mcyclecfgh CSR, and bits 63:32 of minstretcfg are accessed via the minstretcfgh CSR.#
```
aswaterman, line 40: I think I'm OK with this particular case on the basis that these bits aren't accessed through other means as well, but see my comments on other PRs before deciding on the general strategy about "can be" versus "are".

PR #3305, opened 2026-08-12

UmerShahidengr:


src/priv/smctr.adoc
```diff
@@ -460,7 +460,7 @@ recorded, if anything, depends on whether the source privilege mode
 and/or target privilege mode are enabled for recording, and on the transfer type (trap
 or trap return).
 
-[#norm:ctr_trap_enabled]#Traps between enabled privilege modes are recorded as normal.#  [#norm:ctr_trap_disabled_src]#Traps from a disabled privilege mode to an enabled privilege mode are partially recorded, such that the `ctrsource`.PC is 0.# [#norm:ctr_trap_disabled_tgt]#Traps from an enabled mode to a disabled mode, known as external traps, are not recorded by default.#  See <<External Traps>> for how they can be recorded.
+[#norm:ctr_trap_enabled]#Traps between enabled privilege modes are recorded as normal.#  [#norm:ctr_trap_disabled_src]#Traps from a disabled privilege mode to an enabled privilege mode are partially recorded, such that the `ctrsource`.PC is 0.# [#norm:ctr_trap_disabled_tgt]#Traps from an enabled mode to a disabled mode, known as external traps, are not recorded by default.#  See <<External Traps>> for how they are recorded.
```
aswaterman, line 463: The existing wording is correct, as the traps are not always recorded.  Literally, they "can be"; it is incorrect to say that they "are".

allenjbaum: It is actually outside the normative tag as well, so I don't have a problem with the original wording.

aswaterman: OK, I'll leave it to y'all to close this PR if no change is needed.

PR #3306, opened 2026-08-12

UmerShahidengr:
keyword_matches row 1311. WG: move the 3rd sentence to 2nd and let the tag cover sentences 1-2; replace 'opt-in' with 'enable(d)'; replace 'By default' with 'When the xctrctl.xTE bits are clear'.

### DIFF
--- a/src/priv/smctr.adoc
+++ b/src/priv/smctr.adoc
@@ -504,7 +504,7 @@ _Consider an exception that traps from VU-mode to HS-mode, with `vsctrctl`.U=1 a
 
 ====== External Traps
 
-[#norm:exttrap_def]#External traps are traps from a privilege mode enabled for CTR recording to a privilege mode that is not enabled for CTR recording.  By default external traps are not recorded, but privileged software running in the target mode of the trap can opt-in to allowing CTR to record external traps into that mode. The `__x__ctrctl`.__x__TE bits allow M-mode, S-mode, and VS-mode to opt-in separately.#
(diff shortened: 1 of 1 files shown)


src/priv/smctr.adoc
```diff
@@ -504,7 +504,7 @@ _Consider an exception that traps from VU-mode to HS-mode, with `vsctrctl`.U=1 a
 
 ====== External Traps
 
-[#norm:exttrap_def]#External traps are traps from a privilege mode enabled for CTR recording to a privilege mode that is not enabled for CTR recording.  By default external traps are not recorded, but privileged software running in the target mode of the trap can opt-in to allowing CTR to record external traps into that mode. The `__x__ctrctl`.__x__TE bits allow M-mode, S-mode, and VS-mode to opt-in separately.#
+[#norm:exttrap_def]#External traps are traps from a privilege mode enabled for CTR recording to a privilege mode that is not enabled for CTR recording.  The `__x__ctrctl`.__x__TE bits allow M-mode, S-mode, and VS-mode to be enabled separately.#  When the `__x__ctrctl`.__x__TE bits are clear, external traps are not recorded, but privileged software running in the target mode of the trap can enable CTR to record external traps into that mode.
```
aswaterman, line 507: "The TE bits allow M-mode, S-mode, and VS-mode to be enabled separately" sounds like this feature is controlling which privilege modes exist, as in, you can change TE to delete S-mode from existence.  The older wording doesn't create this confusion.

allenjbaum:
Yea, that should be "enabled for CTR recording" In this case,. But also:
this describes what software is doing, "can" is clearly appropriate - but
shouldn't be part of the normative rule tag
The tag covers too much, and the point of this change was to swap the two
sentences so that the tag only covers the normative parts.
The above suggestion would remove the problem you had, but also I think the
"but privileged software..." should start a new sentence (", but
privileged" -> ". privileged") and that last sentence should be outside the
tag (and possibly the first sentence should be also).

aswaterman:
OK, ping me when the PR has been updated and I'll review it again.

PR #3307, opened 2026-08-12

UmerShahidengr:
keyword_matches row 1302. WG: change 'can be' to 'is'; 'while siselect holds' ->
'when siselect'; the state-mapping table is not tagged, so tag this rule.

### DIFF
--- a/src/priv/smcdeleg.adoc
+++ b/src/priv/smcdeleg.adoc
@@ -36,11 +36,11 @@ The Smcdeleg and Ssccfg extensions both depend on the Sscsrind extension.
 The `mcounteren` register allows M-mode to provide the next-lower
 privilege mode with read access to select counters.[#norm:smcdeleg_cde_en]#When the Smcdeleg/Ssccfg extensions are enabled (`menvcfg`.CDE=1), it further allows M-mode to delegate select counters to S-mode.#
 
-The `siselect` (and `vsiselect`) index range 0x40-0x5F is reserved for
+[#norm:smcdeleg_delegated_access]#The `siselect` (and `vsiselect`) index range 0x40-0x5F is reserved for
 delegated counter access. When a counter _i_ is delegated
 (`mcounteren`[_i_]=1 and `menvcfg`.CDE=1), the register state associated
-with counter _i_ can be read or written via `sireg*`, while `siselect` holds
-0x40+__i__. The counter state accessible via alias CSRs is shown in
+with counter _i_ is read or written via `sireg*`, when `siselect` is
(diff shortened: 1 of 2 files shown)


src/priv/smcdeleg.adoc
```diff
@@ -36,11 +36,11 @@ The Smcdeleg and Ssccfg extensions both depend on the Sscsrind extension.
 The `mcounteren` register allows M-mode to provide the next-lower
 privilege mode with read access to select counters.[#norm:smcdeleg_cde_en]#When the Smcdeleg/Ssccfg extensions are enabled (`menvcfg`.CDE=1), it further allows M-mode to delegate select counters to S-mode.#
 
-The `siselect` (and `vsiselect`) index range 0x40-0x5F is reserved for
+[#norm:smcdeleg_delegated_access]#The `siselect` (and `vsiselect`) index range 0x40-0x5F is reserved for
 delegated counter access. When a counter _i_ is delegated
 (`mcounteren`[_i_]=1 and `menvcfg`.CDE=1), the register state associated
-with counter _i_ can be read or written via `sireg*`, while `siselect` holds
-0x40+__i__. The counter state accessible via alias CSRs is shown in
+with counter _i_ is read or written via `sireg*`, when `siselect` is
```
aswaterman, line 42: Why "is" here?  it sounds more like an instruction to software than a statement of ISA fact.

Note that counters can also be read through other means (through their direct CSR addresses), so "is" implies a more direct relationship than seems appropriate.

aswaterman: "While", in standard English, can mean either "whereas" or "when".  Here, it definitely means "whereas" rather than "when".  I'd argue that this word doesn't need to be changed at all, but if it does, "whereas" is the better choice.

jordancarlin: Unless I'm missing something, I think this use of "while" actually implies "when", not "whereas". It is possible to read the register state associated with counter i via sireg only when siselect is 0x40 + i.

aswaterman: You're right, it should be "when".

I think "can be" and "holds" are still better than the proposed wording.  Certainly the first "is" is problematic, for the reason I mentioned earlier.  The second isn't problematic, but it's less clear than "holds", IMO.

aswaterman: Also, I think the comma before "when" should be deleted.

allenjbaum:
Reading or written `sireg*` alias CSRs when `siselect` holds
0x40+__i__ will read or write counter _i_  state.

That's a pure statement of fact without any can/might/may/should/might baggage

PR #3308, opened 2026-08-12

UmerShahidengr:
keyword_matches row 1315. WG: the first sentence is normative and should be tagged; remove the duplicated 'This bit can be set by hardware ...' reference.

### DIFF
--- a/src/priv/smctr.adoc
+++ b/src/priv/smctr.adoc
@@ -725,7 +725,7 @@ an implementation may opt to record CCV=0 for all calls, or those whose parent c
 
 ===== Freeze
 
-When `sctrstatus`.FROZEN=1, transfer recording is inhibited.  This bit can be set by hardware, as described below, or by software.
+[#norm:sctrstatus_frozen_inhibit]#When `sctrstatus`.FROZEN=1, transfer recording is inhibited.#
 
 [#norm:sctrstatus_frozen_set]#When `sctrctl`.LCOFIFRZ=1 and a local-counter-overflow interrupt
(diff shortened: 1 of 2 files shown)


src/priv/smctr.adoc
```diff
@@ -725,7 +725,7 @@ an implementation may opt to record CCV=0 for all calls, or those whose parent c
 
 ===== Freeze
 
-When `sctrstatus`.FROZEN=1, transfer recording is inhibited.  This bit can be set by hardware, as described below, or by software.
+[#norm:sctrstatus_frozen_inhibit]#When `sctrstatus`.FROZEN=1, transfer recording is inhibited.#
```
aswaterman, line 728: Can you explain why the CSC prefers this phrasing?  To me, it seems to delete a useful introductory fact without adding any additional clarity.

allenjbaum:
The reason is to get rid of those pesky. "can be".
The register is described as RW, so obviously SW is able to set the bit,
and the description below has the details of how HW can do it,
so the final sentence seems completely redundant.

aswaterman:
> The reason is to get rid of those pesky. "can be".

The authors of the spec do not share your opinion on this matter.  There may be lingering cases that this phrase is ambiguous, but we do not agree that purging it is sensible, particularly for the common case where it describes what might happen, as a function of time.

jordancarlin:
The normative rule should only cover the first sentence ("When `sctrstatus`.FROZEN=1, transfer recording is inhibited"), but I am fine leaving the second sentence as long as it isn't tagged.

PR #3310, opened 2026-08-12
Describe frm and fflags as standalone CSRs

UmerShahidengr:
keyword_matches row 1268. WG: rewrite to admit the existence of the frm CSR. The fcsr fields frm and fflags are also individually addressable CSRs; reword the fcsr-field-access sentence to say so.

### DIFF
--- a/src/unpriv/f-st-ext.adoc
+++ b/src/unpriv/f-st-ext.adoc
@@ -46,8 +46,9 @@ the original value into integer register _rd_, and then writing a new
 value obtained from integer register _rs1_ into csr:fcsr[].
 
 [[norm:fflags_frm_op_sz_acc]]
-The fields within the csr:fcsr[] can also be accessed individually through
-different CSR addresses, and separate assembler pseudoinstructions are defined
+The Rounding Mode and Accrued Exception Flags fields of csr:fcsr[] are also
+accessible individually as the csr:frm[] and csr:fflags[] CSRs at different CSR
+addresses, and separate assembler pseudoinstructions are defined
 for these accesses. The insn:frrm[] instruction reads the Rounding Mode field
 csr:frm[] (csr:fcsr[] bits 7{endash}5) and copies it into the least-significant three bits of
 integer register _rd_, with zero in all other bits. insn:fsrm[] swaps the value in
--- a/src/unpriv/f-st-ext.adoc
+++ b/src/unpriv/f-st-ext.adoc
@@ -45,17 +45,15 @@ into integer register _rd_. insn:fscsr[] swaps the value in csr:fcsr[] by copyin
 the original value into integer register _rd_, and then writing a new
 value obtained from integer register _rs1_ into csr:fcsr[].
 
-[[norm:fflags_frm_op_sz_acc]]
-The Rounding Mode and Accrued Exception Flags fields of csr:fcsr[] are also
+[#norm:fflags_frm_op_sz_acc]#The Rounding Mode and Accrued Exception Flags fields of csr:fcsr[] are also
 accessible individually as the csr:frm[] and csr:fflags[] CSRs at different CSR
-addresses, and separate assembler pseudoinstructions are defined
-for these accesses. The insn:frrm[] instruction reads the Rounding Mode field
-csr:frm[] (csr:fcsr[] bits 7{endash}5) and copies it into the least-significant three bits of
-integer register _rd_, with zero in all other bits. insn:fsrm[] swaps the value in
-csr:frm[] by copying the original value into integer register _rd_, and then
-writing a new value obtained from the three least-significant bits of integer
-register _rs1_ into csr:frm[]. insn:frflags[] and insn:fsflags[] are defined analogously for the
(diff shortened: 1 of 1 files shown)


src/unpriv/f-st-ext.adoc
jordancarlin: New proposed text
```
The Rounding Mode and Accrued Exception Flags fields of csr:fcsr[] are also
accessible individually as the csr:frm[] and csr:fflags[] CSRs at different CSR
addresses. The csr:frm[] CSR contains the Rounding Mode field 
(csr:fcsr[] bits 7{endash}5) right-justified and zero-extended.
The csr:fflags[] CSR contains the Accrued Exception Flags field 
(csr:fcsr[] bits 4{endash}0) right-justified and zero-extended.
The insn:frrm[] pseudoinstruction performs a insn:csrr[] from
csr:frm[]. ins:fsrm[] similarly performs a ins:csrrw[] to csr:frm[]. 
insn:frflags[] and insn:fsflags[] are defined analogously for csr:fflags[].
```

UmerShahidengr: Updated

src/unpriv/f-st-ext.adoc
```diff
@@ -45,16 +45,15 @@ into integer register _rd_. insn:fscsr[] swaps the value in csr:fcsr[] by copyin
...
-for these accesses. The insn:frrm[] instruction reads the Rounding Mode field
-csr:frm[] (csr:fcsr[] bits 7{endash}5) and copies it into the least-significant three bits of
-integer register _rd_, with zero in all other bits. insn:fsrm[] swaps the value in
-csr:frm[] by copying the original value into integer register _rd_, and then
-writing a new value obtained from the three least-significant bits of integer
-register _rs1_ into csr:frm[]. insn:frflags[] and insn:fsflags[] are defined analogously for the
-Accrued Exception Flags field csr:fflags[] (csr:fcsr[] bits 4{endash}0).
+[#norm:fflags_frm_op_sz_acc]#The Rounding Mode and Accrued Exception Flags fields of csr:fcsr[] are also
+accessible individually as the csr:frm[] and csr:fflags[] CSRs at different CSR
+addresses. The csr:frm[] CSR contains the Rounding Mode field
```
aswaterman, line 50: Delete the phrase "at different CSR addresses"; that's implied by the fact that they're different CSRs.

src/unpriv/f-st-ext.adoc
```diff
@@ -45,16 +45,15 @@ into integer register _rd_. insn:fscsr[] swaps the value in csr:fcsr[] by copyin
...
+(csr:fcsr[] bits 7{endash}5) right-justified and zero-extended.
+The csr:fflags[] CSR contains the Accrued Exception Flags field
+(csr:fcsr[] bits 4{endash}0) right-justified and zero-extended.#
```
aswaterman, line 53: The phrase "right-justified" is ambiguous (who ever formally defined that the LSBs are on the right?) and is not otherwise used in the unpriv spec.  The old wording (least-significant bits) was better.

jordancarlin: We had other wording proposed for this initially, but decided to change it since "right-justified" is already used to describe mtval and stval. If we want to stay away from right-justified, we should change those spots too.

aswaterman: What was your alternative wording?

I see that the phrase is only used in those two places in the entire spec, and so it probably should be removed there, too.  In that case, I was thinking about using something like "The value is loaded into the least-significant bits of `mtval`, with the unused upper bits cleared to zero."

jordancarlin: I don't recall and don't seem to have saved it. That wording for `mtval` sounds good to me. For this one, maybe something like the following:
> The Rounding Mode and Accrued Exception Flags fields of csr:fcsr[] are also
accessible individually as the csr:frm[] and csr:fflags[] CSRs.
[#norm:frm_op_sz_acc]#The csr:frm[] CSR contains the Rounding Mode field 
(csr:fcsr[] bits 7{endash}5) in its least-significant bits with the
unused upper bits cleared to zero.# [#norm:fflags_op_sz_acc]# The csr:fflags[] CSR contains
the Accrued Exception Flags field (csr:fcsr[] bits 4{endash}0) in
its least-significant bits with the unused upper bits cleared to zero.# 
The insn:frrm[] pseudoinstruction performs a insn:csrr[] from
csr:frm[]. ins:fsrm[] similarly performs a ins:csrrw[] to csr:frm[].
insn:frflags[] and insn:fsflags[] are defined analogously for csr:fflags[].

allenjbaum:
Least significant is used 11 times, so perhaps  you are correct, though
right-justified and zero-extended
is a lot more succinct than
  loaded into the least-significant bits of mtval, with the unused upper
bits cleared to zero."

so maybe
 loaded into the least-significant bits of mtval, zero extended.
or maybe just
  is zero extended [to XLEN] bits and loaded into mtval.

aswaterman:
@allenjbaum I agree, "zero-extended and loaded into mtval".  Don't even need to say "zero-extended to XLEN bits".

PR #3318, opened 2026-08-18
Clarify that MXR doesn't affect M-mode pointer masking

### LINKED ISSUE #3313: [Clarification] - Does MXR affect pointer masking when MPRV=0?
## Summary
It is unclear whether MXR=1 disables pointer masking in M-mode when MPRV=0.  Whisper disables pointer masking, while Spike and Sail do not.

 (cc @riscv/arc)

---

## Spec Reference
Privileged 25.2.6:
```
MPRV and SPVP affect pointer masking as well, causing the pointer masking settings of the effective
privilege mode to be applied. When MXR is in effect at the effective privilege mode where explicit memory
access is performed, pointer masking does not apply.

Note that this includes cases where page-based virtual memory is not in effect; i.e., although
MXR has no effect on permissions checks when page-based virtual memory is not in effect, it is
still used in determining whether or not pointer masking should be applied.
```
Privileged 3.1.6.4
```
The MXR (Make eXecutable Readable) bit modifies the privilege with which loads access virtual memory.
When MXR=0, only loads from pages marked readable (R=1 in Figure 67) will succeed. When MXR=1, loads
from pages marked either readable or executable (R=1 or X=1) will succeed. MXR has no effect when page-
based virtual memory is not in effect.
```
---

## Expected vs Observed
Consider the case of an M-mode access with MPRV=0, MXR=1.  It is difficult to parse the spec language to determine whether MXR is in effect at the effective privilege mode, and thus whether pointer masking would apply.

An argument that pointer masking is disabled comes from "Note that this includes cases where page-based virtual memory is not in effect." Whisper behaves this way.

An argument that pointer masking is enabled comes from "MXR modifies the privilege with which loads access virtual memory."  When MPRV = 0, M-mode is accessing physical memory, not virtual memory, so MXR is not in effect at this privilege mode. Spike and Sail behave this way.

aswaterman:
The spec's intent is obvious in that if MXR did affect M-mode pointer masking, then S-mode could control M-mode's addressing behavior by writing MXR, which would be bad.  Nevertheless, this has been a source of confusion.

We don't need the old MXR text that says it is only in effect when paging is enabled, because that's redundant with the text below that says MXR doesn't affect physical memory accesses.

@nadime15 does this look OK?

cc @davidharrishmc 

Resolves #3313

### DIFF
--- a/src/priv/machine.adoc
+++ b/src/priv/machine.adoc
@@ -602,7 +602,8 @@ less privileged than M also sets csr::[mprv]=0.#
 The csr::[mxr] (Make eXecutable Readable) bit modifies the privilege with which loads access virtual memory.
 [#norm:mstatus_mxr_op]#When csr::[mxr]=0, only loads from pages marked readable (R=1 in <<sv32pte>>) will succeed.
 When csr::[mxr]=1, loads from pages marked either readable or executable (R=1 or
-X=1) will succeed.  csr::[mxr] has no effect when page-based virtual memory is not in effect.#
+X=1) will succeed.
+csr::[mxr] only affects loads whose effective privilege mode is less than M.#
 [#norm:mstatus_mxr_rdonly0_no_smode]#csr::[mxr] is read-only 0 if S-mode is not supported or if csr:satp[mode] is read-only 0; otherwise, it must be writable.#
 
 [NOTE]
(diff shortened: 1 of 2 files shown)


nadime15 (approved):
Looks good to me, thanks!

PR #3319, opened 2026-08-18
zvk: fix VSM3ME message expansion pseudocode

skyhgzsh:


nibrunieAtSi5:
It looks like the original pseudo-code was somewhat consistent (although certainly not proper SAIL):
- initial definition of `w[i]` for i 16..23 from operations on `w<i>`
- then bit reverse of `w[i]` into `wi`
- finally concatenation and storage

This PR swaps uses of `w[i]` and `wi`, why not, but also why ?

The original pseudo-code had some issues:
- use of `W[i]` and w[i]` for i 16..23 to mean the same variable 
- use of `Bits(32)` rather than `bits(32)`

And to agree with the AI-review, 
Since in https://github.com/riscv/riscv-isa-manual/blob/main/src/unpriv/zvk.adoc#insnvaeskf2vi, we have
```
      set_velem(vd, EGW=128, i, w[3:0]);
```
we should use `[]`
```
      set_velem(vd, ..., i, w[x:y]);
```
or with `{}`
```
      set_velem(vd, ..., i, {a @ b @ c @ ... });
```
but not a mix of the two.

skyhgzsh:
fixed capitaled bits and slice grammar

skyhgzsh:
The main purpose of this commit is to correct the algorithm of SM3 message expansion.
Previous version of this pseudocode use 'array elements 16:23' as the results of ZVKSH_W, and reverse the 'array elements' to 'distinct variables' and use those variables as the final output and also the source of the later iterations. 
However, according to SM3 spec, the intermediate results from earlier iterations are used 'without' reverse. Reverse is only required for 'Endian Transform'.

nibrunieAtSi5:
I am not sure I understand your comment here @skyhgzsh 

> However, according to SM3 spec, the intermediate results from earlier iterations are used 'without' reverse. Reverse is only required for 'Endian Transform'.

Your PR does not change the way input or outputs are byte-reversed, it seems to me that you are just changing the naming conventions but keeping byte-reversal on inputs AND outputs.

I believe this pseudo-code is aligned with the spike implementation
https://github.com/riscv-software-src/riscv-isa-sim/blob/master/riscv/insns/vsm3me_vv.h

I remember the TG had some conversations about the endianess and reversal for SM3 and maybe this is the point of your comment.

skyhgzsh:
sm3diff

Compare the orange arrow and the yellow one. The source of the latter extension iteration should be the original former iteration results, not reversed results.

nibrunieAtSi5:
@skyhgzsh, I see, thank you for the picture. I believe your are correct and the current pseudo-code is misleading, using values before their definition (and with the byte reverse).

nibrunieAtSi5 (approved):

aswaterman (approved):

PR #3323, opened 2026-08-20
Clarify that debug mode CSRs are inaccessible from M-mode

davidharrishmc:
Per discussion of Allen Baum with Andrew Waterman


aswaterman (approved):
The "should" was inserted to accommodate priv-1.9 implementations around the time that the debug CSRs were introduced.  @allenjbaum and I believe that it "should" no longer be operative... :)

PR #3337, opened 2026-08-23
Clarify that vsatp.ASID is the same width as satp.ASID

### LINKED ISSUE #3335: [Clarification] - Is `ASIDLEN` required to be the same for `satp` and `vsatp`
What part of the ISA is unclear? (cc @riscv/arc)

The satp section of the manual defines ASIDLEN as the number of implemented bits in the ASID field. The vsatp section of the manual does not mention ASIDLEN, but `vsatp` includes an `ASID` field. Do satp and vsatp need to have the same number of implemented ASID bits? Or is there really an SATP_ASIDLEN and a separate VSATP_ASIDLEN? This came up during discussions in the Parameters SIG.

aswaterman:
Fixes #3335

### DIFF
--- a/src/priv/hypervisor.adoc
+++ b/src/priv/hypervisor.adoc
@@ -1417,6 +1417,9 @@ ignored as it is for `satp`, or the fields of `vsatp` are treated as *WARL* in
 the normal way.# [#norm:vsatp_mode_unsupported_v1]#However, when V=1, a write to `satp` with an unsupported
 MODE value _is_ ignored and no write to `vsatp` is effected.#
 
+[#norm:vsatp_asidlen]#The ASID field in the `vsatp` register has the same number of implemented bits
+as that of the `satp` register, i.e., it is ASIDLEN bits wide.#
+
 [[norm:vsatp_v0]]
 When V=0, `vsatp` does not directly affect the behavior of the machine,
 unless a virtual-machine load/store (HLV, HLVX, or HSV) or the MPRV
--- a/normative_rule_defs/hypervisor.yaml
+++ b/normative_rule_defs/hypervisor.yaml
@@ -378,6 +378,8 @@ normative_rule_definitions:
     impl-def-category: WARL
   - name: vsatp_mode_unsupported_v1
     tags: ["norm:vsatp_mode_unsupported_v1"]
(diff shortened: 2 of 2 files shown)


src/priv/hypervisor.adoc
```diff
@@ -1417,6 +1417,9 @@ ignored as it is for `satp`, or the fields of `vsatp` are treated as *WARL* in
 the normal way.# [#norm:vsatp_mode_unsupported_v1]#However, when V=1, a write to `satp` with an unsupported
 MODE value _is_ ignored and no write to `vsatp` is effected.#
 
+The ASID field in the `vsatp` register has the same number of implemented bits
+as that of the `satp` register, i.e., it is ASIDLEN bits wide.
```
jordancarlin: Seems like this should be a normative rule. Could you tag it and add it to `normative_rule_defs/hypervisor.yaml`?

aswaterman: I knew this was a trap

jordancarlin: Haha. Thanks for taking care of it.

src/priv/hypervisor.adoc
```diff
@@ -1417,6 +1417,9 @@ ignored as it is for `satp`, or the fields of `vsatp` are treated as *WARL* in
 the normal way.# [#norm:vsatp_mode_unsupported_v1]#However, when V=1, a write to `satp` with an unsupported
 MODE value _is_ ignored and no write to `vsatp` is effected.#
 
+[#norm:vsatp_asidlen]#The ASID field in the `vsatp` register has the same number of implemented bits
+as that of the `satp` register, i.e., it is ASIDLEN bits wide.#
```
jordancarlin, line 1421: There is no functional difference so this is fine as is, but for future reference, the following format tends to be simpler for tags that span an entire paragraph.

```suggestion
[[norm:vsatp_asidlen]]
The ASID field in the `vsatp` register has the same number of implemented bits
as that of the `satp` register, i.e., it is ASIDLEN bits wide.
```

aswaterman: Yeah, that's cleaner, good to know for future ref.

PR #3343, opened 2026-08-24
Add Zibi extension for branch with immediate

wangpc-pp:
The Zibi extension adds the beqi and bnei conditional branches, which compare an integer register against a small immediate constant instead of a second register. This avoids materializing the constant in a register before branching, reducing register pressure and potential spills.

Both instructions reuse the B-type format: the field that would carry rs2 instead holds a 5-bit cimm operand. The cimm value is not sign-extended; cimm=0 decodes to -1 and values 1 through 31 decode to the corresponding positive integers. The instructions write no integer register, access no memory, and follow standard branch address-misaligned exception behavior.

Spec: https://github.com/riscv/zibi


src/unpriv/zibi.adoc
```diff
@@ -0,0 +1,134 @@
...
+==== Instructions
+
+[#insns-beqi,reftext="Branch if equal immediate"]
+===== insn:beqi[]
+
+Synopsis::
+Branch if equal immediate
+
+Mnemonic::
+beqi _rs1_, _cimm_, _offset_
```
topperc: Elsewhere `cimm` seems to refer to the encoded value. Should we have some way to refer to the -1, 1-31 value that the assembler uses?

src/unpriv/zibi.adoc
```diff
@@ -0,0 +1,134 @@
...
+  {bits: 5,  name: 'imm[4:1|11]',   attr: ['offset[4:1|11]'] },
+  {bits: 3,  name: 'funct3',        attr: ['BEQI:010', 'BNEI:011'] },
+  {bits: 5,  name: 'rs1',           attr: ['src'] },
+  {bits: 5,  name: 'cimm',          attr: ['constant'] },
+  {bits: 7,  name: 'imm[12|10:5]',  attr: ['offset[12|10:5]'] },
+], config:{fontsize: 10}}
+....
+
+[#norm:zibi_cimm_decode]#All _cimm_ encodings are valid. The _cimm_ field
+is not sign-extended; it decodes to an XLEN-bit comparison constant:
```
aswaterman: This description is correct as-written, but I suspect "is not sign-extended" risks being confusing to some readers since one of the patterns corresponds to a sign-extended pattern (with the others being zero-extended).  Maybe just drop the phrase:

`The _cimm_ field encodes an XLEN-bit comparand: pattern 0 encodes the value -1, whereas patterns 1 through 31 encode the values 1 through 31, respectively.`

Or something like that.

wangpc-pp: Good point. I dropped the "is not sign-extended" phrasing and now state the encoding directly: pattern 0 encodes -1, and patterns 1 through 31 encode 1 through 31, respectively.

src/unpriv/zibi.adoc
wangpc-pp: Agreed. I now distinguish the two: cimm names the 5-bit encoded field, and constant names the XLEN-bit value used in the comparison (-1 for pattern 0, and 1 through 31). The assembler-visible operand in the mnemonics is now written as constant (e.g. `beqi rs1, constant, offset`), matching the 'constant' label in the encoding diagram and the decode formula.

aswaterman: You might consider "comparand" instead of "constant" since the latter is pretty generic.  OK with me regardless.

wangpc-pp: Done in https://github.com/riscv/riscv-isa-manual/pull/3343/commits/64d5c08d23a31285813e47574e4407beb98715eb.

rpsene:
@wangpc-pp heads-up: I pushed to `main-zibi` on your fork and rewrote its history, so your local branch is now out of date.

What changed:
- **Two new commits**, both on top of your work:
  - `ee9ab38` sets the Zibi version to 0.9 for the Freeze milestone.
  - `8e31590` replaces the `[stem]` block with a table, because the manual build does not render math and the PDF showed raw LaTeX. The normative text is unchanged.
- **Your three commits are re-signed** so the PR passes the commit signing check. They keep you as the author, and their content and sign-offs are unchanged. Only the committer and the signature are new.

Before you push again, please update your local branch so these commits are not overwritten:

```
git fetch origin
git checkout main-zibi
git reset --hard origin/main-zibi   # or: git pull --rebase, if you have local changes
```

If you add more commits, please sign them (`git commit -S -s`) so the signing check keeps passing. The v0.9 PDF for the Freeze milestone vote is attached to RVS-4691.

src/unpriv/zibi.adoc
```diff
@@ -0,0 +1,129 @@
...
+[NOTE]
+====
+Profiling general benchmarks shows that positive immediate values occur more
+frequently than negative immediate values.
+Comparison with zero can be performed using insn:beq[rs1,x0,offset] or
+insn:bne[rs1,x0,offset].
+====
+
+[#norm:zibi_no_state]#The insn:beqi[] and insn:bnei[] instructions do not
+write any integer register and do not access memory.#
```
aswaterman: It's a little odd to include in the spec what instructions _don't_ do.  Please remove this clause.

PR #3344, opened 2026-08-24
Restore "or vstimecmp" in norm:menvcfg_stce_op2

### LINKED ISSUE #3329: [NormRules] - menvcfg.STCE rule lost "or vstimecmp" when the Sstc text was moved
## Description
norm:menvcfg_stce_op2 currently reads:
  

> When this extension is implemented and STCE in menvcfg is zero, an attempt to access stimecmp in a mode other than M-mode raises an illegal-instruction exception, STCE in henvcfg is read-only zero, and STIP in mip and sip reverts to its defined behavior as if this extension is not implemented.

Before the Sstc chapter was split up this said "an attempt to access stimecmp or vstimecmp". The two words were dropped in #2504, which essentially only moved moved text, so I think it was accidental rather than intentional.

The following was removed from src/sstc.adoc:

> When this extension is implemented and STCE in `menvcfg` is zero, an attempt to access `stimecmp` or `vstimecmp` in a mode other than M-mode raises an illegal-instruction exception, ...

The following was added to src/machine.adoc:

> When this extension is implemented and `STCE` in `menvcfg` is zero, an attempt to access `stimecmp` in a mode other than M-mode raises an illegal-instruction exception, ...

norm:mcounteren_tm_clr moved in the same commit and kept both names ("attempts to access the stimecmp or vstimecmp register while executing in a mode less privileged than M"), so the two rules now disagree about whether vstimecmp is covered.

Spike implements the old wording. stimecmp_csr_t::verify_permissions is shared by stimecmp and vstimecmp, and its first check is on MENVCFG_STCE with the comment "access to (v)stimecmp with MENVCFG.STCE = 0".
 
## Any changes required
Restore "or vstimecmp" in norm:menvcfg_stce_op2, so it reads "an attempt to access stimecmp or vstimecmp in a mode other than M-mode raises an illegal-instruction exception".

(cc @riscv/csc)

nadime15:
The rule lost "or vstimecmp" when the Sstc text was moved in #2504.

Fixes #3329

### DIFF
--- a/src/priv/machine.adoc
+++ b/src/priv/machine.adoc
@@ -2221,12 +2221,12 @@ enabled, see <<smcdeleg>>. When CDE=0, the Smcdeleg and Ssccfg extensions appear
 [#norm:menvcfg_stce_op1]#The Sstc extension adds the `STCE` (STimecmp Enable) bit to `menvcfg` CSR.#
 [#norm:menvcfg_stce_rdonly0]#When the Sstc extension is not implemented, `STCE` is read-only zero.#
 [#norm:menvcfg_stce_op2]#The `STCE` bit enables `stimecmp` for S-mode when set to one.
-When this extension is implemented and `STCE` in `menvcfg` is zero, an attempt to access `stimecmp`
-in a mode other than M-mode raises an illegal-instruction exception, `STCE` in
-`henvcfg` is read-only zero, and `STIP` in `mip` and `sip` reverts to its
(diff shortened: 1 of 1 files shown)


aswaterman (approved):

aswaterman:
@nadime15 Looks like you don't have your signing key uploaded to github?

nadime15:
@aswaterman Yeah I generated a new one and uploaded it. The signing check passes now.

PR #2901, opened 2026-04-14
Add Zisb extension

aswaterman:
This draft PR incorporates the proposed Zisb extension and will not be merged until it has been ratified.

cc @ved-rivos @dkruckemyer-work @gfavor

PR #3352, opened 2026-08-31
Add Ziteb extension

aswaterman:


src/unpriv/ziteb.adoc
```diff
@@ -0,0 +1,240 @@
...
+Additionally, in the course of executing instructions out-of-order, a younger
+instruction has a _speculative dependency_ on an older instruction when at
+least one of the following is true:
+
+* A younger instruction has a _speculative control dependency_ on an older
+control transfer instruction if the older instruction has _not_ executed and
+its execution affects the execution of the younger instruction.
+
+* A younger instruction has a _speculative register dependency_ on an older
+instruction if the older instruction has _not_ executed and its predicted
```
tariqkurd-repo, line 47: I think this would be clearer if it used the phrase "data value prediction" in the definition. That would make the comparison with Aarch64 CSDB simpler:

`The CSDB instruction is a memory barrier instruction that controls speculative execution arising from data value
prediction.`

Are there any differences between TEB.RD and CSDB?

lfiolhais: The biggest different is that the TEB depends also on in-hart uarch state instead of just memory subsystem uarch state. The definition of "signaling instruction" is 

https://github.com/riscv/riscv-isa-manual/blob/3ce933494a287e2c762ce4cbc498fad0bf155f26/src/unpriv/ziteb.adoc?plain=1#L87-L92

src/unpriv/ziteb.adoc
```diff
@@ -0,0 +1,241 @@
...
+An instruction executes when it produces its register and memory results, i.e.,
+the values that will be written into architectural registers or memory.
+For a control-transfer instruction, execution includes determining the next
+instruction in program order.
+For an explicit memory access instruction, execution includes performing
+address translation to determine the physical address of the memory location
+accessed by the instruction.
+
+The program order is the order of instructions resulting from running
+a program on a simple in-order hart.
```
lfiolhais: I'd avoid the usage of the expression "simple in-order hart" as I believe it will generate confusion. I can think of "simple in-order harts" which contain out-of-order elements and would invalidate the mental model that is being explained here. IMO the sentence should be
> The program order is the order of instructions resulting from executing a program one instruction at a time.

aswaterman: Yes, I agree, this text is from an informal sketch of the model and still could be improved.

aswaterman: "One at a time" doesn't say the right thing either, though. That says there's no overlap; it doesn't specify the order.  It needs a  "sequentially" in there, too.

src/unpriv/ziteb.adoc
aswaterman: I think "predicted register result" is already pretty clear and copying the CSDB wording is not a goal.  But I'll leave the decision to @dkruckemyer-work

src/unpriv/ziteb.adoc
lfiolhais: Funny you should mention that. I avoided the word "sequentially" because I was concerned about the argument that uarchs can be sequential but still exhibit speculative out-of-order behavior. It seems we agree on the same interpretation of "sequentially": it is sequential in both architectural and uarch execution. The "one at a time" disambiguates.

lfiolhais:
Apologies on drip-feeding comments but I'm having difficulty finding a slot where I can post all of my thoughts about this proposal. Here is my latest attempt.

Organizational: I'm afraid I've not been following the TEB work very closely since the TG disbanded/stopped meeting, but my impression is that this set of instructions is going to be pursued as an FT. Is this still the case? I assume the goal of this extension is to supersede https://github.com/riscv/riscv-isa-manual/pull/2901, focusing only on the speculation barrier instructions.

**Threat Model**: The extension refers several times to "timing channels", yet it doesn't provide a definition for it. We have had extensive discussion on the fence.t TG about what "timing channels" should be, and the conclusion was that it is any mechanism architectural or otherwise that allows an attacker to measure a timing difference between two uarch events. This means that the "timing channel" can be a read of the time or cycle CSRs, it can be a different process running a counter, or it can even be someone holding a watch in their hand. What is the definition employed here? We also had similar challenges when writing the security model. It would be great if we all agreed on an universal definition.

**Extension name**: the extension is called "Extension for Transient-Execution Barriers" which would imply these barriers affect all types of transient-execution. Halfway through it is specified that these instructions only apply to _transient execution that result in cover timing channels_. Changing the name of the extension to better understand its use would greatly benefit the reader. I'm terribly with names but I'd throw in: "Extension for Transient-Execution Timing Barriers". I think we all agree that non-timing channels are out-of-scope.

**Types of signaling instructions**: when defining "signaling instructions" the text uses "may" multiple times. https://github.com/riscv/riscv-isa-manual/blob/0735845445bf5cc515431c52ca10202dafe217fb/src/unpriv/ziteb.adoc?plain=1#L87-L88
The text then subsequently concretely defines what _is_ a "signaling instruction" https://github.com/riscv/riscv-isa-manual/blob/0735845445bf5cc515431c52ca10202dafe217fb/src/unpriv/ziteb.adoc?plain=1#L89-L92. IMO the best solution would be to drop the "may" in the initial definition. Whether the instruction can be used for data exfiltration is then left to the threat model at play.

**Usage of Covert Channels**: I don't fully grasp why the proposal is only concerned with covert channels. I interpret the difference between a covert- and side-channel by analyzing if the victim execution is willingly cooperating in transmitting information through non-architectural mechanisms. The barriers described herein would be useful for programs that contain a uarch execution which results in a side-channel.

**Definition of Transient-Executing Instructions**: The following definition is provided for an "executing instruction":
https://github.com/riscv/riscv-isa-manual/blob/0735845445bf5cc515431c52ca10202dafe217fb/src/unpriv/ziteb.adoc?plain=1#L22-L28
However, in the literature, transient-execution has a much broader definition. It encompasses all "behavior" outside of providing architectural results. For example, pre-fetching instructions in a pipelined design with an I$ (cache lines are allocated), or updating predictors (predictors are updated before any architectural result is generated or are updated using out-of-order uarch results). Note that both of the previous examples, in certain implementations, don't require any architectural result---right or wrong---to be generated. As a result, the current barrier definition will still yield covert/side timing channels under certain scenarios. Given the definition of transient-execution, I assume that the examples I provided are out-of-scope. Is this correct?

As a broader comment, IMO the extension would be tighter if it defined a microarchitectural consistency model, where execution is one component of the model. A microarchitectural operation depends on a set of loads from different uarch structures. If one of these loads is reliant on non-speculation-verified data, the operation is deemed speculative. A store, in this instance, is defined as the verification of the speculative load, i.e., if the there is a mispeculation a store is triggered to a uarch address. If there are no stores to the same uarch load and all dependent older instructions that would store to the original uarch load addresses are not performed, the uarch is speculation-free. The definition for ordering operations in relation to the barriers would then follow the specification. (this is a crappy explanation I'll try to expand it when I have more time).

iamywang:
@aswaterman Following up on @lfiolhais's comments, I have a few thoughts to share. I am very interested in this area and would be glad to collaborate on it.

> However, in the literature, transient-execution has a much broader definition. It encompasses all "behavior" outside of providing architectural results. For example, pre-fetching instructions in a pipelined design with an I$ (cache lines are allocated), or updating predictors (predictors are updated before any architectural result is generated or are updated using out-of-order uarch results). Note that both of the previous examples, in certain implementations, don't require any architectural result---right or wrong---to be generated. As a result, the current barrier definition will still yield covert/side timing channels under certain scenarios. Given the definition of transient-execution, I assume that the examples I provided are out-of-scope. Is this correct?

I believe that the behaviors you mentioned, such as prefetching and predictor updates, lean more towards changes in microarchitectural states. If they do not trigger out-of-bounds, unauthorized, or stale data accesses, they should be treated as covert channels or timing side channels. Nevertheless, if such state changes subsequently trigger out-of-bounds, unauthorized, or stale data accesses, this situation appears to fit within the scope of transient execution in Ziteb. Additionally, if the software itself already produces timing channels at the ISA level through microarchitectural structures like caches, TLBs, and branch predictors,  Ziteb shouldn't be required to guarantee it remains leak-free under a speculative oracle. I am open to correction if my understanding is flawed.

> As a broader comment, IMO the extension would be tighter if it defined a microarchitectural consistency model, where execution is one component of the model. A microarchitectural operation depends on a set of loads from different uarch structures. If one of these loads is reliant on non-speculation-verified data, the operation is deemed speculative. A store, in this instance, is defined as the verification of the speculative load, i.e., if the there is a mispeculation a store is triggered to a uarch address. If there are no stores to the same uarch load and all dependent older instructions that would store to the original uarch load addresses are not performed, the uarch is speculation-free. The definition for ordering operations in relation to the barriers would then follow the specification. (this is a crappy explanation I'll try to expand it when I have more time).

Do you think a formal speculative semantic model is necessary? For example, under a formal hardware-software security contract such as [Speculative Non-Interference (SNI)](https://ieeexplore.ieee.org/document/9519500), consider a program parameterized by distinct secret data (e.g., the 256 cache slots mentioned in the Ziteb proposal, which can be treated as 256 different secrets). If the program yields indistinguishable observations under a non-speculative ISA model (e.g., QEMU) then it must also yield indistinguishable observations on a real speculative implementation, even if the branch is transiently taken.

I would like to raise an additional point regarding the current draft.

https://github.com/riscv/riscv-isa-manual/blob/0735845445bf5cc515431c52ca10202dafe217fb/src/unpriv/ziteb.adoc?plain=1#L82-L92

1. Would it be beneficial for the proposal to explicitly define the relationship between signaling instructions and their specific transient execution triggers? A mapping table such as the following could help clarify for software developers where and when to insert the barrier:

| Signaling Instruction | Transient Execution | Covert Channel |
| -- | -- | -- |
| conditional branch | branch misprediction | Cache, TLB, ... |
| indirect branch | branch misprediction | Cache, TLB, ... |
| return branch | branch misprediction | Cache, TLB, ... |
| memory access | store-to-load forwarding | Cache, TLB, ... |
| memory access | fault / exception | Cache, TLB, ... |
| register access | fault / exception | Cache, TLB, ... |
| ... | ... | ... |

> Threat Model: The extension refers several times to "timing channels", yet it doesn't provide a definition for it. We have had extensive discussion on the fence.t TG about what "timing channels" should be, and the conclusion was that it is any mechanism architectural or otherwise that allows an attacker to measure a timing difference between two uarch events. This means that the "timing channel" can be a read of the time or cycle CSRs, it can be a different process running a counter, or it can even be someone holding a watch in their hand. What is the definition employed here? We also had similar challenges when writing the security model. It would be great if we all agreed on an universal definition.

2. Maybe including a threat model section would also help clarify the security boundaries and assumptions of the extension.

lfiolhais:
> I believe that the behaviors you mentioned, such as prefetching and predictor updates, lean more towards changes in microarchitectural states. If they do not trigger out-of-bounds, unauthorized, or stale data accesses, they should be treated as covert channels or timing side channels. Nevertheless, if such state changes subsequently trigger out-of-bounds, unauthorized, or stale data accesses, this situation appears to fit within the scope of transient execution in Ziteb.

I disagree with your framing. The definition of out-of-bounds or stale data access is up to the program being executed. The uarch has no way of knowing if it's access is within the expected behavior of the program until the required instructions are executed to assert such a statement. I agree with the detection of unauthorized access. That information is readily available to the uarch and is not program specific.

> Do you think a formal speculative semantic model is necessary? For example, under a formal hardware-software security contract such as [Speculative Non-Interference (SNI)](https://ieeexplore.ieee.org/document/9519500), consider a program parameterized by distinct secret data (e.g., the 256 cache slots mentioned in the Ziteb proposal, which can be treated as 256 different secrets). If the program yields indistinguishable observations under a non-speculative ISA model (e.g., QEMU) then it must also yield indistinguishable observations on a real speculative implementation, even if the branch is transiently taken.

My proposal applies to more components than just speculation structures, and in fact is completely decoupled from it. A speculative semantic model can only exist for specific uarchs with specific speculative structures. It is true that some speculative structures are ubiquitous but they aren't mandated by the ISA. The whole idea of the uarch consistency model is to create a framework where any uarch can be plugged into while also being valid at the ISA level.
 
> I would like to raise an additional point regarding the current draft.
> 
> https://github.com/riscv/riscv-isa-manual/blob/0735845445bf5cc515431c52ca10202dafe217fb/src/unpriv/ziteb.adoc?plain=1#L82-L92
> 
>     1. Would it be beneficial for the proposal to explicitly define the relationship between signaling instructions and their specific transient execution triggers? A mapping table such as the following could help clarify for software developers where and when to insert the barrier:
> 
> 
> Signaling Instruction 	Transient Execution 	Covert Channel
> conditional branch 	branch misprediction 	Cache, TLB, ...
> indirect branch 	branch misprediction 	Cache, TLB, ...
> return branch 	branch misprediction 	Cache, TLB, ...
> memory access 	store-to-load forwarding 	Cache, TLB, ...
> memory access 	fault / exception 	Cache, TLB, ...
> register access 	fault / exception 	Cache, TLB, ...
> ... 	... 	...
> 

I would disagree with this addition. There are many uarch structures that can be used to create a covert-/side-channel. Having an exhaustive list of all structures would be an impossible maintenance burden.

PR #3356, opened 2026-09-01
Clarify that speculative VS-stage PTE accesses can set G-stage D bit

aswaterman:
cc @gfavor @dkruckemyer-work

### DIFF
--- a/src/priv/supervisor.adoc
+++ b/src/priv/supervisor.adoc
@@ -1631,12 +1631,13 @@ Two schemes to manage the A and D bits are defined:
   PTE as part of, and before, conditionally updating the PTE value.
   Updates of the A bit may be performed as a result of speculation, even
   if the associated memory access ultimately is not performed
-  architecturally. However, updates to the D bit, resulting from an
-  explicit store, must be exact (i.e., non-speculative), and observed in
-  program order by the local hart. When two-stage address translation is
-  active, updates to the D bit in G-stage PTEs may be performed by an
-  implicit access to a VS-stage PTE, if the G-stage PTE provides write
-  permission, before any speculative access to the VS-stage PTE.        +
+  architecturally.
+  However, updates to the D bits in VS-stage and G-stage PTEs
+  that result from an explicit store must be exact (i.e. non-speculative) and
+  observed in program order by the local hart.
+  If the G-stage PTE provides
+  write permission, an update to the D bit in a G-stage PTE that results from
+  an implicit access to a VS-stage PTE may be inexact (i.e. speculative). +
                                                                         +
   The PTE update must appear in the global memory order before the
   memory access that caused the PTE update and before any subsequent


src/priv/supervisor.adoc
```diff
@@ -1631,12 +1631,14 @@ Two schemes to manage the A and D bits are defined:
...
-  implicit access to a VS-stage PTE, if the G-stage PTE provides write
-  permission, before any speculative access to the VS-stage PTE.        +
+  architecturally.
+  However, updates to the D bits in VS-stage and G-stage PTEs
+  that result from an explicit store must be exact (i.e. non-speculative) and
+  observed in program order by the local hart.
+  If the G-stage PTE provides
+  write permission, an update to the D bit in a G-stage PTE that results from
+  an implicit access to a VS-stage PTE may be inexact (i.e. speculative) and
+  is observed before the access to the VS-stage PTE.                    +
```
gfavor: Delete 'is' since the G-stage PTE update _may_ be observed before ... (i.e. not guaranteed).

aswaterman: Is your point that it might not be observed at all?

What we are trying to say is that it cannot be observed after; "may" might send the wrong signal here.

To convey both notions, we could say "is ordered before".  @dkruckemyer-work WDYT?

dkruckemyer-work: Or we just stop this sentence at "inexact (i.e. speculative)" and rely on the immediately following sentence to capture the ordering requirement?

aswaterman: Ah, true. That works for me.  @gfavor OK with you?

src/priv/supervisor.adoc
```diff
@@ -1631,12 +1631,14 @@ Two schemes to manage the A and D bits are defined:
...
-  architecturally. However, updates to the D bit, resulting from an
-  explicit store, must be exact (i.e., non-speculative), and observed in
-  program order by the local hart. When two-stage address translation is
-  active, updates to the D bit in G-stage PTEs may be performed by an
-  implicit access to a VS-stage PTE, if the G-stage PTE provides write
-  permission, before any speculative access to the VS-stage PTE.        +
+  architecturally.
+  However, updates to the D bits in VS-stage and G-stage PTEs
+  that result from an explicit store must be exact (i.e. non-speculative) and
+  observed in program order by the local hart.
```
dkruckemyer-work, line 1637: I realize the ship has probably already sailed on this.... but do the D bit updates really need to be in program order with respect to each other? Or just before the explicit stores?

dkruckemyer-work: Oh sorry, it's only by the local hart. Never mind.

aswaterman: But yeah, I agree that would have been overly strict without an obvious benefit.

src/priv/supervisor.adoc
aswaterman: Agreed upon during ARC meeting.

aswaterman (commented):
Approved verbally during ARC meeting.

PR #3359, opened 2026-09-02
Restore Zve floating-point permutation requirements

abhinavagarwal07:
#3031 dropped the Zve floating-point permutation requirements while splitting the vector chapter.

- Restore EEW=32 support for Zve32f and EEW=32/64 for Zve64d.
- Add the corresponding normative-rule definitions.
- Fix the Zve64f union typo.

Zve64f inherits EEW=32 support from Zve32f. No encoding change.


aswaterman (approved):
Oops, thank you.

PR #3364, opened 2026-09-06
Specify `svg` for more asciidoctor diagrams

mlechu:
I noticed some rather pixelated diagrams in the PDF manual and thought it might be an easy fix.

Before:
before

After:
after


aswaterman (approved):

wmat:
I thought this was completed by #1025 a few years ago.  Thanks for the fix, I'll look into making svg the default format.

Issue #3371, opened 2026-09-10
[Clarification] - Is mstatus.SPELP read-only zero if supervisor mode not supported


davidharrishmc:
## Summary
What part of the ISA is unclear? (cc @riscv/arc)

The priv spec is silent about whether mstatus.SPELP is writable or read-only zero when Zicfilp is implemented but supervisor mode is not supported. Other supervisor-related fields such as mstatus.SPIE and SIE are specified to be read-only 0 when supervisor mode is not supported.

I am wondering whether this is intentional (WARL, either writable or ROZ allowed) or an omission that should be corrected.

@riscv/arc 

---

## Expected vs Observed
Sail makes mstatus.SPELP writable when Zicfilp is supported and supervisor mode is not.  Spike and whisper make it read-only zero.

ved-rivos:
SPELP is provided for the purpose of taking traps to S/HS mode. When S/HS is not implemented, traps cannot be taken to these modes, and this bit is not provided - is read-only zero.

PR #3372, opened 2026-09-10
mstatus.SPELP is read-only zero if supervisor mode is not implemented 

davidharrishmc:
Clarification per
https://github.com/riscv/riscv-isa-manual/issues/3371


davidharrishmc:
@aswaterman

aswaterman (approved):

PR #3373, opened 2026-09-14
priv: untag text that is not a hardware normative rule

davidharrishmc:
Removes normative-rule tags from privileged-spec text that is not a hardware
requirement: motivation and purpose paragraphs, spec-authoring policy, software
responsibilities, "Ordinarily"/"typically" hedges, threat-model description of
behaviour *without* the extension, and derived values in an informative table.

Where only part of a tagged region was non-normative the anchor is trimmed rather
than deleted, so the rule that was buried in the paragraph survives (`pma_precise_recom`,
`wfi_op`, `satp_op_sfence_vma`, `hgatp_op_hfence_gvma`, `H_scsrs_nomatch`,
`stateen_implicit_state_update`, `Svpbmt_aliasing_attribute`, `sscratch`).

Two Smctr chapter-overview tags that restate Behavior-section rules are dropped.

Deliberately **not** untagged: tags backing implementation-defined behaviours
(`critical_error`, `rst_and_nmi_addr`, `pma_mo_dyn`, `pma_cache_mm_all_atomics`,
`hsxlen`, `vsxlen`). "might", "can" and "platform-specific" is what an
implementation-defined behaviour reads like; the extracted set still holds 209 of
them, unchanged.

Rules 3307 → 3277. No visible text changes — anchors and YAML only.

**Downstream:** this PR deletes rule names that `riscv-arch-test`'s
`coverpoints/norm/*.yaml` reference by string. Every deletion is recorded, with the
affected downstream file, in the old→new map kept alongside this work.


davidharrishmc (commented):
These rules tighten up tags to only match the normative part of a paragraph, or exclude text that is nonnormative.  Look reasonable to me.

src/priv/smstateen.adoc
```diff
@@ -97,33 +97,32 @@ exception instead of an illegal-instruction exception.#
...
+supervisor-level `sstateen` CSR are applicable. The intention is to allocate bits
 for this purpose starting at the most-significant end, bit 63, through to bit
-32, and then on to the next-higher `mstateen` CSR.# [#norm:mstateen_bit_encroachment]#If the rate that bits are
+32, and then on to the next-higher `mstateen` CSR. If the rate that bits are
 being allocated from the least-significant end for `sstateen` CSRs is
 sufficiently low, allocation from the most-significant end of `mstateen` CSRs may
 be allowed to encroach on the lower 32 bits before jumping to the next-higher
-`mstateen` CSR.# [#norm:sstateen_encroachment_bits_roz]#In that case, the bit positions of "`encroaching`" bits will remain
+`mstateen` CSR. [#norm:sstateen_encroachment_bits_roz]#In that case, the bit positions of "`encroaching`" bits will remain
 forever read-only zeros in the matching `sstateen` CSRs.#
```
jordancarlin: `sstateen_encroachment_bits_roz` should be removed too.

src/priv/svpbmt.adoc
```diff
@@ -81,13 +81,13 @@ optimizations will likely improve performance when applied with adequate
...
+When Svpbmt is used with non-zero PBMT encodings, it is possible for
 multiple virtual aliases of the same physical page to exist
 simultaneously with different memory attributes. It is also possible for
 a U-mode or S-mode mapping through a PTE with Svpbmt enabled to observe
 different memory attributes for a given region of physical memory than a
 concurrent access to the same page performed by M-mode or when
-MODE=Bare. In such cases,
+MODE=Bare. [#norm:Svpbmt_aliasing_attribute]#In such cases,
 the behaviors dictated by the attributes
 (including coherence, which is otherwise unaffected) may be violated.#
```
jordancarlin: This change makes the rule harder to understand because it is no longer clear what "In such cases" refers to.

jordancarlin (approved):

PR #3374, opened 2026-09-14
unpriv: untag text that is not a hardware normative rule

davidharrishmc:
Removes normative-rule tags from unprivileged-spec text that is not a hardware
requirement: what compiled software does in a prologue/epilogue, assembler syntax,
extension-introduction prose, NOTE consequences of rules stated elsewhere, and RVWMO
scope caveats.

Where a region mixed software guidance with a rule the anchor is trimmed rather than
deleted, so the rule survives (`fence_cons_ok`, `csr_fence_ordering`).

`lr_sc_aq_rl_software_rule` is **dropped**, not trimmed. Both halves address software,
and the half that reads like a rule — "`lr.rl` and `sc.aq` instructions are not
guaranteed to provide any stronger ordering than those with both bits clear" — withholds
a guarantee rather than imposing a requirement. It does not forbid stronger ordering, so
no implementation can fail to comply with it.

The four RVWMO PPO category labels ("Overlapping-Address Orderings", "Explicit
Synchronization", "Syntactic Dependencies", "Pipeline Dependencies") are untagged. They
are bullet headings; each numbered condition below them is already a complete clause, and
the antecedent they share ("…and any of the following hold:") is tagged separately as
`rvwmo_ppo_lead-in`. They were tags of the single `rvwmo_ppo` rule and never rule names.

Deliberately **not** untagged: `ee_impl_flex` and `NTL_trap_behavior`, which back
implementation-defined behaviours — hedged phrasing is what those look like. Also kept
`zk_scalar_xreg_op` (the 2-read-1-write constraint is an architectural property, not an
introduction) and the `xlen_definition` / `base_integer_variants` definitions.

Rules 3307 → 3296. No visible text changes — anchors and YAML only.

**Downstream:** this PR deletes rule names that `riscv-arch-test`'s
`coverpoints/norm/*.yaml` reference by string. Every deletion is recorded, with the
affected downstream file, in the old→new map kept alongside this work.


davidharrishmc (commented):
These look reasonable to me.

jordancarlin (approved):

PR #3375, opened 2026-09-14
priv: make tagged regions self-contained and split multi-rule tags

davidharrishmc:
Stacked on #3373 — the first commit here is that PR; review the last two commits.

**Self-contained clauses.** Several tags covered two-to-four-word fragments
(`mnepc_acc` = "read-write register", `Sv39_pte_sz` = "eight bytes each",
`Svnapot_pte_N` = "N=1"). The sentence had been chopped up only to obtain separate
`sz`/`acc`/`op` rules. Each is now one anchor over the whole sentence with the rule
names preserved through the many-to-one YAML form the machine chapter already uses
(`mvendorid_sz_acc_op`). `mml_truth_table` moves off the sentence announcing the
table and onto the table itself.

**Splits.** Tags that bundled unrelated requirements are split at sentence
boundaries so each rule can be cited and tested on its own — `mstatus.TVM` into
`_warl`/`_op`/`_acc` to match TW and TSR, `hstatus.VGEIN` into four, the
`sstatus.SDT` double-trap paragraph into six, `satp` MODE=Bare into three, senvcfg
`CBZE`/`CBCFE`/`CBIE` each into three or four.

Rules 3277 → 3324. No visible text changes — anchors and YAML only.


davidharrishmc (commented):
I have reviewed these edits and find them reasonable.

src/priv/svpbmt.adoc
```diff
@@ -81,13 +81,13 @@ optimizations will likely improve performance when applied with adequate
...
+When Svpbmt is used with non-zero PBMT encodings, it is possible for
 multiple virtual aliases of the same physical page to exist
 simultaneously with different memory attributes. It is also possible for
 a U-mode or S-mode mapping through a PTE with Svpbmt enabled to observe
 different memory attributes for a given region of physical memory than a
 concurrent access to the same page performed by M-mode or when
-MODE=Bare. In such cases,
+MODE=Bare. [#norm:Svpbmt_aliasing_attribute]#In such cases,
 the behaviors dictated by the attributes
 (including coherence, which is otherwise unaffected) may be violated.#
```
jordancarlin, line 92: This change got carried over from a previous PR. Revert. The rule becomes impossible to understand without knowing what case it is referring to.

src/priv/supervisor.adoc
```diff
@@ -1057,15 +1051,14 @@ Implementations are not required to support all MODE settings, and if
 `satp` is written with an unsupported MODE, the entire write has no
 effect; no fields in `satp` are modified.
 
-[[norm:asidlen]]
-The number of ASID bits is UNSPECIFIED and may be zero. The number of implemented
+[#norm:asidlen]#The number of ASID bits is UNSPECIFIED and may be zero.# [#norm:asidlen_discovery]#The number of implemented
 ASID bits, termed _ASIDLEN_, may be determined by writing one to every
 bit position in the ASID field, then reading back the value in `satp` to
-see which bit positions in the ASID field hold a one. The
+see which bit positions in the ASID field hold a one.# [#norm:asid_lsb_first]#The
```
jordancarlin: `asidlen_discovery` seems like a software hint, not a normative rule.

src/priv/svpbmt.adoc
davidharrishmc: Reverted in 15de878d: the tag covers the whole paragraph again, as on main. The same fix is on #3377, #3379 and #3384.

src/priv/supervisor.adoc
davidharrishmc: Agreed. Untagged, and removed the rule definition. Also applied to #3377, #3379 and #3384.

jordancarlin (approved):

PR #3376, opened 2026-09-14
unpriv: split tags that carry several independent rules

davidharrishmc:
Stacked on #3374 — the first commit here is that PR; review the last commit.

Tags that bundled unrelated requirements are split at sentence boundaries so each
rule can be cited and tested on its own: the cache-block size / discovery / uniformity
rules in `cmo`, the `frm` and `fflags` field placements in `f-st-ext` (previously
buried in one paragraph about assembler pseudoinstructions), the V-on-F/D dependency
versus the supported reductions, `Zcmt_fetch`'s three requirements, and `jvt_op`'s
four — including promoting the implementation-defined set of legal `jvt` values to a
proper `impl-def-behavior` rule.

Rules 3296 → 3306. No visible text changes — anchors and YAML only.

Two findings are **not done in this stack at all**, and are recorded as open rather
than deferred: `h_trans_cache` (`cmo.adoc:478`, the anchor still covers only the lead-in)
and the CBO pseudocode blocks (`cmo.adoc:545/599/627`). Both need new anchors on a
figure or a listing rather than a split. One is blocked outright: `csr_side_effects` (`zicsr.adoc:121`) should move onto the table at
92–119, but that table already carries the `[[csrsideeffects]]` xref id and a second
block anchor would replace it.


davidharrishmc (commented):
Looks reasonable to me.

src/unpriv/f-st-ext.adoc
```diff
@@ -45,16 +45,15 @@ into integer register _rd_. insn:fscsr[] swaps the value in csr:fcsr[] by copyin
...
+for these accesses.# [#norm:frm_sz_acc_op]#The insn:frrm[] instruction reads the Rounding Mode field
 csr:frm[] (csr:fcsr[] bits 7{endash}5) and copies it into the least-significant three bits of
-integer register _rd_, with zero in all other bits. insn:fsrm[] swaps the value in
+integer register _rd_, with zero in all other bits.# [#norm:fsrm_op]#insn:fsrm[] swaps the value in
 csr:frm[] by copying the original value into integer register _rd_, and then
 writing a new value obtained from the three least-significant bits of integer
-register _rs1_ into csr:frm[]. insn:frflags[] and insn:fsflags[] are defined analogously for the
-Accrued Exception Flags field csr:fflags[] (csr:fcsr[] bits 4{endash}0).
+register _rs1_ into csr:frm[].# [#norm:fflags_sz_acc_op]#insn:frflags[] and insn:fsflags[] are defined analogously for the
+Accrued Exception Flags field csr:fflags[] (csr:fcsr[] bits 4{endash}0).#
```
jordancarlin: There is significant overlap here with #3310. I recommend reverting the changes to this paragraph and let's get that PR merged instead.

jordancarlin (approved):

PR #3377, opened 2026-09-14
priv: tag normative text that carried no rule

davidharrishmc:
Stacked on #3375 — the first three commits are #3373 and #3375; review the last two.

Adds normative rules for privileged-spec requirements that were stated in the text but
had no tag. The largest gap was `priv/csrs.adoc`, where the whole CSR-access model was
untagged: WARL's definition, the determinism requirement on WLRL/WARL reads after an
illegal write, CSR field modulation and its alias and no-side-effect rules, implicit
reads, width modulation, access to CSRs wider than XLEN, and the high-half alias and
RV64 reservation. Also the PTE U/G/RSW/A/D bit definitions in `supervisor.adoc`, the
`misa` collective-WARL constraint and `mcause`-is-mandatory in `machine.adoc`, and the
state-enable bit-layout figures, which are the only statement of those bit positions.

Eight tables and figures that carry rules had plain block ids; they are renamed to
`norm:` anchors following the convention `machine.adoc` already uses for
`norm:exc_priority` and `norm:mcause_exccode_enc_img`, with their cross-references
updated. A cross-reference to a table renders its caption, so the output is unchanged.

Rules 3324 → 3380. No visible text changes — anchors and YAML only.

One of the new rules is an **implementation-defined behaviour**: `CUSTOM_EXC_PRIORITY`
("the priority of any custom synchronous exceptions is implementation-defined",
`machine.adoc`), taking this volume from 209 to 210. Promoting something to
implementation-defined is a claim about the architecture, not only about tagging, so it
is called out rather than left in the count.

This does **not** exhaust the untagged findings for `src/priv`; the remainder is listed
in the tracking notes, and includes items that are chapter-scale (one rule per CSR
address; the ten-step Sv32 translation algorithm; the hardware A/D update scheme) or
blocked by the anchor syntax.

One such limitation found here and worth documenting in `tagging_normative_rules.adoc`:
an inline `[#norm:x]#…#` region silently produces **no id at all** — no warning — when
it wraps a line immediately after a non-empty bracketed construct such as `csr::[xs]`.
Two anchors hit it; one became a block anchor, the other is left untagged.


davidharrishmc (commented):
Looks reasonable to me.

src/priv/hypervisor.adoc
```diff
@@ -1764,8 +1764,8 @@ HLVX, and HSV. The explicit loads and stores of these instructions
 always act as though V=1 and the nominal privilege mode were
 `hstatus`.SPVP, overriding MPRV.
 
-The `mstatus` register is a superset of the HS-level `sstatus` register
-but is not a superset of `vsstatus`.
+[#norm:mstatus_superset_h]#The `mstatus` register is a superset of the HS-level `sstatus` register
+but is not a superset of `vsstatus`.#
```
jordancarlin: This seems explanatory, not normative.

src/priv/supervisor.adoc
```diff
@@ -523,9 +523,9 @@ an exception when executing in U-mode. Hence, they are effectively
 [NOTE]
 ====
 The setting of a bit in `mcounteren` does not affect whether the
-corresponding bit in `scounteren` is writable. However, U-mode may only
+corresponding bit in `scounteren` is writable. [#norm:scounteren_mcounteren_and]#However, U-mode may only
 access a counter if the corresponding bits in `scounteren` and
-`mcounteren` are both set.
+`mcounteren` are both set.#
```
jordancarlin: We should avoid tagging notes, which are non-normative by definition. If this is the only place this is said, then it probably needs to be moved outside of a note (which should happen in a separate PR).

jordancarlin (changes requested):

src/priv/hypervisor.adoc
davidharrishmc: Agreed. Removed the `mstatus_superset_h` tag and rule in 84e4bdc5.

src/priv/supervisor.adoc
davidharrishmc: Removed the tag from the NOTE here (84e4bdc5). This is the only place the rule is stated, so #3413 moves it out of the NOTE and tags it there.

jordancarlin (approved):

PR #3378, opened 2026-09-14
unpriv: tag normative text that carried no rule

davidharrishmc:
Stacked on #3376 — the first two commits are #3374 and #3376; review the last one.

Nine chapters of ratified extensions had **zero** normative-rule tags. `Ziccid` gets a
rule file covering its Ziccif dependency, the scope of its operational model, the fetch
and execution ordering rules, the fetch-value rule and the FENCE.I operation. The
composition and dependency statements of A, C, Zmmul, Zbkc, Zce, Zba and Zbkb are
tagged, along with the definition of FLEN and the Zfinx/Zdinx incompatibility with F
and D. Composition is already tagged for Zvbb, Zvfh and Zve*, so this makes the
treatment consistent rather than adding a new category.

Also: the six tables that are the entire definition of Zvkn/Zvknc/Zvkng/Zvks/Zvksc/Zvksg;
Zvkt's data-independent-latency requirement and its DIEL definition and exemptions; the
AMO operation in `zaamo`; and the exception/interrupt/trap definitions in `intro`.

Rules 3306 → 3344. No visible text changes — anchors and YAML only.

This does not exhaust the untagged findings for `src/unpriv`; the remainder is tracked
separately and includes chapter-scale items (the RVBNA algorithm, the Zcmp rlist and
stack_adj tables, the vector encoding tables, the Zkt instruction lists) and the
per-instruction `_enc` gap, which is a policy question rather than an editing one.

Worth documenting in `tagging_normative_rules.adoc`: an inline `[#norm:x]#…#` region
silently produces **no id at all** — no warning — when it wraps a line immediately
after certain bracketed constructs such as `ext:m[]` or `csr::[xs]`. `Zmmul_insns` and
`Zvkt_req` are block anchors for that reason.


davidharrishmc (commented):
Looks reasonable to me.

src/unpriv/f-st-ext.adoc
```diff
@@ -45,16 +45,15 @@ into integer register _rd_. insn:fscsr[] swaps the value in csr:fcsr[] by copyin
...
+for these accesses.# [#norm:frm_sz_acc_op]#The insn:frrm[] instruction reads the Rounding Mode field
 csr:frm[] (csr:fcsr[] bits 7{endash}5) and copies it into the least-significant three bits of
-integer register _rd_, with zero in all other bits. insn:fsrm[] swaps the value in
+integer register _rd_, with zero in all other bits.# [#norm:fsrm_op]#insn:fsrm[] swaps the value in
 csr:frm[] by copying the original value into integer register _rd_, and then
 writing a new value obtained from the three least-significant bits of integer
-register _rs1_ into csr:frm[]. insn:frflags[] and insn:fsflags[] are defined analogously for the
-Accrued Exception Flags field csr:fflags[] (csr:fcsr[] bits 4{endash}0).
+register _rs1_ into csr:frm[].# [#norm:fflags_sz_acc_op]#insn:frflags[] and insn:fsflags[] are defined analogously for the
+Accrued Exception Flags field csr:fflags[] (csr:fcsr[] bits 4{endash}0).#
```
jordancarlin: Looks like this got pulled into this PR again. Revert the changes to this paragraph (and the corresponding change to the YAML).

davidharrishmc: Reverted the paragraph and its YAML entries to main.

jordancarlin (approved):

PR #3379, opened 2026-09-14
priv: unify duplicated Sv level rules

davidharrishmc:
Stacked on #3377 — review the last commit.

`Sv39_levels` and `Sv39_LEVELS` are two tags that both say the page table has three
levels, differing only in case. They become one rule with both tags (same for Sv48 and
Sv57). Note this unifies the *rules*, not the tags — both anchors still exist, so #3384
still has to give one of each pair a distinct name.

**Deliberately left alone**, and this is the more interesting half:

- Cross-chapter duplication — MRET/SRET in both `machine` and `hypervisor`, the csrind
  access-control rules in both `smcsrind` and `smstateen`, `spelp`/`mpelp` in three
  chapters. Picking a primary chapter moves rules between rule files; that is a
  modelling decision for the SIG.
- `scountovf_smode_read_access_control` and `Ssqosid_srmcfg_introduced`, both summaries
  of rules stated beside them. Marking a summary as `context` leaves a rule whose *only*
  normative tag belongs to another rule — a shape that appears **zero times** in `main`.
- The mseccfg SSEED/USEED truth table, which restates in tabular form three rules that
  each already carry a self-contained sentence. A `context` tag in this manual supplies
  meaning a rule lacks ("these conditions", "in either case", "such a region"); a
  restatement supplies none.

The real fix for all of these is to delete the redundant rule, which is a decision about
names that `riscv-arch-test` references.

Rules unchanged at 3,380. No visible text changes.


davidharrishmc (commented):
LGTM

UmerShahidengr (approved):
LGTM

PR #3380, opened 2026-09-14
unpriv: move tags onto the rules they name

davidharrishmc:
Stacked on #3378 — review the last commit.

`Zawrs_wrs-nto_stall_exec` and `Zawrs_wrs-sto_stall_duration` were attached to the
motivating narrative rather than to the section that defines the stall. They move onto
the definition, and "these instructions are available in all privilege modes" becomes a
rule of its own.

Rules 3344 → 3345. No visible text changes.

**Left alone**, with the reasoning recorded rather than acted on:

- Three rv32 tags restate the instruction-address-misaligned rules given earlier in the
  same chapter, and the vector chapter tags the `mstatus.SD` sentence a second time with
  byte-identical text. Marking the restating tag as `context` leaves a rule whose *only*
  normative tag belongs to another rule — a shape that appears **zero times** in `main`
  — so the restatement stops being a rule without the duplication going away. Deleting
  the duplicate rules is the real fix, and that is a decision about names other projects
  reference.
- The aq/rl semantics stated three times (informally in `za` and `zaamo`, formally in
  RVWMO's untagged annotation mapping): making RVWMO primary means tagging that mapping
  first, which belongs with the untagged work.
- The chapter-intro summaries in `zbb`/`zbc`/`zbs`/`zbkx`, which have no single
  per-instruction rule to become context on.
- `misaligned_atomicity_granule_size`, which duplicates machine's PMA rules but is the
  sole tag of an implementation-defined behaviour.


davidharrishmc (commented):
LGTM

jordancarlin (approved):

PR #3381, opened 2026-09-14
unpriv: correct two instruction descriptions that name the wrong register

davidharrishmc:
Two instruction descriptions name the wrong register. Both sentences carry
normative-rule tags, so the extracted rules were wrong too.

- `zalasr.adoc` — "This instruction stores 2^width^ bytes of memory **from rs1**
  atomically" omits the source register entirely. A store-release writes the value in
  `rs2` to the address in `rs1`; the load sentence four paragraphs earlier already says
  "loads … from rs1 … and writes the result into rd".
- `zbkx.adoc` — "Elements in **rs1** are replaced by the indexed element in **rs2**" is
  backwards. The sentence directly above it says `rs2` holds indices into `rs1`, and the
  `xperm8`/`xperm4` pages in `zb.adoc` say "The result is each element in rs2 replaced by
  the indexed element in rs1". Two tagged rules contradicted each other.

Text only; no tagging changes. Independent of the normative-rule cleanup PRs.

Found during the tagging audit in riscv/riscv-arch-test#2237.


davidharrishmc:
These both correct genuine errors in the ISA manual.  They affect the meaning of the manual, and need proper review separate from the normative rule effort.

src/unpriv/zbkx.adoc
```diff
@@ -5,7 +5,7 @@
 inside the general purpose registers.#
 _rs1_ is used as a vector of N-bit words, and _rs2_ as a vector of N-bit
 indices into _rs1_.
-[#norm:zbkx_indexing_behavior]#Elements in _rs1_ are replaced by the indexed element in _rs2_, or zero
+[#norm:zbkx_indexing_behavior]#Elements in _rs2_ are replaced by the indexed element in _rs1_, or zero
 if the index into _rs2_ is out of bounds.#
```
aswaterman: Neither description looks correct as nothing is being replaced.

I think the right description is "Each element of the destination is formed by
using the corresponding index element from _rs2_ to select an element from _rs1_, or zero if the index element is out of bounds."

davidharrishmc: Much better wording.  Thank you.

I applied the same description everywhere the instructions are described.  It seems so many descriptions are redundant, but I'd leave streamlining to a bigger refactor.

aswaterman: Yeah, the refactoring is a matter for another day, but thanks for fixing the problematic wording elsewhere.

src/unpriv/zbkx.adoc
```diff
@@ -5,8 +5,9 @@
 inside the general purpose registers.#
 _rs1_ is used as a vector of N-bit words, and _rs2_ as a vector of N-bit
 indices into _rs1_.
-[#norm:zbkx_indexing_behavior]#Elements in _rs1_ are replaced by the indexed element in _rs2_, or zero
-if the index into _rs2_ is out of bounds.#
+[#norm:zbkx_indexing_behavior]#Each element of the destination is formed by using the
```
aswaterman: I think "result" is better than "destination".  I'm OK to merge once that's fixed.

aswaterman (approved):

davidharrishmc:
Updated

PR #3382, opened 2026-09-14
profiles: pilot normative-rule tagging for RVA23

davidharrishmc:
`src/profiles` has **zero** normative-rule tags across six chapters and ~1,580 lines,
although profile chapters state mandates that bind an implementation as firmly as
anything in the other two volumes. Whether those mandates belong in the normative rule
set is a scope decision nobody has made.

Rather than assume an answer, this tags one chapter as a strawman: RVA23U64 and
RVA23S64 each get a rule for the mandatory base (base ISA, endianness, ECALL
behaviour) and a rule gathering the mandatory-extension lists. Those lists are broken
into 13 and 18 separate AsciiDoc lists by the explanatory paragraphs between them,
hence the many-to-one tag lists — that shape is itself part of what the decision has
to accept.

Left untagged on purpose, so the boundary is visible: the optional and transitory
extension lists (what an implementation *may* choose) and the Recommendations sections
(advisory).

Tag names follow the convention in #3383: lower case, since a profile name is not an
extension name, and each list tag is named after the extension it leads with rather than
by a bare number.

Four rules from a 318-line chapter. No visible text changes.

If this shape is acceptable, the remaining five profile chapters follow the same
pattern. If profile mandates are out of scope for the rule set, closing this PR settles
that too.


davidharrishmc:
This doesn't look useful.  The extensions are all defined in the ISA manual now.  The tagging bunches rules together.  I propose not tagging the profiles.

PR #3383, opened 2026-09-14
Make the tag naming convention explicit

davidharrishmc:
The guide's naming section leaves several questions open, and the tree answers each of
them both ways: 377 tags start with a capital and 2,550 do not, with seven prefixes
appearing in both cases; 368 tags use a hyphen where the guide reserves it for
mnemonics; ~140 end in a bare digit, several `_op2` with no `_op1`; and dozens tag
two-to-four-word fragments. The guide's own table shows `Sv48_gigapage` while its prose
said tag names are typically lower case — that sentence is removed here, since it
directly contradicted the rule three sections above it.

This states the rules:

- A hyphen appears only where the standard's own name contains a dot or hyphen.
- A name beginning with an extension name uses that extension's capitalization **as the
  manual renders it** — `src/lib/macros.rb` renders `ext:zvknha[]` as `Zvknha`, so the
  rule is mechanical rather than a matter of taste — and it applies **wherever** an
  extension name appears in a tag, not only at the front. Everything else is lower case;
  upper case stays reserved for implementation-defined behavior rule names, whose tags
  follow the ordinary convention (rule `ENDIANNESS`, tag `norm:endianness`).
- Rules are not distinguished by a bare number.
- A tag covers a complete clause; per-attribute rules are derived in YAML with the
  many-to-one form the machine chapter already uses for `mvendorid_sz_acc_op`.

It also states the three things a mechanical sweep cannot infer and will otherwise
propose again every time: the **ambiguous single-letter extension names** (A, B, D, F, H
and Q keep their capital; M, S, U and V need a per-tag decision, since
`M_implies_Zmmul` is the M extension while `m_mode_mandatory` is machine mode); that
capitalization applies to non-leading tokens; and the **hyphens that are not word
separators** — `R-type`, mnemonics whose dots became hyphens (`c-mop`, `cbo-inval`,
`ntl-all`, `vsha2chl-vv`), the generic vector operand forms, and `vmv-nr-r`. Plus the
two false friends: `ss_push` is shadow stack, not the Ss prefix, and `sh_enc` is the SH
instruction, not the Sh extension.

Finally it documents a gotcha found while applying this: an inline `[#norm:x]#…#` region
whose text wraps immediately after a bracketed construct such as `csr::[xs]` or
`ext:m[]` silently produces **no anchor and no warning**; the first symptom is the rule
tool reporting a reference to a non-existent tag — which only happens at all because a
rule already points at it. A tag added without one would vanish silently.

**This changes no tags.** It is the convention the per-volume renaming PRs (#3384,
#3385) implement, separated so the convention can be settled before ~380 tags are
renamed to match.


wmat (approved):
Reviewed the convention doc — clear, self-consistent, and changes no existing tags as stated. LGTM.

PR #3384, opened 2026-09-14
priv: rename tags to the agreed naming convention

davidharrishmc:
Stacked on #3379, and implements the convention proposed in #3383 — review the last commit.

205 renames, touching 64 files (26 under `src/priv`, the rest `normative_rule_defs/`).

- **Extension names** take their own capitalization as the manual renders it —
  `src/lib/macros.rb` renders `ext:shcounterenw[]` as `Shcounterenw`, so the rule is
  mechanical — and they take it **wherever they appear in a tag**, not only at the
  front. The identifier list comes from the manual's own `ext:`/`extlink:` macros.
- **Everything else is lower case**: `ENDIANNESS`, `FP_`, `NTL_`, `PMA_`, `PMP_`, `PPO_`,
  and the `M_`/`M-mode_` tags, which name the privilege mode rather than the M extension.
- **Hyphens** survive only inside mnemonics and the standard's own hyphenated names. All
  five `vsctrctl-<field>_op` tags become `vsctrctl_<field>_op` (CSR and field, not a
  mnemonic); `cbo-inval_*`, `c-mop_enc`, `vop-vv_*`, `vmv-nr-r_*` and `R-type_operands`
  keep theirs.
- **Counter suffixes** are replaced by the distinguishing condition:
  `mip_stip_stimecmp_op2` → `mip_stip_reflects_stimecmp`; `mtval_vaddr_wr1`/`_wr2` →
  `_wr_faulting_addr`/`_wr_misaligned_addr`; `ctr_behavior_criteria0`–`3` → named
  criteria. Trailing numbers that carry meaning are untouched (`rv32`, `xlen64`,
  `sew32`, `sha256`, `ttype10`, `bit0`).

Six rules have one tag in each volume — `ecall_op1`/`op2`, `ebreak_op1`/`op2`,
`mstatus_sd_op1`/`op2`, and the Zama16b/Ziccamoa/Ziccamoc/Ziccrse PMA rules. **This PR
renames both halves of each**, so no pair is left half-converted and this PR and #3385
never rewrite the same YAML line; `git merge-tree` on the two tips is clean.

Rule names follow their tags, except implementation-defined behavior rule names, which
stay upper case by schema rule.

**Invariant:** 3,380 rules and 210 implementation-defined behaviours before and after —
a rename adds and removes nothing. No visible text changes.

The map was generated mechanically and reviewed by hand twice. The review caught
`ss_push`/`ss_pop` (shadow stack, not the Ss prefix), `sh_enc` (the SH instruction),
mnemonic and vector-operand-form hyphens, three `*_LEVELS` tags colliding with their own
lower-case twins, and — on the second pass — `vsctrctl-bpfrz_op`, the `ntl.all` family,
and a mis-cased `amocas-BH`. The exceptions are now stated in #3383 so the next
mechanical sweep does not re-propose them.

**Downstream:** `riscv-arch-test`'s `coverpoints/norm/*.yaml` reference rule names by
string. An old→new map is kept alongside this work; it records both renames and the rule
names these stacks delete, with the affected downstream file for each.


wmat:
Thanks David, I'll await the reviews.  I hadn't even noticed they were in Draft.

UmerShahidengr (approved):

Issue #3140, opened 2026-05-25
[Clarification] - Smrnmi: can a custom CSR disable RNMI without violating spec?


leo308078:
## Summary
The spec says RNMI "cannot be disabled by software. Specifically, they are not disabled by clearing the mstatus.MIE register."

We want to add a custom CSR that can prevent RNMI from being taken — but only when NMIE=0 (which only hardware can cause). After boot, behavior is 100% standard Smrnmi.

Is this allowed? Or does "cannot be disabled by software" mean absolutely no mechanism can ever gate RNMI? (cc @riscv/arc)

---

## Spec Reference

**Permalink:** https://docs.riscv.org/reference/isa/priv/rnmi.html

_Text Reference: Smrnmi, "RNMI Interrupt Signals" section_

---

## Expected vs Observed
| Expected | Observed |
|---|---|
| Whether custom CSR gating RNMI is compatible with Smrnmi | Unclear — "Specifically" might narrow scope to just MIE, or might be a blanket ban |

---

## Additional Context

Why we need this: automotive CPU, NMI pin could get stuck-at-1 due to external HW fault → CPU trapped in RNMI handler forever with no escape (since SW can't clear NMIE).

Our custom CSR:
- Can only be cleared inside RNMI handler (NMIE=0 required)
- After boot, default = enabled = standard Smrnmi behavior
- Gives firmware a recovery path for this fault scenario

Basically: does "Specifically" mean "the only thing we're saying is MIE won't disable RNMI", or does it mean "for example, MIE won't disable RNMI, and neither will anything else"?

aswaterman:
The nature of custom extensions means we can't provide definitive guidance on them in this forum.    But it sounds to me like your processor would be conformant while your extension is not engaged, and while the extension is engaged, your processor would not be conformant to the Smrnmi extension.  If that's something you're OK with, then go for it.

leo308078:
Thanks so much for the clarification.
If you don't mind, I have two follow-up questions:

1. Just to make sure I understand correctly: the non-conformance is because the Smrnmi spec intends that RNMI cannot be disabled by any software mechanism (not just limited to MIE) — is that right?

[...]

PR #3386, opened 2026-09-14
docs: clarify Smrnmi conformance for RNMI gating

JulienKervarrec:
Relates to #3140.

This documentation-only change clarifies the conformance boundary around RNMI gating:

- Smrnmi architectural RNMIs cannot be disabled by software.
- A custom mechanism that masks RNMIs while active is not conformant to Smrnmi.
- A gate implemented outside the hart is an integration-level platform policy, not a software control defined by Smrnmi.

No ISA behavior or normative requirement is changed; the note makes the distinction explicit for implementers and reviewers. The change is limited to `src/priv/smrnmi.adoc`. No build or test was run for this documentation-only contribution.


aswaterman (commented):
While I agree with what this note says, I'm not wild about adding it.  It's a true statement in general that a custom means of doing X is not a standard-conforming way of doing X.  It would add an awful lot of clutter to comment as such throughout the spec.

JulienKervarrec:
Thanks for taking a look, Andrew. That's fair: the general principle already covers this case, and spelling it out locally would add clutter to the spec. Closing this one.

PR #3387, opened 2026-09-14
Make the implementation-defined marking agree with the rule names

davidharrishmc:
The guide reserves all-upper-case rule names for implementation-defined behaviors, and
`create_normative_rules.py` enforces one direction: an impl-def rule whose name is not
upper case is fatal. The other direction is unchecked — and **twelve upper-case rule
names carry no `impl-def-behavior` flag**. Two of them already declare `kind` and
`instance`, which only make sense for an implementation-defined behavior, so the flag
looks simply forgotten.

**Three are implementation-defined and get the flag:**

| Rule | Why |
|---|---|
| `FLEN` | 32, 64 or 128 depending on which of F/D/Q is implemented; already declared `kind`/`instance` |
| `CTR_DEPTH` | "depends upon both the implementation (the maximum supported depth) and the CTR configuration"; "it is implementation-specific which DEPTH value(s) are supported" |
| `MSTATUS_UXL_LEGAL_VALS_SMODE` | constrains the legal values of a WARL field; already declared `kind`/`instance` |

**The other nine are ordinary rules whose names merely look like the convention.**
Flagging them would assert that an implementation may choose the behaviour, which is not
what the manual says — `MSTATUS_MDT_CLR_MNRET`, `MSTATUS_MXR_SUM_OP_ACC_FAULT`,
`MSTATUS_TSR_ACC` and `MEDELEGH_OMIT_XLEN64` are all determined, and `LMUL`, `VLMAX`,
`EEW`, `EMUL` and `NFIELDS` are derived values and definitions. They are lower-cased
instead, which is what the convention asks for and which matches the tags they already
reference (rule `LMUL` → tag `norm:lmul`).

The invariant now holds both ways: every implementation-defined rule name is upper case,
and every upper-case rule name is an implementation-defined behavior.

Rules unchanged at 3,307; implementation-defined behaviours 209 → 212. No tags change and
no document text changes — `normative_rule_defs/` only.

Independent of the normative-rule cleanup PRs; found while reviewing them.


UmerShahidengr (approved):
LGTM

PR #3388, opened 2026-09-14
Fix ISA manual typos

davidharrishmc:
Fix ISA manual typos
Also modified several British spellings to be consistent with usage of same words elsewhere in the manual.


wmat (approved):
Thanks, David.

PR #3392, opened 2026-09-14
ci: cache rendered diagrams and build the PDF in its own job

rpsene:
Speeds up `isa-build.yml` from about **11.5 min to about 7 min** by building the PDF in its own job and caching rendered diagrams. No change to the content or format of any output, nor to artifact names, release files or the Pages layout.

## Baseline

`isa-build.yml` takes about **11 minutes** (last 8 successful runs: 8–12 min). Timeline of run 34630925556:

| Phase | Time |
|---|---|
| Pull `riscv-docs-base-container-image:latest` | 1m42s |
| HTML + EPUB + norm tags (in parallel with PDF) | 3m21s |
| norm-rules JSON/HTML | 4s |
| **PDF** (finishes last) | **9m43s** after the pull |

Two findings:
- **Diagrams are rendered three times per run and then discarded.** The PDF, HTML and EPUB targets each render all 465 diagrams (368 WaveDrom, 97 bytefield), because asciidoctor-diagram keeps its cache inside the per-target workdir, which the Makefile deletes.
- **The PDF is single-threaded** and competes with three other targets for 4 cores.

Measured locally, PDF only: diagrams are about 12% of the PDF time (222s cold, 195s warm). Removing `-a compress` or `volume-xrefs.rb` made no measurable difference, and both stay.

## Changes

**Makefile**
- Adds `DIAGRAM_CACHE_DIR ?= .diagram-cache`, mounted into the container at `/diagram-cache`.
- Each target gets its own subdirectory, so parallel targets never write the same cache files.
- Passes `-a diagram-cachedir=… -a diagram-cache-images-option`, so the rendered images are kept, not only their metadata.
- Adds `clean-diagram-cache`, and ignores `.diagram-cache`.

**`isa-build.yml`**
- **`build-pdf`** and **`build-html`** (HTML, EPUB, norm tags, norm rules) run **in parallel**.
- Each job restores and saves `.diagram-cache` with `actions/cache`, keyed on `hashFiles('src/**')` with a prefix fallback.
- The container is pulled once per job before `make` starts.
- A new **`publish`** job downloads both jobs' outputs and creates the release and Pages content. `deploy` now depends on it.
- Artifact names, release files and Pages layout are **unchanged**.

## Local check (14-core arm64, full `make -j`)

| | Time | Outputs |
|---|---|---|
| cold cache | 236s | all 6 |
| warm cache | 195s | all 6 |

## CI measurements

| Run | PDF job (pull + build) | HTML, EPUB and rules job (pull + build) | Wall time |
|---|---|---|---|
| **Baseline**, `main` single job (run 34630925556) | PDF finishes last | n/a | **~11.5 min** (build step 686s) |
| Cold, run 34898001589 | 459s = 70s + **382s** | 274s = 86s + **171s** | **7.7 min** |
| Cold, run 34897698084 | 767s = 385s + 371s | 261s = 82s + 165s | 12.8 min (slow image pull) |
| **Warm**, run 34897698084 attempt 2 (cache hit on both jobs) | 426s = 110s + **302s** | 127s = 74s + **41s** | **7.1 min** |

- Splitting the jobs is the largest gain: about 7.7 min even with a cold cache.
- The diagram cache cuts the HTML, EPUB and rules build from about 170s to 41s, and the PDF build from about 375s to 302s.
- The image pull from ghcr.io (70s to 385s) is now the least predictable part.

## Cache key

The key is `hashFiles('src/**')` with a prefix `restore-keys` fallback. The key only decides when a new snapshot is saved: a run whose key misses still restores the latest cache (including `main`'s), and asciidoctor-diagram re-renders only diagrams whose checksum changed. About 56% of the 465 diagrams are inline in `.adoc` files, so keying on `.edn` files alone would let the cache go stale (discussed with @jordancarlin below). Each snapshot is about 20 MB.

## Permissions

The workflow now sets `permissions: contents: read` (CodeQL finding). No job writes through `GITHUB_TOKEN`: the release step uses `GHTOKEN`, and `deploy` keeps its own Pages permissions.


jordancarlin:
> * Each job restores and saves `.diagram-cache` with `actions/cache`, keyed on `hashFiles('src/**')` with a prefix fallback.

Using a hash of the full `src` directory as the key seems like it defeats the benefit of caching. Just about any PR is going to change the `src` directory. What about using a hash of just the `*.edn` (and other image related extensions) as the key?

rpsene:
> Using a hash of the full `src` directory as the key seems like it defeats the benefit of caching.

Good question @jordancarlin. The key only controls when a new cache is *saved*. On a miss, `restore-keys: isa-diagrams-pdf-` restores the most recent cache with that prefix, including the one saved by the latest `main` build, and asciidoctor-diagram re-renders only diagrams whose checksum changed. So a PR that touches `src` still starts from a warm cache.

Keying on `*.edn` alone would save far fewer snapshots, but about 56% of the 465 diagrams are inline in `.adoc` files (205 are in `.edn`). Changes to those would then never make it into a new cache snapshot until an `.edn` file changed, so the cache would get colder over time. Saving a new snapshot whenever `src` changes keeps `main`'s cache current, at about 20 MB per snapshot, with GitHub evicting old ones. I'll post the cold and warm CI timings from this PR here once the runs finish.

jordancarlin:
Ah. I was just going off of the description and hadn't looked at the implementation yet. Sounds fine if you are using a separate restore key.

rpsene:
CI timings from this PR, as promised.

| Run | PDF job (pull + build) | HTML, EPUB and rules job (pull + build) | Wall time |
|---|---|---|---|
| **Baseline**, `main` single job (run 34630925556) | PDF finishes last | n/a | **~11.5 min** (build step 686s) |
| Cold, run 34898001589 | 459s = 70s + **382s** | 274s = 86s + **171s** | **7.7 min** |
| Cold, run 34897698084 | 767s = 385s + 371s | 261s = 82s + 165s | 12.8 min (slow image pull) |
| **Warm**, run 34897698084 attempt 2 (cache hit on both jobs) | 426s = 110s + **302s** | 127s = 74s + **41s** | **7.1 min** |

- **Splitting the jobs** is the largest gain. Even with a cold cache, a run takes about 7.7 min instead of 11.5, because the single-threaded PDF build no longer shares the runner with the other targets.
- **The diagram cache** cuts the HTML, EPUB and rules build from about 170s to 41s, and the PDF build from about 375s to 302s. The rest of the PDF time is page layout.
- With a warm cache, the HTML artifacts are ready in about 2 minutes and the whole run takes about 7.1 min.
- The image pull from ghcr.io varies from 70s to 385s between runs and is now the least predictable part. The second cold run above was slow only because of the pull.

This is one sample per scenario; the build-step times are the more reliable comparison.

wmat (approved):
LGTM

wmat (approved):
Thanks, Rafael.

Issue #29, opened 2017-03-13
C nits


sorear:
* No indication in the tables of which immediates are sign extended (could use simm/nzsimm to make the table self-contained)
* LI is falsely listed as rs1/rd
* SLLI is falsely listed as rd only
* ADD should be rs1/rd

aswaterman:
LI and ADD got fixed last week.  I've just fixed SLLI.

Standardizing on simm/nzsimm for sign-extended immediates and imm/nzimm for zero-extended sounds ok; I don't have time to do that at the moment, but would review a PR.

DSHorner:
I hope to have PR  ready tomorrow.

aswaterman:
Thanks @DSHorner

DSHorner:
zimm is used in the description of ADDI4SPN. I wanted to use uimm consistently in the table for unsigned zero extended values, and revise ADDI4SPN accordingly. But zimm appears to be a directive to the assembler: "and expands to addi rd 0 , x2, zimm[9:2]". (as does nzimm by it's usage in "expands to" descriptions, but with sign extension instead). I've been trying to confirm zimm/nzimm directives in binutils, but it could also be a standard macro in any of the projects. Is it safe to switch to uimm as the variant for this document?

sorear:
@DSHorner I'm pretty sure that's just informal notation, not a directive, and we should be consistent with it.  There are two axes for an immediate: signed/unsigned, and may-be-zero/must-be-nonzero; everything outside C is a "szimm" but C varies.

michaeljclark:
Zero extension of RVC instruction immediate values caught me out while working on an RVC implementation based on the spec.

While zero extension is mentioned in the text description for each RVC encoding with a zero-extended immediate, the zero extension property does not appear in tabular form such as in the immediate mnemonic in each instruction's encoding table or in the instruction set listing. The tables just contain 'imm' and don't indicate whether to sign or zero extend.  I made frequent reference to both while implementing but the zero extension property can only be derived from the text description of the instruction, and not from the tables. 

As @DSHorner mentions, almost all of the Base ISA immediate values are sign extended (excepting shifts and the CSR instruction immediate values), so it is an easy mistake to make. Load and store offsets, including stack relative loads and stores, in the compressed ISA have zero-extended immediate values which is different from the default of sign extension in the Base ISA.

[...]

PR #3393, opened 2026-09-14
ci: add an experimental build with the Ubuntu 26.04 container image

rpsene:
> **Stacked on #3392** (split jobs and diagram cache), because both change `isa-build.yml`. GitHub retargets this PR to `main` when #3392 merges; until then the diff shows only the new job.

## What

Adds a **`build-ubuntu2604`** job that builds every output with the **experimental Ubuntu 26.04 LTS** build container, `ghcr.io/riscv/riscv-docs-base-container-image:experimental-ubuntu2604` (Ruby 3.3, Python 3.14, Node.js 22, OpenJDK 25, TeX Live 2025).

## Why

The goal is to catch **source changes in this repository** that build on the current Ubuntu 22.04 image (`latest`) but break with the newer tools, before `latest` moves to 26.04.

The container project already covers the other direction: every candidate image must build this manual's `main` before it is published (riscv/riscv-docs-base-container-image#27, #29).

## How it is kept harmless

| | |
|---|---|
| **Allowed to fail** | `continue-on-error: true`; not a required check, and no job depends on it |
| **When** | push to `main`, manual runs, and weekly (Tuesday 03:00 UTC, after the images are rebuilt on Monday). **Not on pull requests**, where a failure would still show as a red check to contributors |
| **Outputs** | uploads `riscv-spec-<sha>-ubuntu2604.pdf` for comparison; never released or deployed to Pages |
| **Cache** | its own diagram cache key, `isa-diagrams-ubuntu2604-*` |

## Validation

- The image is published for amd64 and arm64 (`experimental-ubuntu2604` = `ubuntu2604-cc5f427`), and it passes the container project's smoke test and full ISA manual build on both architectures.
- A local build of `be8cb00` with the 26.04 image produced **output identical to the 22.04 image**: all 490 embedded SVGs in `riscv-spec.html`, the PDF text (906 pages), `norm-rules.json` and `riscv-spec-norm-tags.json`.
- This job only runs on `main`, manual runs and the schedule, so this PR's own CI does not exercise it. Once merged, a manual run of the workflow runs it.


wmat (approved):
This LGTM, along with 3392.

jordancarlin (approved):

PR #3398, opened 2026-09-16
ci: queue ISA build runs so GitHub Pages deploys can't overlap

wmat:
## Problem

[Run 34988531518](https://github.com/riscv/riscv-isa-manual/actions/runs/34988531518) failed in the `deploy` job. Two pushes to `main` 31 seconds apart started two builds, and both reached `actions/deploy-pages` at nearly the same time. GitHub Pages returned HTTP 400 because another deployment was still in progress. The site wasn't affected, since the other deploy was for the newer commit and it succeeded.

## Change

Add a workflow-level `concurrency` group to `isa-build.yml`:

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.ref }}
  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

A run for a later push to `main` waits until the earlier run finishes, and queued runs start in the order they were triggered. Deploys can no longer overlap, and the newest commit deploys last.

- **Waiting runs:** a group keeps only its newest waiting run. If several pushes land while a run is in progress, the older waiting runs are cancelled and only the newest one builds and deploys.
- **Event name in the group:** scheduled and manual runs on `main` get their own groups, so a push can't cancel a waiting manual release run.
- **Pull requests:** each PR has a different `ref`, so it gets its own group. A new commit on a PR cancels the run in progress instead of waiting for it.

Trade-off: back-to-back pushes to `main` wait for each other's full run, about 11 minutes including the experimental Ubuntu 26.04 build.

An earlier version of this PR used a job-level concurrency group plus a head-of-main check in `deploy`. It was replaced after review, as described in the PR comments.

## Testing

The YAML parses. Concurrency only matters when two pushes to `main` overlap, so it will first be exercised after merge.


rpsene (commented):
@wmat thanks for digging into this. I checked run 34988531518 and the diagnosis matches: `9b2bc70` and `f3dbf60` were pushed 31s apart, `f3dbf60` started deploying at 15:35:11, and `9b2bc70` hit the 400 at 15:35:53. The concurrency group plus the head check is the right approach. Two things I'd change before merging:

**1. A failed `git ls-remote` silently skips the deploy**

The step has no `shell:`, so it runs as `bash -e {0}` without `pipefail`. If `git ls-remote` fails (network blip, rate limit), `cut` still exits 0, `head_sha` is empty, the step writes `latest=false` and posts "main is at , this run built ...". The run goes green and nothing is deployed. I'd fail loudly instead:

```bash
head_sha=$(git ls-remote "https://github.com/${{ github.repository }}" refs/heads/main | cut -f1)
if [ -z "$head_sha" ]; then
  echo "::error::Could not read the head of main"
  exit 1
fi
```

**2. A skipped deploy still records a successful `github-pages` deployment for the older commit**

GitHub creates the environment deployment when the job starts, before any step runs. You can see it for the failed run: deployment `6462532700` for `9b2bc70` was created at 15:35:44. With this change, a stale run skips only the step, so the job succeeds and that deployment is marked `success` for the older SHA. The Deployments panel would then show the older commit as the latest `github-pages` deployment while the site is serving the newer one.

Moving the check into its own small job avoids that, since `deploy` never starts for a stale run:

```yaml
  check-head:
    if: github.event_name == 'push' && github.ref == 'refs/heads/main'
    needs: publish
    runs-on: ubuntu-latest
    outputs:
      latest: ${{ steps.check-head.outputs.latest }}
    steps:
      - name: Check that this commit is still the head of main
        id: check-head
        run: |
          # same script as now, with the empty check from point 1

  deploy:
    if: needs.check-head.outputs.latest == 'true'
    needs: [publish, check-head]
    concurrency:
      group: github-pages-deploy
      cancel-in-progress: false
```

Ordering stays safe: a newer push needs about 8 minutes to reach deploy, while the older run's deploy starts seconds after its check, and the concurrency group still replaces any older waiting run.

On testing: once merged, re-running the head check (and the deploy that depends on it) of an older `main` run keeps its original SHA, so it takes the skip path. That confirms the logic without waiting for two pushes to collide.

jordancarlin:
Seems like this could be dramatically simplified by just using concurrency at the workflow level instead of at the job level. Then the workflow for the second push won't even start until the first one is completed.

wmat:
Thanks, both of you. I pushed a revision that takes @jordancarlin's approach.

@jordancarlin: The job-level group and head-of-main check are gone. The workflow now has a top-level `concurrency` group with `cancel-in-progress: false`. Queued runs start in trigger order, so the newest push deploys last. I added `github.event_name` to the group name so scheduled and manual runs on `main` aren't in the push queue. Otherwise a push could replace, and so cancel, a waiting manual release run.

@rpsene: Thanks for checking the run and the careful review. Both points came from the head check, which this version removes. Nothing can skip the deploy quietly now, and no run records a `github-pages` deployment for a commit it didn't deploy.

.github/workflows/isa-build.yml
```diff
@@ -30,6 +30,16 @@ on:
...
 
+# GitHub Pages rejects a deployment while another is in progress, so runs for
+# back-to-back pushes to main must not overlap. Runs in a group wait for the one
+# in progress and start in the order they were triggered, so the newest commit
+# deploys last. Only the newest waiting run is kept; older waiting ones are
+# cancelled. The event name keeps scheduled and manual runs on main out of the
+# push queue, so a push can't cancel a waiting release run.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.ref }}
+  cancel-in-progress: false
```
jordancarlin: If the event is a PR, we probably want to cancel in progress runs. Otherwise pushing a new commit to a PR will wait for the previous commit to finish CI before running CI for the new commit. For pushes to main, cancel-in-progress: false seems fine.

wmat: Good point, done. `cancel-in-progress` is now `${{ github.event_name == 'pull_request' }}`, so a new commit on a PR cancels the run in progress. Pushes to `main` still queue, so Pages deploys stay in order.

jordancarlin (approved):

rpsene (approved):
@wmat thanks, this is much simpler, and both of my earlier points go away with the head check. A superseded run is cancelled while it waits, so it never starts `deploy` or records a `github-pages` deployment. Separate groups per event and per PR, plus `cancel-in-progress` only on pull requests, look right to me.

One small wording nit, not blocking: the comment says queued runs "start in the order they were triggered". GitHub's docs say runs in a group are handled first-in-first-out by the time each one started waiting on the group, and that "ordering is not guaranteed". In practice that only matters if two pushes reach the group at almost the same instant, so no code change is needed, but softening the comment would keep it accurate.

aswaterman (approved):

wmat:
@rpsene Thanks for the review. Good catch on the ordering; the comment now says runs "normally start in the order they reached the group" and notes that GitHub doesn't guarantee the order when runs arrive almost at once.

PR #3404, opened 2026-09-21
Change Zifencei to order prior instruction fetches as well as explicit memory accesses

mehnadnerd:
As part of the formalisation work that I, Nils, and Ben Simner have been doing, we discovered that the text in the manual for fence.i was too weak. It only ordered prior explicit accesses, not prior instruction fetches, so the litmus test [miniJit03+fencei](https://github.com/mehnadnerd/herdtools7/blob/dev/brs/rvwork2/catalogue/riscv-ifetch/tests/handwritten/incoherent/IDC.miniJit03%2Bfencei.litmus) would not work.

We therefore want to change the text here to state the prior instruction fetches are ordered before the fence.i as well as prior explicit accesses. We believe this matches the original intent and will not disallow any extant or even unreasonable hardware implementations.

Also of note is that this is equivalent to "all memory accesses that precede", because the other implicit accesses (page table walks and setting A/D bits) are ordered before their explicit accesses. Therefore we may want to write "all memory accesses that precede" instead so future types of implicit accesses (such as tracing) work with this as well.


aswaterman (approved):
For everyone else's context, this omission was found in trying to formally model instruction fetch.  It struck us as obvious and therefore as an unintentional omission rather than a deliberate relaxation.  Furthermore, it is exceedingly implausible to build an implementation that violates this constraint.

